Issue #41, opened 2026-07-09
evo-graph weakness-graph custom UI view (per-algorithm dashboard)


OsherElhadad:
**Wave 3 — LAST. Needs #39 (custom-view mount mechanism) and #40 (evo-graph must write the wiki files the view reads).**

## Problem
evo-graph's value is visual — the weakness graph and the linked markdown wiki. Deliver it as evo-graph's optional custom view via #39's mechanism, reading the wiki files #40 writes.

## Design
- Port `evo-graph/dashboard/` (React SPA: ReactFlow d3-force **weakness graph**, status-colored nodes, `related`-edge links; weakness **detail panel** with `SolutionTimeline` + SSE **live progress** + rendered markdown case file; KPI strip tracking the **primary** metric only) into `skills/algorithms/evograph/dashboard/` as a custom view bundle conforming to #39's contract.
- Its **data contract** = the evo-graph wiki files under the run dir (`wiki/weaknesses/*.md`, `wiki/solutions/**`, `wiki/results/*.json`, agent logs) that #40's engine/skill writes. The view reads those read-only (like evo-graph's backend), mounted as an extra "Weakness graph" tab.
- Reuse the primary/secondary metric distinction from #38 in the KPI strip and metric switcher ("primary — decides accept/reject" vs "secondary — shown only").

## How to solve (order)
1. Adapt evo-graph's frontend components + read-only backend to #39's mount / data-contract shape.
2. Point the view at the run-dir wiki files #40 produces.
3. Wire primary/secondary labels from #38.

## Acceptance
- A `algorithm_skill: evograph` run shows the weakness-graph tab alongside the default tabs.
- Nodes/edges/detail/live-progress render from the run dir.
- Non-evograph runs are unaffected.

## Key files
`skills/algorithms/evograph/dashboard/` (new, ported); integrates via #39's mount point.
Source: `evo-graph/dashboard/{backend/app.py,frontend/src/...}`.

## Dependencies
- **Wave 3.** Needs #39 and #40.

OsherElhadad:
## Reframe — evo-graph writes data, the one dashboard renders it (no ported SPA)

[...]

Issue #40, opened 2026-07-09
Integrate evo-graph as an algorithm (both modes): deterministic engine + agent-mode skill


OsherElhadad:
**Wave 2 — needs #37 (agent-mode contract) and #38 (primary/secondary metrics + GitHub mirror). Blocks the evo-graph UI issue.**

## Problem
evo-graph is a collaborative weakness-graph optimizer with **no deterministic engine** today — it's a single agent skill. We want it as a cap-evolve algorithm in **both** orchestration modes.

## Design
- **Agent mode:** port `evo-graph/skills/evo-graph/SKILL.md` (+ `references/{questions,clustering,graph,dashboard,cost}.md`) into `skills/algorithms/evograph/` as the algorithm's **Agent-mode loop** (per #37's contract): eval train → cluster failures into a weakness graph (Obsidian-style markdown wiki = source of truth) → one solver agent per weakness in its own git worktree scoped to that weakness's `affected_tasks` → merge verified improvements → whole-round revert on regression → repeat until `stop_condition` → final test once. Keep the wiki-files contract intact so the UI issue can read them.
- **Deterministic mode:** write `core/cap_evolve/evograph.py`, a deterministic engine analogous to `gepa.py`/`skillopt.py`, registered in `skills/_registry/manifest.json`. The engine owns loop control, splits, seal, budget; it calls the optimizer agent (via the existing `run_step` / `run-optimizer` machinery in `harness.py`) as the **per-weakness edit proposer**: cluster → fan out one optimizer call per open weakness scoped to its tasks → **gate each on the primary metric averaged over its frozen `affected_tasks`** (uses #38's primary metric) → revert round on whole-train regression. Reuse `selection.py`, `gate.py`, `rundir.py`.
- **GitHub mirror** (from #38): when `github_integration: true`, mirror each weakness as a GitHub issue and ship fixes as PRs with `Closes #n`; wiki stays source of truth (evo-graph `graph.md` semantics).

## How to solve (order)
1. Agent-mode skill first (port evo-graph SKILL.md + references) — fastest path to a working evo-graph under #37.
2. Deterministic `core/cap_evolve/evograph.py` engine + manifest registration + `core/tests/` (use `--name mock` optimizer for a zero-API test).
3. Wire primary-metric weakness gating + GitHub mirror.

## Acceptance
- `algorithm_skill: evograph` runs in **both** modes.
[...]

PR #51, opened 2026-07-13
Dashboard: default template + optional per-algorithm custom view (iframe mount)

### LINKED ISSUE #39: Dashboard: general default view template + optional per-algorithm custom view from the skill
**Wave 1 — parallel with #37 and the intake issue. Blocks the evo-graph UI issue.**

## Problem
The dashboard is generic/data-driven with **zero** per-algorithm branching. We want a **general default template** any algorithm gets, **plus** an optional custom view each algorithm ships inside its own skill (e.g. evo-graph's weakness-graph wiki).

## Design
- **Default template** = today's generic tabs (Progress / Lineage / Trajectories / Git diff / Memory / Insights, driven by `core/cap_evolve/dashboard.py` + `dashboard/frontend`). Formalize it as "the default view every run gets."
- **Custom-view contribution mechanism.** An algorithm skill may bundle an optional UI under a known path (e.g. `skills/algorithms/<algo>/dashboard/`) declaring (a) a **data contract** — the files/JSON the algorithm writes into the run dir that the view reads — and (b) a **view bundle**. The cap-evolve dashboard detects `algorithm_skill` for a run and, when a custom view is present, mounts it as an **extra tab** alongside the default tabs. **No `if algorithm==` branching** in core — a single registration/mount point loads whatever the skill provides (matches cap-evolve's existing data-driven ethos).
- Define the contract precisely so future algorithms can add a view without touching dashboard internals: where the bundle lives, how it is served, what run-dir files it may read, how it is mounted.

## How to solve (order)
1. Document + implement the default-template contract (label the current generic view as the default).
2. Add the custom-view registration/mount point in `dashboard/backend` + `dashboard/frontend` (detect per-run `algorithm_skill`, mount its bundle as a tab if present).
3. Ship a trivial example custom view to prove the mechanism (evo-graph's real view is the follow-up issue).

## Acceptance
- A run whose algorithm ships a custom view shows an extra tab.
- A run whose algorithm ships none shows only the default tabs.
- Core has no per-algorithm conditionals.

## Key files
`dashboard/backend/capevolve_dashboard/` (mount point), `dashboard/frontend/src/` (tab registration), `core/cap_evolve/dashboard.py`, `dashboard/frontend/src/lib/types.ts`.
Inspiration: `evo-graph/dashboard/` (a full custom algorithm dashboard).

## Dependencies
- **Wave 1**, parallel with #37 and the intake issue.
- Blocks the evo-graph UI issue (custom-view mount mechanism).

PrielHaz:
🎥 **Demo video:** https://drive.google.com/file/d/1x_jwrXriJsZuED7Z-Nupv7menJtcPN8e/view

---

# PR: Dashboard — default template + optional per-algorithm custom view (iframe mount)

Closes #39

## What this does

Formalizes today's generic tabs as the **default view every run gets**, and adds a **generic custom-view mount** so an algorithm can ship its own UI as an extra tab — with **no `if algorithm == …` branching in the dashboard**.

The mechanism is data-driven and framework-agnostic:

- An algorithm run may write **`<run_dir>/custom_view.json`**: `{"title": "...", "url": "..."}`.
- The backend exposes it at `GET /api/runs/{id}/custom-view` (`{}` when the run declares none).
- The frontend mounts an extra tab (titled from the file) that embeds `url` in an **`<iframe>`**.

`url` can point at a served static bundle **or the algorithm's own running server** — so an algorithm brings any web UI (any framework, arbitrary HTML/React) and the dashboard just embeds it. Runs that ship no view are unaffected (default tabs only).

## Changes

- **Backend** (`dashboard/backend/capevolve_dashboard/`): new `custom_view.py` (tolerant reader — missing/invalid/`url`-less → `{}`), and a `GET /api/runs/{run_id}/custom-view` route (path-safe via the existing `resolve_run`).
- **Frontend** (`dashboard/frontend/src/`): `CustomView` type; `api.customView()`; `RunDeepDive` fetches it (fails silently → no tab) and, when a `url` is present, appends a `{ id: 'custom', label: title }` tab rendering the iframe. Rebuilt `dist`.

## Why iframe / URL (not React federation)

Keeps the contract trivial and the algorithm UI fully decoupled: no shared build, no React-version coupling. evo-graph (a follow-up) already ships its own FastAPI + React dashboard — it will just point the tab at its own server.

## Tests

- `dashboard/backend/tests/test_custom_view.py` (5): declared view returned; default title; absent → `{}`; invalid/malformed → `{}` (never 500); missing run → 404. Full backend suite: **41 passed**.

## Verification (real UI)

Ran a real toy_calc run, dropped a `custom_view.json` pointing at a demo server, launched the dashboard, and drove it with Playwright: the extra tab appears, clicking it renders the embedded page (confirmed by reading the iframe's content back through the frame), and **all default tabs still render** (no regression). Screenshot captured.

## Not in scope

- **evo-graph's actual weakness-graph view** — arrives in #41 (which points this tab at evo-graph's dashboard and reads the run-dir wiki files #40 writes). This PR is only the generic socket + a proof.


PrielHaz:
### Demo — custom-view iframe tab

A run with `custom_view.json = {title, url}` mounts an extra tab (here "Weakness graph (demo)") that embeds the URL, alongside all the default tabs (no regression):

custom-view dashboard tab

_Verified live with Playwright: the tab renders the embedded page and every default tab still works._

PrielHaz:
📎 **Demo assets on Google Drive** (folder: https://drive.google.com/drive/folders/15CrnKwnzb8ZGayvyFb2KcrpT06a4tiJf)
- Custom-view iframe tab: https://drive.google.com/file/d/1LCdH_ajQVzp6Zey4B8E1SSJfxzJ_tKZN/view

PrielHaz:
🎥 **Demo video** (Google Drive, ~15–25s): https://drive.google.com/file/d/1x_jwrXriJsZuED7Z-Nupv7menJtcPN8e/view

Walks through this PR's work end to end. (Recorded with Playwright; `.webm` — plays in Drive/Chrome.)

OsherElhadad:
## Review — honest take: right ethos, wrong default

First, credit where due: this is small (180 LOC), the reader is tolerantly coded, the URL-scheme allowlist + `sandbox`/`referrerPolicy` on the iframe are careful, and the tests cover the malformed cases. As a *mechanism* it's clean.

My concern is not the code — it's what this makes the **default path** for extending the dashboard. `custom_view.json = {title, url}` → iframe means the easy way for any algorithm to add a view is *"stand up your own web server / ship your own bundle and we'll embed it."* That's what produced #53: an entire second FastAPI backend + second React build + committed minified `dist` + `fastapi/uvicorn/pyyaml` deps + a second process on :7878 — to draw a node graph the main dashboard's stack can already draw. Multiply that by every future algorithm and we have N dashboards in a trench coat, not one extendable dashboard.

### What the one dashboard already has (verified in-tree)
- Backend: path-safe run-dir file reads (`files.py` `_safe_join`, `/api/runs/{id}/tree`, `/file`), SSE (`/stream`), git/memory/rollouts, static mount — **one process**.
- Frontend: `Tabs`/`TabDef` registration, a **status-colored node graph** (`LineageTree.tsx`, `FILL` = seed/accepted/rejected/failed), `KpiStrip`, `recharts`, a query+SSE hook.

So ~80% of any "custom view" is already sitting in the shared dashboard. The iframe throws all of it away and reimplements it per algorithm, in a frame that can't share theme, selection, sizing, or deep-links with the parent.

### The redesign: declarative panels, not embedded apps
Keep #39's data-contract half (algorithm writes files into the run dir) exactly as-is. Replace the *view-bundle* half with a **declarative manifest** the ONE dashboard renders with built-in panel types:

```jsonc
// <run_dir>/views.json  (tolerant reader, same safety as this PR)
{ "tabs": [ { "id": "weakness-graph", "label": "Weakness graph", "panels": [
  { "type": "kpi",      "source": "wiki/results/latest.json", "metric": "primary" },
  { "type": "graph",    "source": "wiki/graph.json", "node": {"status":"status","label":"title"}, "edge": {"field":"related"} },
  { "type": "markdown", "source": "wiki/weaknesses/{slug}.md", "selectFrom": "graph" }
] } ] }
```

- Add ONE endpoint `/api/runs/{id}/views` (reuse this PR's tolerant reader).
- Add a generic `PanelRenderer` switching over a small vocabulary — `graph | kpi | table | markdown | chart | timeline` — reading run-dir files through the **existing** `/file` + `/tree`.
- New algorithm → writes data + a manifest → gets a **native** tab. No branching, no second server, no second build, no extra deps, no iframe. Same data-driven ethos #39 asked for; the data now drives native panels instead of a foreign app.

### Keep this PR — but as the escape hatch, not the headline
The iframe is legitimately the right tool for the genuine 5%: a heavyweight, independently-maintained UI cap-evolve doesn't want to own. So keep `custom_view.json`/iframe, but **document it as "last resort — you own the server and its lifecycle,"** and land the declarative panel contract as the primary mechanism. I'd hold merging this as-is until the panel contract exists, so it lands framed as the fallback rather than the norm.

Happy to implement the `views.json` + `PanelRenderer` slice if we agree on the direction.

PrielHaz:
I'm reworking it your way — the full plan is on #53, in short:

- `<run_dir>/views.json` declares panels (`graph | kpi | table | markdown | chart | timeline`); `GET /api/runs/{id}/views` in **this** backend reads it; a generic `PanelRenderer` in the shared frontend draws them, reusing `LineageTree`'s status-coloured graph, `KpiStrip`, `recharts` and the existing `/file` + `/tree`. No second process, no new deps, no committed bundle.
- evograph ships `graph.json` + `views.json` and zero UI; `skills/algorithms/evograph/dashboard/**` and `scripts/view.py` are deleted.
- You were right that the real gap is one missing primitive (a markdown renderer + a generic graph panel), not a dashboard.

For **this** PR specifically, what I'd like your call on: the iframe mechanism is still the honest escape hatch for the case a view genuinely can't be expressed as panels (an existing third-party UI you don't own). I'd keep it, but demote it in the docs from "how to add a view" to "last resort, and here's why you probably don't want it". Merge it on that basis, or hold it until the panel work lands and re-review both together? I'm fine either way.

CI is green here (8/8) and the branch is rebased on current `main`, so there's nothing mechanical blocking it.

dashboard/frontend/src/routes/RunDeepDive.tsx
```diff
@@ -119,6 +135,32 @@ export function RunDeepDive() {
...
+                        <span aria-hidden="true">↗</span>
+                      </a>
+                    </div>
+                    <iframe
+                      src={customUrl}
+                      title={customView?.title || 'Custom view'}
+                      // The embedded page is data-driven (declared by the run dir), so
+                      // sandbox it: allow it to run scripts and talk to its own origin,
+                      // but deny top-level navigation, popups, and same-origin trust.
+                      sandbox="allow-scripts allow-same-origin allow-forms"
```
OsherElhadad, line 162: `allow-scripts` + `allow-same-origin` together only sandboxes as intended while `customUrl` resolves to a **different** origin than the dashboard (today it does — a separate localhost port). If a deployment ever serves the custom view from the dashboard's own origin, the framed doc can reach up and remove its own `sandbox` attribute — the classic escape. `allow-same-origin` is genuinely needed here (the embedded app uses its own fetch/storage), so I'm not asking you to drop it; just flagging the caveat.

Also the comment above ("deny … same-origin trust") reads as if same-origin is denied, but `allow-same-origin` is set — worth rewording to "the frame keeps access to *its own* origin's storage; it is denied the parent's origin, top-level navigation, and popups."

OsherElhadad (commented):
## Review — dashboard custom-view mount

Clean, well-scoped, and genuinely well-tested. The data-driven `custom_view.json` contract with **no per-algorithm branching** in the dashboard is the right design, and `custom_view.py` is nicely defensive: absent/invalid/non-object/blank-url all collapse to `{}`, and it rejects `javascript:`/`data:`/scheme-relative/relative/`ftp:` so only absolute `http(s)` reaches the iframe `src`. `test_custom_view.py` covers all of those paths incl. the 404. Nice.

### One real merge hazard (not a code bug)
The committed `dashboard/frontend/dist/*` bundle in this PR was built from a branch that predates `main`'s later frontend work (`applyDataBaseOverride` / `isLiveOverride` / `LivePendingError` / lazy `dataBase()` — all present on `main`, absent here). The three-dot diff shows this PR's *source* touches `api.ts` by only +4 lines (good), but the **built** bundle here does not contain `main`'s live-override feature. Merging this `dist` would ship a stale bundle / conflict on `index.html`'s hashed asset refs.

**Recommendation:** rebase on `main` and rebuild `dist` (`cd dashboard/frontend && npm ci && npm run build`) as the final commit before merge — or drop `dist` from the PR and let CI build it. Please confirm the bundle is rebuilt post-rebase so the live-override feature isn't regressed.

### Minor
Inline note below on the iframe sandbox. Everything else LGTM.

OsherElhadad:
Pushed a fix to this branch (`fb2666c`): clarified the iframe `sandbox` comment and documented the `allow-scripts`+`allow-same-origin` cross-origin caveat. The **dist rebuild** item is left for you to own — please rebase on `main` and rebuild `dashboard/frontend/dist` before merge so the committed bundle isn't stale vs. main's live-override frontend work.

PR #52, opened 2026-07-13
evograph: evo-graph as a cap-evolve agent-mode algorithm

PrielHaz:
🎥 **Demo video (evo-graph UI *inside* the cap-evolve dashboard):** https://drive.google.com/file/d/1bjIW38CzJSrSPWdyW0-3KNeTenRk0qES/view

---

# PR: evograph — evo-graph as a cap-evolve agent-mode algorithm

Closes #40

> **Stacked on #37 (PR #50)** and depends on #38 (PR #49). Its diff shows those commits until they merge; review/merge #38 then #37 first.

## What this does

Adds `skills/algorithms/evograph/` — evo-graph ported as a cap-evolve **agent-mode** algorithm. It keeps evo-graph's distinctive machinery (the weakness graph, per-weakness solver worktrees, whole-round revert, and the exact wiki file formats the dashboard reads) but swaps everything else to cap-evolve's conventions.

Selected with `algorithm_skill: evograph` + `orchestration_mode: agent`: cap-evolve runs intake → check → baseline, then hands the loop to the coding agent (per #37), which follows this skill's round loop and seals with `cap-evolve finalize`.

## Fits cap-evolve (replace vs keep)

| stock evo-graph | evograph here |
|---|---|
| own setup Q&A | **cap-evolve intake** (reads `capevolve.yaml`; asks nothing of its own) |
| discover & run the bench | **cap-evolve adapter + harness** (eval writes run-dir rollouts/results) |
| own train/test split | **cap-evolve seeded splits**; test **sealed** until `finalize` |
| own final-test.json + cost prompt | **`cap-evolve finalize`** owns the sealed number (mirrored into the wiki for the view) |
| wiki under `.evograph/` | wiki under the **cap-evolve run dir** (`<run_dir>/wiki/…`) |
| own primary/secondary config | the spec's `metric_primary` / `metrics_display` (#38); gate on primary only |
| standalone dashboard | writes `<run_dir>/custom_view.json` so its weakness-graph tab mounts via #39 |

**Kept:** the weakness-graph round loop (eval train → cluster failures → one solver per weakness in its own worktree scoped to frozen `affected_tasks` → merge → whole-round revert on regression → repeat until `stop_condition`), the wiki formats, and `github_integration` (mirror weaknesses as issues / ship as PRs, mirror-only).

## Agent-mode only (by design)

[...]

PR #50, opened 2026-07-13
Dual-mode orchestration: orchestration_mode (deterministic | agent)

PrielHaz:
🎥 **Demo video (UI):** https://drive.google.com/file/d/1SYKqv3iqbEuDG0mHKmBj42mbLQef4X-g/view

---

# PR: Dual-mode orchestration — `orchestration_mode: deterministic | agent`

Closes #37

> **Stacked on #38 (PR #49).** This branch builds on the #38 metrics/intake work (shared edits to `intake` + `capevolve.yaml`). Until #38 merges, its commits also appear in this diff; they drop out once #38 lands. Review/merge #38 first.

## What this does

Adds a per-run `orchestration_mode`:

- **`deterministic`** (default) — today's behavior, **byte-for-byte unchanged**: `cap-evolve run` sequences intake → check → baseline → algorithm → finalize → report, with code-enforced honesty.
- **`agent`** — the **coding agent the user is talking to drives the optimization loop itself**. It reads the selected algorithm's "Agent-mode loop" prose, makes the edits, and runs evaluations directly — through cap-evolve's own eval/gate/finalize primitives, so the run dir and dashboard stay populated. It does not delegate the search to a separate optimizer agent.

The sealed headline number stays free and unfakeable: agent mode ends by calling the existing `cap-evolve finalize` (`RunDir.reserve_test`/`commit_test`) — no new seal.

## Changes

- **Spec key + intake** (`templates/project/capevolve.yaml`, `skills/phases/intake/{SKILL.md,inputs/INPUTS.md}`): `orchestration_mode: deterministic | agent`, asked in intake (default deterministic).
- **CLI branch** (`core/cap_evolve/cli.py`): `_cmd_run` reads the mode; `--plan-only` surfaces it. In agent mode, cap-evolve runs the hard check + baseline (creating the run dir), then prints a handoff (`{mode, run_dir, algorithm, stop_condition, next}`) and returns — the agent drives from there and seals with `finalize`. The deterministic block is untouched below the branch.
- **Agent-mode driver loop** (`skills/orchestrate/orchestrate/SKILL.md`): the contract the conversational agent follows — drive through cap-evolve primitives (so `events.jsonl`/rollouts/results/snapshots populate and the dashboard renders unchanged), self-police honesty (no test until the end, revert on regression, val-only acceptance), verify each round wrote its run-dir artifacts, re-read `stop_condition`, seal once.
[...]

Issue #38, opened 2026-07-09
Intake: primary vs shown-only metrics, GitHub integration, inspect-first onboarding


OsherElhadad:
**Wave 1 — parallel with #37 and the UI-framework issue. Blocks the evo-graph integration issue.**

## Problem
cap-evolve `score()` returns a single scalar reward. evo-graph distinguishes **one primary metric** (gates accept/reject) from **shown-only secondaries** (display, don't gate), and supports optional **GitHub** for the optimization process. Intake is also a flat questionnaire; evo's `discover` inspects the repo first and asks the *fewest* questions. Adopt these into cap-evolve intake.

## Design
- **Primary vs shown-only metrics.** Extend `core/cap_evolve/types.py::Score` to carry secondary metrics alongside the primary reward, encoded as evo-graph does: `{value, primary, direction}` (exactly one `primary: true`; `direction: higher|lower`). The **gate uses only the primary** (`core/cap_evolve/gate.py` logic unchanged — it already gates on the scalar; formalize that scalar = the primary metric). Secondaries flow through rollouts/results JSON for display only. Intake asks: which metrics to display, which is primary, each one's direction.
  - Reference shape: `evo-graph/skills/evo-graph/references/{questions,graph,dashboard}.md`, `evo-graph/dashboard/frontend/src/lib/{types.ts,metrics.ts}`.
- **GitHub integration.** New spec key `github_integration: true|false`. Intake runs `gh auth status`; if authed, offers to mirror candidates/weaknesses as issues and ship the winner as a PR (`Closes #n`); else offers `gh auth login` or skip. GitHub is optional and **never the source of truth** — the run dir stays authoritative (mirror-only). Ref: `evo-graph/skills/evo-graph/references/{questions,graph}.md`, `evo/plugins/evo/skills/ship/SKILL.md`.
- **stop_condition** (free text, re-read each round) added for agent mode; deterministic mode keeps its budget knobs. Surface the dashboard port (already `dashboard_launch.py`, default 7878).
- **Inspect-first onboarding UX.** Refine `skills/phases/intake/SKILL.md` to inspect the repo/benchmark/agent *before* asking (like `evo/plugins/evo/skills/discover/SKILL.md` Steps 1-3) and ask the **fewest** questions — proposing detected metrics/splits/caps as defaults with multiple-choice + free-text, rather than a flat list. Keep the existing ask-user-if-missing discipline for NEEDED inputs.

## How to solve (order)
[...]

PR #53, opened 2026-07-13
evograph weakness-graph view (mounted in the cap-evolve dashboard)

### LINKED ISSUE #41: evo-graph weakness-graph custom UI view (per-algorithm dashboard)

PrielHaz:
🎥 **Demo video (evo-graph UI *inside* the cap-evolve dashboard):** https://drive.google.com/file/d/1_DTpj907jUrzExeld6nt-3JbZYMoatn8/view

---

# PR: evograph weakness-graph view (mounted in the cap-evolve dashboard)

Closes #41

> **Stacked on #40 (PR #52)** → #37 (#50) → #38 (#49), and composes with #39 (PR #51, the custom-view mount). Its diff shows the stack until those merge; merge order: #38 → #37 → #39 → #40 → #41.

## What this does

Delivers evo-graph's weakness-graph UI as evograph's optional custom view — mounted as a **Weakness graph** tab in the cap-evolve dashboard via the #39 iframe/URL contract, reading the run-dir wiki that #40's skill writes.

- Bundles evo-graph's read-only dashboard under `skills/algorithms/evograph/dashboard/` (FastAPI backend + prebuilt React SPA: ReactFlow/d3 weakness graph with status-colored nodes and `related` edges, KPI strip on the primary metric, metrics-over-rounds chart, round timings, per-weakness detail + `changes.diff` + SSE live progress).
- Adds `scripts/view.py`: points the backend at the cap-evolve **run dir** (`EVOGRAPH_BASE=<run_dir>`), serves the bundled dist, and writes `<run_dir>/custom_view.json` = `{title, url}` so the cap-evolve dashboard mounts the tab (the #39 contract). One command; runs in the background.
- Data contract = the run-dir wiki #40 writes (`wiki/weaknesses/*.md`, `wiki/solutions/**`, `wiki/results/*.json`, `runs/**/agents/*.log`). The view is **read-only**.

Non-evograph runs are unaffected (no `custom_view.json` → no tab).

## Verification (real UI, end-to-end)

Built a run dir with real evo-graph wiki content, launched `view.py` + the cap-evolve dashboard, and drove it with Playwright.

**1. The evograph weakness-graph view (served from the run-dir wiki):**

evograph weakness graph

**2. Embedded as the "Weakness graph" tab inside the cap-evolve dashboard, alongside all default tabs (no regression):**

embedded in cap-evolve tab

Confirmed via Playwright: the embedded evo-graph SPA renders (2 weaknesses, `wrong-tool-arg` completed with a record + `missing-validation` open, joined by the `shared root cause` edge), and the default tabs (Overview … Insights) still render. `/api/graph` returns the nodes/edges and `/api/health` reports `wiki_exists: true`.

## Notes

- The evograph backend sets no frame-blocking headers, so it embeds cleanly in the iframe tab.
- Backend deps (fastapi/uvicorn/pyyaml) install via `dashboard/backend/requirements.txt` (the SKILL notes this one-time step).


PrielHaz:
### Demonstration — weakness detail panel

Clicking a weakness node opens its detail: the primary-metric **solution timeline** (kept 0.85 solution with a new-record ribbon), the **live auto-research log** streamed from the run-dir agent log, the rendered markdown **case file**, and the **Rejected Store Memory** — all read from the run-dir wiki #40 writes.

weakness detail panel

PrielHaz:
📎 **Demo assets on Google Drive** (folder: https://drive.google.com/drive/folders/15CrnKwnzb8ZGayvyFb2KcrpT06a4tiJf)
- evograph weakness graph: https://drive.google.com/file/d/14w4CRU3g5cHfEJCyZJIWdYhuXkZV77X_/view
- Embedded in the cap-evolve tab (end-to-end): https://drive.google.com/file/d/174J65JWYTeAEmfI452wYfzlfUCsCkKdH/view
- Weakness detail panel: https://drive.google.com/file/d/1LnrBrsqsLXnYJ6LHmoRqEiT1AP6-c9Ho/view

PrielHaz:
🎥 **Demo video** (Google Drive, ~15–25s): https://drive.google.com/file/d/1EFFxXCN7wWK1A2Upm31ZA54UicJHciiU/view

Walks through this PR's work end to end. (Recorded with Playwright; `.webm` — plays in Drive/Chrome.)

OsherElhadad:
## Review — request changes: this relocates evo-graph, it doesn't integrate it

The end result in the demo looks great, and the `#40` engine work (writing the wiki into the run dir) is the valuable, keep-worthy half. But as an *architecture* this is the thing #39/#41 were meant to avoid: a **second dashboard** — its own FastAPI backend (`dashboard/backend/{app.py,server.py}`), a second React build with a **committed minified `dist`**, new runtime deps (`fastapi`/`uvicorn`/`pyyaml`), a second server on :7878 launched by `view.py`, plus its own `test_app_security.py` because it now has its own attack surface — glued into cap-evolve through #51's iframe. "evo-graph must not be an external repo with its own dashboard" was the goal; this moves that external dashboard into `skills/` verbatim and iframes it. That's relocation, not integration.

### Cost this imposes, concretely
- Users must `pip install -r requirements.txt` (fastapi/uvicorn/pyyaml) **and** run `view.py` in the background on a second port **and** run the cap-evolve dashboard, and keep them alive together. The main project advertises "runtime deps: 0 (stdlib)"; every algorithm view adding a web stack erodes that.
- The evo-graph SPA reimplements what the shared dashboard already has: `app.py` re-reads the run dir (we already have path-safe `/tree`+`/file`), redraws a status-colored graph (`LineageTree.tsx` already does status colors), a KPI strip (`KpiStrip`), SSE (`/stream`), markdown detail. Duplicate code, duplicate maintenance.
- iframe embedding: off-theme (won't inherit cap-evolve's fonts/tokens), fixed `80vh` with a second scrollbar, no shared selection/deep-linking, an "open in new window" band-aid.
- **Reviewability:** 1976 additions that are mostly a minified bundle can't be meaningfully reviewed. A committed `dist` is a supply-chain and drift hazard.

### What I'd do instead (ties to the #39 redesign I posted)
Deliver the weakness graph as **data + a declarative manifest** rendered by the ONE dashboard:
1. **Drop** `skills/algorithms/evograph/dashboard/**` entirely — backend, frontend, `dist`, `requirements.txt`, and `scripts/view.py`.
2. **Keep** the #40 engine: have evo-graph write `wiki/weaknesses/*.md`, `wiki/results/*.json`, and a small `graph.json` (nodes+`related` edges) into the run dir — read-only data, which it already produces.
3. Add `<run_dir>/views.json` declaring a `graph` + `markdown-detail` + `kpi` tab (see #39).
4. In the **shared** frontend, add the one missing primitive — a relationship/force graph panel (a small component, or a single markdown dep for the case files) — reused by every future algorithm, not just evo-graph.

Result: the "Weakness graph" tab renders natively inside the one dashboard, on-theme, no second server, no extra deps, no committed bundle — and non-evograph runs are still unaffected (no `views.json` → no tab).

### Also
- This PR is stacked on #38→#37→#39→#40; the diff is huge partly because of that. Please split the **engine** (#40) from the **view** (#41) so each is reviewable on its own.
- Whatever we land, the primary/secondary metric distinction from #38 in the KPI strip is the right call — keep it; it maps cleanly onto a `kpi` panel `metric` field.

Net recommendation: **hold**, rework onto the declarative-panel contract, and delete the bundled second app.

PrielHaz:
@OsherElhadad — rebased onto current `main`, conflicts resolved, and CI is green 8/8 (including the CodeQL results gate, which now reports "no new alerts": the path handling in the backend was rewritten so every served path comes out of a directory listing (`child_named`, fixed globs + name match) instead of joining a route param, plus 7 regression tests for the traversal payloads).

But: **I'm accepting your review, not arguing it.** Two things convinced me.

1. Your cost list is accurate. This PR asks the user to `pip install fastapi/uvicorn/pyyaml`, run `view.py` on a second port, keep it alive next to the real dashboard — in a project that advertises 0 runtime deps — and it re-implements `/tree`+`/file`, a status-coloured node graph, a KPI strip, SSE and markdown detail that the shared dashboard already has.
2. Something worse that I hit myself while re-testing this week: `git ls-files skills/algorithms/evograph/dashboard/` returns the backend plus `dist/index.html` + two minified asset files — **and nothing else**. There is no frontend source in the repo. I found a real rendering bug in the embedded view (the final-test panel's secondary metrics overflow their card over the header) and there is no way to fix it in-tree; you'd have to go back to the external repo, rebuild, and re-commit a new opaque bundle. That's your "unreviewable / supply-chain hazard" point, but concretely worse than unreviewable: it's unmaintainable.

So the plan I'm executing, which is your proposal:

- `<run_dir>/views.json` — a declarative manifest an algorithm writes, e.g. `[{ "id": "weakness-graph", "title": "Weakness graph", "panels": [{ "type": "graph", "src": "wiki/graph.json" }, { "type": "kpi", ... }] }]`
- `GET /api/runs/{id}/views` in the existing backend — tolerant reader, same path safety, no new process and no new deps.
- A generic `PanelRenderer` in the shared frontend over a small vocabulary (`graph | kpi | table | markdown | chart | timeline`), reusing `LineageTree`'s graph rendering, `KpiStrip` and `recharts`. Panel data is read through the existing `/file` + `/tree`.
- evograph then writes `graph.json` + `views.json` and ships **zero** UI: `skills/algorithms/evograph/dashboard/**` and `scripts/view.py` get deleted from this PR.
- #51's iframe stays, but demoted in the docs to the last-resort escape hatch, not the way to add a view.

You were right that the gap is one missing primitive (markdown renderer + a generic graph panel), not a whole dashboard.

**Question so I don't waste a round-trip:** do you want me to (a) close/gut this PR and reopen the panel work as a fresh PR on top of #52, or (b) force-push the rework into this branch so the discussion history stays attached? I'd default to (b) unless you say otherwise. Either way #52 (the skill + wiki writer, no UI, no web stack) is independent and ready — I've asked over there whether you can merge it now.

skills/algorithms/evograph/dashboard/backend/app.py
```diff
@@ -0,0 +1,353 @@
...
+        # re-resolve each loop so a log appearing mid-run is picked up
+        nonlocal log
+        while True:
+            cur = log or latest_log_for(slug)
+            if cur and cur.exists():
+                log = cur
+                with cur.open("r", encoding="utf-8", errors="replace") as f:
+                    f.seek(pos)
+                    chunk = f.read()
+                    pos = f.tell()
```
OsherElhadad: **Bug: the stream never picks up a newer log, contradicting the comment on the line above.** Once `log` is set (`log = cur`), `cur = log or latest_log_for(slug)` short-circuits on the truthy `log` and `latest_log_for` is never called again — so a later round's `<slug>.log` for the same weakness is never streamed. The `# re-resolve each loop` comment describes behavior the code doesn't have. It also means a file swap would reuse a stale `pos`.

Re-resolve every iteration and reset `pos` when the file actually changes:

```suggestion
            cur = latest_log_for(slug) or log
            if cur and cur.exists():
                if cur != log:  # a newer round's log appeared -> start from its top
                    log, pos = cur, 0
                with cur.open("r", encoding="utf-8", errors="replace") as f:
                    f.seek(pos)
                    chunk = f.read()
                    pos = f.tell()
```

skills/algorithms/evograph/dashboard/backend/app.py
```diff
@@ -0,0 +1,353 @@
...
+
+
+@app.get("/api/results")
+def results() -> dict[str, Any]:
+    return {"results": load_all_results()}
+
+
+@app.get("/api/run-config")
+def run_config() -> dict[str, Any]:
+    """The run's `.evograph/run-config.json`, if the user dropped one in. Optional and free-form:
```
OsherElhadad: Doc drift: the view reads `BASE / "run-config.json"` (the cap-evolve run dir), not `.evograph/run-config.json` — that's the stock evo-graph layout this PR deliberately moved away from. Reword so the docstring matches the path actually read:

```suggestion
    """The run's `run-config.json` under the run dir, if the user dropped one in. Optional and free-form:
```

skills/algorithms/evograph/dashboard/backend/server.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env python3
```
OsherElhadad: This whole file looks like dead carryover from stock evo-graph. `view.py` is the real entry point and it launches uvicorn directly (`uvicorn.run("app:app", ...)`) — it never calls `server.py`. And this docstring describes a launch path that doesn't exist in cap-evolve ("Launched by bootstrap.py with cwd = .evograph/" — there is no `bootstrap.py` and no `.evograph/` here). Nothing in the PR references it (confirmed via grep). Recommend **deleting `server.py`** to avoid a second, misleading entry point; `view.py` already covers direct launch. If you want a bare `python … --port` runner, keep it but fix the docstring to the run-dir/`view.py` reality.

OsherElhadad (commented):
## Review — evograph weakness-graph view (mounted via #39)

The backend path-safety design is genuinely good: `child_named()` enumerates the fixed parent dir and matches by exact name instead of joining untrusted route params, so `..`/absolute/encoded-slash payloads simply never match. `test_app_security.py` locks that in (traversal on `slug`/`weakness`/`sol_id` + the `parse_rejected` heading filter). Front-matter/JSON parsing is tolerant throughout, `run-config` errors don't leak paths, and setting no frame-blocking headers so it embeds in the #39 iframe is the right call for a read-only localhost view. `view.py`'s pre-flight (refuse to register a tab pointing at a UI that isn't built, unless `--register-only`) is a nice guard.

Three items below — one real behavioral bug (SSE), one dead file, one doc drift.

### Also worth confirming
- `requirements.txt` uses unpinned `>=` — fine for a bundled dev tool, matching the main dashboard's style. No change needed unless you want reproducibility.
- The bundled `frontend/dist` is committed (with a matching `.gitignore` allowlist) — same rebuild-on-source-change caveat as PR #51's dist applies if the SPA source ever changes.

OsherElhadad:
Pushed fixes to this branch (`4ce06a9`): SSE `progress_stream` now re-resolves the log each tick and resets `pos` on file change (was short-circuiting and never streaming a newer round's log); deleted the dead `server.py`; fixed the `run_config` docstring path; and added `Task` to `allowed-tools` to keep the stack consistent with #52.

PR #55, opened 2026-07-16
docs(capabilities): teach how to optimize skills & tools for the AI-agent reader

eranra:
Closes #54.

## Summary
Enhances the four cap-evolve capability skills so each teaches how to optimize its layer of a skill/tool **for the AI-agent reader, not the human**. Docs-only — no infrastructure or code changes.

- **4-layer model** (Description → Snippets → Tools → Implementation) mapped to the capability that owns each layer, with route-to-layer guidance mirroring cap-evolve's propose→evaluate→gate loop.
- **Write-for-agents principle** — new `skill-package/references/writing-for-agents.md` (canonical, with checklist); compact, capability-scoped restatements in `tools`, `mcp-tool`, `system-prompt`.
- **Tool build-and-validate discipline** — new `tools/references/authoring-and-validation.md`: full tool spec fields, strict docstring rules (serve slot-filling), the test/validate pipeline (serve functional correctness), error-case discipline, granularity/composition/observability, and script-interface rules (non-interactive, --help, structured output, exit codes, idempotency, dry-run, truncation-safe, self-contained deps).
- **Description/snippet enrichment** — task decomposition, keyword enrichment, decision trees, failure-handling/recovery blocks, example quality, redundancy removal, and per-layer metrics.
- **`skills/capabilities/SHARED-TEXT.md`** — external maintenance index tracking duplicated passages across capabilities so future edits can be synced.
- Also fixes a pre-existing broken `references/pitfalls.md` pointer in `system-prompt/SKILL.md`.

## Design decisions
- **Per-capability content** (each skill self-contained) rather than a shared cross-referenced file, per the maintainer's choice; duplication is tracked explicitly in `SHARED-TEXT.md`.
- All new long-form content goes in `references/`; `tools/SKILL.md` (already over the body-token budget) received only a single pointer line.

## Sources synthesized
AI Skill Optimization Guide (4-layer model), skillberry-skill-maker tool_making/remediation stages (tool spec + validation pipeline), agentskills.io skill-creation docs, Anthropic 'Writing tools for agents' / OpenAI function-calling / GEPA-DSPy / BFCL-ToolLLM (already cited in the references).

## Testing
Docs-only. Verified: only `.md` files changed; all reference links resolve; body-token budgets under limit for skill-package/mcp-tool/system-prompt (tools was already over-budget pre-change and grew by only the one pointer line). Design spec and implementation plan included under `docs/superpowers/`.

Signed-off-by: Eran Raichstein <eranra@il.ibm.com>

PR #58, opened 2026-07-20
feat: consuming-LLM awareness — target-model profiles

### LINKED ISSUE #57: Make optimization aware of the consuming LLM (target-model profiles)
## Problem

cap-evolve treats the **optimizer LLM** (which proposes edits) as a first-class, configurable concept (`optimizer_model`), but the **consuming/runtime LLM** — the model the agent under test actually reads the capabilities with — is only an adapter wiring detail. The optimizer proposes edits blind to how capable the reader is, and the capability guidance implicitly assumes a strong, modern reader (e.g. "soften MUSTs", "explain the why, models follow reasoning"). That advice is often backwards for a weaker consuming model.

## Ask

Let a user declare the runtime/consuming LLM — by concrete model id **or** an abstract capability tier (`frontier | strong | mid | weak`) — so the optimization process (optimizer prompt + capability guidance) adapts its proposed edits (text, slot-filling, tool docs, enforcement strategy) to fit that reader.

## Scope

- New `capevolve.yaml` fields: `target_model` (id or tier) + `target_profile_file` (override). Blank = today's behavior.
- A profile resolver with built-in tiers (prose "reader briefs") + a name→tier lookup.
- Inject the resolved brief into the optimizer prompt (new `{{TARGET_READER}}` slot).
- Tier-conditional guidance across all four capability skills.
- `cap-evolve check` warns (non-blocking) on a consuming-vs-runner tier mismatch.
- report + dashboard surface the consuming model alongside the optimizer model.
- Honesty core (gate / splits / seal) untouched.

eranra:
Closes #57.

## Summary
Adds a first-class notion of the **consuming/runtime LLM** — the model the agent under test reads the capabilities with — distinct from `optimizer_model` (which proposes edits). Declaring it steers the optimizer prompt and the capability guidance to optimize *for that reader*: a weaker reader gets more explicit rules, worked examples, literal slot-filling docs, and code enforcement; a frontier reader gets leaner prose that explains the *why*.

Declare via two new optional `capevolve.yaml` fields:
- `target_model` — a concrete model id (resolved to a tier via a built-in name→tier map) **or** a tier keyword: `frontier | strong | mid | weak`.
- `target_profile_file` — optional raw-text brief overriding a tier's built-in brief.

Blank `target_model` = **exactly today's behavior** (backward compatible, verified end-to-end via toy_calc).

## What's included
- `core/cap_evolve/target_profile.py` — zero-dependency resolver: built-in tiers (prose "reader briefs") + name→tier lookup. Briefs live as Python data (no YAML-parse fragility).
- New `{{TARGET_READER}}` slot rendered into the optimizer prompt; threaded `capevolve.yaml` → `cli.py` → `hill-climb/run.py` → `hill_climb_loop` → `_focus_instructions`, mirroring the existing `--instructions-file` path.
- Tier-conditional *"Adapting to the reader's capability tier"* guidance in all four capability skills (system-prompt, tools, mcp-tool, skill-package).
- `cap-evolve check` warns (non-blocking) on a consuming-vs-runner tier mismatch, via a new optional `adapter.runner_model()` hook.
- report + dashboard surface the consuming model + tier alongside the optimizer model.
- intake INPUTS.md / SKILL.md + config template document the new inputs.
- tau2-airline example opts in (`gpt-oss-120b`, tier `mid`).

## Design decisions
- **Honesty core untouched** — gate, splits, and the test seal are unchanged. The profile influences *what edits are proposed*, never *whether a candidate is accepted*.
- **Warn, don't block** on mismatch — respects proxy/local-endpoint setups; the adapter hook is optional and the declaration is trusted when absent.
- Built-in tier briefs as Python data (not parsed YAML) to honor the zero-runtime-dependency constraint and avoid the minimal reader's single-line-scalar limitation.

## Testing
- New tests: resolver (8), prompt render + agnostic no-token (2), end-to-end wiring, metadata payload + dashboard surfacing (2), check mismatch warning (3).
- Full suite green: **134 passed**.
- Backward-compat verified end-to-end: `bash examples/toy_calc/run.sh` still yields `baseline_val 0.0 → test_reward 1.0`.
- Config parses under both PyYAML and the zero-dependency minimal reader.

Spec: `docs/superpowers/specs/2026-07-20-consuming-llm-profile-design.md`
Plan: `docs/superpowers/plans/2026-07-20-consuming-llm-profile.md`


OsherElhadad:
## Review — PR #58 (consuming-LLM target-model profiles)

Read the full diff, the spec/plan, and ran a real before/after on tau2-airline (below). Verdict: **a genuinely good, well-scoped idea, cleanly implemented and safely bounded — but shipped with zero evidence it improves outcomes, plus a few concrete rough edges.**

### What's right
- **The premise is sound and literature-backed.** The consuming model *is* conceptually distinct from the optimizer, and optimal prompt style is model-dependent (explicit rules / few-shot help weaker readers; over-constraining can hurt frontier ones). Naming that reader to the optimizer is a real gap to close.
- **Honesty core untouched.** Gate / splits / seal are not modified; the profile only influences *what is proposed*, never *what is accepted*. Correct thing to leave alone.
- **Backward-compatible & tested.** Blank `target_model` → empty `{{TARGET_READER}}` → byte-identical prompt, with a test asserting the token is substituted away. Resolver / render / wiring / metadata / dashboard / check-mismatch all covered.
- **Zero-dependency** tier data as Python (not parsed YAML), `warn-don't-block` on mismatch — both pragmatic.
- Confirmed working end-to-end: the `target_profile` event fires (`mid`), and the injected reader block **measurably changes the optimizer's edit style** in the intended direction (see run below).

### Ruthless part — the real problems
1. **No empirical validation, on a framework whose entire thesis is "don't trust an improvement you didn't measure on a held-out split."** The PR changes optimization behavior and ships no before/after numbers. That's the central irony. My run (below) is an attempt to supply the missing evidence and — even trying — can't, at N=3. That's not an excuse for the PR; it's a reason the PR should include at least one real, powered run before claiming the tier advice helps.
2. **`suggested_num_trials` is dead data.** Every tier defines it (frontier=1 … weak=5), the spec explicitly muses about seeding `num_trials` from it, the run *logs* it (`suggested_num_trials=5` in my after-run) — but **nothing ever consumes it**. `grep -rn suggested_num_trials core/ skills/` outside `target_profile.py`+tests = 0 hits. Either wire it (with the "default the user may override" semantics the spec promised) or delete it; right now it's a promise the data model doesn't keep.
3. **`MODEL_MAP` rots and is silently wrong for known models.** It's a hardcoded id→tier dict that already omits models on the very proxy this example targets — `claude-opus-4-7`, `claude-sonnet-4-6`, `gemini-*`, `gpt-4.1` — all of which fall through to `DEFAULT_TIER="strong"` **with no `resolution_note`** (the note only fires for *fully* unknown ids). A frontier `opus-4-7` silently gets *strong* advice and no warning. A hand-maintained model map is guaranteed stale within weeks. Prefer family-prefix heuristics (`opus`/`gpt-4`/`gemini-*-pro` → frontier) or treat id→tier as best-effort and always emit the note when a bare id resolved via the map, not a keyword.
4. **Only hill-climb is wired.** `gepa` (the flagship sample-efficient algorithm) and `skillopt` render agnostic — set `target_model` under gepa and it does **nothing, silently**. This is flagged as a follow-up, but a half-connected feature with no warning is a footgun; at minimum, `check`/`estimate` should warn "target_model set but algorithm=gepa ignores it."
5. **Injected advice now contradicts the sibling advice in the same file.** `system-prompt/references/concepts.md` still says "soften MUSTs / explain the why / newer models over-comply" *and* now adds "flip it for weaker readers." For a `mid`/`weak` target the optimizer reads both and must reconcile them from the reader block — the exact "conflicting instructions resolve unpredictably" failure that skill itself warns against. Better to *neutralize* the base advice into the tier table than to append a contradiction.
6. Minor: spec/impl drift (`skills/_registry/target_profiles.yaml` + `registry_path` in the design doc were abandoned for the Python module — good call, but the doc wasn't reconciled). The two unused `Path` imports were already cleaned on main.

### Generality
General across **capabilities** (all four skills got tier notes) and **benchmarks/agents** (it's just a prompt slot). **Not** general across **algorithms** (hill-climb only) and the **model map is IBM/gpt-oss-flavored**. The tier *briefs* are reasonable but generic prose whose real effect on any given optimizer model is itself unmeasured.

### Reproducible before/after run (this review)

Same harness both sides. Optimizer `claude-opus-4-6`; runtime/consuming model `gpt-oss-120b` (served via IBM litellm proxy `openai/aws/gpt-oss-120b`, tool-calling verified). Subset: **6 train/val tasks + 6 held-out test tasks**, `num_trials=2`, `max_iterations=3`, hill-climb `all`, paired gate `k_se=0.2`. `before` = clean `origin/main`; `after` = `main` + #58 + #55, with `target_model: gpt-oss-120b` (tier `mid`).

| Metric | BEFORE (main, agnostic) | AFTER (#58+#55, target=mid) |
|---|---|---|
| seed baseline — val | 0.75 ± 0.19 | 0.67 ± 0.21 |
| best candidate — val | 1.00 ± 0.00 | 1.00 ± 0.00 |
| seed baseline — **held-out test** | 0.50 ± 0.22 | 0.33 ± 0.20 |
| **best candidate — held-out test** | **0.33 ± 0.20** | **0.42 ± 0.22** |
| optimizer budget-cap terminations | 2 | 3 |
| `target_profile` event | absent ✓ | `mid`, `suggested_num_trials=5` logged ✓ |

**Read this honestly — there is no detectable score signal at this scale, by design:**
- The *identical* seed capability scored **0.75 val / 0.50 test** in one run and **0.67 val / 0.33 test** in the other. Same code, same tasks, same seed — pure `gpt-oss` + user-sim stochasticity moved the *same* baseline by **0.08 (val) / 0.17 (test)**.
- The after−before best-test gap (**+0.08**) is *smaller than* one eval's stderr (**±0.20**) and smaller than that baseline drift. It is noise. N=3 iterations / 6 tasks / 2 trials cannot resolve a prompt-nudge effect — this is the framework's own "Δ > k·SE" gate telling you the experiment is underpowered.
- Both runs overfit the 6-task val to 1.00 and the honest test seal caught it — the framework's core discipline working as intended.

**What the run *does* show (the useful part) — the feature changes optimizer behavior in the intended direction:**
- BEFORE (agnostic): the opus optimizer added **one** concise `MUST verify reservation` rule.
- AFTER (`mid` reader block active): markedly more **explicit / literal / enumerated** edits — `"you MUST call get_user_details"`, `"Never trust user claims"`, literal status enums (`"available"/"on time"/"flying"/"landed"` vs `"delayed"/"cancelled"`), `"Both conditions must be met: (1)… AND (2)…"`, and in tools: literal enum docstrings (`regular/silver/gold`), example values (`'sara_doe_496'`), and `raise ValueError` code-enforcement guards. That is exactly the mid/weak-tier style the brief prescribes.

So #58 **mechanically does what it claims** — the reader block measurably shifts the edit *style* toward explicit/literal/code-enforced. Whether that style *raises the score* remains **unproven** — by this underpowered run and, more importantly, by the PR itself.

### Bottom line
Approve the *direction* and most of the implementation. Before merge I'd want: (a) wire or delete `suggested_num_trials`; (b) make `MODEL_MAP` misses emit the note (or go heuristic); (c) warn when `target_model` is set under an algorithm that ignores it; (d) at least one *powered* run (the real example's 50×10×10, not my 6×2×3) showing the tier advice moves held-out test — or explicitly frame the feature as "changes edit style, score-impact TBD," which is what the evidence currently supports.

PR #59, opened 2026-07-20
Add OpenShift deployment manifests and guide

AIMikav:
## What & why
Add self-contained OpenShift deployment manifests and guide for running cap-evolve with vLLM-served models on-cluster — no external API keys required.

Two vLLM instances serve distinct roles:
  - **Agent** (Qwen 2.5 7B, 1 GPU) — executes benchmark tasks
  - **Optimizer** (Qwen 2.5 14B, 2 GPUs) — analyzes failures, proposes skill improvements

  Includes: namespace, RBAC, secrets template, vLLM Deployments + Services + Routes,
  runner Job (one-shot) and Deployment, Dockerfile, and a step-by-step README.

  Tested end-to-end on OpenShift — runner pod deployed, `cap-evolve check` passed,
  cross-namespace vLLM connectivity verified.

## Checklist
  - [x] `python -m pytest core/tests -q` passes
  - [x] `python skills/_registry/build_manifest.py skills` is clean (no errors)
  - [x] Any new/changed skill's `scripts/check.py` is green
  - [x] Honesty invariants untouched (gate on val, test sealed) — or explained
  - [x] Docs updated (README/skill SKILL.md) if behavior changed
  
  > Manifests and docs only — no core code changes.


OsherElhadad:
### 🚀 Really nice contribution — thank you!

This is a genuinely valuable addition. A self-contained, on-cluster OpenShift path with both models served by vLLM is exactly what people running in air-gapped / no-external-key environments need, and it's the kind of thing that's easy to *want* and hard to *do well*. The manifests are clean and idiomatic (pinned vLLM image, templated secrets with a "don't commit" warning, `ttlSecondsAfterFinished`, health probes, sensible resource splits), the README is clear and genuinely followable, and it's fully additive with zero core changes. Great work. 🙌

> **Rebase:** not needed to merge — this PR only adds files under `openshift/**`, so there's no conflict with the adapter-template refactor that just landed on `main`. But that refactor changed the model-wiring convention, so the manifests' env vars need a small update to match it (below).

A few things to tighten before merge, mostly around how the runner is wired to the models.

**1. Agent model wiring — adopt the `model_config.py` convention that just landed on `main`.** The manifests set `TAU2_MODEL` / `TAU2_API_BASE` / `TAU2_API_KEY`, but these don't match the adapter that actually runs, so the runner won't reach vLLM as written. `main` now standardizes model wiring on a provider-agnostic `model_config.py` — the tau2 template (`templates/adapters/tau2_bench/adapter.py`, the recommended path) reads generic `MODEL` + `model_config.llm_kwargs()`. So set:

```
MODEL=openai/qwen2.5-7b-instruct
OPENAI_API_BASE=http://vllm-agent:80/v1
OPENAI_API_KEY=dummy
```

This reaches vLLM correctly *and* makes the guide work for every adapter template (jsonl, huggingface, swe-bench…), not just tau2. Two notes: `TAU2_MAX_CONCURRENCY` is fine to keep — that one *is* read by the adapter. And since `cap-evolve check` is offline, a wrong model var passes the check and only surfaces in a real eval. (FWIW the older `examples/tau2_airline` path reads `TAU2_AGENT_MODEL` / `TAU2_USER_MODEL` via `rits.py` — different again — but the template is the path the guide should point at.)

**2. The optimizer model isn't wired to the vLLM optimizer service.** cap-evolve's optimizer is a coding-agent CLI resolved from `skills/optimizers/registry.yaml` (model via `CAPEVOLVE_OPTIMIZER_MODEL`), not a raw OpenAI-compatible call — so `OPTIMIZER_MODEL` / `OPTIMIZER_API_BASE` / `OPTIMIZER_API_KEY` aren't consumed, and the default `optimizer_skill: claude-code` needs `ANTHROPIC_API_KEY`. As-is, the 14B optimizer pod would sit idle and the run still needs an external key, which undercuts the (great) "no external keys" goal. To truly run the optimizer on-cluster you'd point the `generic` optimizer (`CAPEVOLVE_OPTIMIZER_CMD`) at a CLI configured against the vLLM endpoint — happy to help wire an example. Alternatively, scope the guide to "vLLM serves the agent; optimizer = your CLI of choice" and say so explicitly.

**Minor / polish:**
- `vllm-serving.yaml` (agent) Route uses `targetPort: 8000`, but the Service exposes `80`/`http` → `targetPort 8000`; Routes point at the *service* port, so this likely won't resolve. `vllm-optimizer.yaml` correctly uses `targetPort: http` — suggest matching it.
- `cap-evolve-workspace` PVC is defined in both `cap-evolve-runner-deployment.yaml` and `cap-evolve-runner-job.yaml`. Fine to apply both, but `oc delete -f <one>` would drop the shared PVC. Consider a single `pvc.yaml`.
- `secrets.yaml` template is shipped but the README uses inline `oc create secret` — worth referencing the committed template so the two don't drift.
- `nodeSelector: gpu-pool-size: xlarge` (optimizer only) is cluster-specific — a README note ("match your GPU node label") and matching it on the agent would help portability.
- `serviceAccountName: harbor-task` reads like an internal name; something like `cap-evolve-runner` would be more self-documenting.
- The `anyuid` SCC grant is broad — a one-line least-privilege caveat in the README would be good hygiene for a public example.
- **Surface this on the GitHub Pages site.** The `site/` docs now have a page per major topic (Getting started, Architecture, Optimize your own, Adapter templates…) but no deployment page — a short `site/deployment.html` (or `openshift.html`) that surfaces this guide, added to the top nav, would make on-cluster deployment discoverable alongside the rest of the docs.

Again, really solid PR — none of this is structural, mostly making sure the model wiring actually closes the loop end-to-end and tracks the `model_config.py` convention now on `main`. Thanks for pushing this forward! 🙏

AIMikav:
@OsherElhadad I have addressed most of the review comments, I will create separate PR for the GitHub Pages site. Thanks!

Issue #146, opened 2026-07-26
Update Tau2 results accurate on end-to-end page


eranra:
### What happened?

In the https://skillberry-ai.github.io/cap-evolve/run-end-to-end.html there are reports on execution of Tau from multiple environments. The reorts are not accurate and the split between test and train. 

Once completed we need to update also https://dev.to/eranra/optimize-agentic-capabilities-with-agents-3n7j

### Reproduction

Observe the pages

OsherElhadad:
## ✅ Fixed — PR #148

The incorrect split label has been corrected across all four locations (README, docs/RESULTS.md, docs/COMPARISON.md, and the GitHub Pages site) in PR #148: https://github.com/skillberry-ai/cap-evolve/pull/148

**What was wrong:** the held-out run was labelled as `30/10/10` (30 train / 10 val / 10 test) everywhere.
**What is correct:** the run used `train=val=30, test=20` — consistent with the agent-mode run description.

---

@eranra — once the PR is merged, could you also update the dev.to article https://dev.to/eranra/optimize-agentic-capabilities-with-agents-3n7j with the correct split? The τ²-bench airline held-out run used **train=val=30 and test=20** (not 30/10/10), so any mention of that split or the 10-task val/test counts there should be updated to match.

eranra:
> @eranra — once the PR is merged, could you also update the dev.to article https://dev.to/eranra/optimize-agentic-capabilities-with-agents-3n7j with the correct split? The τ²-bench airline held-out run used train=val=30 and test=20 (not 30/10/10), so any mention of that split or the 10-task val/test counts there should be updated to match.

DONE :-)

PR #149, opened 2026-07-27
fix: correct Qwen 14B held-out split labels from 30/10/10 to train=val=30, test=20

OsherElhadad:
## Summary

Follows up on #148 / #146 — applies the same split correction to both Qwen 2.5 14B result sections.

The Qwen runs used the same split as the other held-out runs: **train=val=30, test=20** — not 30 train / 10 val / 10 test.

### `site/results.html`
Both the `#qwen-tools` (tools-only) and `#qwen-all` (all-capabilities) sections:
- Split bullet: `30 train / 10 val / 10 test` → `train=val=30, test=20`
- val row task count: `(10 tasks)` → `(30 tasks)`
- sealed test row task count: `(10 tasks, scored once)` → `(20 tasks, scored once)`

### `docs/RESULTS.md`
The Qwen sections were missing from the markdown docs entirely. Added both:
- **τ²-Bench airline — Qwen 2.5 14B (tools only, held-out)** — split train=val=30, test=20, val 30 tasks, test 20 tasks
- **τ²-Bench airline — Qwen 2.5 14B, all capabilities (held-out)** — same split

Reward values and gain figures are unchanged — only the split description and task counts are corrected.

Relates to #146

### DIFF
--- a/docs/RESULTS.md
+++ b/docs/RESULTS.md
@@ -124,6 +124,56 @@ run head-to-head, and improved the single-trial held-out test +37.5%. A genuine,
 here needs a stronger runner model or tool-level edits (which drove most of the lift in the
 no-holdout run).
 
+---
+
+## τ²-Bench airline — Qwen 2.5 14B (tools only, held-out)
+
+Same benchmark, capability, split, and algorithm as the held-out run above, with a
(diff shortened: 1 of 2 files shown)


AIMikav:
LGTM! approve the changes for the training split

Issue #1, opened 2026-06-16
Executing: `pip install ./core`


eranra:
### What happened?

$ pip install ./core  
Defaulting to user installation because normal site-packages is not writeable
Processing ./core
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  error: subprocess-exited-with-error
  
  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [35 lines of output]
      Traceback (most recent call last):
        File "/home/eranra/.local/lib/python3.10/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 389, in <module>
          main()
        File "/home/eranra/.local/lib/python3.10/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 373, in main
          json_out["return_val"] = hook(**hook_input["kwargs"])
        File "/home/eranra/.local/lib/python3.10/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 143, in get_requires_for_build_wheel
          return hook(config_settings)
        File "/tmp/pip-build-env-8pvzjk4c/overlay/lib/python3.10/site-packages/setuptools/build_meta.py", line 333, in get_requires_for_build_wheel
          return self._get_build_requires(config_settings, requirements=[])
        File "/tmp/pip-build-env-8pvzjk4c/overlay/lib/python3.10/site-packages/setuptools/build_meta.py", line 301, in _get_build_requires
          self.run_setup()
        File "/tmp/pip-build-env-8pvzjk4c/overlay/lib/python3.10/site-packages/setuptools/build_meta.py", line 317, in run_setup
          exec(code, locals())
        File "<string>", line 1, in <module>
        File "/tmp/pip-build-env-8pvzjk4c/overlay/lib/python3.10/site-packages/setuptools/__init__.py", line 117, in setup
          return distutils.core.setup(**attrs)  # type: ignore[return-value]
        File "/tmp/pip-build-env-8pvzjk4c/overlay/lib/python3.10/site-packages/setuptools/_distutils/core.py", line 160, in setup
          dist.parse_config_files()
        File "/tmp/pip-build-env-8pvzjk4c/overlay/lib/python3.10/site-packages/setuptools/dist.py", line 762, in parse_config_files
          pyprojecttoml.apply_configuration(self, filename, ignore_option_errors)
[...]

PR #175, opened 2026-07-28
Add Harbor integration for sandboxed agent optimization with cap-evolve

AIMikav:
## What & why
Add Harbor integration for running cap-evolve optimizations using Harbor (https://www.harborframework.com/) as the sandboxed evaluation backend. Harbor executes benchmark trials (SWE-bench, Terminal-Bench, etc.)
  in isolated Docker containers or OpenShift pods while cap-evolve drives the optimize → evaluate → gate cycle.

  The Harbor adapter (`templates/adapters/harbor/adapter.py`) bridges the two — its `run_batch()` calls harbor run as a subprocess each time cap-evolve needs an evaluation. Users run one command (`cap-evolve run`), and
  the adapter handles Harbor invocation, result parsing, and reward mapping internally.

 ## What's included:
  - `capevolve_harbor/ package` — `harbor run` subprocess wrapper with signal forwarding, job directory parser (reward.json → cap-evolve Score), task packager
  - `emplates/adapters/harbor/` — ready-to-use adapter template with `capevolve.yaml` (skill-package capability) and seed `SKILL.md`
  - `site/harbor.html` — integration guide: how cap-evolve uses Harbor, setup, env vars, OpenShift orchestrator image/RBAC/credentials, config reference
  - `site/harbor-openshift.html `— generic vLLM on OpenShift guide: deploy, GPU sizing, agent connection patterns, manifests
  - Harbor nav link added to all site pages, sitemap updated

  **Tested end-to-end on OpenShift** with SWE-bench + claude-code (Vertex AI) and Qwen 14B (on-cluster vLLM). Harbor created task pods, built environment images via OpenShift Binary Builds, ran claude-code inside the
  pods, and returned results to cap-evolve's gating loop.

## Checklist
  - [x] python -m pytest core/tests -q passes — no core changes
  - [x] python skills/_registry/build_manifest.py skills is clean — no skill changes
  - [x] Any new/changed skill's scripts/check.py is green — N/A (adapter template, not a skill)
  - [x] Honesty invariants untouched (gate on val, test sealed) — adapter uses standard run_batch()/score() contract; gating and sealing are in core, untouched
  - [x] Docs updated — site/harbor.html, site/harbor-openshift.html, nav links, sitemap

### DIFF
--- a/capevolve_harbor/results.py
+++ b/capevolve_harbor/results.py
@@ -57,8 +57,6 @@ def parse_job_dir(
     for trial_dir in sorted(job_dir.iterdir()):
         if not trial_dir.is_dir():
             continue
-        if trial_dir.name in ("config.json", "result.json"):
-            continue
 
         task_id = _extract_task_id(trial_dir)
         if task_ids and task_id not in task_ids:
@@ -111,33 +109,29 @@ def _parse_trial(trial_dir: Path, task_id: str) -> TrialResult:
 def _read_reward(verifier_dir: Path) -> tuple[float, dict]:
     """Read reward from verifier output.
 
-    Harbor reads reward.json first, falling back to reward.txt.
+    Tries reward.json first (``"reward"`` key). Falls back to reward.txt
+    only when reward.json is absent or unparseable — a legitimate 0.0 in
+    reward.json is honoured as-is.
     """
     reward_json: dict = {}
-    reward: float = 0.0
 
     rj_path = verifier_dir / "reward.json"
     if rj_path.exists():
         try:
             reward_json = json.loads(rj_path.read_text(encoding="utf-8"))
             if "reward" in reward_json:
-                reward = float(reward_json["reward"])
-            elif reward_json:
-                first_val = next(iter(reward_json.values()))
-                if isinstance(first_val, (int, float)):
-                    reward = float(first_val)
-        except (json.JSONDecodeError, ValueError, StopIteration):
+                return float(reward_json["reward"]), reward_json
+        except (json.JSONDecodeError, ValueError):
             pass
 
-    if reward == 0.0:
-        rt_path = verifier_dir / "reward.txt"
-        if rt_path.exists():
-            try:
-                reward = float(rt_path.read_text(encoding="utf-8").strip())
-            except ValueError:
-                pass
+    rt_path = verifier_dir / "reward.txt"
+    if rt_path.exists():
+        try:
+            return float(rt_path.read_text(encoding="utf-8").strip()), reward_json
+        except ValueError:
+            pass
 
-    return reward, reward_json
+    return 0.0, reward_json
 
 
 def _read_trajectory(agent_dir: Path) -> Any:
--- a/capevolve_harbor/results.py
+++ b/capevolve_harbor/results.py
@@ -1,0 +1,205 @@
+"""Parse Harbor job directories and map results to cap-evolve types."""
+
+from __future__ import annotations
+
+import json
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import Any
+
+
+@dataclass
+class TrialResult:
+    """One Harbor trial's outcome, parsed from the job directory."""
+
+    task_id: str
+    reward: float = 0.0
+    reward_json: dict = field(default_factory=dict)
+    trajectory: Any = None
+    cost_usd: float = 0.0
+    tokens: int = 0
+    verifier_stdout: str = ""
+    verifier_stderr: str = ""
+    error: str | None = None
+
+
+def parse_job_dir(
+    job_dir: Path,
+    task_ids: list[str] | None = None,
+) -> dict[str, list[TrialResult]]:
+    """Walk a Harbor job directory and extract per-trial results.
+
+    Harbor job directory structure::
+
+        <job-dir>/
+            config.json
+            result.json
+            <trial-name>/
+                config.json
+                result.json
+                agent/
+                    trajectory.json
+                    recording.cast
+                verifier/
+                    reward.txt
+                    reward.json  (optional)
+                    test-stdout.txt
+                    test-stderr.txt
+
+    Returns ``{task_id: [TrialResult, ...]}``.
+    """
+    results: dict[str, list[TrialResult]] = {}
+    job_dir = Path(job_dir)
+
+    if not job_dir.is_dir():
+        return results
+
+    for trial_dir in sorted(job_dir.iterdir()):
+        if not trial_dir.is_dir():
+            continue
+        if trial_dir.name in ("config.json", "result.json"):
+            continue
+
+        task_id = _extract_task_id(trial_dir)
+        if task_ids and task_id not in task_ids:
+            continue
+
+        tr = _parse_trial(trial_dir, task_id)
+        results.setdefault(task_id, []).append(tr)
+
+    return results
+
+
+def _extract_task_id(trial_dir: Path) -> str:
+    """Extract the task ID from a trial directory.
+
+    Tries trial config.json first, falls back to the directory name.
+    """
+    config_path = trial_dir / "config.json"
+    if config_path.exists():
+        try:
+            cfg = json.loads(config_path.read_text(encoding="utf-8"))
+            task_name = cfg.get("task", {}).get("name") or cfg.get("task_name")
+            if task_name:
+                return str(task_name)
+        except (json.JSONDecodeError, KeyError):
+            pass
+    return trial_dir.name
+
+
+def _parse_trial(trial_dir: Path, task_id: str) -> TrialResult:
+    """Parse a single trial directory into a TrialResult."""
+    tr = TrialResult(task_id=task_id)
+
+    verifier_dir = trial_dir / "verifier"
+    if verifier_dir.is_dir():
+        tr.reward, tr.reward_json = _read_reward(verifier_dir)
+        tr.verifier_stdout = _read_text(verifier_dir / "test-stdout.txt")
+        tr.verifier_stderr = _read_text(verifier_dir / "test-stderr.txt")
+
+    agent_dir = trial_dir / "agent"
+    if agent_dir.is_dir():
+        tr.trajectory = _read_trajectory(agent_dir)
+
+    result_path = trial_dir / "result.json"
+    if result_path.exists():
+        tr.cost_usd, tr.tokens = _extract_cost_tokens(result_path)
+
+    return tr
+
+
+def _read_reward(verifier_dir: Path) -> tuple[float, dict]:
+    """Read reward from verifier output.
+
+    Harbor reads reward.json first, falling back to reward.txt.
+    """
+    reward_json: dict = {}
+    reward: float = 0.0
+
+    rj_path = verifier_dir / "reward.json"
+    if rj_path.exists():
+        try:
+            reward_json = json.loads(rj_path.read_text(encoding="utf-8"))
+            if "reward" in reward_json:
+                reward = float(reward_json["reward"])
+            elif reward_json:
+                first_val = next(iter(reward_json.values()))
+                if isinstance(first_val, (int, float)):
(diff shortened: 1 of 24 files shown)


templates/adapters/harbor/adapter.py
```diff
@@ -0,0 +1,283 @@
...
+        ae["ANTHROPIC_MODEL"] = HARBOR_MODEL
+        ae["ANTHROPIC_DEFAULT_SONNET_MODEL"] = HARBOR_MODEL
+        ae["ANTHROPIC_DEFAULT_HAIKU_MODEL"] = HARBOR_MODEL
+        ae["ANTHROPIC_DEFAULT_OPUS_MODEL"] = HARBOR_MODEL
+
+    # Inject the candidate skill content so the agent can use it
+    skill_md = candidate_dir / "SKILL.md"
+    prompt_md = candidate_dir / "prompt.md"
+    if skill_md.exists():
+        ae["CAPEVOLVE_SKILL_CONTENT"] = skill_md.read_text(encoding="utf-8")[:8000]
```
OsherElhadad: 🔴 **Candidate skill never reaches the agent.** `CAPEVOLVE_SKILL_CONTENT` / `CAPEVOLVE_PROMPT_CONTENT` are passed as `--ae` env vars but nothing reads them (grepped the whole repo + `harbor.html` + Harbor's claude-code agent — zero consumers). For a registry dataset the `/capability/` dir is never injected either (`package_dataset` isn't called). So every candidate looks identical to the agent and the val gate measures run-to-run noise, not the edit. This needs a real injection path before the loop can optimize anything.

Also: `[:8000]` silently truncates the artifact being optimized — edits past 8 KB never reach the agent but are still scored.

capevolve_harbor/run.py
```diff
@@ -0,0 +1,160 @@
...
+
+    if jobs_dir and jobs_dir.is_dir():
+        job_dirs = sorted(jobs_dir.iterdir())
+        if job_dirs:
+            result.job_dir = job_dirs[-1]
+
+    return result
+
+
+def _forward_signals(proc: subprocess.Popen) -> None:
```
OsherElhadad: 🔴 **Signal handlers are never restored.** `original_handlers` is stashed locally and dropped on return. After the first `harbor_run`, SIGINT/SIGTERM stay bound to this closure over a dead `proc`, so a later Ctrl-C during the optimization loop is swallowed (send_signal to a finished process) instead of raising `KeyboardInterrupt`. Restore the originals in a `finally` around `proc.communicate`, e.g. store them and re-`signal.signal(sig, original)` after the run.

capevolve_harbor/results.py
```diff
@@ -0,0 +1,205 @@
...
+    reward: float = 0.0
+
+    rj_path = verifier_dir / "reward.json"
+    if rj_path.exists():
+        try:
+            reward_json = json.loads(rj_path.read_text(encoding="utf-8"))
+            if "reward" in reward_json:
+                reward = float(reward_json["reward"])
+            elif reward_json:
+                first_val = next(iter(reward_json.values()))
```
OsherElhadad: 🟡 Fragile: with no `"reward"` key this makes the *arbitrary first* JSON value the reward — a verifier that emits `{"cost":..., "passed":...}` could hand back `cost` as the reward. Prefer an explicit key list, or require `reward`.

capevolve_harbor/results.py
```diff
@@ -0,0 +1,205 @@
...
+                if isinstance(first_val, (int, float)):
+                    reward = float(first_val)
+        except (json.JSONDecodeError, ValueError, StopIteration):
+            pass
+
+    if reward == 0.0:
```
OsherElhadad: 🟡 `if reward == 0.0:` means a *legitimate* 0.0 from `reward.json` (task genuinely failed) triggers a `reward.txt` re-read that can disagree. Gate on "reward.json missing/unparseable", not "reward is 0".

OsherElhadad (commented):
## Review — Harbor integration (PR #175)

Solid, well-documented integration: clean module split (`run` / `tasks` / `results`), thorough docstrings, honesty invariants correctly left untouched (gate/seal stay in core; the adapter only implements the standard contract). Docs are extensive. Most of my concern is a single **correctness gap that likely makes the optimization a no-op on the headline use case**, plus one process-level bug.

### 🔴 High

**1. Candidate edits never reach the sandboxed agent (registry path).** The only mechanism the adapter uses to hand the candidate skill to the agent is the `--ae CAPEVOLVE_SKILL_CONTENT=...` / `CAPEVOLVE_PROMPT_CONTENT=...` env vars in `_build_agent_env` (adapter.py:92). Nothing in this repo, in `harbor.html`, in the seed `SKILL.md`, or in Harbor's `claude-code` agent reads those env vars — I grepped the whole tree, zero consumers. And for a registry dataset (the documented SWE-bench flow) `package_dataset`/`package_task` are never called, so the `/capability/` directory is never injected either. Net effect: every candidate looks identical to the agent, so the val gate is measuring pure run-to-run LLM noise rather than the effect of the edit. This is the thing to nail down before merge — how does an edited `SKILL.md` actually change agent behavior inside the pod?

**2. Signal handlers are installed but never restored** (`run.py:146` `_forward_signals`). Each `harbor_run` calls `signal.signal(SIGINT/SIGTERM, _handler)` and stashes the originals in a local `original_handlers` that is discarded on return — nothing restores them. After the first Harbor run, the process's SIGINT/SIGTERM are permanently bound to a closure over a now-dead `proc`; a later Ctrl-C during the optimization loop just does `proc.send_signal` on a finished process (swallowed by the `ProcessLookupError` guard) instead of raising `KeyboardInterrupt`. cap-evolve calls this many times per run, so Ctrl-C is effectively dead for the rest of the process. Restore the originals in a `finally` (or use a context manager) around `proc.communicate`.

### 🟡 Medium

**3. Reward parsing is fragile** (`results.py`). Two issues in `_read_reward`: (a) with no `"reward"` key it takes `next(iter(reward_json.values()))` (results.py:126) — the *arbitrary first* JSON value becomes the reward, which can silently pick `cost`/`duration`/some other metric; (b) `if reward == 0.0:` then falls back to `reward.txt` (results.py:132), so a *legitimate* 0.0 from `reward.json` triggers a `reward.txt` re-read that can disagree. The fallback should key on "reward.json absent/unparseable", not "reward happens to be 0". Reward here feeds the gate directly, so a mis-read flips accept/reject.

**4. Per-trial `seed` is accepted but never forwarded** (`adapter.py:148` → `harbor_run` at 168; `harbor_run` has no seed param). The adapter docstring/`check.py` are explicit that a stochastic runner MUST forward `seed`. `run_trials` fires N full `harbor run`s that are non-reproducible and, if anyone sets the agent to temperature 0, degenerate to identical trials → SE=0 → "any Δ>0 wins". At minimum forward it as an agent env var; if Harbor can't seed, document that trials rely on sampler nondeterminism.

**5. The optimized artifact is silently truncated to 8000 chars** (`adapter.py:92`, `[:8000]`). Edits the optimizer makes past 8 KB never reach the agent but still get scored — invisible ceiling on optimization. Truncating the very thing under optimization should at least warn.

**6. `capevolve_harbor/tasks.py` (194 lines) is dead in the shipped template** — the adapter imports only `harbor_run, parse_job_dir` and points Harbor at `HARBOR_DATASET` directly for both registry and local paths, so `package_task`/`package_dataset` are never invoked. Either the adapter should package tasks (which would also fix #1 by injecting `/capability/`), or drop the module until it's wired.

### 🟢 Low / nits
- **TOML injection** (`tasks.py:145`): `description` only escapes `"`; a newline or backslash in a task description produces invalid `task.toml`. `name`/instruction unescaped too. Prefer `tomllib`-safe emission or escape `\n`/`\\`.
- **`network_mode = "public"` default** (`tasks.py:73`) — permissive default for sandboxed eval; worth calling out / defaulting to `none` where the benchmark allows.
- **`_last_jobs_dir` is a class attribute** (`adapter.py:106`) — clobbered if two evals ever run concurrently, which is exactly what the materialize/live split exists to enable. Fine for the serial template default; note the ceiling.
- **`import shlex as _shlex` mid-file** (`adapter.py:52`) — move to the top import block.
- **`results.py`**: the `if trial_dir.name in ("config.json","result.json")` check is dead after the preceding `if not trial_dir.is_dir(): continue`.
- **`pyproject.toml`**: `readme = "../README.md"` + `package-dir = {"capevolve_harbor" = "."}` will fail/warn on sdist build (readme outside the package root); fine for editable installs only.

### Tests
No tests ship with the integration. Given #1/#3, a couple of pure unit tests would have caught real bugs cheaply: `parse_job_dir` against a fixture job dir (reward.json/txt precedence, missing files) and `_read_reward` on a metrics-only JSON. Worth adding before this becomes a supported template others copy.

templates/adapters/harbor/adapter.py
```diff
@@ -0,0 +1,283 @@
...
+        prefix = HARBOR_TASK_PREFIX
+        if is_registry and not prefix:
+            prefix = HARBOR_DATASET.split("/")[0]
+
+        harbor_task_names = [
+            f"{prefix}/{t.id}" if prefix and not t.id.startswith(prefix) else t.id
+            for t in tasks
+        ] if is_registry else None
+
+        result = harbor_run(
```
OsherElhadad, line 224: 🟡 `seed` is accepted by `run_batch`/`run_trials` but never forwarded to `harbor_run` (which has no seed param). The adapter contract + `check.py` require stochastic runners to forward it; without it trials are non-reproducible and collapse to SE=0 at temperature 0. Forward it as an agent env var or document the reliance on sampler nondeterminism.

capevolve_harbor/tasks.py
```diff
@@ -0,0 +1,194 @@
...
+            test_script=test_script,
+        )
+
+    return output_dir
+
+
+def _write_task_toml(dest: Path, name: str, description: str, cfg: dict) -> None:
+    content = _TASK_TOML_TEMPLATE.format(
+        name=name,
+        description=description.replace('"', '\\"'),
```
OsherElhadad: 🟢 `description` only escapes `"` — a newline or backslash in a task description produces invalid TOML. `name`/instruction are unescaped too. Escape `\n`/`\\` or emit via a real TOML writer. (Also note: this whole module is currently unused by the shipped adapter, which imports only `harbor_run`/`parse_job_dir`.)

capevolve_harbor/run.py
AIMikav: `_forward_signals` replaced with `_install_signal_forwarding` + `_restore_signal_handlers` in a `finally` block.

templates/adapters/harbor/adapter.py
AIMikav: it was a bug with the registry datasets, now that is rewired to use `--extra-instruction-path`

capevolve_harbor/results.py
AIMikav: rewritten: `reward.txt` fallback only triggers when `reward.json `is absent or unparseable.

capevolve_harbor/results.py
AIMikav: A legitimate 0.0 in reward.json is honoured.

templates/adapters/harbor/adapter.py
AIMikav: `CAPEVOLVE_SEED` passed as an agent env var via --ae

capevolve_harbor/tasks.py
AIMikav: Fixed - added `_toml_escape()` that handles \\, ", and \n. Both name and description are escaped. `tasks.py `is also now wired into the adapter via `package_dataset()` (no longer dead code). Tests in `test_tasks.py` cover the escaping.

AIMikav:
all review comments are addressed, which caused because of the local registry run. Added test cases!

OsherElhadad (approved):

PR #184, opened 2026-07-29
docs(changelog): correct stale 0.46→0.80, add real [0.1.0] section, move features to Added

OsherElhadad:
Closes #101

`CHANGELOG.md` had three defects that made it unusable as the canonical "what shipped" record. All three are fixed; **docs-only, no code touched.**

## 1. The stale honesty defect: `0.46 → 0.80`

**Before** (`CHANGELOG.md:131-132`, under `### Fixed`):

```
- Examples: toy_calc, json_extract, date_tool, skills_bench, tau2_airline
  (real run: 0.46 → 0.80 on 50 tasks).
```

**After** (under `## [0.1.0]` → `### Added`):

```
- Examples: toy_calc, json_extract, date_tool, skills_bench, tau2_airline — the last a
  real 50-task × 10-trial run, val **0.536 → 0.712** (best candidate `cand_0007`,
  +0.176 / +32.8% relative; *fit metric*, `train == val == test`, so the test number
  0.694 pass@1 is not held out). Canonical figures and the committed artifact:
  [`docs/RESULTS.md`](docs/RESULTS.md) and
  [`examples/tau2_airline/run_full/final.json`](examples/tau2_airline/run_full/final.json).
```

### Where the corrected number's source of truth lives

Nothing was invented or rounded. The figure is triple-sourced:

| Source | Evidence |
|---|---|
| **`examples/tau2_airline/run_full/final.json`** (primary — the committed run artifact) | `best_id = cand_0007`, `test.reward = 0.6940000000000001`, `test.pass_k/1 = 0.694`, `test.pass_k/2 = 0.584`, `test.per_task` length 50 |
| **`docs/RESULTS.md:47-53`** (canonical results page, derived from that artifact) | Baseline **0.536**; best candidate `cand_0007` val **0.712** (+0.176 / +32.8% relative); sealed test (fit metric) **0.694** pass@1 |
| **`README.md:109`** (snapshot of RESULTS.md) | `0.536 → 0.712`, **+0.176 / +32.8%** |

Neither `0.46` nor `0.80` appears in any artifact, doc, or run output — they were an orphaned pre-`v0.1.0` claim. The corrected entry additionally states the **split discipline** (`train == val == test`, no holdout) which the old line omitted entirely, so the number can no longer be misread as a generalization claim. `docs/RESULTS.md` was *not* modified — it was already correct; the CHANGELOG was brought into line with it.

## 2. No released-version section → real `## [0.1.0] - 2026-07-27`

[...]

PR #187, opened 2026-07-29
Report pass^k/pass@k as N/A (not 0.0) when k > num_trials

### LINKED ISSUE #112: Report pass^k/pass@k as N/A (not 0.0) when k > num_trials
## Problem

Every default single-trial run reports **`pass^2 = 0.0`**, which reads as "0% reliable" when it actually just means "not enough trials to compute k=2."

- Default `num_trials: 1` (`templates/project/capevolve.yaml:53`).
- Default `ks=(1,2)` in `evaluate_candidate` (`harness.py`).
- `aggregate_scores` computes `pass_k` for every `k` in `ks` regardless of trial count (`loop.py:73-84`); `stats.pass_k` returns `0` when `k > n` (`stats.py:56-90`, correct by definition).
- So `pass_k["2"]` = pass^2 with k=2, n=1 → always `0.0`, surfaced in the dashboard/report as a reliability metric.

## Why it matters

The dashboard and report show `pass^2 = 0.0` on essentially every default run, which is misleading — it looks like the optimized capability is catastrophically unreliable when the metric is simply undefined at `n_trials=1`. Undermines the honest-metrics story.

## What to change

- In `aggregate_scores` (`loop.py:73`), **suppress** (omit) any `k > n_trials` from the emitted `pass_k` / `pass_at_k` dicts, rather than emitting `0.0`.
- Downstream (dashboard KPI strip, ANSI report) already treat a missing key gracefully or should render "N/A"; verify they don't assume the key exists.
- Optionally: when `num_trials == 1`, default `ks=(1,)` so k=2 isn't requested at all.

## How to test

- New test: `aggregate_scores` with `n_trials=1` and `ks=(1,2)` returns `pass_k` containing only `"1"`, not `"2"`.
- Run `toy_calc` (single trial) and confirm the report/dashboard no longer show `pass^2 = 0.0` (either absent or "N/A").
- A multi-trial run (e.g. `num_trials=10`) still reports `pass^2` normally.

- **Parallel-safe.** Small, self-contained core fix. Improves the observability story. Good first issue for someone comfortable with the stats path.

OsherElhadad:
Closes #112

## Problem

Every default single-trial run reported **`pass^2 = 0.0`**. That reads as "0% reliable" when the statistic is simply **undefined** at `n_trials=1`. An honesty bug: it understates a real result.

## Root cause

`core/cap_evolve/loop.py` → `aggregate_scores()` (line 64) — the *single* aggregation site every algorithm, `harness.evaluate_candidate`, `harness.baseline`, `harness.finalize` and `harness.split_result_from_rollouts` route through. It computed `pass_k`/`pass_at_k` for every requested `k` regardless of trial count.

`stats.pass_k` returning `0.0` for `k > n` is correct *by its own definition*, and `stats.pass_at_k` silently clamps `k → n`. The bug was the aggregation site emitting both as if they were measurements. Fixed there, so all callers inherit it — no per-caller guards.

```python
trials_per_task = [len(s.trial_rewards or [s.reward]) for s in scores]
max_k = min(trials_per_task) if trials_per_task else 0
for k in ks:
    if k > max_k:
        continue          # undefined ⇒ omit, never 0.0
```

Ragged splits use the **min** trial count: if any single task has fewer than `k` trials, `pass^k` for the split as a whole is not defined.

## N/A representation

**A missing key is the N/A.** No sentinel value, no `null` in the dict, nothing a consumer can coerce back to `0`.

| surface | representation |
|---|---|
| `final.json` / `baseline.json` / dashboard JSON | key **absent** from `pass_k` / `pass_at_k` (`{"1": 1.0}`, not `{"1": 1.0, "2": 0.0}`) |
| `report.md`, report JSON summary | `pass^1=1.000, pass^2=N/A` |
| React SPA KPI hint | `pass^1 100.0% · pass^2 N/A` |

## Consumer sweep

Grepped every reader of `pass_k` / `pass_at_k` / `test_pass_k` across `core`, `skills`, `dashboard`, `templates`, `examples`, `docs`:

- **`skills/phases/report/scripts/run.py`** — was `f"(pass^k={test.get('pass_k')})"`, dumping the raw dict (`pass^k={'1': 1.0, '2': 0.0}`). Now renders each k with `N/A` for absent ones. Also tolerates the legacy bare-scalar `pass_k` that `report/scripts/check.py` feeds it.
- **`dashboard/frontend/src/lib/types.ts` + `components/KpiStrip.tsx`** — pre-existing type lie: `test_pass_k?: number | null` while the backend has *always* sent a `{k: value}` dict, so `pct(summary.test_pass_k)` rendered **`NaN%`**. Typed as `Record<string, number> | null` and rendered per-k, with absent ks as `N/A`. No crash, no coercion to 0.
- **`core/cap_evolve/dashboard.py:548`** — `"test_pass_k": test.get("pass_k")` passes the dict through unchanged; correct as-is. The self-contained HTML template (`_HTML_TEMPLATE`) does not read `test_pass_k` at all — verified by grep.
- **`render_ansi`** (terminal report) — its KPI strip never included pass^k; nothing to fix.
- **`docs/HONEST_EVAL.md`** — documents the convention alongside the existing variance rules.

Deliberately **not** done: changing the default `ks=(1,2)` to `(1,)` when `num_trials == 1` (issue's "optionally"). Redundant once the aggregation site suppresses undefined ks, and it would fork the decision into every caller that passes `ks`.

No new runtime deps.

### Regression test fails before, passes after

Test: `core/tests/test_pass_k_undefined.py`. Proof by stashing only `core/cap_evolve/loop.py` back to `origin/main`:

```
=== SOURCE REVERTED (loop.py at origin/main) ===
E         {'2': 0.0}
E       AssertionError: assert {'1', '2'} == {'1'}
E         Extra items in the left set:
E         '2'
FAILED core/tests/test_pass_k_undefined.py::test_single_trial_omits_k2 - Asse...
FAILED core/tests/test_pass_k_undefined.py::test_ragged_trials_use_the_min - ...
2 failed, 1 passed in 0.03s
```

```
=== FIX RESTORED ===
...                                                                      [100%]
3 passed in 0.01s
```

### Full suite — 182 passed (179 baseline + 3 new), 0 failed

```
$ PYTHONPATH=/tmp/wt-112/core python -m pytest core/tests -q
182 passed in 66.26s (0:01:06)
```

### compileall

```
COMPILEALL CLEAN (rc=0)
```

### Frontend build

```
> tsc -b && vite build
✓ 2809 modules transformed.
dist/assets/index-b_9cC4ZS.css   16.36 kB │ gzip:   4.51 kB
dist/assets/index-5zepM1pz.js   830.51 kB │ gzip: 249.99 kB
✓ built in 875ms
```

(`dashboard/frontend/dist/` is intentionally committed — `dashboard/frontend/.gitignore` says "dist is intentionally COMMITTED: the backend serves this prebuilt SPA at runtime" — so the rebuilt bundle is in this commit.)

### Real end-to-end run — `examples/toy_calc`, `mock` optimizer, zero API cost

`templates/project/capevolve.yaml` default `num_trials: 1`, default `ks=(1,2)` ⇒ `k > num_trials`.

**BEFORE** (`origin/main`):

```json
{
  "test_reward": 1.0,
  "test_pass_k": {
    "1": 1.0,
    "2": 0.0
  }
}
```
```
- **Held-out test (optimized skills): 1.0**  (pass^k={'1': 1.0, '2': 0.0})
```

**AFTER**:

```json
{
  "test_reward": 1.0,
  "test_pass_k": {
    "1": 1.0
  }
}
```
```
- **Held-out test (optimized skills): 1.0**  (pass^1=1.000, pass^2=N/A)
```

Test reward unchanged at 1.0 — only the undefined statistic stopped lying.


OsherElhadad:
## 🔬 Evidence

Every command and its full literal output. Worktree `/tmp/wt-112` at `origin/main` (`6fca097`), Python `/tmp/ce-venv/bin/python`.

---

### 1. Baseline suite before any edit

<details><summary><code>git worktree add</code></summary>

```
$ git worktree add /tmp/wt-112 -b fix/issue-112-passk-na origin/main
Preparing worktree (new branch 'fix/issue-112-passk-na')
branch 'fix/issue-112-passk-na' set up to track 'origin/main'.
HEAD is now at 6fca097 evograph weakness-graph view (mounted in the cap-evolve dashboard) (#53)
```
</details>

---

### 2. The bug, reproduced on a real run (BEFORE)

```
$ REPO=/tmp/wt-112
$ export CAPEVOLVE_CORE=$REPO/core PYTHONPATH=$REPO/core CAPEVOLVE_SKILLS_DIR=$REPO/skills \
         CAPEVOLVE_TOY_DATA=$REPO/examples/toy_calc \
         CAPEVOLVE_MOCK_SCRIPT=$REPO/examples/toy_calc/mock_script.json
$ D=/tmp/toy-before; mkdir -p $D/.capevolve/project/adapters
$ cp $REPO/examples/toy_calc/adapter.py $D/.capevolve/project/adapters/
$ cp -R $REPO/examples/toy_calc/capability $D/seed_capability
$ cp $REPO/templates/project/capevolve.yaml $D/.capevolve/project/capevolve.yaml
$ python -m cap_evolve.cli run --spec $D/.capevolve/project/capevolve.yaml \
      --project $D/.capevolve/project --run-ts demo
{
  "best_id": "cand_0001",
  "baseline_val": 0.0,
  "test_reward": 1.0,
  "test_baseline_reward": 0.0,
  "test_delta": 1.0,
  "test_pass_k": {
    "1": 1.0,
    "2": 0.0
  },
  "iterations": 3,
  "dashboard_server": "skipped"
}
```

`report.md`:

```
$ cat /tmp/toy-before/.capevolve/run_demo/report.md
# cap-evolve run report — run_demo

- Best candidate: `cand_0001`
- Baseline val: 0.0
- **Held-out test (optimized skills): 1.0**  (pass^k={'1': 1.0, '2': 0.0})
- Held-out test (baseline `seed` skills): 0.0
- **Test improvement (optimized − baseline): +1.0**
- Iterations: 3

Test was scored exactly once on the sealed split, for BOTH the baseline (`seed`) and the optimized skills — the improvement above is on held-out tasks the optimizer never saw.
```

`final.json` — both dicts carry the fake `"2"`:

```
$ python -c "import json;d=json.load(open('/tmp/toy-before/.capevolve/run_demo/final.json'));print(json.dumps({'test_pass_k':d['test']['pass_k'],'test_pass_at_k':d['test']['pass_at_k']},indent=2))"
{
  "test_pass_k": {
    "1": 1.0,
    "2": 0.0
  },
  "test_pass_at_k": {
    "1": 1.0,
    "2": 1.0
  }
}
```

Note `pass_at_k["2"] = 1.0` too — `stats.pass_at_k` clamps `k → n`, so it reported a k=2 capability from one trial. Also suppressed.

---

### 3. Second, latent bug found during the frontend sweep

`dashboard/frontend/src/lib/types.ts:91` declared `test_pass_k?: number | null`, but `core/cap_evolve/dashboard.py:548` has always sent the **dict**. So `pct(summary.test_pass_k)`:

```
$ node -e "const v={'1':1.0,'2':0.0}; console.log('frontend pct(v) ->', (v*100).toFixed(1)+'%')"
frontend pct(v) -> NaN%
```

The SPA KPI hint read `pass^k NaN%`. Fixed in the same sweep (typed as `Record<string, number>`, rendered per-k).

---

### 4. Fail-before / pass-after proof

Only `core/cap_evolve/loop.py` is stashed back to `origin/main`; the new test file stays.

```
$ git stash push -- core/cap_evolve/loop.py
$ PYTHONPATH=/tmp/wt-112/core python -m pytest core/tests/test_pass_k_undefined.py -q
=== SOURCE REVERTED (loop.py at origin/main) ===
E         {'2': 0.0}
E         Use -v to get more diff

core/tests/test_pass_k_undefined.py:14: AssertionError
________________________ test_ragged_trials_use_the_min ________________________

    def test_ragged_trials_use_the_min():
        # One task with a single trial makes k=2 undefined for the split as a whole.
        scores = [
            Score(task_id="t1", reward=1.0, trial_rewards=[1.0, 1.0]),
            Score(task_id="t2", reward=1.0, trial_rewards=[1.0]),
        ]
        r = aggregate_scores("val", scores, ks=(1, 2))
>       assert set(r.pass_k) == {"1"}
E       AssertionError: assert {'1', '2'} == {'1'}
E
E         Extra items in the left set:
E         '2'
E         Use -v to get more diff

core/tests/test_pass_k_undefined.py:38: AssertionError
FAILED core/tests/test_pass_k_undefined.py::test_single_trial_omits_k2 - Asse...
FAILED core/tests/test_pass_k_undefined.py::test_ragged_trials_use_the_min - ...
2 failed, 1 passed in 0.03s
```

```
$ git stash pop
$ PYTHONPATH=/tmp/wt-112/core python -m pytest core/tests/test_pass_k_undefined.py -q
=== FIX RESTORED ===
...                                                                      [100%]
3 passed in 0.01s
```

`test_multi_trial_still_reports_k2` passes in *both* states — it is the guard that the fix does not over-suppress.

---

### 5. Full suite: 182 passed, 0 failed (179 baseline + 3 new)

```
$ PYTHONPATH=/tmp/wt-112/core python -m pytest core/tests -q
182 passed in 66.26s (0:01:06)
```

---

### 6. compileall

```
COMPILEALL CLEAN (rc=0)
```

---

### 7. Frontend build

```
  npm audit fix --force

Run `npm audit` for details.

> frontend@0.0.0 build
> tsc -b && vite build

transforming...✓ 2809 modules transformed.
rendering chunks...
computing gzip size...
dist/assets/index-b_9cC4ZS.css   16.36 kB │ gzip:   4.51 kB
dist/assets/index-5zepM1pz.js   830.51 kB │ gzip: 249.99 kB

✓ built in 875ms
[plugin builtin:vite-reporter]
(!) Some chunks are larger than 500 kB after minification. Consider:
- Using dynamic import() to code-split the application
...
```

`tsc -b` passing is itself the proof the `Record<string, number>` retype has no other unhandled consumer — TypeScript would have errored at any remaining site treating `test_pass_k` as a number. The chunk-size note is pre-existing.

---

### 8. Real end-to-end run (AFTER) — same command, same fixture, zero API cost

```
$ D=/tmp/toy-after   # fresh dir, identical setup to §2
$ python -m cap_evolve.cli run --spec $D/.capevolve/project/capevolve.yaml \
      --project $D/.capevolve/project --run-ts demo
{
  "best_id": "cand_0001",
  "baseline_val": 0.0,
  "test_reward": 1.0,
  "test_baseline_reward": 0.0,
  "test_delta": 1.0,
  "test_pass_k": {
    "1": 1.0
  },
  "iterations": 3,
  "dashboard_server": "skipped"
}

=== report.md AFTER ===
# cap-evolve run report — run_demo

- Best candidate: `cand_0001`
- Baseline val: 0.0
- **Held-out test (optimized skills): 1.0**  (pass^1=1.000, pass^2=N/A)
- Held-out test (baseline `seed` skills): 0.0
- **Test improvement (optimized − baseline): +1.0**
- Iterations: 3

Test was scored exactly once on the sealed split, for BOTH the baseline (`seed`) and the optimized skills — the improvement above is on held-out tasks the optimizer never saw.
```

### Before → after diff on the one thing that changed

| | before | after |
|---|---|---|
| `test_reward` | `1.0` | `1.0` (unchanged) |
| `test_pass_k` JSON | `{"1": 1.0, "2": 0.0}` | `{"1": 1.0}` |
| `test_pass_at_k` JSON | `{"1": 1.0, "2": 1.0}` | `{"1": 1.0}` |
| `report.md` | `pass^k={'1': 1.0, '2': 0.0}` | `pass^1=1.000, pass^2=N/A` |
| SPA KPI hint | `pass^k NaN%` | `pass^1 100.0% · pass^2 N/A` |

OsherElhadad:
## 🔍 Review — PR #187

**Verdict: APPROVE WITH NITS** — plus one *mandatory merge-time action* (rebuild `dist/` after this lands alongside #192; see Merge-order note).

Every claim in the evidence comment reproduced byte-for-byte, including the committed `dist/` bundle. The guard is in the right place, the math is right at every boundary I probed, and no consumer anywhere in the repo breaks on key-absence. Findings below are latent-defect and test-coverage grade, not correctness-of-this-fix grade.

---

### Blocking

None against this PR's own diff.

---

### Non-blocking

**1. `skills/phases/report/scripts/run.py:73-76` — the k range is hardcoded `(1, 2)`, so a non-default `ks` silently drops measured values and fabricates unmeasured ones.**
`pk_str` iterates `for k in (1, 2)` regardless of what `ks` was actually requested. With `ks=(1,3)` (nothing forbids it — `ks` is a plain kwarg on `aggregate_scores`/`evaluate`/`finalize`, and `gepa.py:184` already passes a non-default `ks=(1,)`):
```
report.md renders: pass^1=0.900, pass^2=N/A   <- pass^3=0.400 SILENTLY DROPPED, pass^2 N/A INVENTED
```
Consequence: a measured `pass^3` never reaches `report.md`, and the report asserts a k=2 N/A that was never requested — a new honesty defect of exactly the class #112 is about. Today's defaults hide it because `ks=(1,2)` everywhere the test split is scored.
Fix: iterate the union of the keys actually present with the `ks` actually requested, or just `for k in sorted(int(x) for x in pk)` and drop the fabricated-N/A entirely:
```python
pk_str = ", ".join(f"pass^{k}={pk[k]:.3f}" for k in sorted(pk, key=int))
```

**2. `dashboard/frontend/src/components/KpiStrip.tsx:22` — same defect, same fix.**
`for (let k = 1; k <= Math.max(maxK, 2); k++)` hardcodes a floor of 2, so the KPI hint fabricates a `pass^2 N/A` chip on a run where k=2 was never requested:
```
KPI hint: pass^1 90.0% · pass^2 N/A · pass^3 40.0%   <- pass^2 N/A is FABRICATED
```
Fix: drop `Math.max(maxK, 2)` → iterate the sorted present keys and render only genuine gaps (`1..maxK`), or render only what's there.

**3. `core/cap_evolve/stats.py:65-66` — the root cause is still armed and still publicly exported.**
`pass_k` returns a plausible-looking `0.0` for the undefined input `k > n`, and `core/cap_evolve/__init__.py:23,55` re-exports it as public API. The chokepoint argument does hold for internal use — I confirmed the ONLY non-test caller in the repo is `loop.py:85`, and all four production call sites (`harness.py:309,350`, `gepa.py:184`, `harness.py:1948`) route through `aggregate_scores` — so the guard is not leaky *today*. But `test_core.py:102` enshrines `pass_k([1.0], 2) == 0.0` as intended behavior, so a third-party adapter or a future skill that calls `cap_evolve.pass_k` directly re-derives #112 with the test suite's blessing.
Fix: make the undefined input unrepresentable — `return None` for `k > n` (loop.py already `continue`s on those k, so nothing downstream sees the `None`), or `raise ValueError`. Update `test_core.py:102` to assert the new contract.

**4. `core/tests/test_pass_k_undefined.py:20-28` — the over-suppression guard does not cover the `k == n` boundary, which is exactly where an off-by-one would live.**
`test_multi_trial_still_reports_k2` uses `n=3, k=2`, so `k == max_k` is never exercised. I mutated the guard to over-suppress the *defined* `k == n` case and the whole suite stayed green:
```
=== MUTANT 4: `if k > max_k or (max_k >= 2 and k == max_k):` ===
core/tests/test_pass_k_undefined.py ... 3 passed
core/tests -q ......................... 181 passed (only the pre-existing port flake failed)
```
So the claim that `test_multi_trial_still_reports_k2` "is the guard that the fix does not over-suppress" is only half true — it catches a wholesale off-by-one (`k >= max_k`, which I also mutated and confirmed fails 2 tests) but not the boundary-only one. The live behavior is correct (`n=2, k=2` → `{'1': 0.5, '2': 0.0}`), it's just unguarded.
Fix: one more `Score(task_id="t1", reward=1.0, trial_rewards=[1.0, 1.0])` with `ks=(1,2)` asserting `"2" in r.pass_k`.

**5. `.github/workflows/ci.yml:59-63` — CI builds the SPA but never checks the committed bundle against the build output.**
`npm ci && npm run build` runs, then nothing compares. A future PR that edits `dashboard/frontend/src/` without rebuilding ships a stale `dist/` and CI stays green — and since the backend serves the prebuilt SPA at runtime (per `dashboard/frontend/.gitignore:11-13`), that stale bundle is what users actually execute. Not this PR's fault (its bundle is correct), but it is the systemic hazard that makes committed-`dist/` PRs risky, and #187 + #192 both hit it simultaneously.
Fix: append `git diff --exit-code -- dashboard/frontend/dist` to the build job.

**6. `docs/HONEST_EVAL.md:31` overclaims "every surface renders a missing `k` as `N/A` / `—`".**
Two surfaces render nothing at all rather than N/A: the self-contained `dashboard.html` template never reads `test_pass_k` (I confirmed — `core/cap_evolve/dashboard.py` mentions it only at the docstring `:35` and the summary emit `:548`; the rendered KPI at `:910` has no pass^k hint), and `skills/phases/evaluate/scripts/run.py:36` dumps `result.to_dict()` as raw JSON. Absence is fine for both; the sentence just isn't accurate.
Fix: "and human surfaces render a missing `k` as `N/A` / `—`".

---

### Nits

**7. `core/cap_evolve/loop.py:77` — `max_k` is computed with `min()`.** The name reads as "the largest k" (which it is, semantically: the largest *usable* k) but the code says `min`, which costs a double-take at 3am. `max_usable_k` or `min_trials` would read straight.

**8. `core/cap_evolve/loop.py:82-83` — the guard is upper-bound only, so `k <= 0` still leaks a fake `0.0`.** `aggregate_scores(..., ks=(0,1))` → `{'0': 0.0, '1': 1.0}`. Nobody passes 0 today. `if k < 1 or k > max_k: continue` closes it for free.

**9. No CHANGELOG entry.** #184 is concurrently fixing CHANGELOG staleness; an honesty-semantics change to a reported metric is worth a line under Fixed.

**10. `core/cap_evolve/loop.py:85-86` — the inner `if (s.trial_rewards or [s.reward])` filter is dead.** `trial_rewards or [s.reward]` is never falsy, so the guard can never exclude anything. Pre-existing, not introduced here; noting it because it reads as if it handles a zero-trial task and it does not (a task with `trial_rewards=[]` is counted as 1 trial — verified: `aggregate_scores` on such a score yields `{'1': 0.0}`, and mixed with a 2-trial task it suppresses `pass^2`, which is the conservative-correct outcome but arrives by accident rather than by that guard).

---

### Consumer audit

Grepped `pass_k`, `pass_at_k`, `pass^`, `passk`, `pass@` across `core/`, `skills/`, `dashboard/` (backend + frontend + committed bundle), `site/`, `docs/`, `examples/`, `templates/`, and every JSON fixture. There is **no** `.get(..., 0)`, no `|| 0`, no `["2"]` subscript, and no arithmetic-on-possibly-absent anywhere outside the PR's own test. Full list of real read sites:

| Consumer `file:line` | How it reads the value | Survives key-absence |
|---|---|---|
| `core/cap_evolve/loop.py:41-42` (`to_dict`) | passes the dict through | ✅ |
| `core/cap_evolve/loop.py:55-56` (`from_dict`) | `dict(d.get("pass_k") or {})` | ✅ (see caveat below) |
| `core/cap_evolve/dashboard.py:548` | `test.get("pass_k")` → into JSON summary, no arithmetic | ✅ |
| `core/cap_evolve/dashboard.py:910` (embedded `dashboard.html` KPI) | never reads `test_pass_k` at all | ✅ (nothing to break — verified, the author's claim holds) |
| `skills/phases/report/scripts/run.py:63` | `test.get("pass_k")` → JSON summary passthrough | ✅ |
| `skills/phases/report/scripts/run.py:70-76` | `pk.get(str(k)) is None` → `"N/A"` | ✅ (but see Non-blocking 1) |
| `skills/phases/evaluate/scripts/run.py:36` | `json.dumps(result.to_dict())` | ✅ |
| `dashboard/frontend/src/components/KpiStrip.tsx:16-27` | `pk[String(k)]`, explicit `v == null → 'N/A'` | ✅ (but see Non-blocking 2) |
| `dashboard/frontend/src/lib/types.ts:92` | type only | ✅ |
| `dashboard/backend/**` | no reference at all (grep empty) | ✅ n/a |
| `site/**`, `docs/**` | prose only; the only numeric mentions (`site/run-end-to-end.html:80`, `docs/REPRODUCE_tau2.md:23`) are the 10-trial τ²-bench run, legitimately defined | ✅ |
| `core/tests/test_dashboard.py:62` | fixture `{"pass_k": {"1": 0.8}}` — already dict-shaped, already k=1-only | ✅ |
| `examples/*/run_full/*.json` | all fixtures are 3- or 10-trial runs; **zero** contain a `pass_k` entry equal to `0.0` | ✅ no stale fixture needs regenerating |

**Answer to risk #1: NO consumer breaks on key-absence.** Key-omission is a safe N/A representation here, precisely because every read site already goes through `.get()`/`== null`. The hazard you flagged (`d["pass_2"]` KeyError / `d.get("pass_2", 0)` zero-fallback) does not exist anywhere in the tree today — the surface is small enough (2 Python readers, 1 TS reader) that the pattern is defensible. It does rely on reviewer discipline forever, which is why Non-blocking 3 matters.

One residual, **pre-existing on `main`**: `loop.py:55` does `dict(d.get("pass_k") or {})`, which raises `TypeError: 'float' object is not iterable` on a legacy run dir that stored a bare scalar `pass_k`. `skills/phases/report/scripts/check.py:51` writes exactly that shape (`{"test": {"reward": 0.8, "pass_k": 0.7}}`). The PR *does* defend against the scalar shape in `report/scripts/run.py:71-72` — so the author clearly knew the legacy shape exists — but left `from_dict` unhardened. Not a regression; `check.py`'s path reads raw JSON, not `from_dict`, so it passes (verified: report `check.py` returns clean). Worth a one-line `isinstance` in `from_dict` while you're here.

---

### Is the fix at the right layer?

**`loop.py` is a legitimate chokepoint, so the fix is not misplaced — but it is one layer too high to be permanent.**

Verified: `aggregate_scores` is the *only* non-test consumer of `stats.pass_k`/`stats.pass_at_k` in the repo, and all four production entry points funnel through it (`harness.evaluate:309`, `harness.split_result_from_rollouts:350`, `harness.finalize` → `:1948/:1954`, `gepa:184`). So there is no sibling caller left broken — the one-guard-in-the-shared-aggregator diff genuinely covers every path today.

And `loop.py` is arguably the *only* place the guard can live correctly, because it is the only layer that knows the **actual** per-task trial count. `harness.split_result_from_rollouts:346-350` reconstructs `trial_rewards` from whatever rollout files exist on disk, which on a resumed or partially-failed run need not equal the configured `num_trials`. A guard keyed on the config value would be wrong there. That's also the strongest argument for the author's design.

The part I don't accept is "`stats.pass_k` returning 0.0 for k>n is correct by its own definition." pass^k for k>n is not zero, it's undefined; returning a well-formed float for an undefined input is the actual root cause, and it is exported publicly (`__init__.py:55`) with a test blessing it (`test_core.py:102`). Verdict: **right layer for the aggregation, wrong layer for the invariant.** Land as-is, then follow up making `stats.pass_k` return `None`/raise for `k > n` — `loop.py` already `continue`s on those k so the follow-up is a two-line change with no downstream churn.

On the `min()` choice for ragged splits (risk #2): **`min` is correct and I'd reject `max` or per-task suppression.** `pass_k` in `aggregate_scores` is a *mean across tasks* (`loop.py:88`). Under `max`, a split where task A has 5 trials and task B has 1 would average A's real `pass^2` against B's structural `0.0` — the exact #112 bug, laundered through a mean, and *harder* to spot because the number is no longer a suspicious round 0.0. Per-task suppression would make the mean's denominator vary silently per k, so `pass^1` and `pass^2` would be computed over different task sets and stop being comparable — the drop from `pass^1` to `pass^2` is precisely the reliability signal (`skills/phases/report/SKILL.md:36-37`), and it's meaningless across different denominators. `min` discards A's valid measurement, yes, but a split-level statistic that isn't defined for the whole split isn't a split-level statistic. Conservative *and* correct here.

Boundary probes, all clean, no crash and no wrong number:
```
empty rollouts:                {} {}
k == n exactly (n=2, k=2):     {'1': 0.5, '2': 0.0}     <- DEFINED, correctly kept
zero-trial task:               {'1': 0.0}
mixed 2-trial + zero-trial:    {'1': 0.5}               <- pass^2 suppressed
ks=(1,) single trial:          {'1': 1.0}
ks=(0,1):                      {'0': 0.0, '1': 1.0}     <- see Nit 8
```

On suppressing `pass_at_k` (risk #3): **correct, and I verified the clamping claim.** `stats.py:84` is literally `k = min(k, n)`, so `pass_at_k([r], 2)` returns `pass@1` under a `pass@2` label — a mislabeled number, which is worse than an absent one. The author's reasoning is sound. (Charitably, `pass@n` is a valid *lower bound* on `pass@k` since capability is monotone in k — but nothing in the emitted dict says "lower bound," and `pass_at_k["2"] = 1.0` off one trial is not a bound anyone would read as one.) I found no other derived statistic with the same k>n defect: `combined_stderr`, `bootstrap_ci`, and `aggregate` all degrade honestly (`stats.py:44-46,96-99` return `0.0`/`(0.0,0.0)` on `n==0`, which is a genuinely empty input, not an undefined k).

On the skipped `ks=(1,)`-when-`num_trials==1` item (risk #8): **correctly skipped, and the guard is strictly better than what the issue asked for.** A `num_trials`-keyed default cannot see the ragged/resumed case that `split_result_from_rollouts` produces; the trial-count-keyed guard can. Adding both would be two mechanisms enforcing one invariant, with the weaker one able to drift.

---

### dist/ bundle integrity

**The committed bundle matches source, byte-for-byte, reproducibly. This one is clean.**

`dist/` is confirmed intentionally tracked — `dashboard/frontend/.gitignore:11-13` has `# dist` commented out with an explicit rationale, and root `.gitignore:29-30` re-includes it via `!dashboard/frontend/dist/**`.

```
$ md5 dist/assets/*.js dist/assets/*.css dist/index.html      # committed
MD5 (dist/assets/index-5zepM1pz.js) = 5901b8d5927830cb88d9b317bdd2efc7
MD5 (dist/assets/index-b_9cC4ZS.css) = d8c7578b0cfdedd4a9f9143a7b92765a
MD5 (dist/index.html)                = bb5a97151600734529f3e3999f65f143

$ npm ci && npm run build
✓ 2809 modules transformed.
dist/assets/index-b_9cC4ZS.css   16.36 kB │ gzip:   4.51 kB
dist/assets/index-5zepM1pz.js   830.51 kB │ gzip: 249.99 kB
✓ built in 739ms

$ md5 dist/assets/*.js dist/assets/*.css dist/index.html      # rebuilt
MD5 (dist/assets/index-5zepM1pz.js) = 5901b8d5927830cb88d9b317bdd2efc7   # identical
MD5 (dist/assets/index-b_9cC4ZS.css) = d8c7578b0cfdedd4a9f9143a7b92765a   # identical
MD5 (dist/index.html)                = bb5a97151600734529f3e3999f65f143   # identical

(empty)
$ diff -rq /tmp/dist-committed dashboard/frontend/dist
IDENTICAL
```
`dist/index.html` correctly points at the new hash (`assets/index-5zepM1pz.js`), the stale `index-g1UdrsC9.js` is deleted from tracking, and the new bundle carries the fix (`grep -c "N/A"` → 1; `main`'s `index-g1UdrsC9.js` instead carries the buggy literal `pass^k ` template). No runtime staleness hazard **from this PR**. The hazard is real for the merge, below.

---

### The second bug (`test_pass_k` NaN%) — both halves verified independently

Half one — the type lied:
```
$ git show origin/main:dashboard/frontend/src/lib/types.ts | sed -n 91p
  test_pass_k?: number | null
$ git show origin/main:.../KpiStrip.tsx | sed -n 37p
  <Kpi ... hint={summary.test_pass_k != null ? `pass^k ${pct(summary.test_pass_k)}` : undefined}>
```
Half two — the backend always sent a dict: `core/cap_evolve/dashboard.py:548` is `"test_pass_k": test.get("pass_k")`, and `pass_k` has been `dict` since `loop.py:27` (`pass_k: dict = field(default_factory=dict)`), i.e. since well before this PR. So the declared type never matched the wire.

Reproduced the user-visible symptom against the real `format.ts:6-9` implementation (`pct` guards `null` and `NaN` → `'—'`, but a dict is neither `null` nor `NaN` until multiplied):
```
$ node -e "... pct({'1':1.0,'2':0.0}) ..."
OLD (pct on dict): pass^k NaN%
```
Confirmed. On `tsc -b` as sufficient proof of no other consumer: **not sufficient on its own** — a widening from `number | null` to `Record<string, number> | null` would type-error at any *other* arithmetic site, so `tsc` clean does prove no other site does arithmetic; it does not prove no site does something type-compatible-but-wrong (e.g. `String(pk)` → `"[object Object]"`, or a `pk == null` truthiness check). I grepped rather than trusting it: `KpiStrip.tsx:16,54` and `types.ts:92` are the *only* three references to `test_pass_k` in the entire frontend source. So the conclusion holds; the reasoning behind it didn't.
```
$ npx tsc -b   →   TSC CLEAN rc=0
```

---

### Merge-order note

Fetched every open PR. Your three named branches don't exist under those names; the live equivalents are **#194 `perf/issue-119-reduce-cache`**, **#192 `fix/issue-120-cdn-font`**, **#191 `feat/issue-116-follow-tail`**.

**#192 `fix/issue-120-cdn-font` — HARD CONFLICT, and a shipping hazard if resolved carelessly. Highest concern.**
It also commits a rebuilt `dashboard/frontend/dist/`, from a source tree that does **not** contain #187's `passKHint`:
```
$ git merge-tree $(git merge-base origin/main origin/fix/issue-120-cdn-font) HEAD origin/fix/issue-120-cdn-font
changed in both
  base   ... dashboard/frontend/dist/index.html
  our    ... dashboard/frontend/dist/index.html
  their  ... dashboard/frontend/dist/index.html

$ git show origin/fix/issue-120-cdn-font:.../dist/assets/index-zqmQoE4e.js | grep -o 'pass\^k \|N/A'
pass^k          # ← the OLD buggy render, no N/A
$ git show HEAD:.../dist/assets/index-5zepM1pz.js | grep -o 'pass\^k \|N/A'
N/A             # ← the fix
```
Both branches introduce a *differently-hashed* JS bundle and both rewrite `dist/index.html` to point at their own hash. Git will conflict on `dist/index.html` and leave both `.js` files present. **Resolving that conflict by picking either side ships a bundle missing the other PR's fix** — take #192's side and this PR's `pass^k NaN%` → `N/A` fix silently does not reach users at runtime even though the merge is green, because the backend serves the prebuilt SPA. (#192 hits `skills/algorithms/evograph/dashboard/frontend/dist/` too, same shape.)
→ **Whichever lands second MUST `npm ci && npm run build` and commit the regenerated `dist/` as the conflict resolution.** Never hand-resolve `dist/`. This is also why Non-blocking 5 (a `git diff --exit-code dist` gate in CI) is worth doing before either merges.

**#194 `perf/issue-119-reduce-cache` — textual conflict unlikely, no semantic overlap.** It touches `dashboard/frontend/src/lib/types.ts` but at `:261` (`StreamEvent.snapshot` payload), ~170 lines from #187's `:92`. It does **not** touch `dist/` at all — which means #194 itself will land with a stale bundle relative to its own `types.ts` change (type-only, so harmless at runtime, but it's the same CI gap).

**#191 `feat/issue-116-follow-tail` — no overlap** (`cli.py`, `eventstream.py`, `backend/stream.py`, `CHANGELOG.md`). Only a trivial `CHANGELOG.md` conflict if you act on Nit 9.

Recommended order:
1. **#187** (this PR) — smallest surface, `dist/` verified reproducible, unblocks the honesty story.
2. **#192** — rebase onto the new `main`, then `npm ci && npm run build` and commit the regenerated `dist/` (do not hand-merge `index.html`). Verify the merged bundle contains **both** `N/A` and the no-CDN font change before merging.
3. **#194** — independent; ideally add the `git diff --exit-code -- dashboard/frontend/dist` CI step here since it already owns the dashboard-perf surface.
4. **#191** — independent, any time.

---

### Verification I re-ran

**Full suite — 182, exactly as claimed (179 baseline + 3 new):**
```
$ cd /tmp/rv-187 && PYTHONPATH=/tmp/rv-187/core /tmp/ce-venv/bin/python -m pytest core/tests -q
182 passed in 64.43s (0:01:04)
```
(Later reruns in the same session showed `test_dashboard_launch.py::test_maybe_launch_spawns_when_available` failing with `assert 'http://127.0.0.1:7882' == 'http://127.0.0.1:7878'`. Diagnosed as a **pre-existing environment flake, unrelated to this PR** — `lsof -nP -iTCP:7878 -sTCP:LISTEN` showed a stray `Python` process holding 7878, so `maybe_launch` correctly walked to a free port and the test's hardcoded expectation broke. Reproduces identically with `loop.py` reverted to `origin/main`.)

**Fail-before / pass-after (`loop.py` only reverted, new test kept):**
```
$ git checkout origin/main -- core/cap_evolve/loop.py
=== SOURCE REVERTED ===
$ pytest core/tests/test_pass_k_undefined.py -q
FAILED core/tests/test_pass_k_undefined.py::test_single_trial_omits_k2
FAILED core/tests/test_pass_k_undefined.py::test_ragged_trials_use_the_min
2 failed, 1 passed in 0.03s

$ git checkout HEAD -- core/cap_evolve/loop.py
=== RESTORED ===
...                                                                      [100%]
3 passed in 0.02s
```
2 failed → 3 passed. Claim holds, and `test_multi_trial_still_reports_k2` does pass in both states as stated.

**Mutation-probing the guard (my addition — the PR did not do this):**
```
=== MUTANT `if k >= max_k:` (wholesale off-by-one) ===        2 failed, 1 passed   ✅ caught
=== MUTANT `if k >= max_k and max_k > 1:` ===                 2 failed, 1 passed   ✅ caught
=== MUTANT `if k > max_k or (max_k >= 2 and k == max_k):` ===  3 passed / 181 passed ❌ NOT caught
```
→ Non-blocking 4.

**Real zero-API `toy_calc` run with the `mock` optimizer — the headline claim, reproduced:**
```
$ python -m cap_evolve.cli run --spec .../capevolve.yaml --project ... --run-ts demo
{
  "best_id": "cand_0001",
  "test_reward": 1.0,
  "test_pass_k": {
    "1": 1.0
  },                          ← was {"1": 1.0, "2": 0.0}
  "iterations": 3
}

$ cat .capevolve/run_demo/report.md
- **Held-out test (optimized skills): 1.0**  (pass^1=1.000, pass^2=N/A)

$ python -c "... final.json ..."
{
  "test_pass_k":    {"1": 1.0},     ← was {"1": 1.0, "2": 0.0}
  "test_pass_at_k": {"1": 1.0}      ← was {"1": 1.0, "2": 1.0}
}
```

**`compileall` + the report skill's own smoke check:**
```
$ python -m compileall -q core skills   →   COMPILEALL rc=0
$ python skills/phases/report/scripts/check.py   →   all checks pass, incl.
    "renders dashboard.html", "self-contained (no CDN / network)",
    "secret redaction holds", "ANSI terminal report mode"
```

**Frontend:**
```
$ npx tsc -b   →   TSC CLEAN rc=0
$ npm ci && npm run build   →   ✓ built in 739ms, bundle identical to committed (see dist/ section)
```

**Skill registry manifest** — rebuilt per `CONTRIBUTING.md` step 3, no diff, so the PR did not leave it stale:
```
wrote skills/_registry/manifest.json (20 skill(s))
$ git status --porcelain
(empty)
```

**Security / generalizability (risk #9):** no secrets, no absolute paths, no host assumptions in the diff. The guard reads only `len(trial_rewards)` off the `Score` objects the adapter's rollouts produced, so it holds for a third-party adapter at any trial count including ragged and zero — `harness.split_result_from_rollouts:346-350` reconstructs real per-task counts from disk and feeds the same guard, so resume is covered too. `redact` (`dashboard.py:67-88`) still walks the summary that carries `test_pass_k`; unchanged.

**Docs sweep (risk #7):** `grep -rn "pass\^2 = 0\|\"2\": 0.0"` across `docs/ site/ README.md RUN.md CHANGELOG.md examples/` returns **only** the new explanatory line in `docs/HONEST_EVAL.md:33`. No doc, site page, or example fixture documents the old behavior or shows a `pass^2 = 0.0`. The two numeric pass^2 mentions (`site/run-end-to-end.html:80`, `docs/REPRODUCE_tau2.md:23` — `pass² 0.584`) are the 10-trial τ²-bench run, legitimately defined. Confirmed the self-contained `dashboard.html` template needs no change: it never reads `test_pass_k`.

OsherElhadad:
## 🔧 Review fixes

Pushed `2a34dbd`… (see HEAD). All 10 findings answered below. Two were more serious than "non-blocking" suggests and are fixed properly; nothing is declined outright.

---

### 1. `report/scripts/run.py` + `KpiStrip.tsx` hardcode `k in (1, 2)` — FIXED IN BOTH ✅

You're right that this is the same defect as #112, mirrored: with a non-default `ks` the report **dropped a measured `pass^3`** and **fabricated a `pass^2=N/A` nobody requested**. And it's reachable today, since `gepa.py:184` passes a non-default `ks`.

Both surfaces now render exactly the ks that are present, in numeric order.

```python
# skills/phases/report/scripts/run.py
pk_str = ", ".join(f"pass^{k}={float(pk[k]):.3f}" for k in sorted(pk, key=int))
```
```ts
// KpiStrip.tsx — exported so it's directly testable
export function passKHint(pk: RunSummaryDetail['test_pass_k']): string | undefined {
  if (pk == null || typeof pk !== 'object') return undefined
  const ks = Object.keys(pk)
    .filter((k) => Number.isFinite(Number(k)) && pk[k] != null)
    .sort((a, b) => Number(a) - Number(b))
  if (ks.length === 0) return undefined
  return ks.map((k) => `pass^${Number(k)} ${pct(pk[k])}`).join(' · ')
}
```

Note the `N/A` literal is gone from both. That is deliberate and it *strengthens* the honesty property: the only source of truth for "which ks were requested" is `final.json`, and `final.json` records the ks that were **measured**, not the ks that were **asked for**. So any `N/A` rendered from that data is a guess. Omitting the k is the honest representation — same argument as the core fix.

New tests, one per surface:
```
core/tests/test_report_pass_k_render.py   (drives the REAL report script end-to-end)
  test_nondefault_ks_renders_measured_k3_and_never_invents_k2
  test_single_trial_renders_only_k1
  test_legacy_scalar_pass_k_still_renders

dashboard/frontend/src/test/passKHint.test.ts
  renders a non-default ks without inventing pass^2
  renders only k=1 for a single-trial run
  sorts numerically, not lexically      ← {'10':…, '2':…} was lexically wrong too
  is undefined when there is nothing measured
```
`npm test` was not wired into CI at all, so the TS test would never have run — added a vitest step to the frontend job (see finding 5).

---

### 2. `stats.pass_k` root cause — FIXED, not deferred ✅

Not too invasive: I verified `loop.aggregate_scores` is the only non-test production caller, and it already `continue`s on exactly the undefined `k`, so `None` never reaches a consumer. `None` over `raise` because raising would make the public function hostile to the exploratory use the export exists for.

```python
def pass_k(trial_rewards, k, threshold=1.0) -> float | None:
    ...
    if k < 1 or k > n:
        return None      # was: return 0.0
```
`test_core.py:102` no longer blesses the fabricated value:
```python
def test_pass_k_k_gt_n_is_undefined_not_zero():
    assert pass_k([1.0], 2) is None
    assert pass_k([], 1) is None
    assert pass_k([1.0], 0) is None      # k < 1 is undefined too
    assert pass_k([1.0], 1) == 1.0       # the k == n boundary stays DEFINED
    assert pass_k([0.0], 1) == 0.0       # a genuinely measured zero is still 0.0
```
The last two lines exist so this test also pins the boundary from the `stats` side, and so "undefined" can never be conflated with "really measured 0".

`pass_at_k` deliberately still returns `0.0`/clamps — its clamp is at `k = min(k, n)` *after* the `n == 0` check, and `aggregate_scores` never calls it with an out-of-range k now. Changing it would be churn with no reachable caller.

---

### 3. `k == n` boundary untested — FIXED, and I re-ran your mutation ✅

You were right that `test_multi_trial_still_reports_k2` (n=3, k=2) never touches the boundary. Added `k == n`, `k == n+1`, `k=1/n=1`, `k <= 0`, and a non-default-`ks` case.

**Your mutant, applied to the new guard:**
```
$ # if k < 1 or k > max_usable_k or (max_usable_k >= 2 and k == max_usable_k):
=== MUTANT: over-suppress the DEFINED k == n boundary ===
FAILED core/tests/test_pass_k_undefined.py::test_k_equals_n_exactly_is_reported
FAILED core/tests/test_pass_k_undefined.py::test_k_equals_n_plus_one_is_suppressed
FAILED core/tests/test_pass_k_undefined.py::test_nondefault_ks_reports_only_what_was_asked
3 failed, 6 passed in 0.03s
```
**Restored:**
```
$ grep -n "if k < 1" core/cap_evolve/loop.py
96:        if k < 1 or k > max_usable_k:
$ pytest core/tests/test_pass_k_undefined.py -q
9 passed in 0.02s
```
3 failed → 9 passed. The mutant that previously slipped through the whole 182-test suite is now caught by three independent tests.

---

### 4. CI cannot detect a stale bundle — FIXED (minimal drift check, defers to #188) ✅

Implemented the minimal check here and referenced #188 for the broader committed-`dist/` design problem. One thing worth flagging: `git diff --exit-code` **alone is not sufficient**, because a changed bundle hash *renames* the file, so the fresh output lands as **untracked**, not modified — `git diff` would report clean. The check covers both:

```yaml
- name: Fail if the committed dist/ bundle is stale
  run: |
    if ! git diff --exit-code -- dashboard/frontend/dist; then
      echo "::error::dashboard/frontend/dist is stale. Run 'cd dashboard/frontend && npm ci && npm run build' and commit the result."
      exit 1
    fi
    untracked="$(git ls-files --others --exclude-standard -- dashboard/frontend/dist)"
    if [ -n "$untracked" ]; then
      echo "::error::the fresh build produced files missing from the committed dashboard/frontend/dist:"
      echo "$untracked"
      exit 1
    fi
```

**Proof it actually fires.** First attempt with a comment-only source edit did NOT trip it — minification strips comments, so the bundle was byte-identical. Re-ran with a real behavioral change (`const DASH = '—'` → `'DRIFT_PROBE'` in `format.ts`):

```
A) fresh bundle (as committed)
$ bash drift-check
dist/ is fresh
rc=0

B) src edited, rebuilt, dist NOT committed
$ npm run build
dist/assets/index-BYskQRcX.js   830.46 kB
$ bash drift-check
::error::dashboard/frontend/dist is stale. Run 'cd dashboard/frontend && npm ci && npm run build' and commit the result.
rc=1

C) source restored + rebuilt + committed
$ bash drift-check
rc=0
```

Also added the missing `npm test` step — vitest existed in `package.json` but CI never invoked it.

---

### 5. `docs/HONEST_EVAL.md` overclaim — FIXED ✅
Rewritten to be accurate, and extended to name the mirrored defect finding 1 was about:
> …`stats.pass_k` returns `None` rather than a plausible `0.0` for such a `k`, and no surface renders a `k` it did not measure — human surfaces (`report.md`, the dashboard KPI strip) list exactly the measured ks, and the JSON surfaces simply omit the key. Reporting `pass^2 = 0.0` on a single-trial run would read as "0% reliable" when it only means "not enough trials"; inventing a `pass^2 = N/A` on a run where `ks` never included 2 is the same defect mirrored.

### 6. Nit — `max_k` computed with `min()` — FIXED ✅
Renamed `max_usable_k`. Your reading was right; the name cost a double-take.

### 7. Nit — guard is upper-bound only, `k <= 0` leaks a fake `0.0` — FIXED ✅
`if k < 1 or k > max_usable_k: continue`, in both `loop.py` and `stats.pass_k`. Covered by `test_nonpositive_k_is_suppressed` and the `pass_k([1.0], 0) is None` assertion.

### 8. Nit — no CHANGELOG entry — FIXED ✅
Added under `[Unreleased] → Fixed`, covering the core suppression, the `stats.pass_k` hardening, both hardcoded-`(1,2)` surfaces, the `NaN%` type bug, and the CI drift gate. (Trivial conflict with #191 if it also touches CHANGELOG — take both hunks.)

### 9. Nit — dead inner `if (s.trial_rewards or [s.reward])` filter — FIXED ✅
Deleted. You're right it can never be falsy, and leaving it in implies a zero-trial branch that does not exist. The zero-trial task is still handled correctly, but now by the `min(trials_per_task)` guard on purpose rather than by accident: `trial_rewards=[]` falls back to `[reward]` (1 trial), so a mixed 2-trial + zero-trial split suppresses `pass^2` — the conservative-correct outcome, reached deliberately.

### 10. Pre-existing `from_dict` `TypeError` on a legacy scalar `pass_k` — FIXED HERE, not deferred ✅
Cheaper to fix than to track, and `report/scripts/check.py:51` writes that exact shape today:
```python
def _as_k_dict(v) -> dict:
    if isinstance(v, dict):
        return dict(v)
    if isinstance(v, (int, float)) and not isinstance(v, bool):
        return {"1": float(v)}
    return {}
```
Covered by `test_from_dict_tolerates_legacy_scalar_pass_k`. (`bool` excluded because `isinstance(True, int)`.)

---

## Non-default `ks` end-to-end proof

`ks` is not a CLI flag, so I drove a **real** zero-API `toy_calc` + `mock` run at `num_trials: 3`, then re-aggregated **that run's actual on-disk rollouts** through the real `harness.split_result_from_rollouts(ks=(1,3))` and re-rendered through the real report script.

```
$ python -m cap_evolve.cli run --spec …/capevolve.yaml --project … --run-ts demo --dashboard off
{
  "best_id": "cand_0001",
  "test_reward": 1.0,
  "test_pass_k": { "1": 1.0, "2": 1.0 },
  "iterations": 3
}

$ harness.split_result_from_rollouts(rd, "FINAL", split="test", ks=(1, 3))
  pass_k   : {"1": 1.0, "3": 1.0}
  pass_at_k: {"1": 1.0, "3": 1.0}

$ python skills/phases/report/scripts/run.py --run-dir … --no-dashboard
$ grep "Held-out test (optimized" report.md
- **Held-out test (optimized skills): 1.0**  (pass^1=1.000, pass^3=1.000)

$ grep -c "pass\^2" report.md
0        ← never requested, never fabricated (was: pass^3 dropped, pass^2=N/A invented)
```
And the original single-trial case still holds:
```
$ grep "Held-out test (optimized" /tmp/toy-after/.capevolve/run_demo/report.md
- **Held-out test (optimized skills): 1.0**  (pass^1=1.000)
```

---

## Verification

**Full suite — 191 passed, 0 failed** (182 on this branch + 9 new). No port-7878 flake (#200) in either run:
```
$ PYTHONPATH=/tmp/fx-187/core python -m pytest core/tests -q
........................................................................ [ 37%]
........................................................................ [ 75%]
...............................................                          [100%]
191 passed in 62.12s (0:01:02)
```
```
$ python -m compileall -q core skills                 → rc=0
$ python skills/phases/report/scripts/check.py        → ok: true, problems: []
    incl. "renders dashboard.html", "self-contained (no CDN / network)",
          "secret redaction holds", "ANSI terminal report mode"
$ python skills/_registry/build_manifest.py skills    → 20 skill(s), no diff
```
**Frontend:**
```
$ npx tsc -b --noEmit          → rc=0
$ npm test  (vitest)           → 14 files, 49 tests passed
$ npm run build                → ✓ built in 729ms
    dist/assets/index-EltQq6s9.js   830.45 kB
```
**`dist/` re-verified reproducible after the `KpiStrip.tsx` change** — rebuilt, committed, then rebuilt again from clean:
```
$ git status --porcelain -- dashboard/frontend/dist
R  dist/assets/index-5zepM1pz.js -> dist/assets/index-EltQq6s9.js
M  dist/index.html
(both staged & committed; a subsequent fresh build leaves the tree clean — drift check rc=0)
```
The new bundle no longer contains the `N/A` literal, by design (finding 1): the KPI strip renders only measured ks.

---

## ⚠️ Merge hazard — read before merging #187 or #192

**#192 (`fix/issue-120-cdn-font`) is a hard conflict with this PR AND a silent shipping hazard.** It commits its own rebuilt `dashboard/frontend/dist/` from a tree that does **not** contain #187's `passKHint`, so its bundle still carries the old `pass^k` render with no honest-N/A handling. Both branches introduce a differently-hashed `.js` and both rewrite `dist/index.html` to point at their own hash.

**Resolving that conflict by picking either side ships a bundle missing one of the two fixes — even on a green merge**, because the backend serves the prebuilt SPA at runtime. Take #192's side and this PR's fix silently never reaches users. (#192 hits `skills/algorithms/evograph/dashboard/frontend/dist/` too, same shape.)

**→ Whichever lands second MUST `cd dashboard/frontend && npm ci && npm run build` and commit the regenerated `dist/` as the conflict resolution. Never hand-resolve `dist/index.html`.** Then verify the merged bundle contains **both** fixes before merging.

Recommended order: **#187 → #192 → #194 → #191**. The CI drift check added here (finding 4) will now *catch* this class of mistake instead of letting it through green — which is a reason to land #187 first.

Issue #125, opened 2026-07-25
Publish pip install cap-evolve to PyPI (remove clone-and-pip-./core requirement)


OsherElhadad:
## Problem

Every install path requires `git clone` + `pip install ./core`. There is no `pip install cap-evolve` (acknowledged as future work in `docs/ROADMAP.md:54-56`). For a project with a polished site, badges, and "Made at" logos, "clone the repo" is a maturity mismatch, and it blocks the shortest possible quickstart.

## Why it matters

The core is pure stdlib with **zero runtime dependencies** — it is the easy case for packaging, and a published wheel would let the quickstart be `pip install cap-evolve && cap-evolve ...` with no clone. It also enables the `doctor` and CLI story to stand on their own, and makes the "two minutes to first run" claim literally true without a checkout.

## What to change

1. Finalize package metadata in `core/pyproject.toml` (name `cap-evolve-core` / distribution name, version, description, URLs, classifiers, license, README long-description). Confirm the CLI entry point (`cap-evolve`) is declared.
2. Decide the public distribution name (`cap-evolve` vs `cap-evolve-core`) and reconcile with the README's `pip install ./core`.
3. Add a release workflow (`.github/workflows/`) that builds sdist+wheel and publishes to PyPI on a tag (Trusted Publishing / OIDC preferred over a stored token). See `RELEASING`-style conventions.
4. Publish an initial `0.x` (coordinate with the CHANGELOG-restructure issue so the released version section is real).
5. Update README/INSTALL/site quickstarts to `pip install cap-evolve` once live (keep the `./core` path for from-source dev).

## How to test

- `pip install cap-evolve` in a clean venv → `cap-evolve version` works, `cap-evolve check`/`run` work on `toy_calc`.
- Tagging a release triggers the publish workflow (dry-run to TestPyPI first).
- The dashboard package (`cap-evolve-dashboard`) install story is documented alongside (optional extra).

## Dependencies / parallelization

- **Coordinate with** the CHANGELOG-restructure issue (needs a real released version). Otherwise **parallel-safe.** Distribution/DX; no core logic changes.

PR #181, opened 2026-07-29
fix(docs): unify adapter contract as 3 required methods + hooks, fix stale implement-and-check signatures

OsherElhadad:
Closes #103

## Ground truth

Established by reading the code, not the docs:

- **`core/cap_evolve/adapter.py`** — exactly **3 `@abstractmethod`s**:
  `tasks(split)`, `run_target(task, ctx, *, seed=0)`, `score(task, rollout)`.
  Defaulted hooks: `materialize(candidate_dir, edits=None)`, `live(candidate_dir)`,
  `apply(candidate_dir, edits=None)`. Optional/feature-detected: `trajectories(split, ctx=None)`,
  `runner_model()`, and `run_batch` / `run_trials` (probed via `hasattr` in `harness.py:201-202`).
- **`core/cap_evolve/check.py`** — the gate checks: no IMPLEMENT-ME stubs, `tasks('val')`
  non-empty + stable, deterministic `score()`, and probes **`materialize()`** into a temp
  copy (`check.py:146-165`) — deliberately *not* `apply()`/`live()`, so the host is never mutated.

Canonical phrasing adopted everywhere: **"3 required methods + defaulted hooks (+ optional)."**

## Surfaces fixed

| file | old | new |
|---|---|---|
| `core/cap_evolve/adapter.py:3-12` | "we use 4, cleaner" then a flat list of **5** methods | "we require **3**, plus defaulted hooks"; list split into 3 required / 3 defaulted hooks / 4 optional |
| `skills/phases/implement-and-check/SKILL.md:3` (frontmatter description) | "implementing the 4 adapter methods" | "implementing the 3 required adapter methods plus any defaulted hooks that need overriding" |
| `skills/phases/implement-and-check/SKILL.md:27` | "Implement the **four** adapter methods" | "Implement the **3 required** adapter methods" |
| `skills/phases/implement-and-check/SKILL.md:30` | `run_target(task, capability)` (signature does not exist) | `run_target(task, ctx, *, seed=0)` + note to forward `seed` when stochastic |
| `skills/phases/implement-and-check/SKILL.md:34` | `apply(capability, edit)` — listed as a **required** method (signature does not exist) | removed from the required list; replaced by a defaulted-hooks paragraph with real signatures `materialize(candidate_dir, edits=None)` / `live(candidate_dir)` / `apply(candidate_dir, edits=None)` + optional extras |
[...]

Issue #127, opened 2026-07-25
Epic: cap-evolve full review vs Arbor — 28 self-contained improvement issues (#99–#126)


OsherElhadad:
Tracking issue for a ruthless, full review of cap-evolve, benchmarked against a direct competitor (**Arbor**, `RUC-NLPIR/Arbor`, **Apache-2.0**). Research was done by reading both codebases end to end (core engine, algorithms, skills/plugins, CLI, dashboard, docs, site).

> **License note.** Arbor is Apache-2.0. Everything below is **technique/idea inspiration only** — no Arbor source, prompt strings, or assets are to be copied. Ideas and algorithms aren't copyrightable; the individual issues describe *what* to build and cite *our own* file:line, and are implemented clean-room. Where an issue references prior art it does so neutrally ("prior art in this space demonstrates…"), never "copy X".

## The honest headline

cap-evolve's **substance is strong**: the honesty core (seeded splits, sealed test, val-only significance gate, variance-aware stats, git-versioned candidates, redaction) is careful and well-tested, and the React dashboard is genuinely professional. The engine has **no benchmark special-casing** in its executable paths — it's more general than it feels.

Where it loses to the competitor is **not** substance — it's **packaging, live observability, onboarding, and internal consistency**:
- **Silent runs.** A classic `cap-evolve run` prints *nothing* to the terminal for hours (#116). The competitor makes "watch it evolve" literally true; our tagline is *"watch capability evolve."*
- **Discarded signal.** The dashboard already collects a live event stream and throws it away (#117); the terminal has none (#116).
- **Flagship algorithms run blind.** GEPA & SkillOpt get a *thinner* optimizer context than the basic hill-climb, and their CLI flags are silently dropped (#109) — plus dirty snapshots (#110) and a cache that hollows out GEPA's own reflection (#111).
- **Honesty inconsistency** — the one thing the brand can't afford: the headline number is presented as sealed while its artifact isn't committed (#99), results diverge across README ⊂ RESULTS.md ⊂ site (#100), and a stale number survives in the CHANGELOG (#101).
- **"Skills feel not-good/general"** is real but **misattributed**: the bones are coherent; the surface is docs drift (3-vs-4 methods #103, undocumented default gate mode #104), an unfinished rebrand (#105), a capability's playbook leaking into the generic intake phase (#106), and skills that blow the framework's *own* size lint (#107).
[...]

Issue #121, opened 2026-07-25
Add cap-evolve doctor install/health diagnostic


OsherElhadad:
## Problem

There is no install/health self-diagnostic. The most common first-run failures — `cap-evolve: command not found` after `./install.sh` (which intentionally doesn't `pip install` the core), a shadowing second install, "trapped in the wrong venv", missing/misplaced host skill dirs (`install.sh:38-40` admits several host dirs are "best-guess"), and missing provider credentials — all surface as opaque errors with no guided fix.

## Why it matters

The `install.sh` + `pip install ./core` two-step is a predictable stumble (see the docs review: a first-timer who runs only `./install.sh` then `cap-evolve` hits `command not found`). Prior art in this space (Arbor, Apache-2.0) demonstrates the value of a single `doctor` command that introspects the install and preempts the most common support load — reporting the running executable vs the first-on-PATH one (shadowed installs), venv traps, git presence, and the credential surface. We can implement the same *idea* from scratch (technique inspiration only, no code copied).

## What to change

Add **`cap-evolve doctor`** that checks and prints pass/fail with actionable fixes:
1. Is `cap-evolve` on PATH? Does the resolved executable match the intended install (detect a shadowing second install)?
2. Is the core package importable (`import cap_evolve`)? Which venv/interpreter?
3. Is `git` available (the default version store needs it)?
4. Provider credentials surface (env + any config) — present/absent, without printing secret values (reuse the existing redaction in `dashboard.py`).
5. For plugin/host installs: are the skills where the selected host expects them? Flag "best-guess" host dirs explicitly.
6. Exit non-zero if any hard check fails (CI-friendly).

## How to test

- New test: `cap-evolve doctor` in a clean venv with the core installed exits 0 and reports git + import OK.
- A deliberately broken setup (no git, or core not importable) exits non-zero with a clear message.
- Secrets are never printed (assert redaction on a fake key).

## Dependencies / parallelization

- **Parallel-safe.** New self-contained CLI subcommand in `cli.py`. Pairs well with the "verified vs best-guess host labels" docs note and the PyPI-publish issue (all reduce install friction) but is independent.

OsherElhadad:
[...]

PR #190, opened 2026-07-29
feat(config): provider-scoped credential resolution + auto provider probing (closes #134)

### LINKED ISSUE #134: Config ergonomics: provider-scoped credential resolution + auto provider probing
## Problem / opportunity

Provider/model configuration is spread across env vars + `capevolve.yaml` with no explicit, documented resolution order and no protection against a subtle footgun: reusing a **user-level key for a different provider** than the project selected. There's also no `auto` provider probing.

## Why it matters

Prior art in this space (Arbor, Apache-2.0 — inspiration only) uses an **explicit, provider-scoped config precedence** (CLI flag > project config > user config > built-ins) that **refuses to cross provider boundaries** — it won't silently apply a user-level key from provider A when the project selected provider B. It also offers an `auto` provider mode that probes which API shape (e.g. Responses vs chat-completions) a base URL speaks. Both eliminate a class of hard-to-debug "why is it calling the wrong endpoint / using the wrong key" failures. cap-evolve's `model_config.py` is close; it needs the explicit precedence + provider-scoping + `auto`.

## What to change

1. Define and **document** the config precedence (CLI > project `capevolve.yaml` > user config > built-in defaults) in one place (docs + `model_config.py` docstring).
2. Make key/credential resolution **provider-scoped**: don't apply a credential configured for a different provider than the one selected; error clearly instead.
3. Add an `auto` provider mode that probes the endpoint's API shape (optional, behind a flag) for OpenAI-compatible base URLs.
4. Fold in version-from-git-tag packaging as part of **#125** (PyPI) rather than duplicating it here.

## How to test

- New test: project selects provider B, only a provider-A key is present → clear error, no silent cross-use.
- New test: precedence resolves as documented (CLI overrides project overrides user).
- `auto` mode probes and selects the correct API shape against a mocked endpoint.

- Complements **#133** (quickstart) and **#121** (doctor). **Parallel-safe.** Small-to-medium; no algorithm changes.

OsherElhadad:
Closes #134.

Provider/model config gets an explicit, documented precedence and **provider-scoped**
credential resolution, plus a `provider: auto` mode that probes which provider is
actually usable. New module: `core/cap_evolve/model_config.py` (stdlib only, zero new
runtime deps — one resolution table, no plugin framework).

## Precedence (documented in one place)

Applied **per field** (`provider`, `provider_base_url`, `provider_credential_env`) —
highest layer that sets a non-empty value wins:

**CLI flag > project `capevolve.yaml` > user config (`~/.capevolve/config.yaml`, or
`$CAPEVOLVE_CONFIG`) > built-in default.**

Lives in `model_config.PRECEDENCE`, the module docstring, and
`docs/INSTALL.md#resolution-order-provider--credentials`. A test asserts the doc and
the constant agree, so they can't drift.

## Per-provider env vars

Each provider reads **only its own** vars (first set wins):

| provider | credential env vars | base URL env | built-in base URL |
|---|---|---|---|
| `anthropic` | `ANTHROPIC_API_KEY`, `ANTHROPIC_AUTH_TOKEN` | `ANTHROPIC_BASE_URL` | `https://api.anthropic.com` |
| `openai` | `OPENAI_API_KEY` | `OPENAI_BASE_URL` | `https://api.openai.com/v1` |
| `gemini` | `GEMINI_API_KEY`, `GOOGLE_API_KEY` | `GEMINI_BASE_URL` | `https://generativelanguage.googleapis.com/v1beta` |
| `rits` | `RITS_API_KEY` | `RITS_API_URL` | — (required) |
| `watsonx` | `WATSONX_APIKEY`, `WATSONX_API_KEY` | `WATSONX_URL` | — (required) |
| `moonshot` | `MOONSHOT_API_KEY`, `KIMI_API_KEY` | `MOONSHOT_BASE_URL` | `https://api.moonshot.ai/v1` |
| `github-copilot` | `COPILOT_GITHUB_TOKEN`, `GH_TOKEN`, `GITHUB_TOKEN` | `COPILOT_BASE_URL` | `https://api.githubcopilot.com` |
| `bob` | `BOBSHELL_API_KEY`, `BOB_API_KEY` | `BOBSHELL_URL` | — (required) |
| `cursor` | `CURSOR_API_KEY` | `CURSOR_BASE_URL` | — (required) |
| `factory` | `FACTORY_API_KEY` | `FACTORY_BASE_URL` | — (required) |
| `mock` | none (offline) | — | — |

A key for provider A is **never** applied to provider B: `_scoped_credential_env`
reads a single `PROVIDERS[provider]["env"]` row and nothing else. Even an explicit
`provider_credential_env` pointing at another provider's var is refused
(`_owner_of` check) rather than honoured. When nothing resolves, the error names the
vars the selected provider accepts, says which *other* providers do have credentials
and that they won't be reused, and restates the precedence.

## Why a credential can never reach the wrong endpoint

This was the main risk in the issue, so it is structural rather than a convention:

1. **URL, auth header, and env var live in the same table row.** `PROVIDERS[p]` owns
   `base_url` / `base_url_env` / `probe_path` / `header` / `env` together. There is no
   code path that takes the credential from one row and the URL from another.
2. **One function reads a credential value.** `probe()` is the only place
   `os.environ[cred_env]` is dereferenced. It looks up `row = PROVIDERS[provider]`,
   gets `cred_env` from that same row via `_scoped_credential_env(provider)`, and
   builds the header from `row["header"]` against `row`'s own base URL.
3. **No credential ⇒ no request.** If the row's own var isn't set, `probe` returns
   `(False, "no <p> credential (VARS) set")` *before* constructing a request, so
   `auto` walking candidates cannot send provider A's token at provider B's URL while
   scanning.
4. **Unknown provider is refused before any I/O**, so a typo'd name can't fall through
   to some default endpoint.
5. **A caller-supplied `base_url` is only ever the resolved base URL for that same
   provider** (`_resolve_base_url(provider, …)` — its own `*_BASE_URL` env or its own
   built-in), and `auto` passes each candidate only its own resolved URL.

`test_probe_never_sends_credential_to_another_providers_endpoint` asserts this at the
wire level: with only `ANTHROPIC_API_KEY` set it intercepts every `urlopen`, checks the
anthropic probe hits `api.anthropic.com` carrying the token, and that the openai /
gemini / rits / watsonx probes make **zero** requests.
`test_auto_probe_does_not_leak_credential_across_candidates` fails the test from inside
the fake transport if a request with the token ever targets a foreign host.

## No-leak guarantee

- `Resolved` stores the env var **NAME**, never the value. Any `repr`/`str`/`to_dict`/
  `json.dumps`/`describe` of a resolved config is secret-free *by construction*, not by
  remembering to redact. Presence is reported as `credential_present: true`.
- Nothing in the module prints, logs, or embeds a credential value — not a prefix, not a
  length.
- Probe failure reasons (which can carry a library-formatted URL) are passed through
  `_safe_reason`: strip `?api_key=…`-style query creds, then the repo's **existing**
  `dashboard.redact` (reused, not reimplemented), then truncate.
- Nothing is written to `events.jsonl` / run artifacts / the dashboard beyond the
  secret-free `{"step": "provider", …}` line.

Tests proving it: `test_resolved_object_carries_no_credential_value`,
`test_probe_failure_reason_is_scrubbed`,
`test_cli_run_transcript_never_prints_credential` (spawns a real `cap-evolve run`
subprocess with a fake token exported and asserts neither the token nor the fragment
`FAKETOKEN` appears in stdout+stderr), `test_no_leak_on_the_nothing_resolves_path`,
`test_describe_is_a_single_secret_free_line`.

## `provider: auto`

Walks `AUTO_ORDER`, selects the first provider whose **own** credential is set, skips
one that needs a base URL it lacks (e.g. `rits` without `RITS_API_URL`), and with
`--probe-provider` additionally requires that candidate's endpoint answer. The chosen
provider comes with a secret-free `reason` naming the env var that was set, why it won,
and what was skipped.

## Scope / coordination

Touched `harness.py`: **nothing** — no regions changed. `specfile.py`: unchanged (only
read). `cli.py`: two localized regions in `_cmd_run` — the argparse block (4 new
`--provider*` flags) and one resolution block just before `--plan-only` returns, whose
result is echoed as a `provider` key in the plan and as a `{"step": "provider"}` line
before the check gate. Version-from-git-tag is #125's, not here. No algorithm, gate,
split, or seal code touched; `--probe-provider` is opt-in so no run performs network I/O
that didn't before.

## Verification

```
$ PYTHONPATH=/tmp/wt-134/core python -m pytest core/tests -q
........................................................................ [ 70%]
.............................................................            [100%]
205 passed in 63.67s (0:01:03)
```
179 baseline + 26 new, 0 failed.

```
$ python -m compileall -q core skills && echo "COMPILEALL CLEAN"
COMPILEALL CLEAN
```

Real end-to-end, zero API cost (`examples/toy_calc`, `mock` optimizer):
```
{"step": "provider", "provider": "mock", "credential_env": "", "credential_present": false,
 "base_url": "", "sources": {"provider": "built-in default (from optimizer_skill)"},
 "reason": "provider 'mock' from built-in default (from optimizer_skill)"}
{
  "run_dir": ".capevolve/run_demo", "best_id": "cand_0001",
  "baseline_val": 0.0, "test_reward": 1.0, "test_delta": 1.0, "iterations": 3
}
```

Same run with `provider: auto` and a fake token exported as both `ANTHROPIC_API_KEY`
and `OPENAI_API_KEY` — resolution reported, run completed, and the token appears in
**no** stream or artifact:
```
{"step": "provider", "provider": "anthropic", "credential_env": "ANTHROPIC_API_KEY",
 "credential_present": true, "reason": "auto selected 'anthropic': ANTHROPIC_API_KEY is
 set and it is the highest-priority provider with its own credential; credential present
 (no probe requested)."}
baseline_val 0.0 -> test_reward 1.0

$ grep -c "FAKELEAKPROBE" /tmp/leak_stdout.txt /tmp/leak_stderr.txt
/tmp/leak_stderr.txt:0
/tmp/leak_stdout.txt:0
$ grep -rl "FAKELEAKPROBE" $D | wc -l
0
```

Failure path (`--provider openai` with only an anthropic key present) — actionable, no
value echoed:
```
{"step": "provider", "error": "no credential found for provider 'openai' (selected via
CLI flag). Set one of: OPENAI_API_KEY. Credentials for anthropic ARE present but belong
to a different provider and are never reused — either export a openai credential, or
select that provider explicitly (provider: anthropic) / use provider: auto.
Precedence: CLI flag > project capevolve.yaml > user config > built-in default."}
```

One live probe (base URL + token read from `~/.claude/settings.json`, exported, never
printed):
```
probe(anthropic) ok = True
reason           = HTTP 200
token in reason? = False   token len leaked? False
-- credential-less rows make NO request:
   openai (False, 'no openai credential (OPENAI_API_KEY) set')
   gemini (False, 'no gemini credential (GEMINI_API_KEY/GOOGLE_API_KEY) set')
   rits (False, 'no rits credential (RITS_API_KEY) set')
```

Full command transcripts in a follow-up 🔬 Evidence comment.

### DIFF
--- a/core/cap_evolve/cli.py
+++ b/core/cap_evolve/cli.py
@@ -127,6 +127,17 @@ def _cmd_run(argv):
     p.add_argument("--dashboard", choices=("auto", "report-only", "off"), default=None,
                    help="live dashboard: auto (default, launch at run start), report-only, or off")
     p.add_argument("--dashboard-port", type=int, default=None, help="dashboard server port (default 7878)")
+    # Provider/credential flags — the TOP precedence layer (see model_config.PRECEDENCE).
+    p.add_argument("--provider", default=None,
+                   help="provider whose credential to use (e.g. anthropic|openai|gemini|rits|"
+                        "watsonx|mock), or 'auto' to pick the one that actually has a credential")
+    p.add_argument("--provider-base-url", default=None, help="override the provider's base URL")
+    p.add_argument("--provider-credential-env", default=None,
+                   help="env var NAME holding the credential (must belong to the selected "
+                        "provider; cross-provider names are refused)")
+    p.add_argument("--probe-provider", action="store_true",
+                   help="with --provider auto: also probe each candidate's OWN endpoint with "
+                        "its OWN credential before selecting it")
     args = p.parse_args(argv)
 
     skills_dir = Path(args.skills_dir) if args.skills_dir else _find_skills_dir()
@@ -229,8 +240,25 @@ def _cmd_run(argv):
     else:
         sequence = ["intake", "implement-and-check", "baseline", algorithm_name, "finalize", "report"]
 
+    # Provider/credential resolution — decided BEFORE anything is spent. The precedence
+    # (CLI > project spec > user config > built-in) and the provider-scoping rule live in
+    # model_config; here we only surface its secret-free report. `mock`/offline optimizers
+    # need no credential, so a credential is required only when a provider is named.
+    from . import model_config as _mc
+    _cli_layer = {k: v for k, v in (("provider", args.provider),
+                                    ("base_url", args.provider_base_url),
+                                    ("credential_env", args.provider_credential_env)) if v}
+    try:
+        _provider = _mc.resolve(
+            cli=_cli_layer, project=spec,
+            require_credential=bool(_cli_layer.get("provider") or spec.get("provider")),
+            probe_fn=_mc.probe if args.probe_provider else None).to_dict()
+    except _mc.CredentialError as e:
+        _provider = {"error": str(e)}      # message names env VARS, never values
+
     if args.plan_only:
         print(json.dumps({"skills_dir": str(skills_dir), "workdir": str(workdir), "spec": spec,
+                          "provider": _provider,
                           "optimizer": optimizer_name, "optimizer_cmd": opt_cmd,
                           "algorithm": algorithm_name, "focus": algorithm_focus,
                           "target_model": spec.get("target_model", ""),
@@ -244,6 +272,10 @@ def _cmd_run(argv):
                                      "optimizer_max_turns": spec.get("optimizer_max_turns", 0)},
                           "sequence": sequence}, indent=2))
         return 0
+
+    print(json.dumps({"step": "provider", **_provider}))
+    if "error" in _provider:
+        return 1
 
     # Hard gate: cap-evolve check must pass before any budget is spent (intake is the
     # user's job before `run`; here we enforce the check half of implement-and-check).
--- a/core/cap_evolve/cli.py
+++ b/core/cap_evolve/cli.py
@@ -257,7 +257,11 @@ def _cmd_run(argv):
         _provider = {"error": str(e)}      # message names env VARS, never values
 
     if args.plan_only:
-        print(json.dumps({"skills_dir": str(skills_dir), "workdir": str(workdir), "spec": spec,
+        # `spec` is echoed verbatim here, so it carries `provider_base_url` — route the
+        # whole plan through the shared redactor (secrets AND endpoint confidentiality).
+        from .dashboard import redact as _redact
+        print(json.dumps(_redact(
+                         {"skills_dir": str(skills_dir), "workdir": str(workdir), "spec": spec,
                           "provider": _provider,
                           "optimizer": optimizer_name, "optimizer_cmd": opt_cmd,
                           "algorithm": algorithm_name, "focus": algorithm_focus,
@@ -270,7 +274,7 @@ def _cmd_run(argv):
                                      "max_usd": spec.get("max_usd", 0.0),
                                      "max_optimizer_usd": spec.get("max_optimizer_usd", 0.0),
                                      "optimizer_max_turns": spec.get("optimizer_max_turns", 0)},
-                          "sequence": sequence}, indent=2))
+                          "sequence": sequence}), indent=2))
         return 0
 
     print(json.dumps({"step": "provider", **_provider}))
--- a/core/cap_evolve/dashboard.py
+++ b/core/cap_evolve/dashboard.py
@@ -50,6 +50,7 @@ import os
 import re
 import shutil
 import subprocess
+import urllib.parse
 from pathlib import Path
 
 # ---------------------------------------------------------------------------
@@ -90,6 +91,96 @@ _INLINE_KV_RE = re.compile(
 
 _REDACTED = "«redacted»"
 
+# ---------------------------------------------------------------------------
+# Endpoint confidentiality
+# ---------------------------------------------------------------------------
+# Threat model, widened after an internal gateway base URL was printed verbatim into a
+# public PR comment (PR #190). Two things are wrong with treating a URL as harmless
+# diagnostic context:
+#
+# 1. ``https://user:token@host/path`` is a legal URL, so a credential can ride inside
+#    a field nobody classified as secret. Userinfo is stripped UNCONDITIONALLY, from
+#    every URL, public or not — there is no case where it should be rendered or stored.
+# 2. A *custom* base URL is itself confidential. The hostname is precisely what names
+#    internal infrastructure, so showing "the host only" leaks the whole thing; the
+#    path is the least interesting part. We therefore replace the value with
+#    ``<custom>`` and report the *source* of the setting instead, which is what a user
+#    debugging precedence actually needs. Well-known public endpoints (the built-in
+#    defaults) carry no information about anyone's network and stay legible.
+#
+# The rule lives here, in the shared helper, so every consumer (``model_config``'s
+# provider step, the reducer, ``dashboard.html``, and ``doctor``) inherits it.
+
+#: Public endpoints that are safe to print verbatim. Kept as a literal set rather than
+#: a heuristic: "is this host public?" has no safe default answer, so an endpoint is
+#: legible only by being on this list. Mirrors the built-in defaults in
+#: ``model_config.PROVIDERS`` (a test pins the two together).
+PUBLIC_BASE_URLS: frozenset[str] = frozenset({
+    "https://api.anthropic.com",
+    "https://api.openai.com/v1",
+    "https://generativelanguage.googleapis.com/v1beta",
+    "https://api.moonshot.ai/v1",
+    "https://api.githubcopilot.com",
+})
+
+#: What a non-public base URL renders as.
+CUSTOM_URL = "<custom>"
+
+# Credentials also hide in query params (``?api_key=...``); scrubbed alongside userinfo.
+_URL_QUERY_CRED_RE = re.compile(
+    r"([?&](?:key|api[_-]?key|access_token|token|password|sig|signature)=)[^&\s\"']+", re.I)
+
+_URL_IN_TEXT_RE = re.compile(r"\b[a-zA-Z][a-zA-Z0-9+.\-]*://[^\s\"'<>]+")
+
+# Keys whose value is an endpoint, so the non-default-is-secret rule applies to the
+# whole value rather than just its userinfo.
+_URL_KEY_RE = re.compile(r"(base[_\-]?url|_url$|^url$|endpoint|api[_\-]?url|host)", re.I)
+
+
+def strip_url_userinfo(url: str) -> str:
+    """Remove ``user:password@`` (and credential-shaped query params) from ``url``.
+
+    Unconditional and lossy on purpose: userinfo is a credential, so it is dropped
+    rather than masked-in-place. Non-URL strings are returned unchanged.
+    """
+    text = str(url)
+    if "://" not in text:
+        return text
+    try:
+        parts = urllib.parse.urlsplit(text)
+    except ValueError:
+        return _URL_QUERY_CRED_RE.sub(r"\1" + _REDACTED, text)
+    if parts.username or parts.password:
+        host = parts.hostname or ""
+        if parts.port:
+            host = f"{host}:{parts.port}"
+        text = urllib.parse.urlunsplit((parts.scheme, host, parts.path,
+                                        parts.query, parts.fragment))
+    return _URL_QUERY_CRED_RE.sub(r"\1" + _REDACTED, text)
+
+
+def safe_url(url: str, *, public: frozenset[str] | set[str] | None = None) -> str:
+    """Render ``url`` for output: userinfo always stripped, non-public value hidden.
+
+    A URL matching :data:`PUBLIC_BASE_URLS` (ignoring a trailing slash) is returned
+    verbatim so ordinary setups stay debuggable; anything else becomes
+    :data:`CUSTOM_URL`. Empty stays empty.
+    """
+    clean = strip_url_userinfo(url).strip()
+    if not clean:
+        return ""
+    allow = PUBLIC_BASE_URLS if public is None else public
+    return clean if clean.rstrip("/") in {u.rstrip("/") for u in allow} else CUSTOM_URL
+
+
+def safe_urls_in_text(text: str) -> str:
+    """Apply :func:`safe_url` to every URL embedded in free text (probe reasons, errors).
+
+    Needed because an exception message ("HTTPError ... https://host/v1/models") is a
+    plain string, not a field with a URL-shaped key.
+    """
+    return _URL_IN_TEXT_RE.sub(lambda m: safe_url(m.group(0).rstrip(".,;)")), str(text))
+
 
 def _key_is_secret(key: str) -> bool:
     k = str(key)
@@ -97,7 +188,10 @@ def _key_is_secret(key: str) -> bool:
 
 
 def _scrub_value(val: str) -> str:
-    out = _INLINE_KV_RE.sub(lambda m: m.group(1) + _REDACTED, val)
+    # Userinfo first: a credential inside a URL must go before the shape-based passes,
+    # which don't know that `//user:pw@` is a secret. Unconditional — public or not.
+    out = _URL_IN_TEXT_RE.sub(lambda m: strip_url_userinfo(m.group(0)), val)
+    out = _INLINE_KV_RE.sub(lambda m: m.group(1) + _REDACTED, out)
     for rx in _SECRET_VALUE_RES:
         out = rx.sub(_REDACTED, out)
     return out
@@ -107,7 +201,11 @@ def redact(obj):
     """Recursively redact secrets from ``obj`` before it reaches an artifact.
 
     - Dict values under a secret-looking key are replaced wholesale.
-    - String values are scanned for secret-shaped tokens and masked in place.
+    - Dict values under an endpoint-looking key (``base_url``, ``*_url``, ``endpoint``)
+      go through :func:`safe_url`: userinfo stripped, and a non-public endpoint hidden
+      behind :data:`CUSTOM_URL` (see the "Endpoint confidentiality" note above).
+    - String values are scanned for secret-shaped tokens and masked in place, and any
+      URL they embed has its userinfo stripped.
     - Lists/tuples/dicts are walked recursively. Scalars pass through.
 
     Pure, returns a new structure; the caller's object is untouched.
@@ -117,6 +215,8 @@ def redact(obj):
         for k, v in obj.items():
             if _key_is_secret(k) and v not in (None, "", 0):
                 out[k] = _REDACTED
+            elif _URL_KEY_RE.search(str(k)) and isinstance(v, str) and v:
+                out[k] = safe_url(v)
             else:
                 out[k] = redact(v)
         return out
--- a/core/cap_evolve/model_config.py
+++ b/core/cap_evolve/model_config.py
@@ -29,6 +29,14 @@ Secret handling — non-negotiable:
   not a length. Presence/absence only (``credential_present``).
 * A probe sends a credential ONLY to the base URL of the same :data:`PROVIDERS` row
   the credential came from (see :func:`probe`), so a token can't reach a third party.
+* **Endpoints are confidential too.** URL *userinfo* (``https://user:token@host/``) is a
+  credential and is stripped the moment a base URL is resolved, so it can never be
+  rendered or stored. And a base URL that is not one of the well-known public defaults
+  is treated as sensitive: :meth:`Resolved.to_dict` and :func:`describe` report
+  ``base_url: <custom>`` plus ``base_url_source`` (which precedence layer set it)
+  instead of the value, because the hostname is exactly what identifies someone's
+  internal infrastructure. The rule itself lives in ``dashboard.safe_url`` so every
+  consumer — this module, the reducer, ``dashboard.html``, ``doctor`` — inherits it.
 
 Stdlib only, zero runtime deps.
 """
@@ -40,7 +48,7 @@ import re
 from dataclasses import dataclass, field
 from pathlib import Path
 
-from .dashboard import redact
+from .dashboard import CUSTOM_URL, redact, safe_url, safe_urls_in_text, strip_url_userinfo
 from .specfile import read_yaml
 
 # ---------------------------------------------------------------------------
@@ -159,11 +167,18 @@ class Resolved:
     ``credential_env`` is the *name* of the environment variable that carries the
     secret; read it at the point of use. That keeps every repr/log/JSON of this
     object secret-free by construction rather than by remembering to redact.
+
+    ``base_url`` is the real endpoint (needed to make a request) but is
+    ``repr=False``: a custom endpoint is confidential, so it must not appear in a
+    traceback, a log line, or a JSON dump. Use :attr:`base_url_display` /
+    :meth:`to_dict` for anything a human or a file will see. Userinfo has already
+    been stripped by :func:`_resolve_base_url`, so even the in-memory value is
+    credential-free.
     """
 
     provider: str
     credential_env: str = ""          # env var NAME, "" when nothing resolved
-    base_url: str = ""
+    base_url: str = field(default="", repr=False)   # real endpoint; never rendered
     sources: dict = field(default_factory=dict)   # field -> precedence layer
     reason: str = ""                  # human note (why `auto` picked this)
 
@@ -171,12 +186,23 @@ class Resolved:
     def credential_present(self) -> bool:
         return bool(self.credential_env)
 
+    @property
+    def base_url_display(self) -> str:
+        """Safe rendering: public defaults verbatim, anything else ``<custom>``."""
+        return safe_url(self.base_url)
+
+    @property
+    def base_url_source(self) -> str:
+        """Which precedence layer set the base URL — the actionable half of the value."""
+        return self.sources.get("base_url", "built-in default" if self.base_url else "")
+
     def to_dict(self) -> dict:
         return {
             "provider": self.provider,
             "credential_env": self.credential_env,      # a NAME, not a value
             "credential_present": self.credential_present,
-            "base_url": self.base_url,
+            "base_url": self.base_url_display,          # "<custom>" unless well-known
+            "base_url_source": self.base_url_source,
             "sources": dict(self.sources),
             "reason": self.reason,
         }
@@ -315,13 +341,19 @@ def resolve(cli: dict | None = None, project: dict | None = None,
 
 
 def _resolve_base_url(provider: str, values: dict, sources: dict) -> str:
+    """The endpoint to use, with URL userinfo stripped at the single point of resolution.
+
+    Stripping here (rather than at each render site) means no downstream caller can
+    ever hold a base URL carrying ``user:token@`` — the credential is gone before the
+    value is stored on :class:`Resolved`, logged, or handed to :func:`probe`.
+    """
     row = PROVIDERS[provider]
     if values.get("base_url"):
-        return values["base_url"]
+        return strip_url_userinfo(values["base_url"])
     env_name = row["base_url_env"]
     if env_name and os.environ.get(env_name):
         sources["base_url"] = f"environment ({env_name})"
-        return os.environ[env_name]
+        return strip_url_userinfo(os.environ[env_name])
     if row["base_url"]:
         sources["base_url"] = "built-in default"
     return row["base_url"]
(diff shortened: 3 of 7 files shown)


OsherElhadad:
## 🔬 Evidence

All commands run in worktree `/tmp/wt-134` (branch `feat/issue-134-provider-creds`, base `origin/main` 6fca097), interpreter `/tmp/ce-venv/bin/python`. No real credential value appears anywhere below.

### 1. New security tests, verbosely (26/26)

```
$ PYTHONPATH=/tmp/wt-134/core python -m pytest core/tests/test_model_config.py -v
cachedir: .pytest_cache
rootdir: /private/tmp/wt-134/core
configfile: pyproject.toml
plugins: anyio-4.14.2
collecting ... collected 26 items

core/tests/test_model_config.py::test_provider_b_selected_with_only_provider_a_key_errors PASSED [  3%]
core/tests/test_model_config.py::test_explicit_cross_provider_credential_env_is_refused PASSED [  7%]
core/tests/test_model_config.py::test_scoped_lookup_only_reads_own_row PASSED [ 11%]
core/tests/test_model_config.py::test_probe_never_sends_credential_to_another_providers_endpoint PASSED [ 15%]
core/tests/test_model_config.py::test_auto_probe_does_not_leak_credential_across_candidates PASSED [ 19%]
core/tests/test_model_config.py::test_resolved_object_carries_no_credential_value PASSED [ 23%]
core/tests/test_model_config.py::test_probe_failure_reason_is_scrubbed PASSED [ 26%]
core/tests/test_model_config.py::test_cli_run_transcript_never_prints_credential PASSED [ 30%]
core/tests/test_model_config.py::test_no_leak_on_the_nothing_resolves_path PASSED [ 34%]
core/tests/test_model_config.py::test_precedence_cli_over_project_over_user PASSED [ 38%]
core/tests/test_model_config.py::test_empty_layer_does_not_shadow_lower_layer PASSED [ 42%]
core/tests/test_model_config.py::test_user_config_file_is_read PASSED    [ 46%]
core/tests/test_model_config.py::test_base_url_precedence PASSED         [ 50%]
core/tests/test_model_config.py::test_precedence_documented_in_module_and_docs PASSED [ 53%]
core/tests/test_model_config.py::test_auto_picks_the_provider_that_has_a_credential PASSED [ 57%]
core/tests/test_model_config.py::test_auto_prefers_the_documented_order PASSED [ 61%]
core/tests/test_model_config.py::test_auto_skips_a_candidate_whose_probe_fails PASSED [ 65%]
core/tests/test_model_config.py::test_auto_skips_provider_needing_a_base_url_it_lacks PASSED [ 69%]
core/tests/test_model_config.py::test_nothing_resolves_is_actionable PASSED [ 73%]
core/tests/test_model_config.py::test_unknown_provider_errors_clearly PASSED [ 76%]
core/tests/test_model_config.py::test_probe_refuses_unknown_provider_before_any_request PASSED [ 80%]
core/tests/test_model_config.py::test_existing_spec_without_provider_key_infers_from_optimizer PASSED [ 84%]
core/tests/test_model_config.py::test_mock_optimizer_needs_no_credential PASSED [ 88%]
core/tests/test_model_config.py::test_legacy_spec_with_no_provider_and_no_creds_does_not_break PASSED [ 92%]
core/tests/test_model_config.py::test_toy_calc_style_run_still_resolves PASSED [ 96%]
core/tests/test_model_config.py::test_describe_is_a_single_secret_free_line PASSED [100%]

============================== 26 passed in 0.88s ==============================
```

Wrong-endpoint tests: `test_probe_never_sends_credential_to_another_providers_endpoint` (intercepts every `urlopen` — the anthropic probe hits `api.anthropic.com` carrying the token; the openai/gemini/rits/watsonx probes make ZERO requests) and `test_auto_probe_does_not_leak_credential_across_candidates` (fails from inside the fake transport if any request with the token targets a foreign host).

No-leak tests: `test_resolved_object_carries_no_credential_value`, `test_probe_failure_reason_is_scrubbed`, `test_cli_run_transcript_never_prints_credential` (real `cap-evolve run` subprocess with a fake token exported; asserts neither the token nor the fragment `FAKETOKEN` is in stdout+stderr), `test_no_leak_on_the_nothing_resolves_path`, `test_describe_is_a_single_secret_free_line`.

### 2. Full suite — 179 baseline + 26 new = 205 passed, 0 failed

```
$ PYTHONPATH=/tmp/wt-134/core python -m pytest core/tests -q
........................................................................ [ 70%]
.............................................................            [100%]
205 passed in 66.95s (0:01:06)
```

### 3. compileall

```
$ python -m compileall -q core skills && echo COMPILEALL CLEAN
COMPILEALL CLEAN
```

### 4. Real end-to-end, zero API cost — `examples/toy_calc` + `mock` optimizer

```
$ D=/tmp/wt134-e2e-a; mkdir -p $D/.capevolve/project/adapters
$ cp examples/toy_calc/adapter.py $D/.capevolve/project/adapters/
$ cp -R examples/toy_calc/capability $D/seed_capability
$ cp templates/project/capevolve.yaml $D/.capevolve/project/capevolve.yaml   # provider: "" (unset)
$ python -m cap_evolve.cli run --spec $D/.capevolve/project/capevolve.yaml --project $D/.capevolve/project --run-ts demo
{"step": "provider", "provider": "mock", "credential_env": "", "credential_present": false, "base_url": "", "sources": {"provider": "built-in default (from optimizer_skill)"}, "reason": "provider 'mock' from built-in default (from optimizer_skill)"}
{
  "best_id": "cand_0001",
  "baseline_val": 0.0,
  "test_reward": 1.0,
  "test_baseline_reward": 0.0,
  "test_delta": 1.0,
  "test_pass_k": {
    "1": 1.0,
    "2": 0.0
  },
  "iterations": 3,
  "dashboard_server": "http://127.0.0.1:7879"
}

```

### 5. Same e2e with `provider: auto` and a FAKE token exported — full artifact leak scan

```
$ export ANTHROPIC_API_KEY="sk-FAKELEAKPROBE0123456789abcdef" OPENAI_API_KEY="sk-FAKELEAKPROBE0123456789abcdef"
$ sed 's/^provider: ""/provider: auto/' templates/project/capevolve.yaml > $D/.capevolve/project/capevolve.yaml
$ python -m cap_evolve.cli run --spec ... --project ... --run-ts demo > out.txt 2> err.txt; echo exit=$?
exit=0
{"dashboard": "http://127.0.0.1:7880"}
{"step": "provider", "provider": "anthropic", "credential_env": "ANTHROPIC_API_KEY", "credential_present": true, "base_url": "https://api.anthropic.com", "sources": {"provider": "auto probe", "base_url": "built-in default"}, "reason": "auto selected 'anthropic': ANTHROPIC_API_KEY is set and it is the highest-priority provider with its own credential; credential present (no probe requested)."}
{
  "best_id": "cand_0001",
  "baseline_val": 0.0,
  "test_reward": 1.0,
  "test_baseline_reward": 0.0,
  "test_delta": 1.0,
  "test_pass_k": {
    "1": 1.0,
    "2": 0.0
  },
  "iterations": 3,
  "dashboard_server": "http://127.0.0.1:7881"
}

--- stderr (empty) ---
$ grep -c "FAKELEAKPROBE" out.txt err.txt
err.txt:0
out.txt:0
$ find $D -name events.jsonl -o -name dashboard.html | wc -l   # artifacts existed to scan
2
$ grep -rl "FAKELEAKPROBE" $D | wc -l   # every artifact: events.jsonl, dashboard.html, candidates, rollouts
0
```

### 6. Failure path — `--provider openai` while only an anthropic credential is present

```
$ unset OPENAI_API_KEY   # ANTHROPIC_API_KEY still set to the fake token
$ python -m cap_evolve.cli run --spec ... --project ... --provider openai; echo exit=$?
{"step": "provider", "error": "no credential found for provider 'openai' (selected via CLI flag). Set one of: OPENAI_API_KEY. Credentials for anthropic ARE present but belong to a different provider and are never reused \u2014 either export a openai credential, or select that provider explicitly (provider: anthropic) / use provider: auto. Precedence: CLI flag > project capevolve.yaml > user config > built-in default."}
exit=1

$ # cross-provider credential_env, refused rather than honoured:
$ python -m cap_evolve.cli run ... --provider openai --provider-credential-env ANTHROPIC_API_KEY
{"step": "provider", "error": "provider_credential_env='ANTHROPIC_API_KEY' belongs to provider 'anthropic', but the selected provider is 'openai'. Credentials are provider-scoped and are never reused across providers. Set one of: OPENAI_API_KEY."}
exit=1

$ # nothing resolves at all:
$ env -i ... python -m cap_evolve.cli run ... --provider auto
{"step": "provider", "error": "provider: auto found no usable provider \u2014 none of the known providers has its own credential set. Export one of: anthropic: ANTHROPIC_API_KEY/ANTHROPIC_AUTH_TOKEN; openai: OPENAI_API_KEY; gemini: GEMINI_API_KEY/GOOGLE_API_KEY; rits: RITS_API_KEY; watsonx: WATSONX_APIKEY/WATSONX_API_KEY; moonshot: MOONSHOT_API_KEY/KIMI_API_KEY; github-copilot: COPILOT_GITHUB_TOKEN/GH_TOKEN/GITHUB_TOKEN; bob: BOBSHELL_API_KEY/BOB_API_KEY; cursor: CURSOR_API_KEY; factory: FACTORY_API_KEY. Precedence: CLI flag > project capevolve.yaml > user config > built-in default."}
exit=1
```

### 7. One live probe (base URL + token read from `~/.claude/settings.json`, exported, never printed)

```
$ python _probe_live.py     # reads ONLY env.ANTHROPIC_BASE_URL + env.ANTHROPIC_AUTH_TOKEN
probe(anthropic) ok = True
reason              = HTTP 200
base_url used       = <REDACTED — internal gateway base URL>
token value in reason? False | token length leaked? False
-- rows with no credential of their own make NO request at all:
    openai (False, 'no openai credential (OPENAI_API_KEY) set')
    gemini (False, 'no gemini credential (GEMINI_API_KEY/GOOGLE_API_KEY) set')
    rits (False, 'no rits credential (RITS_API_KEY) set')
    watsonx (False, 'no watsonx credential (WATSONX_APIKEY/WATSONX_API_KEY) set')
```

A single real request, to the anthropic row's own base URL with the anthropic row's own credential. The four credential-less rows return their refusal without constructing a request — the wrong-endpoint guarantee at runtime.



---

> **Edited by the orchestrator:** an internal gateway base URL was redacted from the live-probe evidence above. Probe endpoints belong in local logs, not a public PR comment. See the follow-up note on this PR.

OsherElhadad:
## ⚠️ Orchestrator note — redaction applied, and a gap this PR should close

I edited the `## 🔬 Evidence` comment above to redact an **internal gateway base URL** that the live-probe section printed verbatim into this public PR. No API token was exposed — the credential-scrubbing this PR implements did its job on the *secret*. But the base URL was printed as ordinary diagnostic context, and it named internal infrastructure.

I've verified the string appears **nowhere else** in this epic (scanned the bodies and all comments of #127 and every PR #181–#195: zero hits).

### The point that matters for this PR's design
This is exactly the class of leak `model_config.py` should prevent, and currently doesn't. The threat model treats the **credential value** as the only secret, so `base_url` is carried and rendered in the clear — visible in the `{"step": "provider", ...}` line, in `Resolved`, and in probe diagnostics. That's fine for `https://api.anthropic.com`. It is not fine for a private endpoint, which is precisely what `provider_base_url` exists to configure.

Two further reasons it's load-bearing here, not cosmetic:
1. **A base URL can itself contain a credential.** `https://user:token@host/...` is a legal URL, and userinfo would flow straight through a code path that assumes URLs are non-sensitive.
2. **These lines are written to run artifacts**, not just the terminal — `events.jsonl` and `dashboard.html` are committed or shared, so an internal hostname propagates into evidence directories and PR attachments by default.

### Requested change
- Strip userinfo from any URL before it is rendered or stored, unconditionally.
- Treat a **non-default** `base_url` as sensitive by default: show the provider and the *source* of the setting (`"base_url": "<custom>"` / `from user config`) rather than the value. Keep well-known public endpoints legible.
- Extend the existing no-leak tests to assert a custom base URL (and a URL with embedded userinfo) does not appear in stdout/stderr, `events.jsonl`, `dashboard.html`, or a probe failure reason — the current canary only tracks the token.
- `docs/TROUBLESHOOTING.md` should say plainly that run artifacts may carry endpoint config, so users know what they're sharing.

The `dashboard.redact` helper this PR already reuses is the natural place for the URL rule, so #121 (`doctor`, which also reports credential presence) inherits it.

Not a criticism of the security work here, which is otherwise the most thorough in this batch — the wrong-endpoint interception tests are genuinely good. It's a threat-model boundary that was drawn one notch too narrow, found by tripping over it.

OsherElhadad:
## 🔒 Endpoint-confidentiality hardening

Closes the gap in the orchestrator note. Commit `5ed91ee`. Stdlib only (`urllib.parse`), no new deps, no existing guarantee weakened, `harness.py` / `specfile.py` untouched.

### Updated threat model

Before: the credential **value** was the only secret. `base_url` was diagnostic context, printed and stored in the clear.

Now two more things are sensitive:

| Thing | Why | Rule |
|---|---|---|
| URL **userinfo** (`https://user:token@host/`) | It *is* a credential, hiding in a field nobody classified as secret — so it bypassed every token-shaped scrub in this PR | Stripped **unconditionally**, public endpoint or not |
| A **non-default** `base_url` | The hostname names someone's internal infrastructure. `provider_base_url` exists precisely to point at a private gateway | Value replaced with `<custom>`; the *source* is reported instead |

Deliberately **not** sensitive: the built-in public defaults (`https://api.anthropic.com`, `https://api.openai.com/v1`, …). They identify nobody, and over-redacting them would make the provider step useless for the 95% case.

### The userinfo rule

`dashboard.strip_url_userinfo()` — stdlib `urllib.parse.urlsplit` / `urlunsplit`, drops `username`/`password` (and masks credential-shaped query params: `?api_key=`, `?token=`, `?sig=`). Applied at the **single point of resolution**, `model_config._resolve_base_url`, so no downstream caller ever holds a userinfo-bearing URL — not `Resolved`, not `probe`, not a traceback. The credential is gone from the value we *keep*, not merely hidden at render time. It also runs inside `redact()` on every free-text string, so a URL echoed by an exception message loses its userinfo too.

### Default vs custom policy — and is the bare hostname safe?

**No, the bare hostname is not safe, and it's the least safe part.** I considered showing `host` and hiding only the path. Rejected:

1. The hostname is the *whole* secret. `gateway.internal.acme-corp.example` discloses that the org exists, that it runs a private LLM gateway, and its naming convention — enough to fingerprint and to target. The path (`/v1`) is boilerplate that discloses nothing.
2. It's the reconnaissance-useful half. An internal hostname is a DNS/scan target and, on a VPN-reachable network, an attack surface. A path is not.
3. "Is this host public?" has no safe default answer, so I did **not** write a heuristic (no "looks like a corporate TLD" guessing). An endpoint is legible **only** by being on an explicit allowlist, `dashboard.PUBLIC_BASE_URLS`, which mirrors the `PROVIDERS` built-in defaults — and a test pins the two together so they can't drift.

So: `"base_url": "<custom>"`, `"base_url_source": "user config"`. The source is the actionable half — someone debugging precedence needs to know *which layer* won, not what the string is; they can read their own env var on their own machine. `Resolved.base_url` is additionally `repr=False`, so a traceback or a stray `print(res)` can't render it either.

`--plan-only` also echoes the raw spec (which *holds* `provider_base_url`), so that now goes through the same redactor.

### `doctor` (#193) inherits it for free — confirmed

I read `origin/feat/issue-121-doctor` and did **not** edit any of its files. `doctor.py:39` is `from .dashboard import redact`, and it calls `redact(rep.to_dict())` at both output sites (`format_report`, line 319, and `--json`, line 345). Because the URL rule lives in `redact` rather than in `model_config`, `doctor` gets it with zero lines of its own. Verified by copying its file in unmodified, exercising it, and removing it:

```
doctor --json (redacted):
{
  "checks": [
    {
      "name": "credentials",
      "detail": "gateway probe of https://fake-internal-canary.example.invalid/v1 timed out",
      "present": ["ANTHROPIC_API_KEY", "ANTHROPIC_BASE_URL"],
      "base_url": "<custom>"
    }
  ]
}
hostname leaked in the url-keyed field?  False
userinfo token leaked anywhere?          False
```

(Its `_RUNNER_ENV_GROUPS` already lists `ANTHROPIC_BASE_URL` as a **name** only and reports presence as `set (hidden)`, so it never reads the value — the shared rule covers the case where it later adds a url-keyed field or bakes an endpoint into a `detail` string.)

---

## Verification

### 1. Suite

```
$ PYTHONPATH=/tmp/fx-190/core python -m pytest core/tests -q
........................................................................ [ 33%]
........................................................................ [ 67%]
....................................................................     [100%]
212 passed in 63.87s (0:01:03)
```

212 passed, **0 failed** (was 205 claimed; baseline on this branch before my change measured 204 passed + 1 pre-existing environment-dependent failure in `test_dashboard_launch.py::test_maybe_launch_spawns_when_available`, which passes in the full-suite run — +7 net are the new endpoint tests).

```
compileall: clean
```

### 2. The money evidence — real zero-API `examples/toy_calc` + `mock` run

Configured with a **userinfo-bearing custom base URL** in the spec layer *and* the env layer. Fake canary values throughout; **no real gateway URL was used as a test value anywhere, and no live probe was run** (deliberately — the whole point is to stop internal endpoints reaching output, and nothing here needs a real request; the wrong-endpoint guarantees are covered by the mocked-`urlopen` tests below).

```
$ echo 'provider: mock'                                                                      >> $D/.capevolve/project/capevolve.yaml
$ echo 'provider_base_url: "https://u:CANARYTOKEN123456@fake-internal-canary.example.invalid/v1"' >> $D/.capevolve/project/capevolve.yaml
$ export ANTHROPIC_BASE_URL="https://u:CANARYTOKEN123456@fake-internal-canary.example.invalid/v1"
$ export ANTHROPIC_API_KEY="sk-FAKELEAKPROBE0123456789abcdef"
$ python -m cap_evolve.cli run --spec ... --project ... --dashboard off --run-ts demo > out.txt 2> err.txt
exit=0

$ grep '"step": "provider"' out.txt
{"step": "provider", "provider": "mock", "credential_env": "", "credential_present": false, "base_url": "<custom>", "base_url_source": "project capevolve.yaml", "sources": {"provider": "project capevolve.yaml", "base_url": "project capevolve.yaml"}, "reason": "provider 'mock' from project capevolve.yaml"}
```

Artifacts existed to scan (otherwise the greps prove nothing):

```
$ find $D/.capevolve/run_demo -type f | wc -l
     145
$ find $D/.capevolve/run_demo -name events.jsonl -o -name dashboard.html
.../.capevolve/run_demo/dashboard.html
.../.capevolve/run_demo/events.jsonl
```

**Zero hits, custom hostname:**

```
$ grep -rn "fake-internal-canary" $D/.capevolve/run_demo out.txt err.txt; echo "exit=$?"
exit=1 (1 = zero hits)
```

**Zero hits, embedded userinfo token:**

```
$ grep -rn "CANARYTOKEN123456" $D/.capevolve/run_demo out.txt err.txt; echo "exit=$?"
exit=1 (1 = zero hits)
```

**Zero hits, the API token (the original guarantee, unweakened):**

```
$ grep -rn "FAKELEAKPROBE" $D/.capevolve/run_demo out.txt err.txt; echo "exit=$?"
exit=1 (1 = zero hits)
```

Per-file counts on the two artifacts that get shared, plus stdout/stderr:

```
$ grep -rc "fake-internal-canary" $D/.capevolve/run_demo/events.jsonl $D/.capevolve/run_demo/dashboard.html out.txt err.txt
out.txt:0
.../run_demo/events.jsonl:0
err.txt:0
.../run_demo/dashboard.html:0
```

And the **entire** working dir, excluding only the config file I wrote by hand as the input:

```
$ grep -rl "fake-internal-canary" $D | grep -v 'project/capevolve.yaml'; echo "exit=$?"
exit=1 (1 = only our own input config)
```

This is pinned as a test (`test_full_run_artifacts_never_contain_a_custom_endpoint_or_its_userinfo`) which spawns a real `cap-evolve run` and walks `run_dir.rglob("*")`, asserting the artifacts it needs actually exist before scanning them. It also covers `--plan-only`.

**Proof the leak was real** — the same call on the parent commit, before the fix:

```
$ git stash && python -c 'resolve(provider_base_url="https://u:CANARYTOKEN123456@fake-internal-canary.example.invalid/v1")'
BEFORE this fix, to_dict() emitted:
{'provider': 'mock', ..., 'base_url': 'https://u:CANARYTOKEN123456@fake-internal-canary.example.invalid/v1', ...}
```

Hostname **and** embedded credential, verbatim, into `{"step": "provider", ...}` and from there into `events.jsonl`.

### 3. Public defaults still legible (not over-redacted)

```
$ for name, row in PROVIDERS: print(name, row['base_url'], '->', safe_url(row['base_url']))
  anthropic        https://api.anthropic.com                            -> https://api.anthropic.com
  openai           https://api.openai.com/v1                            -> https://api.openai.com/v1
  gemini           https://generativelanguage.googleapis.com/v1beta     -> https://generativelanguage.googleapis.com/v1beta
  moonshot         https://api.moonshot.ai/v1                           -> https://api.moonshot.ai/v1
  github-copilot   https://api.githubcopilot.com                        -> https://api.githubcopilot.com

$ describe(resolve(provider="anthropic"))
provider: anthropic | credential: ANTHROPIC_API_KEY (present) | base_url: https://api.anthropic.com | provider 'anthropic' from project capevolve.yaml
```

Every built-in default renders verbatim — pinned by `test_every_builtin_default_base_url_is_on_the_public_allowlist`, which fails if the allowlist and `PROVIDERS` ever drift.

Custom, side by side, showing the source survives and `repr` is clean:

```
to_dict           : {'provider': 'anthropic', 'credential_env': 'ANTHROPIC_API_KEY', 'credential_present': True, 'base_url': '<custom>', 'base_url_source': 'CLI flag', 'sources': {...'base_url': 'CLI flag'}, ...}
repr(Resolved)    : Resolved(provider='anthropic', credential_env='ANTHROPIC_API_KEY', sources={...}, reason="...")
describe()        : provider: anthropic | credential: ANTHROPIC_API_KEY (present) | base_url: <custom> (from CLI flag) | ...
userinfo survives in .base_url?          False
.base_url actually used for requests:    https://fake-internal-canary.example.invalid/v1
```

Last line is the point of stripping at resolution: the value used to *make the request* is credential-free, and `repr` omits it entirely.

### 4. Probe still works, still refuses cross-provider endpoints

The pre-existing wrong-endpoint tests, re-run unchanged, plus the new endpoint canaries:

```
$ pytest core/tests/test_model_config.py -v -k "endpoint or cross or foreign or probe or leak or scoped"
test_explicit_cross_provider_credential_env_is_refused                       PASSED
test_scoped_lookup_only_reads_own_row                                        PASSED
test_probe_never_sends_credential_to_another_providers_endpoint              PASSED
test_auto_probe_does_not_leak_credential_across_candidates                   PASSED
test_probe_failure_reason_is_scrubbed                                        PASSED
test_no_leak_on_the_nothing_resolves_path                                    PASSED
test_probe_failure_reason_hides_a_custom_endpoint                            PASSED
test_redact_helper_hides_endpoints_so_doctor_inherits_the_rule               PASSED
test_full_run_artifacts_never_contain_a_custom_endpoint_or_its_userinfo      PASSED
test_auto_skips_a_candidate_whose_probe_fails                                PASSED
test_probe_refuses_unknown_provider_before_any_request                       PASSED
11 passed
```

One existing assertion changed, in the strictly-stronger direction: `test_probe_failure_reason_is_scrubbed` used to assert `"redacted" in why` for a URL carrying `?api_key=`. The whole non-public URL is now replaced, so the reason reads `URLError: <urlopen error failed connecting to <custom>` — the host, the path *and* the credential param all go at once. The test now asserts `CUSTOM_URL in why and "x.invalid" not in why and "api_key" not in why`.

### 5. Docs

`docs/TROUBLESHOOTING.md` gains a `base_url: <custom>` entry (explaining why your endpoint isn't shown, and that you should read your own env var locally) and a **"What a run directory contains before you share it"** section: credentials and custom endpoint URLs are redacted, but a run dir still records *that* a custom base URL was in use and which layer supplied it — and it carries task data, prompts, capability source and model outputs entirely unredacted, because those are the point of the artifact. Skim `events.jsonl` before attaching. A custom hostname appearing in one is a bug worth reporting.

### Files touched

| File | Change |
|---|---|
| `core/cap_evolve/dashboard.py` | `strip_url_userinfo`, `safe_url`, `safe_urls_in_text`, `PUBLIC_BASE_URLS`, `CUSTOM_URL`; `redact` now hides endpoint-keyed values and strips userinfo from URLs in free text |
| `core/cap_evolve/model_config.py` | strip userinfo in `_resolve_base_url`; `base_url` → `repr=False`; `base_url_display` / `base_url_source`; `to_dict` + `describe` + `_safe_reason` use the shared rule |
| `core/cap_evolve/cli.py` | `--plan-only` echoes the raw spec → routed through `redact` |
| `core/tests/test_model_config.py` | +7 endpoint no-leak tests incl. the real-run artifact scan; one assertion strengthened |
| `docs/TROUBLESHOOTING.md` | `<custom>` entry + run-artifact disclosure section |

Issue #144, opened 2026-07-25
Terminal/TUI robustness: no-TTY/CI degradation ladder + crash/forensic log


OsherElhadad:
## Problem / opportunity

Once cap-evolve gains a live terminal view (see **#116** `--follow`), it needs to render correctly across the messy reality of terminals — CI (no TTY), Windows, non-UTF-8 locales, resizes — and fail *legibly* when it crashes. Today there's no live TUI at all and no forensic path when the CLI dies (a failure is diagnosed by hand-reading JSON files).

## Why it matters

Prior art in this space (Arbor, Apache-2.0 — inspiration only) invests in TUI robustness that makes the tool feel solid: a **degradation ladder** (rich interactive view → plain Rich → line mode) auto-selected for no-TTY/CI/Windows/known-bad terminals; **layout height-budgeting** so a live view never overflows or duplicates frames on resize; and a **crash log** (with a rolling recent-input trace) written to a known path so "it just exited" becomes a filable bug instead of a mystery. These are the details that separate a demo TUI from a dependable one.

## What to change

1. When adding the live view (#116), implement a **degradation ladder**: full live view on a capable TTY; a plain periodic-print mode for CI/no-TTY/`--no-live`; never emit control sequences into a non-TTY (so CI logs stay clean).
2. **Height-budget** the live layout so it never overflows or double-renders on terminal resize.
3. On an unhandled crash/interrupt, write a **forensic log** (traceback + last events) to a known location (e.g. under the run dir or a `~/.cache/cap-evolve/crashes/` path) and print one line pointing to it.
4. Honor `NO_COLOR` and detect width (the existing report already does some of this in `dashboard.py:_term_width` — reuse it).

## How to test

- Running with output piped to a file (no TTY) produces clean, control-sequence-free line output, not a garbled live view.
- Simulated resize doesn't overflow/duplicate the layout.
- An injected crash writes a forensic log and prints its path; exit code is non-zero.

## Dependencies / parallelization

- **Depends on / co-develop with #116** (`--follow`/live view) — this is the robustness half of that feature. Complements **#121** (`doctor`). **Parallel-safe** with docs/dashboard-web issues.

OsherElhadad:
[...]

Issue #138, opened 2026-07-25
Dashboard phase-pipeline + evidence header (sparkline + live token burn)


OsherElhadad:
## Problem / opportunity

The dashboard shows candidates, costs, and a best-curve, but there's no at-a-glance answer to the four questions a user has *during* a run: **what stage are we in, what's happening right now, is it improving, and what's the honest evidence so far.** The stage/phase, a "now" focal line, a best-score sparkline, a live token-burn rate, and an evidence summary (baseline vs val vs sealed-test vs best + delta + metric direction) aren't presented as a compact header.

## Why it matters

Prior art in this space (Arbor, Apache-2.0 — inspiration only) puts exactly this in a compact header: a **phase pipeline with the current stage lit**, a **`now …` line** with time-in-state, a **best-score sparkline** with ↑/↓ metric direction, a **live token-burn meter**, and an **evidence line** (baseline / val / sealed-test / best + computed improvement). It makes the honesty story *visible* rather than buried across tabs, and answers "is this working / what's it costing" without navigation. cap-evolve emits all the underlying events already (`baseline`, `evaluate`, `step`, `gepa_*`, `budget_warning`, `finalize`) and tracks `run_dir.spent` — this is presentation, not new data.

## What to change

1. Add a compact **header/status strip** to the dashboard DeepDive (and the self-contained HTML): current phase (with the pipeline stages and the active one lit), a `now`/time-in-state line, a best-score sparkline with metric-direction arrow, and a live token/cost burn readout.
2. Add an **evidence line**: metric direction (higher/lower better), baseline, best-on-val, sealed test (once finalized), and the computed delta — the honest-eval story, always visible.
3. Reuse the live stream (from #117) so the header updates in near-real-time.

## How to test

- Frontend test: given a mocked run + stream, the header renders the active phase, sparkline, evidence line, and burn readout.
- The evidence line matches `baseline.json`/`final.json` values and updates on new events.
- Reduced-motion honored (no forced sparkline animation).

## Dependencies / parallelization

- Builds on **#117** (live stream rendering) and benefits from **#118** (stall state in the header) and **#119** (perf so the header updates cheaply). Dashboard-only. **Parallel-safe** with backend/core issues.

OsherElhadad:
[...]

Issue #122, opened 2026-07-25
Demo-first onboarding: shareable, scrubbable run-replay artifact ("watch it before you configure")


OsherElhadad:
## Problem

cap-evolve has no **"see it work before you configure anything"** on-ramp, and no shareable, self-narrating artifact of *what happened during a run over time*. Today: the free proof is `toy_calc` (great, but you still clone + venv + pip install first), and the self-contained `dashboard.html` shows the *final* reduced state, not an animated play-by-play of the search.

## Why it matters

The single biggest adoption barrier for a compute/credential-heavy tool is "I have to set it all up before I can tell if it's worth it." Prior art in this space (Arbor, Apache-2.0) demonstrates two high-leverage moves we can implement independently and cleanly (technique inspiration only — no code/assets copied):
1. a **zero-setup replay** of a recorded run in the live view (`replay --demo`), and
2. a **scrubbable, animated run artifact** — a self-contained HTML page that animates candidates being proposed → accepted/pruned → best-so-far over the run's timeline, with play/pause/scrub/speed controls.

cap-evolve already has everything needed: `events.jsonl` is an append-only, timestamped event log, and there's already a static-export bridge (`export_static.py`) + a self-contained HTML template. This is about adding a *time dimension* and a *no-clone entry point*.

## What to change

1. **`cap-evolve replay <run>` (and `--demo`)**: re-emit a recorded `events.jsonl` on a virtual clock into the existing dashboard/SSE pipeline (with a max-gap clamp so a long eval pause doesn't stall playback). Ship one bundled recorded run so `replay --demo` works with zero config.
2. **Animated shareable export**: extend `export_static.py` / the self-contained HTML so the exported page can *animate* the run over its virtual timeline (candidates appearing, accept/reject/prune status, best-so-far frontier advancing) with play/pause/scrub/speed — a single self-contained file linkable in issues/PRs/README.
3. Link the animated demo from the README/site "try it" section as the first thing a visitor sees.

## How to test

- `cap-evolve replay --demo` opens the dashboard and plays the bundled run end-to-end with no project configured.
- The animated HTML export opens with no server, no network, and scrubbing to time T shows the correct set of candidates/best-so-far for T.
- Reduced-motion is honored (no forced animation when the OS requests reduced motion).

## Dependencies / parallelization

[...]

Issue #118, opened 2026-07-25
Distinguish a stalled/hung run from idle/done (SSE 5-min idle + coarse status heuristic)


OsherElhadad:
## Problem

A **stalled/hung run masquerades as finished** in the dashboard, and the hub status heuristic is coarse enough to leave crashed runs "live" forever.

- The SSE stream closes after ~5 minutes idle (`dashboard/backend/capevolve_dashboard/app.py:116`). A hung optimizer (no new events) makes the dashboard flip to "idle" and stop — which reads like *completion*, not a *hang*.
- The `_status` heuristic (`runs.py:43-51`) is coarse: a run with zero candidates is `failed`, otherwise `live` until test is sealed. A run that produced 1 candidate then crashed shows as `live` forever on the hub.

## Why it matters

During a long, expensive run the most important question is "is it still making progress, or is it stuck / dead?" Today the UI can't distinguish stalled from done from crashed, so a user may either wait forever on a dead run or assume a hung run finished successfully.

## What to change

1. Introduce an explicit **"stalled / no progress"** state: if no event has been appended to `events.jsonl` for N minutes (configurable) and the run isn't finalized, mark it `stalled` (distinct from `live` and `done`) in `_status` and show it distinctly in the Hub and DeepDive.
2. Keep the SSE connection alive with a heartbeat/keepalive comment rather than silently closing at 5 min, or on close send a "stalled — no events for Xm" frame instead of an ambiguous idle close.
3. Detect crashed runs: if the process/heartbeat is gone and the run isn't finalized, mark `crashed`/`incomplete`, not `live`.

## How to test

- Backend test: a fixture run dir whose `events.jsonl` mtime is older than the stall threshold and has no `finalize` event → `_status` returns `stalled` (not `live`/`done`).
- A finalized run still reports `done`; a zero-candidate run still reports `failed`.
- Manually: kill a run mid-loop; the Hub shows `stalled`/`crashed`, not `live` or a clean finish.

## Dependencies / parallelization

- Complements the terminal `--follow` issue (same "is it hung?" question, different surface). **Parallel-safe.** Consider a shared heartbeat concept: emitting a periodic `heartbeat` event during long optimizer/eval calls would make stall detection trivial for both the terminal and the dashboard.

OsherElhadad:
[...]

PR #191, opened 2026-07-29
feat(observability): live terminal progress via `--follow` and `cap-evolve tail`

### LINKED ISSUE #116: Classic cap-evolve run is silent for its entire duration — add --follow / cap-evolve tail
## Problem (flagship observability issue)

A classic `cap-evolve run` is **completely silent in the terminal for its entire duration.** A multi-hour hill-climb / GEPA run shows a blank terminal until a phase returns, then prints one JSON blob.

- The core prints nothing: `grep -c 'print(' loop.py harness.py gepa.py skillopt.py` → all `0`. Only `cli.py` prints, and only `print(json.dumps(...))` at phase boundaries/errors.
- `cli.py:219-220` runs every phase with `subprocess.run(cmd, capture_output=True, ...)`, so baseline / the algorithm loop / finalize / report all have stdout+stderr swallowed until they return (`cli.py:427` prints the final blob).
- No progress bar, no `logging`, no `--verbose`, no ETA, no per-iteration line.
- The rich ANSI KPI strip + cumulative-best chart (`dashboard.py:687-784`) is **post-run only** (report phase).

So the only live observability is the web dashboard. Terminal-only users — CI, ssh, headless, air-gapped — get nothing, and a **hung run is indistinguishable from a working one.**

## Why it matters

"Watch capability evolve" is the tagline, yet the default terminal experience shows nothing evolving. Prior art in this space (e.g. Arbor, Apache-2.0) demonstrates the value of a live terminal view: a phase/stage line, best-score-so-far, per-candidate accept/reject, and a live token/cost meter — the four questions a user has during a long run (*what stage / what's happening / is it improving / what's it costing*). We can implement the same *idea* cleanly from our own `events.jsonl` (no external code — technique inspiration only).

## What to change

1. Add **`cap-evolve run --follow`** (and/or a standalone **`cap-evolve tail <run>`**) that tails `events.jsonl` as it's written and prints human-readable lines: stage transitions, `baseline`, each `step`/`gepa_val_gate`/`skillopt_step` with candidate id + val + accept/reject + reason, `budget_warning`, `optimizer_error`, `finalize`. This is the terminal equivalent of the existing SSE endpoint — the byte-offset tail logic already exists in `dashboard/backend/.../stream.py` and can be shared.
2. Optionally stream phase subprocess stderr (tee, or run without `capture_output` when `--verbose`) so a hung baseline is visible.
3. Add a live token/cost line from `run_dir.spent` (already accumulates continuously).
4. Keep default output quiet-ish; `--follow`/`--verbose` opts into the stream.

## How to test

- New test: run `toy_calc` with `--follow`; assert the terminal receives at least one per-candidate line and a finalize line (capture stdout).
- Manually: a long run now prints progress; killing the optimizer mid-run shows the last event, not a frozen blank screen.
- Default (no `--follow`) behavior unchanged for scripts that parse the final JSON.

- **Shares the `events.jsonl` tail logic** with the dashboard SSE (`stream.py`) — factor it into the core so both use it. **Parallel-safe** with the SPA-side observability issues (different surfaces). Highest user-visible ROI in Group D — recommend leading with it.

OsherElhadad:
**Closes #116**

## Before → After

**Before.** A classic `cap-evolve run` is silent for its entire duration. Every phase runs under `subprocess.run(..., capture_output=True)`, so baseline / the algorithm loop / finalize / report all have stdout+stderr swallowed until they return — a multi-hour hill-climb shows a blank terminal, then one JSON blob. `grep -c 'print(' loop.py harness.py gepa.py skillopt.py` → all `0`. A hung run is indistinguishable from a working one, and the only live observability is the web dashboard (nothing for CI / ssh / headless / air-gapped).

**After.** Real `cap-evolve run --follow` over `examples/toy_calc` with the `mock` optimizer (zero API cost), captured verbatim:

```text
[00:26:43] splits frozen  train=4 val=2 test=2 (test sealed)
[00:26:43] eval val/seed  reward=0.0000 ±0.0000  0.0s
[00:26:43] baseline  val=0.0000 ±0.0000
[00:26:44] eval val/cand_0001  reward=1.0000 ±0.0000  0.0s
[00:26:44] gate warning (paired): combined/paired SE is 0 (likely n_trials=1 or identical trials)
[00:26:44] ACCEPT  cand_0001  val=1.0000 (parent 0.0000)  — paired Δ̄=+1.0000 > 0 (SE=0 → STRICT fallback, warned; n=2)
[00:26:45] eval val/cand_0002  reward=1.0000 ±0.0000  0.0s
[00:26:45] reject  cand_0002  val=1.0000 (parent 1.0000)  — paired Δ̄=+0.0000 <= 0 (SE=0 → STRICT fallback, warned; n=2)
[00:26:45] eval val/cand_0003  reward=1.0000 ±0.0000  0.0s
[00:26:45] reject  cand_0003  val=1.0000 (parent 1.0000)  — paired Δ̄=+0.0000 <= 0 (SE=0 → STRICT fallback, warned; n=2)
[00:26:46] eval test/FINAL  reward=1.0000 ±0.0000  0.0s
[00:26:46] eval test/FINAL_seed  reward=0.0000 ±0.0000  0.0s
[00:26:46] FINALIZE  test=1.0000 (baseline 0.0000, Δ+1.0000)  best=cand_0001
```

On a real TTY these are colored (cyan baseline / green ACCEPT / dim reject / bold FINALIZE); piped or in CI they are plain text. A live `[$0.0123 · 4.2k tok]` cost meter is appended once a run actually spends.

## What changed

- **`cap-evolve run --follow`** — live progress on a daemon thread started *before* baseline creates the run dir, so the first events (`splits`, `baseline`) are never missed. Progress goes to **stderr**, so stdout stays the machine-readable final JSON that scripts parse (`cap-evolve run --follow > result.json` still works). The follower can never raise into the run.
- **`cap-evolve tail [run_dir]`** — attaches to an existing or ongoing run (default: newest `run_*` under `--base`). It *waits* for the run dir to appear, so you can attach before the run creates it. `--from-start` replays history; `--idle-timeout` bounds the wait.
- **Default behavior unchanged**: without `--follow`, `cap-evolve run` writes 0 bytes to stderr and the same JSON to stdout as before.

## Shared tail helper (for #118, #122, #138, #144)

The byte-offset tail is now **one** module in core, stdlib only, no new deps. The dashboard's SSE route (`dashboard/backend/capevolve_dashboard/stream.py`) imports from it instead of owning a copy, so the terminal and the web view read the same typed event stream and can never disagree.

**File:** `core/cap_evolve/eventstream.py`

```python
read_new_events(path: Path, offset: int) -> tuple[list[dict], int]
# One incremental read. Seeks to `offset` (cost is O(new bytes)); a partial
# trailing line is left unconsumed so a half-written record is never parsed.
# Returns (events, new_offset). Missing file -> ([], offset).

follow_events(path: Path, *, offset: int = 0, poll: float = 0.5,
              stop_kinds: tuple[str, ...] = ("finalize",),
              idle_timeout: float | None = 300.0,
              should_stop: Callable[[], bool] | None = None) -> Iterator[dict]
# Blocking generator yielding event dicts as they are appended. Waits for the
# file to appear. Stops after an event whose kind is in `stop_kinds`, after
# `idle_timeout` s of silence (None = forever), or when `should_stop()` is true
# (draining anything written in the race window first).

format_event(ev: dict, totals: dict | None = None) -> str | None
# One human-readable line, or None for bookkeeping kinds worth hiding.
# Never emits ANSI. `totals` is a caller-owned dict that accumulates the
# live cost/token meter across events.

render_line(ev: dict, totals: dict | None = None, *, color: bool = False) -> str | None
# format_event + optional ANSI styling (accept=green / reject=dim / error=red).

use_color(stream=None) -> bool   # True only on a real TTY without NO_COLOR.
```

Suggested reuse: **#118** (stall detection) — drive it off `follow_events` with a `should_stop` deadline and inspect `evaluate`/`step` timestamps; **#122** (replay) — `read_new_events(path, 0)` for the full ordered log, `format_event` to render it; **#138**/**#144** — extend `format_event`/`render_line` rather than adding a second renderer.

## Verification

`core/tests/test_eventstream.py` adds 21 tests: incremental/partial/missing-file reads, `follow_events` live-append + finalize-stop + idle-timeout + stop-signal drain, per-kind formatting, the accumulating cost meter, "`format_event` never emits ANSI", `use_color` for non-TTY/`NO_COLOR`, the `tail` CLI (missing base, `--from-start`, attach-to-ongoing), the `tail` subcommand registration, a real end-to-end `cap-evolve run --follow` subprocess over `toy_calc` asserting plain-text stderr progress + intact stdout JSON, and an identity assertion that the dashboard SSE and the CLI share one function.

```
........................................................................ [ 36%]
........................................................................ [ 72%]
........................................................                 [100%]
200 passed in 66.43s (0:01:06)
```

(179 before this PR + 21 new.)

```
compileall OK
```

Dashboard SSE unaffected by the refactor — its own suite, unchanged:

```
$ cd dashboard/backend && PYTHONPATH=../../core:. python -m pytest tests -q
42 passed, 1 warning in 2.37s
```

…and the live SSE route still serves events, verified against a finished run dir with a record appended while connected:

```
stream: 200 text/event-stream; charset=utf-8
   event: snapshot
   data: {"run_id": "run_demo", "path": "/private/tmp/e2e1/.capevolve/run_demo", "graph": ...
   event: event
   data: {"t": 1785360348.322879, "kind": "step", "candidate": "sse_probe", "accept": true, "val": 1.0}
   event: event
   data: {"t": 1785360348.322898, "kind": "finalize", "test_reward": 1.0}
   event: done
   data: {"run_id": "run_demo"}
```

`cap-evolve tail` attaching **before** the run existed, streaming the same events:

```text
waiting for /tmp/e2e2/.capevolve/run_live/events.jsonl …
[00:26:43] splits frozen  train=4 val=2 test=2 (test sealed)
[00:26:43] baseline  val=0.0000 ±0.0000
[00:26:44] ACCEPT  cand_0001  val=1.0000 (parent 0.0000)  — paired Δ̄=+1.0000 > 0 (SE=0 → STRICT fallback, warned; n=2)
[00:26:45] reject  cand_0002  val=1.0000 (parent 1.0000)  — paired Δ̄=+0.0000 <= 0 (SE=0 → STRICT fallback, warned; n=2)
[00:26:46] FINALIZE  test=1.0000 (baseline 0.0000, Δ+1.0000)  best=cand_0001
```

No-TTY / CI safety — a full run with both streams redirected to files:

```
$ grep -c $'\033' /tmp/e2e3.err /tmp/e2e3.out
/tmp/e2e3.out:0
/tmp/e2e3.err:0
```

Full command transcripts are in the 🔬 Evidence comment.

## House rules

Zero new runtime deps in core (`eventstream.py` is stdlib: `json`, `os`, `sys`, `time`, `pathlib`, `typing`). New tests live in `core/tests/`. `harness.py` is untouched — the run already logged every event needed, which is why the diff is small.


OsherElhadad:
## 🔬 Evidence

Every command below was run in a clean worktree off `origin/main` (`6fca097`). Interpreter: a venv with only cap-evolve's own deps (+`fastapi`/`httpx` for the dashboard suite, which is a dashboard **dev** extra — core gained no deps).

### 1. Full core test suite

```
........................................................................ [ 36%]
........................................................................ [ 72%]
........................................................                 [100%]
200 passed in 85.51s (0:01:25)
```

Baseline on `origin/main` was 179; this PR adds 21 → **200 passed, 0 failed**.

### 2. The new test module on its own

```
$ PYTHONPATH=core python -m pytest core/tests/test_eventstream.py -v
core/tests/test_eventstream.py::test_read_new_events_incremental PASSED  [  4%]
core/tests/test_eventstream.py::test_read_new_events_leaves_partial_line PASSED [  9%]
core/tests/test_eventstream.py::test_read_new_events_missing_file PASSED [ 14%]
core/tests/test_eventstream.py::test_follow_events_stops_on_finalize_and_replays PASSED [ 19%]
core/tests/test_eventstream.py::test_follow_events_idle_timeout_on_empty PASSED [ 23%]
core/tests/test_eventstream.py::test_follow_events_picks_up_appends_live PASSED [ 28%]
core/tests/test_eventstream.py::test_follow_events_should_stop_drains_tail PASSED [ 33%]
core/tests/test_eventstream.py::test_format_event_covers_the_key_kinds PASSED [ 38%]
core/tests/test_eventstream.py::test_format_event_skips_bookkeeping_kinds PASSED [ 42%]
core/tests/test_eventstream.py::test_format_event_shows_unknown_kinds PASSED [ 47%]
core/tests/test_eventstream.py::test_cost_meter_accumulates_across_events PASSED [ 52%]
core/tests/test_eventstream.py::test_format_event_never_emits_ansi PASSED [ 57%]
core/tests/test_eventstream.py::test_render_line_color_on_and_off PASSED [ 61%]
core/tests/test_eventstream.py::test_use_color_false_for_non_tty_and_no_color PASSED [ 66%]
core/tests/test_eventstream.py::test_events_path_prefers_run_ts_and_ignores_pre_existing PASSED [ 71%]
core/tests/test_eventstream.py::test_cmd_tail_missing_base PASSED        [ 76%]
core/tests/test_eventstream.py::test_cmd_tail_from_start_plain_when_piped PASSED [ 80%]
core/tests/test_eventstream.py::test_cmd_tail_attaches_to_an_ongoing_run PASSED [ 85%]
core/tests/test_eventstream.py::test_tail_is_a_registered_subcommand PASSED [ 90%]
core/tests/test_eventstream.py::test_run_follow_end_to_end_pipes_clean_progress PASSED [ 95%]
core/tests/test_eventstream.py::test_dashboard_stream_reexports_the_shared_helper PASSED [100%]

============================== 21 passed in 4.35s ==============================
```

### 3. compileall

```
$ python -m compileall -q core skills; echo "rc=$?"
rc=0
```
........................................................................ [ 70%]
.............................................................            [100%]
205 passed in 81.14s (0:01:21)
```

### 3. compileall

```
$ python -m compileall -q core skills && echo "COMPILEALL CLEAN"
COMPILEALL CLEAN
```

### 4. Dashboard: SSE unaffected by the refactor

```
$ cd dashboard/backend && PYTHONPATH=../../core:. python -m pytest tests -q
42 passed, 1 warning in 1.72s
```

The pre-existing `tests/test_stream.py` (which asserts the byte-offset tail semantics) passes **unmodified** against the re-exported core helper:

```
$ cd dashboard/backend && PYTHONPATH=../../core:. python -m pytest tests/test_stream.py -v
tests/test_stream.py::test_sse_format PASSED                             [ 33%]
tests/test_stream.py::test_read_new_events_incremental PASSED            [ 66%]
tests/test_stream.py::test_read_new_events_ignores_partial_trailing_line PASSED [100%]

============================== 3 passed in 0.02s ===============================
```

And the live SSE route still serves events, against the finished run dir from the `--follow` run in §5 (a `step` + `finalize` were appended while the client was connected):

```
$ python - <<EOF   # TestClient over create_app("/tmp/e2e1/.capevolve")
GET /api/runs -> [{"run_id": "run_demo", "path": "/tmp/e2e1/.capevolve/run_demo", "algorithm": null, "status": "done", "best_val": 1.0, "baseline_val": 0.0, "delta_pct": null, "iterations": 3, "total_usd": 0.0, "mtime...
stream: 200 text/event-stream; charset=utf-8
   event: snapshot
   data: {"run_id": "run_demo", "path": "/private/tmp/e2e1/.capevolve/run_demo", "graph": {"nodes": [{"id": "seed", "parent": null, "
   event: event
   data: {"t": 1785360348.322879, "kind": "step", "candidate": "sse_probe", "accept": true, "val": 1.0}
   event: event
   data: {"t": 1785360348.322898, "kind": "finalize", "test_reward": 1.0}
   event: done
   data: {"run_id": "run_demo"}
```

> Note: the 6 fastapi-importing dashboard test modules error out on a bare checkout because `fastapi` is not installed by default. That is **pre-existing on `origin/main`** (verified in the main checkout before touching anything) and unrelated to this PR; installing the dashboard `dev` extra makes all 42 pass, as above.

### 5. Real end-to-end `cap-evolve run --follow` (zero API cost, `mock` optimizer)

Setup mirrors `examples/toy_calc/run.sh` (`toy_calc` adapter + `templates/project/capevolve.yaml`, `optimizer_skill: mock`):

```bash
set -euo pipefail
REPO=/tmp/wt-116
export CAPEVOLVE_CORE="$REPO/core" PYTHONPATH="$REPO/core" CAPEVOLVE_SKILLS_DIR="$REPO/skills"
export CAPEVOLVE_TOY_DATA="$REPO/examples/toy_calc" CAPEVOLVE_MOCK_SCRIPT="$REPO/examples/toy_calc/mock_script.json"
D="${1:-$(mktemp -d -t toy_calc.XXXXXX)}"
mkdir -p "$D/.capevolve/project/adapters"
cp "$REPO/examples/toy_calc/adapter.py" "$D/.capevolve/project/adapters/"
cp -R "$REPO/examples/toy_calc/capability" "$D/seed_capability"
cp "$REPO/templates/project/capevolve.yaml" "$D/.capevolve/project/capevolve.yaml"
echo "WORKDIR=$D"
```

```
$ python -m cap_evolve.cli run --spec /tmp/e2e1/.capevolve/project/capevolve.yaml \
    --project /tmp/e2e1/.capevolve/project --run-ts demo --follow --dashboard off

# --- STDERR (live progress, appearing line-by-line as the run works) ---
[00:38:23] splits frozen  train=4 val=2 test=2 (test sealed)
[00:38:23] eval val/seed  reward=0.0000 ±0.0000  0.0s
[00:38:23] baseline  val=0.0000 ±0.0000
[00:38:24] eval val/cand_0001  reward=1.0000 ±0.0000  0.0s
[00:38:24] gate warning (paired): combined/paired SE is 0 (likely n_trials=1 or identical trials)
[00:38:24] ACCEPT  cand_0001  val=1.0000 (parent 0.0000)  — paired Δ̄=+1.0000 > 0 (SE=0 → STRICT fallback, warned; n=2)
[00:38:24] eval val/cand_0002  reward=1.0000 ±0.0000  0.0s
[00:38:24] gate warning (paired): combined/paired SE is 0 (likely n_trials=1 or identical trials)
[00:38:24] reject  cand_0002  val=1.0000 (parent 1.0000)  — paired Δ̄=+0.0000 <= 0 (SE=0 → STRICT fallback, warned; n=2)
[00:38:25] eval val/cand_0003  reward=1.0000 ±0.0000  0.0s
[00:38:25] gate warning (paired): combined/paired SE is 0 (likely n_trials=1 or identical trials)
[00:38:25] reject  cand_0003  val=1.0000 (parent 1.0000)  — paired Δ̄=+0.0000 <= 0 (SE=0 → STRICT fallback, warned; n=2)
[00:38:26] eval test/FINAL  reward=1.0000 ±0.0000  0.0s
[00:38:26] eval test/FINAL_seed  reward=0.0000 ±0.0000  0.0s
[00:38:26] FINALIZE  test=1.0000 (baseline 0.0000, Δ+1.0000)  best=cand_0001

# --- STDOUT (unchanged: the machine-readable final JSON) ---
{
  "best_id": "cand_0001",
  "baseline_val": 0.0,
  "test_reward": 1.0,
  "test_baseline_reward": 0.0,
  "test_delta": 1.0,
  "test_pass_k": {
    "1": 1.0,
    "2": 0.0
  },
  "iterations": 3,
}

```

On a real TTY the same run is colored. Captured through `script(1)` (`repr()` of the escape-bearing lines):

```
$ script -q out.txt bash run_follow.sh   # then: [l for l in lines if '[' in l]
ANSI-colored lines on a real TTY: 5
'\x1b[36m[00:28:48] baseline  val=0.0000 ±0.0000\x1b[0m'
'\x1b[32m[00:28:49] ACCEPT  cand_0001  val=1.0000 (parent 0.0000)  — paired Δ̄=+1.0000 > 0 (SE=0 → STRICT fallback, warned; n=2)\x1b[0m'
'\x1b[2m[00:28:49] reject  cand_0002  val=1.0000 (parent 1.0000)  — paired Δ̄=+0.0000 <= 0 (SE=0 → STRICT fallback, warned; n=2)\x1b[0m'
'\x1b[2m[00:28:50] reject  cand_0003  val=1.0000 (parent 1.0000)  — paired Δ̄=+0.0000 <= 0 (SE=0 → STRICT fallback, warned; n=2)\x1b[0m'
'\x1b[1m[00:28:51] FINALIZE  test=1.0000 (baseline 0.0000, Δ+1.0000)  best=cand_0001\x1b[0m'
```

Cyan baseline, green ACCEPT, dim reject, bold FINALIZE — and **zero** escapes on the piped run above.

### 6. `cap-evolve tail` attaching to a run IN PROGRESS

Second run. The tail is started **first**, before the run dir exists, then the run is launched with **no** `--follow` — proving `tail` is a fully standalone attach path and that the two surfaces render the same events.

```
$ python -m cap_evolve.cli tail /tmp/e2e2/.capevolve/run_live --from-start --idle-timeout 60 &
$ python -m cap_evolve.cli run --spec /tmp/e2e2/.capevolve/project/capevolve.yaml \
    --project /tmp/e2e2/.capevolve/project --run-ts live --dashboard off > run.json 2> run.err

# --- tail output ---
waiting for /tmp/e2e2/.capevolve/run_live/events.jsonl …
[00:26:43] splits frozen  train=4 val=2 test=2 (test sealed)
[00:26:43] eval val/seed  reward=0.0000 ±0.0000  0.0s
[00:26:43] baseline  val=0.0000 ±0.0000
[00:26:44] eval val/cand_0001  reward=1.0000 ±0.0000  0.0s
[00:26:44] gate warning (paired): combined/paired SE is 0 (likely n_trials=1 or identical trials)
[00:26:44] ACCEPT  cand_0001  val=1.0000 (parent 0.0000)  — paired Δ̄=+1.0000 > 0 (SE=0 → STRICT fallback, warned; n=2)
[00:26:45] eval val/cand_0002  reward=1.0000 ±0.0000  0.0s
[00:26:45] gate warning (paired): combined/paired SE is 0 (likely n_trials=1 or identical trials)
[00:26:45] reject  cand_0002  val=1.0000 (parent 1.0000)  — paired Δ̄=+0.0000 <= 0 (SE=0 → STRICT fallback, warned; n=2)
[00:26:45] eval val/cand_0003  reward=1.0000 ±0.0000  0.0s
[00:26:45] gate warning (paired): combined/paired SE is 0 (likely n_trials=1 or identical trials)
[00:26:45] reject  cand_0003  val=1.0000 (parent 1.0000)  — paired Δ̄=+0.0000 <= 0 (SE=0 → STRICT fallback, warned; n=2)
[00:26:46] eval test/FINAL  reward=1.0000 ±0.0000  0.0s
[00:26:46] eval test/FINAL_seed  reward=0.0000 ±0.0000  0.0s
[00:26:46] FINALIZE  test=1.0000 (baseline 0.0000, Δ+1.0000)  best=cand_0001

# --- the run itself, WITHOUT --follow: still silent (this is the "before") ---
$ wc -c < run.err
       0
$ cat run.json
{
  "run_dir": ".capevolve/run_live",
  "best_id": "cand_0001",
  "baseline_val": 0.0,
  "test_reward": 1.0,
  "test_baseline_reward": 0.0,
  "test_delta": 1.0,
  "test_pass_k": { "1": 1.0, "2": 0.0 },
  "iterations": 3,
  "dashboard": ".capevolve/run_live/dashboard.html"
}
```

`waiting for … events.jsonl …` is the attach-before-creation case: an earlier draft rejected a not-yet-existent run dir with `not a run dir`, which broke exactly the scenario the issue asks for. Caught by this transcript, fixed, re-run above.

Default behavior is provably unchanged for scripts: **0 bytes** of stderr without `--follow`, identical stdout JSON.

### 7. No-TTY / CI safety (piped to files)

Third run, `--follow`, both streams redirected:

```
$ python -m cap_evolve.cli run ... --run-ts piped --follow --dashboard off > e2e3.out 2> e2e3.err
rc=0
$ grep -c $'\033' e2e3.err e2e3.out
e2e3.out:0
e2e3.err:0
```

Byte-level proof on one progress line (only the UTF-8 `±` = `302 261`, no `033`):

```
$ sed -n '3p' e2e3.err | od -c | head -3
0000000    [   0   0   :   2   7   :   0   8   ]       b   a   s   e   l
0000020    i   n   e           v   a   l   =   0   .   0   0   0   0
0000040  302 261   0   .   0   0   0   0  \n
```

`--follow` progress goes to **stderr**, so `cap-evolve run --follow > result.json` keeps producing a parseable JSON file — asserted in `test_run_follow_end_to_end_pipes_clean_progress`, which runs the whole pipeline as a subprocess and checks `json.loads(proc.stdout)["test_reward"] == 1.0` plus `"\033" not in proc.stderr`.

### 8. Files touched

| File | Change |
|---|---|
| `core/cap_evolve/eventstream.py` | **new** — the shared events.jsonl tail + renderer (stdlib only) |
| `core/cap_evolve/cli.py` | `--follow` flag + follower thread, new `tail` subcommand, usage/docstring |
| `dashboard/backend/capevolve_dashboard/stream.py` | tail logic removed; `read_new_events` now re-exported from core |
| `core/tests/test_eventstream.py` | **new** — 21 tests |
| `docs/GETTING_STARTED.md` | new "Watch a run live in the terminal" section |
| `CHANGELOG.md` | Unreleased → Added |

`harness.py`, `rundir.py`, `loop.py`, `gepa.py`, `skillopt.py` are **untouched** — the run already logged every event needed, so no engine change was required.

---

## 📌 For #118, #122, #138, #144 — reuse this, don't re-implement it

The events.jsonl tail is now **one** module: **`core/cap_evolve/eventstream.py`** (stdlib only, no new deps). It is consumed by both `cap-evolve run --follow`/`cap-evolve tail` and the dashboard's SSE route (`dashboard/backend/capevolve_dashboard/stream.py` re-exports `read_new_events` from it), and `core/tests/test_eventstream.py::test_dashboard_stream_reexports_the_shared_helper` asserts the identity so a future copy-paste fork fails CI.

```python
from cap_evolve import eventstream

read_new_events(path: Path, offset: int) -> tuple[list[dict], int]
follow_events(path: Path, *, offset: int = 0, poll: float = 0.5,
              stop_kinds: tuple[str, ...] = ("finalize",),
              idle_timeout: float | None = 300.0,
              should_stop: Callable[[], bool] | None = None) -> Iterator[dict]
format_event(ev: dict, totals: dict | None = None) -> str | None
render_line(ev: dict, totals: dict | None = None, *, color: bool = False) -> str | None
use_color(stream=None) -> bool
```

- **#118 (stall detection)** — `follow_events(..., idle_timeout=N)` already returns on N seconds of silence; that return *is* the stall signal. For a richer rule, pass `should_stop` and compare the `t` field of the last `evaluate`/`step` event against a deadline. Don't add a second poller.
- **#122 (replay)** — `read_new_events(path, 0)` yields the whole ordered log in one call; `format_event` renders each record. A replay is a `for` loop over those two, with `time.sleep` on the `t` deltas if you want real-time pacing.
- **#138 / #144** — extend `format_event` (per-kind bodies) and `render_line` (styling) rather than introducing a parallel renderer. `format_event` is guaranteed ANSI-free (`test_format_event_never_emits_ansi`) and `use_color()` is already the single TTY/`NO_COLOR` decision point, which is the hook #144 should harden.

Unknown event kinds fall through to a generic `kind k=v k=v` line rather than being dropped, so a new event type added by any of these issues is visible in the terminal before anyone writes a formatter for it.

OsherElhadad:
## 🔍 Review — PR #191

The feature works and the shared-helper refactor is the right call. But this is a **hub** PR and it ships three defects that will propagate: the follower thread **dies silently on a malformed event** (reintroducing exactly the "silent run" bug #116 fixes), unsanitised event text is **written straight to the terminal** (ANSI/OSC injection), and the cost meter **double-counts**. The partial-line handling — the risk you flagged as most likely — is **correct**; I hammered it and could not break it.

---

### Blocking

**B1. `core/cap_evolve/eventstream.py:169` — one malformed event kills the follower thread; the rest of the run is invisible again.**
`format_event` does `float(ev.get("t") or time.time())` with no guard, and `render_line`/`format_event` call `ev.get(...)` assuming `ev` is a dict. Any of these raises:

```
$ # t is not a number
{'kind':'baseline','t':'not-a-number'} -> ValueError: could not convert string to float: 'not-a-number'
{'kind':'baseline','t':{'nested':1}}   -> TypeError: float() argument must be a string or a real number, not 'dict'
{'kind':'baseline','t':1e300}          -> OverflowError: timestamp out of range for platform time_t
$ # a non-dict JSON record (read_new_events happily returns these — see B1b)
'"just a string"' -> AttributeError: 'str' object has no attribute 'get'
'[1,2,3]' / '42' / 'null' -> AttributeError
```

`_spawn_follower`'s worker wraps the whole loop in `except Exception: pass` (`core/cap_evolve/cli.py:105-106`), so the thread **dies without a word** and the user watches a blank terminal for the rest of a multi-hour run — the literal bug #116 exists to fix. Demonstrated end to end:

```
$ # follower running; write baseline, then one poisoned record, then 2 good ones
[00:51:39] baseline  val=0.1000 ±?
follower thread alive after poison? False
# the step/finalize events that followed were never printed, and nothing said why
```

`log_event` uses `json.dumps(rec, default=str)` (`core/cap_evolve/rundir.py:395`), so a non-serialisable field already becomes a string — a `t` that isn't a float is reachable, not hypothetical. And #138/#144 will add event kinds with new field types straight into this path.

*Fix (two lines, both in the helper so every caller is covered):* in `format_event`, `if not isinstance(ev, dict): return None`, and wrap the timestamp: `try: ts = time.strftime(...) except (TypeError, ValueError, OverflowError, OSError): ts = "--:--:--"`. Separately, in `cli.py:105`, replace the bare `pass` with one line to stderr — `print(f"[follow] progress stream stopped: {e!r}", file=sys.stderr)` — so a dead follower is *never* indistinguishable from a silent run. That is the whole point of the feature.

**B1b. `core/cap_evolve/eventstream.py:60` — `read_new_events` returns non-dict records.** `json.loads("42")` / `"null"` / `'[1,2]'` all succeed, so `events` can contain non-dicts that every consumer (the CLI **and** the dashboard SSE route, which forwards them into `sse_format`) assumes are dicts. *Fix:* `obj = json.loads(line); if isinstance(obj, dict): events.append(obj)`.

**B2. `core/cap_evolve/eventstream.py:230` — terminal escape-sequence injection from event payloads.** Nothing is sanitised before rendering. Field values reach the terminal verbatim:

```
$ format_event({"kind":"optimizer_error","candidate":"c1","error":"boom\033]0;PWNED\007\033[2J\033[31mfake red"})
'[00:44:23] OPTIMIZER ERROR  c1: boom\x1b]0;PWNED\x07\x1b[2J\x1b[31mfake red'
$ format_event({"kind":"brand_new","payload":"\033[2J\033[Hcleared screen"})
'[00:44:23] brand_new payload=\x1b[2J\x1b[Hcleared screen'
$ # newline injection forges a whole extra progress line:
$ format_event({"kind":"brand_new","payload":"a\nFINALIZE  test=1.0000 (baseline 0.0000) best=FAKE"})
'[00:44:23] brand_new payload=a\nFINALIZE  test=1.0000 (baseline 0.0000) best=FAKE'
```

This is not theoretical: `optimizer_error.error` is the **optimizer agent CLI's own stderr**, relayed through `run-optimizer`'s `stderr_tail` (`skills/optimizers/run-optimizer/scripts/run.py:301` → `harness.py:483` → `harness.py:1271`), and `step.reason` is gate text. An LLM-authored capability or a hostile model response can emit `\033]0;…\007` (set window title), `\033[2J` (clear screen), or `\033]52;c;<b64>\007` (OSC 52 clipboard write). Note `test_format_event_never_emits_ansi` (`core/tests/test_eventstream.py:143`) tests only *hard-coded clean* events, so it gives false assurance here.

*Fix:* one sanitiser applied to the finished line in `format_event`'s return, so every caller and every future kind is covered:
```python
_CTRL = {c: None for c in range(0x20) if c not in (0x09,)} | {0x7f: None}
...
return f"[{ts}] {body}{meter}".translate(_CTRL)
```
Do it in `format_event`, not `render_line` — the SSE/JSON path and #144's ladder both need it. This also fixes the newline-forgery case.

**B3. `core/cap_evolve/eventstream.py:146-153` — the cost meter double-counts, over-reporting spend by ~2x.** `_accrue` sums `cost_usd`, `opt_cost_usd` **and** `usd` from one event, and `tokens` + `opt_tokens`. But `harness.py:311` logs `evaluate(cost_usd=X, tokens=T)` and then `harness.py:1323` logs `step(cost_usd=X, tokens=T, opt_cost_usd=O, opt_tokens=OT)` for the *same* eval — `step` re-states the runner cost the `evaluate` event already reported:

```
TRUE spend: runner $0.50 + optimizer $0.20 = $0.70 ; tokens 1500
METER says: {'usd': 1.2, 'tokens': 2500} -> [...] ACCEPT cand_0001 [...]  [$1.2000 · 2.5k tok]
```

$1.20 shown for $0.70 actually spent, on a **budget** readout in a tool whose selling point is honest accounting. `test_cost_meter_accumulates_across_events` (`test_eventstream.py:131`) passes because it feeds two synthetic `step`s and never the real `evaluate`+`step` pair. *Fix:* accrue runner cost from `evaluate` only and optimizer cost from `step` only — key the accrual on `kind`, e.g. `if kind == "evaluate": usd += cost_usd; tokens += tokens` / `elif kind in ("step","gepa_val_gate","skillopt_step"): usd += opt_cost_usd; tokens += opt_tokens`. Add the real two-event sequence as the test.

**B4. `core/cap_evolve/cli.py:97-104` — with stderr closed, `--follow` corrupts the stdout JSON contract.** `print(line, file=sys.stderr)` under `2>&-`: CPython gives fd 2 to the next `open()`, so progress lines land **interleaved in stdout**, breaking the exact guarantee the PR advertises:

```
$ cap-evolve run ... --follow --dashboard off >cl.out 2>&-   ; rc=0
$ python -c "json.load(open('/tmp/cl.out'))"
JSON PARSE FAILED: JSONDecodeError Expecting ',' delimiter: line 1 column 3 (char 2)
$ # 12 progress lines before the JSON blob, 3 more after it
$ # same invocation WITHOUT --follow: parsed ok 1.0
```

`2>&-` is common in cron/CI wrappers. *Fix:* in `_spawn_follower`, capture `err = sys.stderr` at spawn time and bail if it is not a live, distinct stream — cheapest guard: `if not hasattr(sys.stderr,'fileno') or sys.stderr.fileno() != 2: return` before starting the thread, or verify `os.fstat(2)` differs from `os.fstat(1)`.

---

### Non-blocking

**N1. `core/cap_evolve/cli.py:132-133,145` — a typo'd run dir pretends to wait for 5 minutes and exits 0.** The comment claims `--idle-timeout` makes this sane. Measured:

```
$ cap-evolve tail /tmp/does/not/exist/run_typo     # default idle-timeout=300
waiting for /tmp/does/not/exist/run_typo/events.jsonl …
STILL RUNNING after 10s -> will pretend-wait 300s on a typo
$ # and the exit code cannot be distinguished from success:
finished-run rc = 0
typo-path   rc = 0
```

A path whose **parent directory doesn't exist** can never be a run dir. *Fix:* if `root.parent` does not exist, print `no such run dir: <root>` and `return 2`; and return a distinct code (e.g. 3) when `follow_events` exits on idle-timeout having yielded no events, so scripts can tell "timed out" from "run finished".

**N2. `core/cap_evolve/eventstream.py:63` — a corrupt (non-partial) line is silently dropped.** `except json.JSONDecodeError: continue` discards a genuinely truncated/interleaved record with no signal, on both surfaces:
```
$ '{"kind":"a"}\nNOT JSON AT ALL\n{"kind":"finalize"}' -> ['a','finalize']
```
Silently losing audit-log records in an honesty-focused tool is a bad default. *Fix:* count skips and yield/log one `{"kind":"log_corruption","dropped":N}` note, or at minimum a stderr warning from the CLI callers.

**N3. `core/cap_evolve/eventstream.py:44-46` — offset integrity on truncate/rewrite silently skips records.** Rewriting the file shorter and then re-growing it past the stale offset loses everything before it:
```
read1 ['a','b'] offset=26
after truncate (size 13): events=[] offset=26   # stuck
after regrow to 3 records:  events=['x'] offset=39   # 'z' and 'y' lost forever
```
Rare for an append-only log, but #122's replay will drive this helper over files it rewrites. *Fix:* `if p.stat().st_size < offset: offset = 0` (treat shrink as rotation and re-read).

**N4. `core/cap_evolve/cli.py:381` — `--resume --follow` replays the entire prior log before showing new progress.** `_spawn_follower` always starts at `offset=0`, so resuming a run with 10k prior events dumps all of them. `cap-evolve tail` gets this right (`cli.py:145` uses `st_size` unless `--from-start`); the follower has no equivalent. *Fix:* when `resume_ts` is set, seed the offset from the existing `events.jsonl` size.

**N5. `core/cap_evolve/eventstream.py:106` — `use_color()` defaults to `sys.stdout` but every real caller passes the stream it writes to.** A future caller that forgets the argument gets a wrong decision (colorising stderr based on stdout's TTY-ness). Make the parameter required, or default to `sys.stderr` for the follow path.

**N6. `render_line` is a thin wrapper over `format_event` and both are public (`__all__`, `eventstream.py:36`).** The PR description tells #138 to "extend `format_event` (bodies) and `render_line` (styling)" — that's a coherent split, but the redundancy invites #144 to call the wrong one and lose sanitisation. Once B2 lands in `format_event` the split is defensible; keep it, but document that `format_event` is the only place text is made terminal-safe.

**N7. `docs/GETTING_STARTED.md:46-77` — the documented cost meter output is wrong.** `[$0.1240 · 8.3k tok]` on the ACCEPT lines will be roughly double the true spend until B3 is fixed; the doc bakes in the bug. Also `tail` is documented only here — `docs/INSTALL.md:29` lists `cap-evolve dashboard` and `README.md` has no subcommand list, so `tail` should get a line wherever `dashboard` is listed.

**N8. `PYTHONIOENCODING=ascii` mangles progress into `—`/`\xb1` escapes.** All 15 lines survive (the follower does *not* die), but the output is illegible:
```
[00:52:03] ACCEPT  cand_0001  val=1.0000 (parent 0.0000)  — paired Δ̄=+1.0000 > 0 …
```
Issue #144 explicitly names non-UTF-8 locales, so this is legitimately #144's job — flagging so it isn't lost.

---

### Nits

- `core/tests/test_eventstream.py:1-9` — `io`, `os`, `shutil`, `threading` imported at module top for a couple of tests; harmless, but `os`/`shutil` are only used by the e2e test.
- `core/cap_evolve/cli.py:133` — `--idle-timeout -5` is accepted and behaves like `0`/wait-forever (`args.idle_timeout or None` turns `0` into `None`, and a negative value trips the `>=` immediately). Clamp to `>= 0`.
- `core/cap_evolve/eventstream.py:73` — `stop_kinds=("finalize",)` as a **default** means every consumer inherits "stop at finalize". Correct for the CLI, wrong for a dashboard/replay consumer that wants the tail after finalize. It's overridable, so fine — but see the hub note below.

---

### Hub API fitness

The decision to put one tail in `core/cap_evolve/eventstream.py` and re-export from `dashboard/.../stream.py` is right and I'd keep it. Per-issue:

**#118 (stalled vs hung vs done) — ⚠️ needs an API change; the current signature cannot express it.**
`follow_events` has **three** exits — `stop_kinds` hit, `idle_timeout` elapsed, `should_stop()` true — and all three are a **bare `return`** (`eventstream.py:83,89,95`; verified: 3 returns, no sentinel, no exception). The consumer that just got `StopIteration` **cannot tell which happened**, which is precisely the distinction #118 exists to make. `should_stop` also can't help: it's a zero-arg `Callable[[], bool]` with no access to the last event's `t`, so #118 must maintain its own parallel timestamp state — the "don't add a second poller" advice in the PR description doesn't actually hold. And the hard-coded `idle_timeout=300.0` default pre-decides #118's threshold in the shared helper; #118's issue text says the threshold is *configurable*, so this default will have to be revisited anyway.
*Recommended now, while it's cheap:* make the reason observable — either `return`/set a `reason` attribute on a small result object, or yield a final sentinel `{"kind": "_follow_end", "reason": "stop_kind"|"idle"|"should_stop"}`. And drop the `300.0` default to `None` (the CLI already passes its own), so the helper stops legislating for #118.

**#122 (replay, scrubbable) — ✅ partly fit; one gap.** `read_new_events(path, 0)` returns the whole finished log deterministically in one call, ordered — verified on a real run dir (17 events, correct order). That covers forward replay and `format_event` renders each record, so #122's CLI is a loop over two existing functions. **Gap:** scrubbing *backwards* has nothing here — there's no index from time→offset, so #122 will read the whole file into memory and index it itself. Acceptable (that's a replay concern, not a tail concern), but N3 (shrink→offset stuck) should be fixed first, and B1b matters more here because a replay over an arbitrary user-supplied log is a wider trust boundary than a live tail.

**#138 (dashboard header/sparkline/burn) — ⚠️ fit for the stream, NOT for the burn meter.** It consumes the SSE route, which the refactor leaves working (verified live below), so the stream side is fine. But the "live token burn" readout is the **same** accounting as B3, and if #138 reuses `_accrue` it inherits the ~2x over-count. Fix B3 before #138 starts, and expose `_accrue` (or a documented `accrue_totals`) as public so #138 does not re-derive it — right now it's private, so #138 will fork the arithmetic. Also note `format_event` returns `None` for `minibatch`/`target_profile`/`gepa_merge_*` (`eventstream.py:214-218`); #138 needs those for phase detection, so it must consume raw events, not `format_event`.

**#144 (no-TTY ladder + crash log) — ⚠️ the seam exists but is not yet safe to build on.** `use_color()` is a single, correct TTY/`NO_COLOR` decision point and is exactly the hook #144 should harden — good. `render_line(color=...)` is a clean styling seam. **But** #144's headline requirement is "never emit control sequences into a non-TTY", and B2 means the current code emits attacker/model-controlled control sequences on **every** path including piped output — #144 would be building its ladder on top of an unsanitised renderer. B2 belongs in this PR, not #144. Secondary: N8 (non-UTF-8) and the crash log are genuinely #144's; the `except Exception: pass` at `cli.py:105-106` is the natural place for #144's forensic hook, so leave a comment there rather than a bare `pass`.

**Net:** the module is the right shape; three small changes here (reason-on-exit, `idle_timeout=None` default, sanitisation in `format_event`) stop all four downstream issues from having to work around it.

---

### Concurrency audit

| Failure mode | Result |
|---|---|
| **Partial-line reads** | ✅ **CORRECT.** Torture-tested with a writer dribbling **7 bytes at a time, `flush()`+`fsync()` after each**, so every read landed mid-record: `events received: ['baseline','step','finalize'] / count == 3: True`. The `rfind(b"\n")` + `offset + last_nl + 1` arithmetic never advances past a partial record and never parses one. Nothing lost, nothing corrupted. This is right. |
| **Truncation / rotation** | ⚠️ Offset gets stuck on shrink and skips records on re-grow — N3. |
| **Deleted+recreated run dir** | ✅ `read_new_events` returns `([], offset)` on a missing file, and `follow_events` keeps polling; a recreated file re-grows past the stale offset (same skip caveat as N3). |
| **Exit drain (`done()` at 5 points)** | ✅ All five `_cmd_run` returns route through `done()` (`cli.py:413,453,533,556,560`) — checked every `return` in the function. The `should_stop` drain does re-read after the signal, and events emitted just before exit arrive: the real `--follow` run printed all 15 renderable lines including `FINALIZE`, which is the **last** record written (`harness.py:1966` returns immediately after). **No hang:** `done()` cost `0.312s` worst case (one `poll` interval) and `0.000s` when the run dir never appeared; `join(timeout=3.0)` bounds it. |
| **Follower thread death** | ❌ **BLOCKING — B1.** Dies silently on a malformed event; the rest of the run goes dark with no message. This is the reported bug re-entering through the fix. |
| **Ctrl-C / SIGINT** | ✅ No traceback, clean exit, stdout JSON intact — and **identical to `main`'s behavior** without `--follow` (both `rc=0` and complete the run; the toy run is too fast for the signal to land mid-phase, but the follower adds no new SIGINT path). `_cmd_tail` catches `KeyboardInterrupt` → `130`, correct. |
| **stderr → /dev/null** | ✅ `rc=0`, stdout JSON parses. |
| **stderr closed (`2>&-`)** | ❌ **BLOCKING — B4.** Progress lands in stdout; JSON unparseable. |

---

### UX verdict on `--follow`

**Genuinely good — this is the strongest part of the PR.** Real `toy_calc`/`mock` transcript, both streams piped (§ below for the literal output): 15 lines that answer all four of #116's questions — stage (`splits frozen`), what's happening now (`eval val/cand_0002`), is it improving (`ACCEPT cand_0001 val=1.0000 (parent 0.0000)` vs `reject … Δ̄=+0.0000 <= 0`), and the sealed result (`FINALIZE test=1.0000 … best=cand_0001`). The accept/reject line carrying the **gate reason** is the detail that makes it legible rather than decorative, and unknown kinds falling through to `kind k=v` means a new event type is visible before anyone writes a formatter. On a real TTY the colour split (green accept / dim reject / bold finalize / cyan baseline) is the right emphasis.

Three honest UX gripes:
1. **"is it stuck?" is still not answered.** During a 20-minute optimizer call the terminal is silent — indistinguishable from a hang, which is #116's stated motivation ("a hung run is indistinguishable from a working one"). #116 point 3 asks for a live cost line; a periodic heartbeat/elapsed line (which #118's issue text explicitly suggests) would close this. Not blocking — the per-candidate lines are a real improvement — but the flagship problem is only ~70% solved.
2. **The `gate warning (paired): combined/paired SE is 0` line fires on every single iteration** and is pure noise at this signal-to-noise ratio; consider de-duplicating repeated identical warnings.
3. **`±?`** renders when `stderr` is absent (`baseline val=0.1000 ±?`) — reads like an error rather than "not applicable".

The cost meter — the thing that would make a long expensive run feel controlled — **did not appear at all** in the zero-cost toy run (correct, `_meter` returns `''`), so its only exercise is the doubled synthetic test. See B3.

---

### Merge-order note

Verified with `git merge-tree` against `origin/main`:

| Branch | Overlap | Conflict? |
|---|---|---|
| `feat/issue-121-doctor` | `core/cap_evolve/cli.py` | ❌ **CONFLICT (1 region)** |
| `fix/issue-101-changelog` (PR #184) | `CHANGELOG.md` | ❌ **CONFLICT (1 region)** |
| `fix/issue-102-counts` (PR #189) | `CHANGELOG.md` | ✅ **clean** |
| `feat/issue-133-quickstart` | — | **branch does not exist on origin** (nothing to sequence) |

**#121 conflict is one line, mechanical.** Both branches edit the same `usage:` string in `main()`:
```
<<<<<<< 191:  print("usage: cap-evolve {version|splits|check|run|estimate|dashboard|tail} [args]",
=======
>>>>>>> 121:  print("usage: cap-evolve {version|splits|check|doctor|run|estimate|dashboard} [args]",
```
Resolution is `{version|splits|check|doctor|run|estimate|dashboard|tail}`. Their other `cli.py` edits don't collide: #121 inserts `_cmd_doctor` at `main:440` and one `COMMANDS` key; #191 inserts at `main:64` (`_events_path`/`_spawn_follower`/`_cmd_tail`), `main:126`/`main:270` (inside `_cmd_run`), and a different `COMMANDS` key. Git resolves the `COMMANDS` dict and the module docstring cleanly — only the usage string needs hands.

**#101 conflict is the `## [Unreleased]` block** — #184 rewrites 74/62 lines of `CHANGELOG.md`; #191 inserts a new `### Added` subsection at the top of Unreleased. Trivial, but it's a rewrite so resolve deliberately.

**Recommended order:** `#189 (102)` → **`#191`** → `#184 (101)` → `#121`.
Rationale: #189 is clean against #191, so order is free — take it first. Land #191 next (largest surface; the four downstream issues are blocked on the helper). Then #184 rebases its CHANGELOG rewrite onto the already-present `--follow` entry, which is far easier than the reverse. #121 last, since its conflict is a one-token usage-string edit its author can fix in seconds. **All of this is moot until the blocking findings are addressed** — #191 should not be the first of the four to land in its current state.

---

### Verification I re-ran

**Core suite — the PR's "200 passed" does NOT reproduce; 199 pass, 1 fails.**
```
$ cd /tmp/rv-191 && PYTHONPATH=core python -m pytest core/tests -q
1 failed, 199 passed in 65.92s (0:01:05)
```
**Not this PR's fault** — the identical failure reproduces on `origin/main`, so it is pre-existing (a port-binding assumption, `http://127.0.0.1:7882` vs expected):
```
$ cd /tmp/rv-main && PYTHONPATH=core python -m pytest core/tests/test_dashboard_launch.py -q
1 failed, 6 passed in 0.04s
```
Flagging only because the PR's evidence block claims a clean 200; the honest number in an environment with something on that port is 199 + 1 pre-existing.

**"fastapi collection failure is pre-existing on origin/main" — ⚠️ claim NOT VERIFIED as stated; it's false in the environment I could construct.** With `fastapi` installed, `origin/main` collects **all 42** dashboard tests with zero errors:
```
$ cd /tmp/rv-main && python -m pytest dashboard/backend/tests -q --collect-only
42 tests collected in 0.25s
```
The 6 modules error out only when `fastapi` is **absent**, which is an environment property, not a property of `main` — so "pre-existing on origin/main" is the wrong framing for the same observation. The substantive claim (unrelated to this PR) does hold. Both suites pass on the PR:
```
$ cd /tmp/rv-191 && python -m pytest dashboard/backend/tests -q
42 passed, 1 warning in 1.93s
$ cd /tmp/rv-191/dashboard/backend && PYTHONPATH=. python -m pytest tests/test_stream.py -q   # clean venv, core NOT on path
3 passed in 0.54s
```
**No import cycle, no new packaging dependency:** `capevolve_dashboard/__init__.py:1` already imports `_bootstrap`, which puts in-repo `core/` on `sys.path`, and `pyproject.toml` already declares `cap-evolve-core` as a hard dependency — so the re-export adds nothing new. Confirmed the unmodified `test_stream.py` byte-offset assertions pass against the re-exported core helper.

**Live SSE still works after the refactor** (records appended mid-connection to a real run dir):
```
$ TestClient(create_app("/tmp/e2e-a/.capevolve")) -> /api/runs 200
stream: 200 text/event-stream; charset=utf-8
   event: snapshot   data: {"run_id": "run_a", "path": "…/run_a", "graph": {"nodes": [{"id": "seed", …
   event: event      data: {"t": 1785362039.939492, "kind": "step", "candidate": "sse_probe", "accept": true, "val": 1.0}
   event: event      data: {"t": 1785362039.939508, "kind": "finalize", "test_reward": 1.0}
   event: done       data: {"run_id": "run_a"}
```

**compileall:** `python -m compileall -q core skills; rc=0`

**E2E `--follow`, both streams piped (reproduces the PR's §5 and §7 exactly):**
```
$ cap-evolve run --spec … --run-ts a --follow --dashboard off >a.out 2>a.err   ; rc=0
[00:47:29] splits frozen  train=4 val=2 test=2 (test sealed)
[00:47:29] eval val/seed  reward=0.0000 ±0.0000  0.0s
[00:47:29] baseline  val=0.0000 ±0.0000
[00:47:30] eval val/cand_0001  reward=1.0000 ±0.0000  0.0s
[00:47:30] gate warning (paired): combined/paired SE is 0 (likely n_trials=1 or identical trials)
[00:47:30] ACCEPT  cand_0001  val=1.0000 (parent 0.0000)  — paired Δ̄=+1.0000 > 0 (SE=0 → STRICT fallback, warned; n=2)
[00:47:30] reject  cand_0002  val=1.0000 (parent 1.0000)  — paired Δ̄=+0.0000 <= 0 (…)
[00:47:31] reject  cand_0003  val=1.0000 (parent 1.0000)  — paired Δ̄=+0.0000 <= 0 (…)
[00:47:32] eval test/FINAL  reward=1.0000 ±0.0000  0.0s
[00:47:32] FINALIZE  test=1.0000 (baseline 0.0000, Δ+1.0000)  best=cand_0001
--- STDOUT --- {"run_dir": ".capevolve/run_a", "best_id": "cand_0001", "test_reward": 1.0, …}
$ grep -c $'\033' a.err a.out
a.err:0
a.out:0
```
**Byte-level no-ANSI check (`od -c`), only the UTF-8 `±` = `302 261`:** confirmed on the `baseline` line — reproduces the PR's §7.
**Without `--follow`:** `wc -c < c.err` → `0` bytes stderr, stdout JSON identical. ✅ default behavior genuinely unchanged.

**Test quality (21 new tests):** the partial-line test (`test_eventstream.py:57`) is real and correct; `test_follow_events_picks_up_appends_live`/`_should_stop_drains_tail` do exercise threading; `test_run_follow_end_to_end_pipes_clean_progress` is a genuine subprocess e2e asserting no-ANSI + parseable stdout — that one is worth its weight. `test_dashboard_stream_reexports_the_shared_helper` **does** work as a fork-guard: it asserts function *identity* (`is`), so re-adding a local copy in `stream.py` fails it. Grep for coverage of the paths I broke: `_spawn_follower` **0**, `done()`/exit-drain **0**, malformed/non-dict events **0**, ANSI injection **0**, file truncation **0**, `KeyboardInterrupt` **0**, real `evaluate`+`step` cost pair **0**. The most critical untested path is **the follower thread itself** — every test drives `follow_events` directly or `_cmd_tail`; nothing tests the daemon thread that is the actual feature, which is why B1 shipped.

OsherElhadad:
## 🔧 Review fixes

Every blocking finding is fixed, tested, and demonstrated live. All 13 non-blocking findings + nits are addressed or explicitly declined below. Commit `009235b`.

**Suite: 217 passed, 0 failed** (179 on `main` → 217 here; +38 in `test_eventstream.py`, up from 21). `test_dashboard_launch.py::test_maybe_launch_spawns_when_available` **passed** in my worktree — you saw it fail because port 7878 was held by a sibling agent, which is #200, not this PR. `python -m compileall -q core skills` → `rc=0`. Dashboard: `42 passed`.

---

### Blocking

**B1 — a malformed event killed the follower. FIXED.** Two halves, both at the shared choke point:

`format_event` is now total. `if not isinstance(ev, dict): return None`, and the timestamp goes through `_timestamp()`, which catches `TypeError/ValueError/OverflowError/OSError` and degrades to `--:--:--`. And `cli.py` no longer swallows: the guard **reports**.

```python
except Exception as e:  # noqa: BLE001 — observability must never break the run
    # ...but it must never go dark in silence either. #144: this is the hook for
    # the forensic crash log; the user-visible line below is the minimum.
    print(f"[follow] live progress stopped: {e!r} — the run continues; "
          f"use `cap-evolve tail` or the dashboard to watch it", file=err, flush=True)
```

Live: the real daemon thread, fed your exact poison sequence plus a non-dict and an unparseable line, then good events.

```
[01:23:02] baseline  val=0.1000 ±?
[--:--:--] baseline  val=? ±?
[01:23:02] WARNING  1 unreadable record(s) skipped in events.jsonl
[--:--:--] null_t
[01:23:02] WARNING  1 unreadable record(s) skipped in events.jsonl
[01:23:03] ACCEPT  cand_0001  val=1.0000 (parent 0.1000)
[01:23:03] FINALIZE  test=1.0000 (baseline 0.0000, Δ+1.0000)  best=cand_0001
follower thread alive after poison? False    ← exited on FINALIZE, not on the poison
```

The `ACCEPT` and `FINALIZE` after the poison **render**. Tests: `test_format_event_survives_every_malformed_record`, `test_poisoned_record_does_not_kill_the_follower`, `test_follower_thread_reports_instead_of_dying_silently` (monkeypatches `render_line` to raise and asserts the stderr line — the follower-thread path you correctly noted had **0** coverage).

**B1b — non-dict records escaped into consumers. FIXED.** Filtered at the source in `read_new_events`, so the CLI, the SSE route, and #118/#122/#138/#144 all get a validated stream.

```
$ read_new_events over '42\nnull\n[1,2]\n"str"\ntrue\n{"kind":"finalize"}\n'
[{'kind': 'finalize'}, {'kind': 'log_corruption', 'dropped': 5}]
all isinstance dict: True
```

**B2 — terminal escape injection. FIXED.** One `sanitize()` applied to the **finished line** in `format_event`, so every field of every kind — present and future — is covered on both the terminal and the JSON path. `render_line` delegates, so neither entry point can return an unsafe line.

**B3 — the cost meter double-counted. FIXED.** See the arithmetic section below.

**B4 — `2>&-` corrupted the stdout JSON. FIXED.** `_stderr_is_usable()` runs at spawn time; `_spawn_follower` returns `(None, None)` and following is disabled rather than falling back to stdout. `sys.stderr is None` (what CPython actually gives you under `2>&-`) and the fd-2-reused case are both caught.

```
$ sh -c 'exec 2>&-; python -m cap_evolve.cli run … --follow --dashboard off > cl.out'
rc=0
$ python -c "import json;print('json.loads OK ->', json.load(open('cl.out'))['test_reward'])"
json.loads OK -> 1.0
$ grep -c 'baseline  val=' cl.out
0
```

Test: `test_run_follow_with_stderr_closed_keeps_stdout_valid_json` (real `exec 2>&-` in a subshell, not a mocked stream) + `test_stderr_unusable_disables_following`.

---

### Escape-injection proof

All three of your demonstrated attacks, rendered through the shipped `format_event`:

```
--- OSC title injection
   repr : '[01:22:39] OPTIMIZER ERROR  c1: boom]0;PWNED[2J[31mfake red'
   lines: 1  ESC: 0  BEL: 0  NL: 0
--- screen clear
   repr : '[01:22:39] brand_new payload=[2J[Hcleared screen'
   lines: 1  ESC: 0  BEL: 0  NL: 0
--- newline-forged line
   repr : '[01:22:39] brand_new payload=a⏎FINALIZE  test=1.0000 (baseline 0.0000) best=FAKE'
   lines: 1  ESC: 0  BEL: 0  NL: 0
```

Zero `\033`, zero `\007`, zero newlines, **one line each** — the forged `FINALIZE` is now visibly a payload on the `brand_new` line, not a second progress line. `PWNED` and `boom` survive as inert printable text, because an unreadable error message is its own defect. Newlines collapse to `⏎` rather than vanishing so a multi-line optimizer error stays legible.

Also covered: 8-bit CSI (`\x9b`), DEL (`\x7f`), the whole C1 range `0x80–0x9f`, and TAB deliberately preserved. Tests: `test_escape_injection_from_event_payloads_is_neutralised`, `test_sanitize_strips_c0_c1_and_keeps_tab`, `test_render_line_is_also_escape_safe_with_color` (asserts `color=True` adds *exactly* the styling escapes and none from the payload). Your point that this belongs here, not #144, is right — #144 can now harden `use_color()` on top of a renderer that is already safe.

---

### Cost meter

`_accrue` summed `cost_usd` + `opt_cost_usd` + `usd` off **every** event. Reading the harness: `evaluate` logs runner spend (`harness.py:307-312`) and the following `step` **re-states** the same `cost_usd`/`tokens` alongside the optimizer's own (`harness.py:1322-1328`). The authoritative source per dollar:

| spend | authoritative event | fields | mirrors |
|---|---|---|---|
| runner | `evaluate` only | `cost_usd` / `tokens` | `Spent.usd` |
| optimizer | `step` / `gepa_val_gate` / `skillopt_step` only | `opt_cost_usd` / `opt_tokens` | `Spent.optimizer_usd` |
| intake | `intake` only | `usd` / `tokens` | `Spent.intake_usd` |

which reproduces `Spent.total_usd = usd + optimizer_usd + intake_usd` (`rundir.py:137`) by construction. Live, against a real `RunDir` fed exactly the harness's two-event sequence for your $0.50 runner + $0.20 optimizer:

```
[01:22:45] eval val/cand_0001  reward=1.0000 ±?  1.0s  [$0.5000 · 1.0k tok]
[01:22:45] ACCEPT  cand_0001  val=1.0000 (parent 0.0000)  [$0.7000 · 1.5k tok]

run's recorded total : runner $0.50 + optimizer $0.20 + intake $0.00 = $0.70  (1500 tok)
meter                : $0.70  (1500 tok)
MATCH: True
```

$0.70 for $0.70 — was $1.20. Test `test_cost_meter_matches_the_runs_recorded_total` builds a real `RunDir`, calls `update_spent` + `log_event` the way the harness does, and asserts `meter == rd.spent.total_usd`, so it fails if either side drifts. `test_cost_meter_accumulates_across_events` now feeds the real `evaluate`+`step` pair (with `step` re-stating `cost_usd`) instead of two synthetic `step`s.

**`_accrue` is now public `accrue_totals(ev, totals)`** with the double-count reasoning in its docstring — #138's burn line should call it, not re-derive it.

---

### Hub API — what changed for #118 / #122 / #138 / #144

```python
read_new_events(path, offset) -> tuple[list[dict], int]      # dict-only; shrink → re-read from 0
follow_events(path, *, offset=0, poll=0.5,
              stop_kinds=("finalize",),
              idle_timeout=None,                             # was 300.0
              should_stop: Callable[[dict | None], bool] | None = None,
              ) -> Iterator[dict]                            # last record is the _follow_end sentinel
format_event(ev, totals=None, *, skip_kinds=BOOKKEEPING_KINDS) -> str | None
render_line(ev, totals=None, *, color=False, **kw) -> str | None
accrue_totals(ev, totals) -> None                            # NEW, public (was _accrue)
sanitize(text) -> str                                        # NEW
use_color(stream) -> bool                                    # stream now REQUIRED
FOLLOW_END = "_follow_end"                                   # NEW
BOOKKEEPING_KINDS = (...)                                    # NEW, documented
```

**#118 (stalled vs idle vs done)** — the three bare `return`s are gone. The last record `follow_events` yields is always a sentinel naming which exit fired:

```python
{"kind": "_follow_end", "reason": "stop_kind", "last_kind": "finalize"}
{"kind": "_follow_end", "reason": "idle", "idle_seconds": 300.0}
{"kind": "_follow_end", "reason": "should_stop"}
```

`reason == "idle"` **is** your stall signal, distinguishable from a finished run. `should_stop` is now `Callable[[dict | None], bool]` — it receives the last event seen, so a stall rule can read its `t` without a second poller. And the `idle_timeout=300.0` module default is **gone** (now `None`): each caller owns its threshold, so nothing here pre-decides #118's configurable value. `format_event` returns `None` for the sentinel, so existing renderers ignore it; `_cmd_tail` consumes it for its exit code.

**#122 (replay)** — N3 fixed: `read_new_events` treats a shrunk file as rotation (`if size < offset: offset = 0`) and re-reads, so a rewritten log no longer sticks or silently skips. Combined with B1b that makes a replay over an arbitrary user-supplied log safe, which is the wider trust boundary you flagged.

**#138 (header/sparkline/burn)** — `format_event(ev, skip_kinds=())` renders **every** kind including `minibatch`/`target_profile`/`gepa_merge_*`, so phase detection no longer needs to bypass the renderer; `BOOKKEEPING_KINDS` is the documented set. And `accrue_totals` is public with correct arithmetic, so the burn meter does not inherit the 2x.

**#144 (no-TTY ladder + crash log)** — every rendered line is already sanitised, so the ladder is built on a safe renderer. `use_color(stream)` now **requires** its stream (N5). The `except Exception` in `_spawn_follower` carries a `# 144:` comment marking it as the forensic-crash-log hook, and already prints a user-visible line.

---

### Non-blocking

**N1 — typo'd run dir pretend-waits and exits 0. FIXED.** A path whose parent isn't a directory can never be a run dir → `no such run dir: <root>`, `return 2`. Idle timeout with zero events rendered → `return 3`. Live:

```
$ cap-evolve tail /tmp/does/not/exist/run_typo
no such run dir: /tmp/does/not/exist/run_typo          rc=2
$ cap-evolve tail .capevolve/run_never --idle-timeout 1
timed out after 1s with no events from …/run_never/events.jsonl   rc=3
$ cap-evolve tail <finished run> --from-start                     rc=0
```

**N2 — corrupt line silently dropped. FIXED.** `read_new_events` counts skips (unparseable **and** non-dict) and appends `{"kind": "log_corruption", "dropped": N}`, which renders as `WARNING  N unreadable record(s) skipped in events.jsonl` (yellow on a TTY) and reaches the dashboard too. Visible in the B1 transcript above.

**N3 — offset stuck on truncate/rewrite. FIXED.** See #122 above. `test_read_new_events_rereads_after_truncation`.

**N4 — `--resume --follow` replays the whole prior log. FIXED.** `_spawn_follower` takes an `offset`; on `--resume` it is seeded from the existing `events.jsonl` size, matching what `tail` already did.

**N5 — `use_color()` defaulted to `sys.stdout`. FIXED.** The parameter is now required.

**N6 — `render_line` / `format_event` redundancy. Kept, documented.** As you concluded, the split is defensible once B2 lands in `format_event`. Both the module docstring and `render_line`'s own docstring now state that `format_event` is the **only** place text is made terminal-safe and that `render_line` is styling only, and `render_line` forwards `**kw` so it can never be the "lesser" entry point.

**N7 — documented meter output baked in the bug. FIXED.** `docs/GETTING_STARTED.md` halved to the true figures and gained a paragraph on what the meter counts (and that event text is sanitised). `tail` is now also mentioned in `docs/INSTALL.md` next to `cap-evolve dashboard`, plus the exit codes are documented in the `cli.py` usage docstring.

**N8 — `PYTHONIOENCODING=ascii` mangling. DECLINED — agreed, #144's.** You called it correctly: no line is lost, the follower doesn't die, and the encoding ladder is #144's headline. Fixing it here would mean picking an ASCII fallback vocabulary that #144 will want to own.

### Nits

1. **Unused-ish test imports. DECLINED (now moot).** `os`/`shutil` are used by the shared `_toy_project()` helper, which both e2e tests (the original and the new `2>&-` one) now call, so they're genuinely module-level. Import order alphabetised.
2. **`--idle-timeout -5` accepted. FIXED.** `p.error("--idle-timeout must be >= 0 (0 = wait forever)")`. `test_cmd_tail_rejects_negative_idle_timeout`.
3. **`stop_kinds=("finalize",)` default. Kept.** It's the right default for both terminal surfaces and is overridable; a dashboard/replay consumer passes `stop_kinds=()`. Unlike `idle_timeout`, this one doesn't pre-decide a downstream issue's configurable knob, so it isn't legislating.

---

### Correction to my earlier claim

You're right and I was wrong. My evidence block said the fastapi collection failures are "**pre-existing on `origin/main`**". That framing is inaccurate: with `fastapi` installed, `origin/main` collects **all 42 dashboard tests with zero errors** — as you verified. The 6 modules error out **only when `fastapi` is absent**, which is a property of the *environment*, not of `main`. The substantive point stands (unrelated to this PR; the dashboard `dev` extra makes all 42 pass), but the accurate statement is: *these 6 modules require `fastapi`, which is a dashboard dev extra and is not installed in a bare checkout — install `dashboard/backend[dev]` and all 42 pass on both `main` and this branch.*

---

### Re-verification

```
217 passed in 66.50s (0:01:06)

$ python -m compileall -q core skills; echo "rc=$?"
rc=0

$ cd dashboard/backend && PYTHONPATH=../../core:. python -m pytest tests -q
42 passed, 1 warning in 1.90s
```

**Live SSE after the `read_new_events` change** (a `step`, a bare `42`, and a `finalize` appended mid-connection):

```
GET /api/runs -> 200
stream: 200 text/event-stream; charset=utf-8
   event: snapshot   data: {"run_id": "run_demo", "path": "…/run_demo", "graph": {"nodes": [{"id": "seed", …
   event: event      data: {"t": 1785363864.577114, "kind": "step", "candidate": "sse_probe", "accept": true, "val": 1.0}
   event: event      data: {"t": 1785363864.57713,  "kind": "finalize", "test_reward": 1.0}
   event: done       data: {"run_id": "run_demo"}
```

The bare `42` never reached the client — filtered at the source, exactly what B1b asked for.

**Real zero-API `toy_calc` + `mock` `--follow`, both streams piped** — unchanged and still clean:

```
[01:23:23] splits frozen  train=4 val=2 test=2 (test sealed)
[01:23:23] eval val/seed  reward=0.0000 ±0.0000  0.0s
[01:23:23] baseline  val=0.0000 ±0.0000
[01:23:24] eval val/cand_0001  reward=1.0000 ±0.0000  0.0s
[01:23:24] gate warning (paired): combined/paired SE is 0 (likely n_trials=1 or identical trials)
[01:23:24] ACCEPT  cand_0001  val=1.0000 (parent 0.0000)  — paired Δ̄=+1.0000 > 0 (SE=0 → STRICT fallback, warned; n=2)
[01:23:25] reject  cand_0002  val=1.0000 (parent 1.0000)  — paired Δ̄=+0.0000 <= 0 (…)
[01:23:25] reject  cand_0003  val=1.0000 (parent 1.0000)  — paired Δ̄=+0.0000 <= 0 (…)
[01:23:26] eval test/FINAL  reward=1.0000 ±0.0000  0.0s
[01:23:26] FINALIZE  test=1.0000 (baseline 0.0000, Δ+1.0000)  best=cand_0001

$ grep -c $'\033' v1.err v1.out
v1.out:0
v1.err:0
$ grep 'baseline  val' v1.err | od -c | head -3
0000000    [   0   1   :   2   3   :   2   3   ]       b   a   s   e   l
0000020    i   n   e           v   a   l   =   0   .   0   0   0   0
0000040  302 261   0   .   0   0   0   0  \n
```

Only the UTF-8 `±` (`302 261`). **`tail` attaching before the run dir exists**, run launched *without* `--follow`:

```
$ cap-evolve tail .capevolve/run_live --from-start --idle-timeout 60 &
$ cap-evolve run … --run-ts live --dashboard off > run.json 2> run.err
waiting for /tmp/v3/.capevolve/run_live/events.jsonl …
[01:24:03] splits frozen  train=4 val=2 test=2 (test sealed)
…
[01:24:05] FINALIZE  test=1.0000 (baseline 0.0000, Δ+1.0000)  best=cand_0001
tail rc=0

$ wc -c < run.err
       0          ← default behavior still provably unchanged for scripts
```

### Files touched

| File | Change |
|---|---|
| `core/cap_evolve/eventstream.py` | dict filter + shrink re-read + corruption count; `_follow_end` sentinel; event-aware `should_stop`; `idle_timeout` default → `None`; `sanitize`; total `format_event` + `skip_kinds`; `accrue_totals` (public, corrected); `use_color(stream)` required |
| `core/cap_evolve/cli.py` | `_stderr_is_usable`; follower reports instead of swallowing; resume offset; `tail` exit codes 2/3; negative-timeout clamp; usage docstring |
| `core/tests/test_eventstream.py` | 21 → **38** tests (malformed records, follower-thread death, non-dict filtering, truncation, all three escape attacks, real-`RunDir` cost equality, real `2>&-` e2e, tail exit codes, `skip_kinds`) |
| `docs/GETTING_STARTED.md` | corrected meter figures; meter + sanitisation + exit codes documented |
| `docs/INSTALL.md` | `--follow` / `tail` listed alongside `cap-evolve dashboard` |
| `CHANGELOG.md` | hardening paragraph under the existing `--follow` entry |

`harness.py`, `rundir.py`, `loop.py`, `gepa.py`, `skillopt.py` and `dashboard/backend/…/stream.py` remain untouched — the cost fix is a read of what the harness already logs, not a change to it.

Issue #123, opened 2026-07-25
Site: add mobile nav (hamburger) + shared-chrome include to end 9-file copy-paste/cache-buster drift


OsherElhadad:
## Problem

The GitHub Pages site (`site/`) has one clear responsive defect and a documented maintenance hazard that is already causing drift.

1. **No mobile nav.** Below 860px the nav just becomes a horizontally-scrollable strip (`site/style.css:620-623`) with 6 links + GitHub + theme toggle — no hamburger/menu. On a phone the links overflow into a scroll region that's easy to miss.
2. **Shared chrome is copy-pasted across 9 HTML files** (nav, footer, `<head>` theme/font block). `site/README.md:59-61` openly documents this as a manual "update every file" hazard, and it has **already drifted** (per-page OG/twitter descriptions differ subtly).
3. **Manual cache-buster** `?v=YYYYMMDD` on every CSS/JS link must be hand-bumped in 9 files per edit (`site/README.md:62-64`) — same rot risk (issue #87 in history was literally a missed `?v` bump).

## Why it matters

The site is the best-executed surface of the project and the first impression for many visitors. A broken mobile nav and guaranteed-to-drift duplicated chrome undercut that polish, and the manual cache-buster has already caused a shipped bug.

## What to change

1. Add a real mobile nav (hamburger toggle) for < ~860px. Keep it dependency-free (small vanilla JS + CSS) to match the site's zero-framework ethos.
2. Kill the 9-file copy-paste: introduce a tiny build/include step (or a small JS partial injected at runtime) for nav/footer/`<head>` so shared chrome lives in one place. Keep the "zero build step" spirit if possible (runtime JS include is acceptable).
3. Automate the cache-buster (derive `?v` from a build timestamp / file hash) so it can't be forgotten.

## How to test

- On a < 860px viewport (or device emulation) the nav collapses to a working menu; all destinations reachable.
- Editing the nav in one place updates all 9 pages (or the runtime include renders identical chrome everywhere).
- OG/twitter/canonical meta are consistent across pages (or intentionally per-page from a single source).

## Dependencies / parallelization

- **Parallel-safe** with all other issues (site-only). The shared-chrome refactor should land before further per-page content edits to avoid re-introducing drift.

Issue #119, opened 2026-07-25
Dashboard re-reduces the full event log on every request/SSE tick — cache by mtime; drop dead SSE snapshot


OsherElhadad:
## Problem

The dashboard backend **re-folds the entire event log on every request and every SSE tick**, and ships a dead SSE snapshot payload.

- `runs.load_run` → `reduce_run` re-reads and re-folds the whole `events.jsonl` + all rollout files on every call (`dashboard/backend/capevolve_dashboard/runs.py:38-40`).
- The SPA refetches the full reduced run on **every** live event (400ms-debounced `invalidateQueries` — `RunDeepDive.tsx:50-56`), so a long run re-reduces the full log ~twice a second.
- `diff_candidate` calls `reduce_run` + `build_diffs` **per candidate** (`trajectories.py:58-63`). The static exporter already noted this is "far too slow" and reduces once (`export_static.py:76-77`), but the live path never got that fix.
- The SSE `snapshot` frame serializes the full reduced run (`app.py:103`), but the client ignores its payload and just marks the connection open (`useRunStream.ts:75`), then separately refetches. Wasted CPU/bandwidth on every connect.
- No pagination anywhere: `/api/runs` returns all runs; `list_rollouts` globs + JSON-parses every rollout file in a split each call (`trajectories.py:23-37`).

## Why it matters

For long runs and large `.capevolve` dirs this is quadratic-ish waste that makes the "real-time" dashboard laggy exactly when it matters most (a big, long run). It's also the reason the SPA feels like poll-on-event rather than true streaming.

## What to change

1. **Memoize `reduce_run` by `events.jsonl` mtime+size** (and invalidate on change). This single cache fixes the per-request and per-SSE-tick re-fold and the per-candidate `diff_candidate` re-reduce.
2. Either **drop the SSE `snapshot` payload** (send just an "open" frame) or **use it** to seed the client query cache and skip the first refetch.
3. Paginate `/api/runs` and `list_rollouts` for large dirs.
4. (Stretch) Have the SPA apply incremental updates from stream events instead of full re-fetch on each event.

## How to test

- Backend test: two consecutive `load_run` calls with an unchanged event log hit the memoized reduction (assert the fold runs once, e.g. via a counter/spy); a changed log re-folds.
- Measure: on a run with many candidates, live event handling no longer triggers a full re-reduce each tick.
- Existing backend tests (9 files) still pass.

## Dependencies / parallelization

[...]

PR #192, opened 2026-07-29
Remove the CDN font import from the SPA (offline/air-gapped; zero-CDN ethos)

### LINKED ISSUE #120: Remove the CDN font import from the SPA (breaks offline/air-gapped; contradicts zero-CDN ethos)
## Problem

The dashboard SPA imports fonts from a CDN, which breaks offline/air-gapped use and contradicts the project's stated zero-CDN, offline-first ethos.

- `dashboard/frontend/src/index.css:1` — `@import url('https://fonts.googleapis.com/css2?family=Fira+Code...&family=Fira+Sans...')`.
- The self-contained `dashboard.html` (`core/cap_evolve/dashboard.py` `_HTML_TEMPLATE`) deliberately avoids all CDNs; the SPA silently reintroduces one.

In an air-gapped eval environment (a common setting for agent benchmarks) the UI loads with fallback fonts after a network stall, and the "runtime deps: 0 / offline" story quietly breaks.

## Why it matters

Small but on-brand: the project markets zero runtime deps and an offline self-contained dashboard. A blocking CDN font request in the SPA is an easy, embarrassing inconsistency to close, and it causes a real UX stall in air-gapped environments.

## What to change

- Bundle Fira Sans / Fira Code via the Vite build (self-host the woff2 files as an asset), **or** drop the webfont and use the system font stack.
- Remove the `@import` from `index.css:1`.

## How to test

- Build the SPA, load it with the network disabled (or DevTools offline): fonts render correctly with no request to `fonts.googleapis.com`.
- `grep -rn "googleapis\|fonts.gstatic\|cdn" dashboard/frontend/src` returns nothing.

- **Parallel-safe.** Trivial, self-contained. Good first issue.

OsherElhadad:
Closes #120

## Problem

`dashboard/frontend/src/index.css:1` `@import`ed Fira Sans / Fira Code from `fonts.googleapis.com`. A CSS `@import` of a stylesheet is **render-blocking**, so every SPA load made a cross-origin request before painting. In an air-gapped eval environment (a common setting for agent benchmarks) that stalls until the request times out, then falls back to system fonts anyway — so the network round-trip bought literally nothing and cost a visible delay.

Meanwhile `core/cap_evolve/dashboard.py`'s `_HTML_TEMPLATE` deliberately ships zero external resources, and `core/tests/test_dashboard.py` asserts it. The SPA quietly broke the same guarantee with no guard to catch it — which is exactly how this drifted in.

## Every CDN reference removed

I grepped the whole repo, not just the file the issue named. Three shipped surfaces carried it:

| File | What was there |
|---|---|
| `dashboard/frontend/src/index.css` | `@import url('https://fonts.googleapis.com/css2?family=Fira+Code…&family=Fira+Sans…')` |
| `dashboard/frontend/tailwind.config.js` | `fontFamily.sans/mono` naming `"Fira Sans"` / `"Fira Code"` |
| `dashboard/frontend/dist/*` | rebuilt — the old committed bundle had the `@import` inlined |
| `skills/algorithms/evograph/dashboard/frontend/dist/index.html` | 2 `<link rel="preconnect">` + 1 font `<link rel="stylesheet">` |
| `skills/algorithms/evograph/dashboard/frontend/dist/assets/*.css` | 9 `Fira` font-family declarations |
| `examples/tau2_airline/run_full/ui/assets/*.css` | the `@import` + 5 `Fira` declarations |

The two prebuilt bundles matter as much as the source: **both are served at runtime** (`evograph/dashboard/backend/app.py` mounts its `dist/` via `StaticFiles`; the tau2 export is a documented standalone static site). Fixing only `src/` would have left the actually-shipped artifacts still calling out to Google.

They are patched **in place** rather than regenerated. The evograph SPA has **no source in this repo** — only `dist/` is committed — so in-place is the only option there, and I used it for the tau2 snapshot too rather than swapping in a fresh 830 kB JS bundle that could drift from the frozen snapshot's data shapes. The edit is textual and narrow: drop the `@import`, substitute the font-family lists. Both files are verified to still parse as valid CSS below.

## Replacement font stack

No webfont, no vendored binary — native system stacks:

```css
/* sans */ ui-sans-serif, system-ui, -apple-system, 'Segoe UI', Roboto,
           'Helvetica Neue', Arial, 'Noto Sans', sans-serif
/* mono */ ui-monospace, SFMono-Regular, 'SF Mono', Menlo, Consolas,
           'Liberation Mono', monospace
```

**Why not self-host the woff2?** The issue offered both options. Fira was decorative, not load-bearing: the previous declarations *already* listed `ui-sans-serif, system-ui` as fallbacks, so any user behind a firewall was seeing system fonts regardless. Self-hosting means committing ~10 font weights of binary to git and wiring a font-asset pipeline into the Vite build — a real maintenance cost for a typeface nothing depends on. The `ponytail:` comment in `index.css` records the upgrade path if a future design ever needs Fira's exact metrics. No build dependency was added.

## Why offline/air-gapped now works

The shipped bundle contains **zero** external URLs, so there is no request to block on and no timeout to wait out. `ui-sans-serif`/`ui-monospace` resolve to the OS UI font on every platform, so the UI paints immediately with correct typography instead of after a network stall. The `tabular-nums` treatment on score columns is a `font-variant-numeric` property, not a Fira feature, so numeric columns still don't jitter.

## Regression guard

`core/tests/test_dashboard.py` gained `test_shipped_spa_bundles_have_no_external_cdn_reference`, extending the existing self-contained-HTML assertion to the SPA source, all three shipped bundles, and 5 banned CDN hosts. Verified to actually fail when the `@import` is reintroduced (output below) — a guard that can't fail is not a guard.

## Left for #123

`site/*.html` carries the same Google Fonts links — lines 21-23 of all 9 pages (2 preconnects + 1 stylesheet), plus `Fira` in `site/style.css`. **Deliberately untouched:** #123 is refactoring that exact shared `<head>` block into a single include across all 9 files, so editing these lines here would guarantee a conflict. Coordinated directly with the #123 implementer, who is applying the same system-font stack as part of that refactor.

### Build

```

> frontend@0.0.0 build
> tsc -b && vite build

transforming...✓ 2809 modules transformed.
rendering chunks...
computing gzip size...
dist/assets/index-DJzo6AHL.css   16.45 kB │ gzip:   4.51 kB
dist/assets/index-zqmQoE4e.js   830.26 kB │ gzip: 249.87 kB

✓ built in 787ms
```

(The >500 kB chunk-size hint is pre-existing and unrelated.)

### Zero-hit grep over the BUILT bundle

The key evidence — clean source proves nothing if the shipped artifact still calls out:

```
$ grep -rcE "googleapis|gstatic|Fira" dashboard/frontend/dist/
dashboard/frontend/dist/index.html:0
dashboard/frontend/dist/favicon.svg:0
dashboard/frontend/dist/assets/index-DJzo6AHL.css:0
dashboard/frontend/dist/assets/index-zqmQoE4e.js:0
dashboard/frontend/dist/icons.svg:0
grep-exit=1        # 1 = no matches anywhere
```

Repo-wide, excluding `site/` (issue #123):

```
$ grep -rniE "googleapis|gstatic|unpkg|jsdelivr|cdnjs" . | grep -v node_modules | grep -v package-lock | grep -v '^site/'
exit=1 (1 = zero hits outside site/)
```

### Tests

```
$ PYTHONPATH=/tmp/wt-120/core python -m pytest core/tests -q
....................................                                     [100%]
180 passed in 63.13s (0:01:03)
```

179 baseline + 1 new guard, 0 failed.

### The new guard actually fails on regression

```
$ # reintroduce the @import, then:
$ python -m pytest core/tests/test_dashboard.py -q -k cdn
core/tests/test_dashboard.py:275: AssertionError
FAILED core/tests/test_dashboard.py::test_shipped_spa_bundles_have_no_external_cdn_reference
1 failed, 10 deselected in 0.02s

$ # restored:
1 passed, 10 deselected in 0.02s
```

### Patched-in-place bundles still parse as valid CSS

```
$ node -e "postcss.parse(...)"
examples/tau2_airline/run_full/ui/assets/index-vwBf6wRK.css -> parsed OK, rules: 273 | bytes: 16457
skills/algorithms/evograph/.../index-juGJTXVt.css           -> parsed OK, rules: 432 | bytes: 39091
```

### UI renders — served, screenshotted, network-verified

Served the tau2 static export (the real SPA with a completed 10-iteration run) and loaded it in headless Chrome. The dashboard renders correctly in system fonts: sidebar + logo, `run_full` with a green "done" badge, all 12 KPI tiles reading correctly (BEST 71.2%, BASELINE 53.6%, Δ +32.8%, SEALED TEST 69.4%, COST $148.43, TOKENS 288.6K, ACCEPTED 5 / REJECTED 5 / FAILED 0, WALL CLOCK 10h 23m), the full tab bar, and the "Fitness over iterations" chart with all 11 accepted/rejected points and the amber champion star. Numeric columns stay aligned via `tabular-nums`. No layout breakage, no clipping. (No FOUT/FOIT claim is made from the screenshot — a settled screenshot cannot show a transient reflow; the actual argument is structural: with no webfont there is no swap period to reflow through.)

Every request the page made, from the server's access log:

```
"GET / HTTP/1.1" 200
"GET /assets/index-vwBf6wRK.css HTTP/1.1" 200
"GET /assets/index-ez7sAMzg.js HTTP/1.1" 200
"GET /cap-evolve-logo.png HTTP/1.1" 200
"GET /data/runs_run_full.json HTTP/1.1" 200
"GET /favicon.png HTTP/1.1" 200
```

Six requests, all local. **Re-run under genuine DNS denial** (`--host-resolver-rules="MAP * ~NOTFOUND, EXCLUDE localhost"`): the dashboard renders identically and still makes exactly those six local requests. Control under the same flag: `https://fonts.googleapis.com/css2?family=Fira+Code` returns Chrome's `DNS_PROBE_POSSIBLE` error interstitial instead of CSS, proving resolution really was denied.

Also confirmed the live backend picks up the rebuild — `resolve_static_dir()` → `dashboard/frontend/dist`, serving the new hashed `index-zqmQoE4e.js` / `index-DJzo6AHL.css`.

### House rules

- Test added for the change (`core/tests/`) — required by CONTRIBUTING.
- `python -m compileall core skills` → exit 0.
- `npx tsc -b --noEmit` → exit 0.
- No new runtime or build dependency; honesty/gate/seal logic untouched.
- `npx eslint src` reports 7 problems (6 errors / 1 warning) spanning **6 files** — `components/CountUp.tsx:21:7`, `components/DiffRows.tsx:6:14`, `components/FileTree.tsx:24:9` (warning), `components/GitDiff.tsx:24:44`, `lib/useRunStream.ts:63:3`, `main.tsx:18:7`, `routes/Hub.tsx:31:7`. None of those files is touched by this PR, and the identical set appears on `origin/main`. (An earlier revision of this body wrongly attributed all 7 to `Hub.tsx` — corrected.)


OsherElhadad:
## 🔬 Evidence

Every command and its literal output. Worktree: `/tmp/wt-120` on `fix/issue-120-cdn-font` off `origin/main` @ `6fca097`.

---

### 1. Baseline: tests before any change

```
$ cd /tmp/wt-120 && PYTHONPATH=/tmp/wt-120/core /tmp/ce-venv/bin/python -m pytest core/tests -q
179 passed in 65.35s (0:01:05)
```

```
$ node -v && npm -v
v25.1.0
11.6.2
```

---

### 2. Discovery: every CDN reference in the repo

```
$ grep -rniE "googleapis|gstatic|unpkg|jsdelivr|cdnjs" . | grep -v node_modules | grep -v package-lock
dashboard/frontend/dist/assets/index-b_9cC4ZS.css:1:@import "https://fonts.googleapis.com/css2?family=Fira+Code:wght@400;500;600;700&family=Fira+Sans:wght@300;4...
dashboard/frontend/src/index.css:1:@import url('https://fonts.googleapis.com/css2?family=Fira+Code:wght@400;500;600;700&family=Fira+Sans:wght@300;400;500;600;70...
examples/tau2_airline/run_full/ui/assets/index-vwBf6wRK.css:1:@import "https://fonts.googleapis.com/css2?family=Fira+Code:wght@400;500;600;700&family=Fira+Sans:...
site/architecture.html:21:  <link rel="preconnect" href="https://fonts.googleapis.com">
site/architecture.html:22:  <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
site/architecture.html:23:  <link href="https://fonts.googleapis.com/css2?family=Fira+Code...&family=Fira+Sans...&display=swap" rel="stylesheet">
   ... (identical 3-line block in all 9 site/*.html: index, architecture,
        run-end-to-end, adapter-templates, getting-started, agent-orchestration,
        results, benchmarks, optimize-your-own)
skills/algorithms/evograph/dashboard/frontend/dist/index.html:7:    <link rel="preconnect" href="https://fonts.googleapis.com" />
skills/algorithms/evograph/dashboard/frontend/dist/index.html:8:    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
skills/algorithms/evograph/dashboard/frontend/dist/index.html:10:      href="https://fonts.googleapis.com/css2?family=Fira+Code:wght@400;500;600;700&family=Fira...
```

So: the file the issue named, **plus** two committed prebuilt bundles served at runtime, plus `site/`.

Confirming the prebuilt bundles are genuinely shipped, not dead weight:

```
$ grep -n "dist\|StaticFiles\|index.html" skills/algorithms/evograph/dashboard/backend/app.py
25:from fastapi.staticfiles import StaticFiles
340:    _dist: Path = DIST
342:    if (_dist / "assets").is_dir():
343:        app.mount("/assets", StaticFiles(directory=str(_dist / "assets")), name="assets")
347:        return FileResponse(str(_dist / "index.html"))
```

And that `dist/` is committed on purpose, not a stale artifact:

```
$ cat dashboard/frontend/.gitignore
# dist is intentionally COMMITTED: the backend serves this prebuilt SPA at runtime
# so the live dashboard works with no node/build step (see dashboard_launch). Keep it tracked.
# dist
```

The evograph SPA has **no source in-repo** — only `dist/` — which is why it is patched in place:

```
$ ls -a skills/algorithms/evograph/dashboard/frontend/
.  ..  dist
```

Font-family occurrences to substitute in the frozen bundles:

```
$ grep -oE 'Fira (Sans|Code)' <each file> | sort | uniq -c
examples/tau2_airline/run_full/ui/assets/index-vwBf6wRK.css:  3 Fira Code, 2 Fira Sans, 1 @import
skills/algorithms/evograph/.../index-juGJTXVt.css:            7 Fira Code, 2 Fira Sans, 0 @import
```

---

### 3. Source fix

`dashboard/frontend/src/index.css` — `@import` deleted, both stacks replaced; `dashboard/frontend/tailwind.config.js` — `fontFamily.sans/mono` de-Fira'd. Full diff in the PR.

### 4. Frozen bundles patched in place

```
$ python <strip script>
examples/tau2_airline/run_full/ui/assets/index-vwBf6wRK.css changed | residual Fira: 0 | residual googleapis: 0
skills/algorithms/evograph/dashboard/frontend/dist/assets/index-juGJTXVt.css changed | residual Fira: 0 | residual googleapis: 0
```

```
$ python <strip font links from evograph index.html>
changed | residual: 0
```

Resulting `<head>` — 5 lines gone, nothing else disturbed:

```html
<!doctype html>
<html lang="en" class="dark">
  <head>
    <meta charset="UTF-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
    <title>EvoGraph</title>
    <script type="module" crossorigin src="./assets/index-BGSNdw8n.js"></script>
    <link rel="stylesheet" crossorigin href="./assets/index-juGJTXVt.css">
  </head>
```

Both patched CSS files still parse (a bad regex on a 39 kB minified file would silently corrupt a rule):

```
$ node -e "const postcss=require('postcss') ... postcss.parse(css)"
../../examples/tau2_airline/run_full/ui/assets/index-vwBf6wRK.css -> parsed OK, rules: 273 | bytes: 16457
../../skills/algorithms/evograph/dashboard/frontend/dist/assets/index-juGJTXVt.css -> parsed OK, rules: 432 | bytes: 39091
```

---

### 5. Build

```
$ cd dashboard/frontend && npm ci
   (completed; only the pre-existing `npm audit fix --force` advisory notice)

$ npm run build

> frontend@0.0.0 build
> tsc -b && vite build

transforming...✓ 2809 modules transformed.
rendering chunks...
computing gzip size...
dist/assets/index-DJzo6AHL.css   16.45 kB │ gzip:   4.51 kB
dist/assets/index-zqmQoE4e.js   830.26 kB │ gzip: 249.87 kB

✓ built in 787ms
[plugin builtin:vite-reporter]
(!) Some chunks are larger than 500 kB after minification. Consider: ...
```

The chunk-size hint is pre-existing (unrelated to fonts).

---

### 6. Zero-hit grep over the BUILT bundle — the key evidence

```
$ grep -rcE "googleapis|gstatic|Fira" dashboard/frontend/dist/
dashboard/frontend/dist/index.html:0
dashboard/frontend/dist/favicon.svg:0
dashboard/frontend/dist/assets/index-DJzo6AHL.css:0
dashboard/frontend/dist/assets/index-zqmQoE4e.js:0
dashboard/frontend/dist/icons.svg:0
grep-exit=1
```

Every shipped file: 0. `grep-exit=1` = no matches anywhere in the tree.

Repo-wide, excluding `site/` (issue #123):

```
$ grep -rniE "googleapis|gstatic|unpkg|jsdelivr|cdnjs" . | grep -v node_modules | grep -v package-lock | grep -v '^site/'
exit=1 (1 = zero hits outside site/)
```

---

### 7. Regression guard, and proof it can fail

```
$ PYTHONPATH=/tmp/wt-120/core python -m pytest core/tests/test_dashboard.py -q
11 passed in 0.10s
```

Re-adding the `@import` to `src/index.css`:

```
$ python -m pytest core/tests/test_dashboard.py -q -k cdn
core/tests/test_dashboard.py:275: AssertionError
FAILED core/tests/test_dashboard.py::test_shipped_spa_bundles_have_no_external_cdn_reference
1 failed, 10 deselected in 0.02s
```

Restored:

```
1 passed, 10 deselected in 0.02s
```

---

### 8. Full suite + static checks

```
$ PYTHONPATH=/tmp/wt-120/core python -m pytest core/tests -q
....................................                                     [100%]
180 passed in 63.13s (0:01:03)
```

179 baseline → 180 (the new guard). 0 failed.

```
$ python -m compileall -q core skills   ->  compileall-exit=0
$ npx tsc -b --noEmit                   ->  tsc-exit=0
```

`eslint`, with the pre-existing failures isolated by stashing my changes:

```
$ npx eslint src                      # with my changes
/private/tmp/wt-120/dashboard/frontend/src/routes/Hub.tsx
  31:7  error  Expected an assignment or function call and instead saw an expression
✖ 7 problems (6 errors, 1 warning)

$ git stash && npx eslint src          # clean origin/main
  31:7  error  Expected an assignment or function call and instead saw an expression
✖ 7 problems (6 errors, 1 warning)
```

Byte-identical. All in `src/routes/Hub.tsx`, which this PR does not touch.

---

### 9. Rendered UI

Backend resolves the rebuilt bundle:

```
$ python -c "from capevolve_dashboard.server import resolve_static_dir; ..."
static dir: /private/tmp/wt-120/dashboard/frontend/dist
index.html refs: ['<script type="module" crossorigin src="/assets/index-zqmQoE4e.js"></script>',
                  '<link rel="stylesheet" crossorigin href="/assets/index-DJzo6AHL.css">']
```

`uvicorn` isn't in the venv, so I rendered the **tau2 static export** instead — the same React SPA, built from the same `src/`, carrying a real completed 10-iteration run. Better evidence than an empty run dir: it exercises charts, KPI tiles, and tabular numerals with actual data.

```
$ cd examples/tau2_airline/run_full/ui && python -m http.server 8129 &
$ curl -sI http://localhost:8129/ | head -3
HTTP/1.0 200 OK
Server: SimpleHTTP/0.6 Python/3.14.2

$ curl -s http://localhost:8129/assets/index-vwBf6wRK.css | grep -c "googleapis"
0
```

```
$ "Google Chrome" --headless --window-size=1440,1000 --virtual-time-budget=8000 \
    --screenshot=/tmp/shot120/dash.png "http://localhost:8129/#/runs/run_full"
67560 bytes written to file /tmp/shot120/dash.png
```

**What I observed in the screenshot:** the dashboard renders cleanly in system fonts. Left sidebar with the cap·evolve logo, Runs / Compare nav, "watch capability evolve" footer. Header `run_full` with a green ✓ done badge. All 12 KPI tiles legible and correctly aligned — BEST **71.2%** (amber), BASELINE 53.6%, Δ VS BASELINE +32.8%, SEALED TEST **69.4%** (amber, `pass^k NaN%` sub-label), COST (OPT+RUN) $148.43, TOKENS 288.6K, ACCEPTED 5 (green), REJECTED 5 (red), FAILED 0, FRONTIER 1, WALL CLOCK 10h 23m, $/+1% $4.53. Full tab bar (Overview / Cost / Phases / Lineage / Trajectories / Iterations / Git diffs / Memory / Files / Insights) with Overview underlined active. "Fitness over iterations" chart renders all 11 points across iterations 0-10 — green-ringed accepted, red rejected, amber ★ champion at iteration 7 — with correct 0/25/50/75/100 y-axis and gridlines. Numeric values stay column-aligned (`tabular-nums` is a font-variant property, not a Fira feature). No layout breakage, no clipped text, no fallback-font reflow.

**Every request the page made** (server access log — the definitive list):

```
::1 - - [30/Jul/2026 00:27:18] "GET / HTTP/1.1" 200 -
::1 - - [30/Jul/2026 00:27:18] "GET /assets/index-vwBf6wRK.css HTTP/1.1" 200 -
::1 - - [30/Jul/2026 00:27:18] "GET /assets/index-ez7sAMzg.js HTTP/1.1" 200 -
::1 - - [30/Jul/2026 00:27:18] "GET /cap-evolve-logo.png HTTP/1.1" 200 -
::1 - - [30/Jul/2026 00:27:18] "GET /data/runs_run_full.json HTTP/1.1" 200 -
::1 - - [30/Jul/2026 00:27:18] "GET /favicon.png HTTP/1.1" 200 -
```

Six requests, all local, zero font requests.

Chrome net-log capture of the same load:

```
$ "Google Chrome" --headless --log-net-log=... --net-log-capture-mode=IncludeSensitive ...
$ python <extract URLs from net-log>
fonts.googleapis / gstatic hits: 0
```

The only external URLs in the net-log were Chrome's own browser telemetry (`clients2.google.com/service/update2/crx`, `clientservices.googleapis.com/chrome-variations/seed`, safebrowsing, accounts) — browser-initiated, not page-initiated, and absent from the server log because the page requested nothing beyond the six local assets above.

---

### 10. `site/` left for #123

Untouched by this PR: lines 21-23 of all 9 `site/*.html` (2 preconnects + 1 font stylesheet each), plus `Fira` in `site/style.css`. #123 collapses that duplicated `<head>` block into a single shared include across all 9 files, so touching these lines here is a guaranteed conflict. Coordinated with the #123 implementer, who is applying the identical system-font stack as part of the refactor.

### 11. Commit authorship

```
$ git log -1 --format='%an <%ae> / %cn <%ce>'
Osher Elhadad <Osher.Elhadad@ibm.com> / Osher Elhadad <Osher.Elhadad@ibm.com>
```

No `Co-Authored-By`, no generated-with line.

OsherElhadad:
## 🔍 Review — PR #192

The core fix is right: `src/index.css` loses the `@import`, the main SPA `dist/` is a *real* rebuild that I reproduced byte-for-byte, both hand-patched CSS bundles parse cleanly and the edits are surgically confined to `font-family:` value lists, and the new guard genuinely fails when the `@import` returns. Nothing here blocks merge. But one shipped, served bundle still hard-codes `Fira Code` (untouched by this PR), the guard is fully hard-coded and silently stops guarding on any path rename, and the PR's eslint claim misattributes 6 of 7 findings to a single file.

---

### Blocking

None.

---

### Non-blocking

**1. `skills/algorithms/evograph/dashboard/frontend/dist/assets/index-BGSNdw8n.js` — 4 surviving `Fira Code` references in a served bundle (byte offsets 843736, 843850, 1033530, 1210211).**
The PR de-Fira'd the evograph *CSS* bundle and stripped the `<link>` tags from its `index.html`, but the sibling JS bundle in the same `dist/` still ships Recharts/React-Flow inline styles:

```js
fontFamily:"Fira Code, monospace"   // ×4: tooltip contentStyle, legend wrapperStyle, edge labelStyle
```

This file is served by `skills/algorithms/evograph/dashboard/backend/app.py:343`. Consequence: functionally harmless offline (`Fira Code` is now unresolvable so it falls through to `monospace` — no network request, since the `@import`/`<link>` that would have fetched it are gone), but it is dead, misleading state in a shipped artifact, and it means the evograph chart tooltips/edge labels render in a *different* monospace fallback chain than the one the CSS now specifies (`ui-monospace, SFMono-Regular, 'SF Mono', Menlo, Consolas, 'Liberation Mono', monospace`). On a Linux box with no `Menlo`/`Consolas`, chart labels and surrounding CSS-styled mono text will pick different faces. Also: the guard test at `core/tests/test_dashboard.py:256-276` walks `.js` files under that exact directory, so if anyone later re-adds a CDN there it *would* be caught — but the stale `Fira` itself is invisible to the guard (it bans hosts, not font names).
**Fix:** same regex pass the author already ran, extended to the JS bundle — replace `"Fira Code, monospace"` with the project's mono stack (or just `"ui-monospace, monospace"`), 4 occurrences, in the same commit. Alternatively defer to the evograph-rebuild issue in finding 2 and note it in the PR body so it isn't lost.

**2. `skills/algorithms/evograph/dashboard/frontend/` contains only `dist/` — a shipped, served bundle with no source in the repo. Confirmed:**

```
$ ls -a skills/algorithms/evograph/dashboard/frontend/
.  ..  dist
$ find . -name "*.tsx" -path "*evograph*"
(no output)
```

Consequence: 1.2 MB of JS and 39 kB of CSS are mounted at runtime (`app.py:343-354`) and cannot be rebuilt, audited, dependency-scanned, or patched except by regex — which is exactly why this PR had to hand-edit it. That is a reproducibility/supply-chain problem independent of fonts. **Not this PR's job to fix. File a separate issue** ("vendor or restore the evograph frontend source, or document it as a vendored third-party bundle"). Flagging so it's on the record.

**3. PR body §8 misattributes the eslint findings.** The body says "All in `src/routes/Hub.tsx`, which this PR does not touch." The 7 problems actually span 6 files:

```
src/components/CountUp.tsx:21:7      error  setState synchronously within an effect
src/components/DiffRows.tsx:6:14     error  react-refresh/only-export-components
src/components/FileTree.tsx:24:9     warning react-hooks/exhaustive-deps
src/components/GitDiff.tsx:24:44     error  setState synchronously within an effect
src/lib/useRunStream.ts:63:3         error  Cannot access refs during render
src/main.tsx:18:7                    error  react-refresh/only-export-components
src/routes/Hub.tsx:31:7              error  @typescript-eslint/no-unused-expressions
```

The *substance* of the claim holds — I verified the identical 6-error/1-warning set on `origin/main`, and this PR touches none of those files. But the stated evidence is wrong, and in a repo whose brand is honesty about verification output that matters. **Fix:** correct the sentence to "all 7 in 6 files, none touched by this PR" (see §Verification below for the main-branch run).

**4. The offline proof is circumstantial, not air-gapped.** The server access log + Chrome net-log show zero `googleapis`/`gstatic` hits, and I reproduced both. But absence of a request on a *networked* machine is weaker than the issue's stated test ("load it with the network disabled"). Consequence: low — the grep-level evidence (zero `googleapis` bytes in any shipped file) is actually the stronger proof and it's already in the PR, so a network-disabled run would only confirm what static analysis proves. **Fix (optional, cheap):** `"Google Chrome" --headless --host-resolver-rules="MAP * ~NOTFOUND"` on the same URL, or serve inside `unshare -n`. One command, closes the loop against the issue's literal acceptance criterion.

**5. `--virtual-time-budget` screenshots can't prove "no fallback-font reflow."** The PR reports "no fallback-font reflow" from a single headless screenshot. A screenshot taken after the page settles cannot show a *transient* reflow. Consequence: cosmetic-only, and moot here — with no webfont there is no swap period at all, so there is nothing to reflow. The claim is true for the wrong reason. **Fix:** drop the clause; the real argument is "no webfont ⇒ no FOUT/FOIT by construction."

---

### Nits

**6. `dashboard/frontend/src/index.css:39,47` vs `tailwind.config.js:23-31` duplicate the two font stacks verbatim.** If someone edits one, the other silently drifts (the `body`/`.tnum` rules and the `font-sans`/`font-mono` utilities would disagree). **Fix:** define `--font-sans` / `--font-mono` CSS vars in `:root` and reference them from both — which is exactly what sibling PR #123 does in `site/style.css`. Worth aligning, but not worth blocking on.

**7. Monospace fallback is adequate for tabular alignment, confirmed.** `.tnum` at `src/index.css:46-48` keeps `font-variant-numeric: tabular-nums`, which is a font *feature* independent of `Fira Code`, and every face in the new stack (`SF Mono`, `Menlo`, `Consolas`, `Liberation Mono`, generic `monospace`) is fixed-pitch by definition. The screenshot confirms column alignment across all 12 KPI tiles. No finding — recording it because the task asked.

---

### Hand-patched bundles

**Verdict on the regex approach: sound, and better-evidenced than most hand-edits I see.** I diffed both bundles character-by-character against `origin/main` and every single edit is a substring replacement *inside* a `font-family:` declaration value — no rule boundary, no selector, no `{`/`}`/`;` was crossed:

- `examples/.../index-vwBf6wRK.css`: 1 leading `@import` statement deleted (bytes 0-137, cleanly terminated by `;` before `*,:before,...`) + 5 `Fira Sans,`/`Fira Code,` value substitutions.
- `evograph/.../index-juGJTXVt.css`: 8 value substitutions, no `@import` present (it loaded fonts via `<link>` in `index.html` instead).

Both parse, with **zero** surviving `@import` at-rules:

```
../../examples/tau2_airline/run_full/ui/assets/index-vwBf6wRK.css -> OK nodes: 260 @import: 0 bytes: 16457
../../skills/algorithms/evograph/dashboard/frontend/dist/assets/index-juGJTXVt.css -> OK nodes: 419 @import: 0 bytes: 39091
dist/assets/index-DJzo6AHL.css -> OK nodes: 261 @import: 0 bytes: 16457
```

No dangling fragments, no `.map` disagreement — **there are no source maps in any of the three `dist/assets/` directories** (`grep -rl sourceMappingURL` over all three: no hits), so the source-map-drift concern doesn't apply. And the strongest confirmation: the hand-patched tau2 CSS is a **near-exact match to the freshly-*built* dashboard CSS** — same 16457 bytes, differing only in 24 quote characters (`'Segoe UI'` vs `"Segoe UI"`, minifier quote style) plus one pre-existing `.h-[80vh]` utility the tau2 snapshot's build didn't emit. The regex landed the same result a rebuild would have.

**Verdict on mutating `examples/tau2_airline/run_full/` — I disagree with the author, mildly. That file should have been left alone.**

`run_full/` is an evidence directory: `README.md:173-174` and `docs/TROUBLESHOOTING.md:73` both point readers at it as *the* committed proof of the 53.6% → 71.2% result, and `TAU2_COMMIT.txt` pins the benchmark commit for reproducibility. The honest framing of a run artifact is "this is what the run produced," and once a maintainer hand-edits bytes inside it — even cosmetically, even provably safely — that sentence stops being literally true, and the next person auditing the directory has to take a reviewer's word for "it was only fonts."

Two things keep this at non-blocking rather than blocking:
1. **The mutation is provably confined to presentation.** No result data moved — `git diff origin/main...HEAD -- examples/` returns exactly one path, the CSS file. All 96 JSON files under `ui/data/` (rollouts, splits, diffs, memory, git log), `final.json`, and `demo.cast` are byte-identical.
2. **The alternative is defensible but not obviously better.** Leaving it means a served, user-facing snapshot keeps making a CDN request, which is precisely the bug — and the repo's own `README` tells users to `python3 -m http.server` that directory, so it's live UI, not cold storage.

So: the fix is correct in outcome and I'm not asking for a revert. But **the honest move is to record the mutation, not just to make it.** The claim "this directory is untouched run output" is now false and nothing in the repo says so. **Recommended (non-blocking):** add a line to `examples/tau2_airline/run_full/` — a `PATCHES.md` or an appended note in `TAU2_COMMIT.txt` — stating: *"`ui/assets/index-vwBf6wRK.css` was edited post-run (#120) to remove a CDN font `@import`; presentation only, no result data altered."* Three lines, and the artifact's integrity story stays airtight. If instead you'd rather keep the artifact bit-frozen, revert that one file and accept the historical CDN reference — but then the guard test must exclude the path, which weakens it. I'd take the patch-plus-provenance-note over the revert.

---

### Guard robustness

`core/tests/test_dashboard.py:250-276`. **Non-vacuous — I reproduced the failure — but hard-coded and narrow.**

Proof it fails on reintroduction (re-added the `@import` to `src/index.css:1`, ran, restored):

```
== A: @import reintroduced in src/index.css
core/tests/test_dashboard.py:275: AssertionError
FAILED core/tests/test_dashboard.py::test_shipped_spa_bundles_have_no_external_cdn_reference
1 failed, 10 deselected in 0.02s
== restored:
1 passed, 10 deselected in 0.02s
```

**What it catches.** Any occurrence of the 5 banned host strings (`fonts.googleapis.com`, `fonts.gstatic.com`, `cdn.jsdelivr.net`, `unpkg.com`, `cdnjs.cloudflare.com`) in any `.css/.js/.jsx/.ts/.tsx/.html` file under 5 enumerated paths. Because it's a plain substring scan, **syntax is irrelevant** — it catches `@import`, `<link>`, `url()`, `preconnect`, `fetch()`, a string inside minified JS, equally. That's the right design choice. It correctly covers all three shipped bundles plus `src/` plus the dev-root `index.html`. It has an anti-vacuity assert (`checked > 0`).

**What it misses — four gaps, in severity order:**

1. **Any CDN host not in the 5-item list passes.** Both probes green:
```
B  typekit.net @import (unlisted host):      1 passed, 10 deselected
B2 fonts.bunny.net url() (unlisted host):    1 passed, 10 deselected
```
`fonts.bunny.net` is the single most likely thing a well-meaning contributor reaches for as a "privacy-friendly Google Fonts replacement" — and this guard waves it through. **Fix:** ban the *shape*, not the vendor. Regex for `https?://` inside `@import`/`url(`/`href=`/`src=` with a small allowlist (`w3.org` for the SVG namespace, `localhost`, `127.0.0.1`) — that inverts the logic from blacklist to whitelist and closes the whole class.

2. **A fourth bundle added later is not guarded.** `targets` is a literal 5-element list; nothing enumerates `**/dist/` or `**/ui/`. Probe:
```
D new 4th bundle at skills/algorithms/newthing/dashboard/frontend/dist/assets/a.css
  containing @import "https://fonts.googleapis.com/...":   1 passed, 10 deselected
```
**Fix:** derive targets by glob — `REPO.glob("**/dist")` + `REPO.glob("**/ui")`, minus `node_modules` and `site/` — so new bundles are covered by construction.

3. **The anti-vacuity assert is too weak to survive a rename.** `checked > 0` passes as long as *any one* target resolves. I moved the entire evograph `dist/` aside; the test stayed green:
```
F evograph dist path MOVED (simulating a rename):   1 passed, 10 deselected
```
So a future directory reorg silently drops that bundle from coverage while the suite stays green — the exact failure mode the docstring's *"did the paths move?"* was meant to prevent. **Fix:** assert each target exists (`assert t.exists(), t`), or assert a per-target floor on `checked`.

4. **`dashboard.py`'s inline template is not covered by the new guard** — but it is already covered by the pre-existing `test_render_html_self_contained_and_parseable` (`test_dashboard.py:171-173`), which is stricter (`'href="http'`, `"<link"`, `"cdn."`). I injected a CDN `<link>` into `_HTML_TEMPLATE` to confirm the division of labour:
```
C dashboard.py inline template, new guard only:       1 passed, 10 deselected
C dashboard.py inline template, ALL dashboard tests:  1 failed, 10 passed
```
Correctly caught, by the older test. No gap — noting it so nobody "fixes" this by duplicating coverage.

Net: ship it as-is (it stops the exact regression in #120), and open a follow-up for gaps 1-3. Gap 1 is the one I'd actually prioritise.

---

### External-resource audit

Independent grep over shipped/served artifacts (not just source), all `href`/`src`/`@import`/`url()`/bare-URL forms, `.css/.js/.html/.svg/.json/.py`:

| Artifact | External hosts found |
|---|---|
| `dashboard/frontend/src/` | none |
| `dashboard/frontend/dist/` (all 5 files) | none |
| `dashboard/frontend/index.html` (dev root) | none |
| `core/cap_evolve/dashboard.py` `_HTML_TEMPLATE` | none |
| `examples/tau2_airline/run_full/ui/` (incl. 96 data JSON, 2 SVG) | none |
| `skills/algorithms/evograph/dashboard/frontend/dist/index.html` | none |
| `skills/algorithms/evograph/.../index-juGJTXVt.css` | none |
| `skills/algorithms/evograph/.../index-BGSNdw8n.js` | none (but 4 stale `Fira Code` font *names* — finding 1) |
| `presentation/` | `skillberry-ai.github.io` (self-hosted docs link in an `<a href>`, not a subresource — fine) |
| `templates/` | none |
| `site/*.html`, `site/style.css` | `fonts.googleapis.com`, `fonts.gstatic.com` — **deliberately deferred to #123** |

**No external CDN reference survives outside `site/`.** Confirmed the excluded set is exactly what the author claims:

```
$ grep -rniE "googleapis|gstatic|unpkg|jsdelivr|cdnjs|cloudflare|fontawesome|bootstrapcdn|cdn\." \
    --include='*.css' --include='*.js' --include='*.ts' --include='*.tsx' --include='*.html' --include='*.py' --include='*.md' . \
  | grep -v node_modules | grep -v package-lock | grep -v '^\./site/'
core/tests/test_dashboard.py:174:  for marker in ('src="http', 'href="http', "<link", "cdn.", "fetch("):
core/tests/test_dashboard.py:264:  banned = ("fonts.googleapis.com", "fonts.gstatic.com", "cdn.jsdelivr.net",
core/tests/test_dashboard.py:265:            "unpkg.com", "cdnjs.cloudflare.com")
skills/phases/report/scripts/check.py:117: for marker in ('src="http', 'href="http', "<link", "cdn.", "fetch("):
```

All 4 remaining hits are guard code listing the hosts it bans — not references. Only `site/` is excluded, and #123 (already on `origin/feat/issue-123-site-chrome`) removes those 27 lines and adopts the *identical* stacks introduced here (`--font-sans: ui-sans-serif, system-ui, -apple-system, "Segoe UI", Roboto, ...`). Coordination confirmed real, not claimed.

---

### Merge-order note

| Branch | Overlap with #192 | Conflict |
|---|---|---|
| `fix/issue-112-passk-na` (#187) | `dashboard/frontend/dist/index.html` | **YES — real, `changed in both`** |
| `perf/issue-119-reduce-cache` | none | clean |
| `feat/issue-123-site-chrome` | none (`site/` only) | clean |
| `feat/issue-117-event-ticker` | branch does not exist on origin | n/a |

`git merge-tree` against #112:

```
changed in both
  base   100644 f8e01c04... dashboard/frontend/dist/index.html
  our    100644 21d1fcff... dashboard/frontend/dist/index.html
  their  100644 af0ce0f4... dashboard/frontend/dist/index.html
```

Exactly the `dist/`-by-construction collision epic #127 predicts. #112 also rebuilt `dist/` (`index-5zepM1pz.js` + the **old, CDN-carrying** `index-b_9cC4ZS.css`) and touches `KpiStrip.tsx` / `types.ts`, so a naive resolution can resurrect the `@import` — I verified #112's committed CSS still starts with `@import "https://fonts.googleapis.com/...`.

**Recommended order: merge #192 first, then rebase #187/#112 on top and rebuild `dist/`.** Rationale: #192's `dist/` is provably reproducible from source (verified below), so whoever rebases just reruns `npm ci && npm run build` and commits — no hand-resolution of a hashed-filename conflict, and the CSS `@import` cannot come back because the source no longer has it. Merging #112 first inverts the burden onto this PR and risks a resolution that keeps `index-b_9cC4ZS.css`. Bonus: after #192 lands, the new guard *mechanically blocks* a bad #112 resolution — the suite goes red if `index-b_9cC4ZS.css` survives. That's a good reason to land the guard early. #119 and #123 can go in any order.

---

### Verification I re-ran

Worktree `/tmp/rv-192` @ `46e332e`, base `origin/main` @ `6fca097`.

**Full suite — 180 tests, 1 environment-caused failure (not this PR):**
```
$ cd /tmp/rv-192 && PYTHONPATH=/tmp/rv-192/core /tmp/ce-venv/bin/python -m pytest core/tests -q
1 failed, 179 passed in 68.24s (0:01:08)
```
Total is 180 as claimed (179 baseline + 1 new guard). The failure is **not** attributable to this PR:
```
$ PYTHONPATH=... pytest core/tests/test_dashboard_launch.py -q -k spawns
E  AssertionError: assert 'http://127.0.0.1:7882' == 'http://127.0.0.1:7878'

$ lsof -nP -iTCP:7878 -sTCP:LISTEN
Python  20104 osherelhadad  6u  IPv4 ...  TCP 127.0.0.1:7878 (LISTEN)

$ cd /tmp/rv-192-main && pytest core/tests/test_dashboard_launch.py -q -k spawns   # origin/main
1 failed, 6 deselected in 0.05s
```
`maybe_launch` calls `_free_port(port)` (`dashboard_launch.py:78`) and skips the squatted 7878 → 7882. Fails identically on `origin/main`; the PR touches neither file (`git diff --stat origin/main...HEAD -- core/cap_evolve/dashboard_launch.py core/tests/test_dashboard_launch.py` → empty). Pre-existing test/environment coupling in my sandbox (another agent holding 7878) — worth its own issue that the test should monkeypatch `_free_port`, but out of scope here.

**Guard test in isolation:**
```
$ pytest core/tests/test_dashboard.py -q
...........  [100%]   11 passed in 0.07s
```

**`dist/` reproducibility — this is the claim I most wanted to check, and it holds exactly:**
```
$ cd /tmp/rv-192/dashboard/frontend && npm ci && npm run build
dist/assets/index-DJzo6AHL.css   16.45 kB │ gzip:   4.51 kB
dist/assets/index-zqmQoE4e.js   830.26 kB │ gzip: 249.87 kB
✓ built in 1.03s

(empty)
```
Fresh build is **byte-identical** to the committed `dist/` — same hashes, no drift. `dist/index.html` references `/assets/index-zqmQoE4e.js` + `/assets/index-DJzo6AHL.css`; the old `index-g1UdrsC9.js`/`index-b_9cC4ZS.css` are deleted and I found no stale reference to either anywhere in the tree.

**Static checks:**
```
$ npx tsc -b --noEmit      -> tsc-exit=0
$ python -m compileall -q core skills   -> compileall-exit=0
```

**eslint, branch vs `origin/main` — identical 7 problems across 6 files (see finding 3):**
```
branch (/tmp/rv-192):        ✖ 7 problems (6 errors, 1 warning)
origin/main (/tmp/rv-192-main): ✖ 7 problems (6 errors, 1 warning)
   CountUp.tsx:21:7, DiffRows.tsx:6:14, FileTree.tsx:24:9, GitDiff.tsx:24:44,
   useRunStream.ts:63:3, main.tsx:18:7, Hub.tsx:31:7   — same in both
```

**CSS parse + `@import` count on all three bundles** — output in §Hand-patched bundles above; 3/3 parse, 0 `@import`.

**Rendered UI, independently reproduced:**
```
$ cd examples/tau2_airline/run_full/ui && python -m http.server 8231 &
$ curl -s http://localhost:8231/assets/index-vwBf6wRK.css | grep -c googleapis
0
$ "Google Chrome" --headless --window-size=1440,1000 --virtual-time-budget=9000 \
    --screenshot=/tmp/shot192/dash.png "http://localhost:8231/#/runs/run_full"
67574 bytes written to file /tmp/shot192/dash.png

server access log:
  GET /  |  GET /assets/index-vwBf6wRK.css  |  GET /assets/index-ez7sAMzg.js
  GET /cap-evolve-logo.png  |  GET /data/runs_run_full.json  |  GET /favicon.png
```
Six requests, all local, zero font requests — matches the PR exactly. I inspected the screenshot: renders correctly in system fonts. All 12 KPI tiles legible and aligned (BEST 71.2%, BASELINE 53.6%, Δ +32.8%, SEALED TEST 69.4% w/ `pass^k NaN%` sub-label, COST $148.43, TOKENS 288.6K, ACCEPTED 5, REJECTED 5, FAILED 0, FRONTIER 1, WALL CLOCK 10h 23m, $/+1% $4.53). Full tab bar present, "Fitness over iterations" plots all 11 points across iterations 0-10 with the amber ★ champion at 7 and correct 0/25/50/75/100 axis. Numerals column-aligned. No clipping, no layout breakage. The rendering claim reproduces.

**Guard blind-spot probes** — B/B2/C/D/F outputs in §Guard robustness above. All restored; `git status --porcelain` clean afterward.

OsherElhadad:
## 🔧 Review fixes

Pushed as `8d1805f`. Every finding answered below; 5 fixed in code, 2 acknowledged as no-change with the reasoning stated. Plus a hardening pass on the guard that goes beyond the review.

---

### 1. ✅ FIXED — 4 surviving `Fira Code` inline styles in the served evograph JS bundle

Verified first, then substituted:

```
$ grep -o 'fontFamily:"[^"]*"' skills/algorithms/evograph/dashboard/frontend/dist/assets/index-BGSNdw8n.js | sort | uniq -c
   4 fontFamily:"Fira Code, monospace"
   1 fontFamily:"font-family"
```

Byte-level substitution (`fontFamily:"Fira Code, monospace"` → the project's mono stack), same approach as the CSS pass:

```
replaced: 4 | residual Fira: 0 | bytes 1216985 -> 1217249
```

```
$ node --check skills/algorithms/evograph/dashboard/frontend/dist/assets/index-BGSNdw8n.js
$ echo $?
0
$ grep -c Fira .../index-BGSNdw8n.js
0
$ grep -o 'fontFamily:"ui-monospace[^"]*"' .../index-BGSNdw8n.js | sort | uniq -c
   4 fontFamily:"ui-monospace, SFMono-Regular, 'SF Mono', Menlo, Consolas, 'Liberation Mono', monospace"
```

**The bundle still parses** (`node --check` exit 0) and the JS now resolves through the *same* fallback chain the sibling CSS specifies — the Linux-without-Menlo divergence you described is gone.

### 2. ✅ NOTED — evograph `frontend/` has only `dist/`, no source

Agreed, and out of scope here. Recorded for a follow-up issue ("vendor or restore the evograph frontend source, or document it as a vendored third-party bundle"). Not fixed in this PR.

### 3. ✅ FIXED — eslint attribution corrected in the PR body

You were right; the body said all 7 were in `Hub.tsx`. Machine-readable run on this branch:

```
$ npx eslint src -f json | <extract>
src/components/CountUp.tsx:21:7    error    react-hooks/set-state-in-effect
src/components/DiffRows.tsx:6:14   error    react-refresh/only-export-components
src/components/FileTree.tsx:24:9   warning  react-hooks/exhaustive-deps
src/components/GitDiff.tsx:24:44   error    react-hooks/set-state-in-effect
src/lib/useRunStream.ts:63:3       error    react-hooks/refs
src/main.tsx:18:7                  error    react-refresh/only-export-components
src/routes/Hub.tsx:31:7            error    @typescript-eslint/no-unused-expressions

✖ 7 problems (6 errors, 1 warning)
```

Exactly your list. PR body §Verification now reads "7 problems across **6 files**, none touched by this PR", enumerates all six paths, and notes the earlier misattribution.

### 4. ✅ FIXED — and I did the stronger test rather than restating

Genuine DNS denial, not absence-of-request on a networked box:

```
$ "Google Chrome" --headless=new --host-resolver-rules="MAP * ~NOTFOUND, EXCLUDE localhost" \
    --virtual-time-budget=9000 --window-size=1440,1000 \
    --screenshot=/tmp/shot192fx/dash.png "http://localhost:8341/#/runs/run_full"
-rw-r--r--  67640  dash.png
```

Server access log for that DNS-denied load — every request the page made:

```
::1 - - [30/Jul/2026 01:11:49] "GET / HTTP/1.1" 200 -
::1 - - [30/Jul/2026 01:11:49] "GET /assets/index-ez7sAMzg.js HTTP/1.1" 200 -
::1 - - [30/Jul/2026 01:11:49] "GET /assets/index-vwBf6wRK.css HTTP/1.1" 200 -
::1 - - [30/Jul/2026 01:11:49] "GET /cap-evolve-logo.png HTTP/1.1" 200 -
::1 - - [30/Jul/2026 01:11:49] "GET /data/runs_run_full.json HTTP/1.1" 200 -
::1 - - [30/Jul/2026 01:11:49] "GET /favicon.png HTTP/1.1" 200 -
```

The screenshot renders fully: all 12 KPI tiles (BEST 71.2%, BASELINE 53.6%, Δ +32.8%, SEALED TEST 69.4% w/ `pass^k NaN%`, COST $148.43, TOKENS 288.6K, ACCEPTED 5, REJECTED 5, FAILED 0, FRONTIER 1, WALL CLOCK 10h 23m, $/+1% $4.53), full tab bar, and all 11 fitness points with the amber ★ at iteration 7.

**Control — proof the DNS rule was actually in effect** (otherwise this is the same circumstantial evidence again). Same flag, external URL:

```
$ "Google Chrome" --headless=new --host-resolver-rules="MAP * ~NOTFOUND, EXCLUDE localhost" \
    --dump-dom "https://fonts.googleapis.com/css2?family=Fira+Code"
$ grep -oE 'DNS_PROBE[A-Z_]*' out
DNS_PROBE_POSSIBLE
DNS_PROBE_STARTED
$ grep -oE '<title>[^<]*' out
<title>fonts.googleapis.com
```

Chrome's DNS-failure interstitial instead of CSS. Resolution genuinely denied, and the dashboard rendered anyway. The issue's literal acceptance criterion is now met.

### 5. ✅ FIXED — dropped the unprovable reflow claim

You're right that a settled `--virtual-time-budget` screenshot cannot show a transient reflow. The clause is gone from the PR body, replaced with the structural argument: with no webfont there is no swap period to reflow through. The claim is no longer sourced from the screenshot.

### 6. ✅ FIXED (nit) — the two font stacks are now defined once

`src/index.css` `:root` gains `--font-sans` / `--font-mono`; `body` and `.tnum` use `var(--font-sans)` / `var(--font-mono)`; `tailwind.config.js` `fontFamily` is now `sans: 'var(--font-sans)'`, `mono: 'var(--font-mono)'`. The duplication that could drift is gone. Confirmed in the rebuilt bundle:

```
$ grep -o 'var(--font-[a-z]*)' dashboard/frontend/dist/assets/index-cVkeQUOc.css | sort | uniq -c
   3 var(--font-mono)
   2 var(--font-sans)
$ grep -c Fira dashboard/frontend/dist/assets/*.css dashboard/frontend/dist/assets/*.js
index-cVkeQUOc.css:0
index-DXi969oT.js:0
```

### 7. ✅ NO CHANGE (nit) — tabular alignment

Agreed, no action needed. `font-variant-numeric: tabular-nums` is a font feature independent of `Fira Code`, and every face in the new stack is fixed-pitch. The DNS-denied screenshot above re-confirms column alignment across all 12 tiles.

---

### `run_full/` provenance note — done, no revert

Took your recommendation exactly: keep the patch (the directory is live UI per `README.md:173`), record the mutation. New `examples/tau2_airline/run_full/PATCHES.md`:

> This directory is the committed output of the tau2-airline run (53.6% -> 71.2%). It is served as live UI (`README.md:173` tells users to `python3 -m http.server` here), so presentation-only fixes are applied in place rather than left broken. Every such edit is recorded below. **No result data has ever been modified** — all JSON under `ui/data/`, `final.json`, and `demo.cast` are byte-identical to what the run produced.
>
> | Date | File | Change | Issue |
> |---|---|---|---|
> | 2026-07-30 | `ui/assets/index-vwBf6wRK.css` | Removed the Google Fonts CDN `@import` and replaced the `Fira Sans`/`Fira Code` font-family values with system stacks, so the snapshot renders offline/air-gapped. Presentation only. | #120 (PR #192) |

Re-confirmed the scope claim it makes:

```
$ git diff --stat origin/main...HEAD -- examples/tau2_airline/run_full/ui/data \
    examples/tau2_airline/run_full/final.json examples/tau2_airline/run_full/demo.cast
(empty)
$ ls examples/tau2_airline/run_full/ui/data | wc -l
96
```

All 96 data JSONs, `final.json`, `demo.cast`: untouched.

---

### Strengthened guard — all three gaps closed, with before/after

The guard was a 5-host **denylist** over 5 **hard-coded** paths with a `checked > 0` floor. That is the same shape as the count-drift bug this epic keeps finding: the next CDN nobody thought of passes. Rewritten to ban the *shape* over *discovered* targets with real floors.

**Gap 1 — unlisted host. BEFORE (your probes): both green. AFTER: both fail.**

```
===== GAP 1: unlisted host (fonts.bunny.net) =====
E   AssertionError: dashboard/frontend/src/index.css loads an external subresource: https://fonts.bunny.net/css?family=Inter
1 failed, 10 deselected in 0.06s
----- restored: 1 passed, 10 deselected in 0.05s

===== GAP 1b: unlisted host (typekit.net) =====
E   AssertionError: dashboard/frontend/src/index.css loads an external subresource: https://use.typekit.net/abc.css
1 failed, 10 deselected in 0.06s
----- restored: 1 passed, 10 deselected in 0.04s
```

Mechanism: a regex for `https?://` in a subresource position (`@import`, `url(`, `href=`, `src=`, `import from`) with an allowlist of `{localhost, 127.0.0.1, www.w3.org, w3.org}` — inverted from denylist to allowlist, so the whole class is closed, not five vendors.

**Gap 2 — a bundle added later. BEFORE: green. AFTER: fails.**

```
===== GAP 2: brand-new 4th bundle with a CDN import (in no hard-coded list) =====
E   AssertionError: skills/algorithms/newthing/dashboard/frontend/dist/assets/a.css loads an external subresource: https://fonts.googleapis.com/css2?family=Fira+Code
1 failed, 10 deselected in 0.06s
----- restored: 1 passed, 10 deselected in 0.05s
```

Mechanism: targets are `REPO.glob("**/dist") + REPO.glob("**/ui")` minus `node_modules` and `site/`. New bundles are covered by construction. Currently discovers 5 dirs / 66 files (was 5 fixed paths).

**Gap 3 — a moved bundle. BEFORE: green (you proved it). AFTER: fails.**

Worth noting: glob alone does **not** fix this — it just stops finding the moved dir. My first attempt still went green:

```
===== GAP 3 (glob only): evograph dist/ MOVED aside =====
1 passed, 10 deselected in 0.04s     <-- still wrong
```

So the known served bundles are *also* named and asserted to exist. Glob adds coverage; the explicit list prevents silent loss of it:

```
===== GAP 3 retest: evograph dist/ MOVED aside =====
E   AssertionError: guard target vanished (renamed/moved?): .../skills/algorithms/evograph/dashboard/frontend/dist
1 failed, 10 deselected in 0.06s
----- restored: 1 passed, 10 deselected in 0.05s

===== GAP 3b: tau2 run_full/ui MOVED aside =====
E   AssertionError: guard target vanished (renamed/moved?): .../examples/tau2_airline/run_full/ui
1 failed, 10 deselected in 0.05s
----- restored: 1 passed, 10 deselected in 0.05s
```

Plus two count floors so an emptied bundle also fails loudly: `len(bundles) >= 4` and `checked >= 60` (replacing `checked > 0`).

Gap 2 re-verified after the Gap 3 fix — the new bundle is still auto-covered, so the explicit list did not turn the glob into dead code. `git status --porcelain` clean after every probe.

---

### Verification (literal output)

**Full suite — 180/180 passed. #200 did not reproduce this run** (port 7878 was free; no stash needed):

```
$ cd /tmp/fx-192 && PYTHONPATH=/tmp/fx-192/core /tmp/ce-venv/bin/python -m pytest core/tests -q
....................................                                     [100%]
180 passed in 62.13s (0:01:02)
```

Guard in isolation:

```
$ pytest core/tests/test_dashboard.py -q
...........  [100%]   11 passed in 0.14s
```

**Static checks:**

```
$ /tmp/ce-venv/bin/python -m compileall -q core skills   ->  compileall-exit=0
$ npx tsc -b --noEmit                                    ->  tsc-exit=0
$ node --check .../index-BGSNdw8n.js                     ->  node-check-exit=0
```

**Build + `dist/` reproducibility.** The CSS-var change to `index.css`/`tailwind.config.js` legitimately changes the CSS hash (`index-DJzo6AHL.css` → `index-cVkeQUOc.css`; JS re-hashed to `index-DXi969oT.js` from the index.html reference). The committed `dist/` is the fresh build, and rebuilding is byte-stable:

```
$ npm ci && npm run build
dist/assets/index-cVkeQUOc.css   16.34 kB │ gzip:   4.54 kB
dist/assets/index-DXi969oT.js   830.26 kB │ gzip: 249.87 kB
✓ built in 803ms

$ shasum -a256 dist/assets/* dist/index.html > /tmp/h1 ; npm run build ; shasum -a256 ... > /tmp/h2 ; diff /tmp/h1 /tmp/h2
IDENTICAL across rebuilds
4422137ca82a7345eb85a4fe1d1aa002f356414d6c568d4236914b645c8c5e1c  dist/assets/index-DXi969oT.js
2d2cbf3d396c7d1243da042ec08b139eea047464fe67045ee12143df5eaa8b93  dist/assets/index-cVkeQUOc.css
1ddc008f947fd87bfa4de5d43b4680c841278b21fb90b5c611b2bfa2d4ccc7bf  dist/index.html

$ git diff --name-only dashboard/frontend/dist     # after rebuild, post-commit
(empty)
```

**Repo-wide external-resource scan — `site/` remains the sole exclusion:**

```
$ grep -rniE "googleapis|gstatic|unpkg|jsdelivr|cdnjs|cloudflare|fontawesome|bootstrapcdn|fonts\.bunny|typekit" \
    --include='*.css' --include='*.js' --include='*.ts' --include='*.tsx' --include='*.html' --include='*.py' --include='*.md' . \
  | grep -v node_modules | grep -v package-lock | grep -v '^\./site/'
core/tests/test_dashboard.py:258:    CDNs someone already thought of (fonts.bunny.net, typekit.net, the next one).
```

One hit, and it is a docstring in the guard naming the hosts it is *not* a denylist for — not a reference. Note the previous two guard-code hits (the `banned` tuple) are gone, since there is no host list any more. `site/` still has its 9 × 3 lines, deferred to #123 / PR #196.

Independent subresource scan across all globbed bundle dirs — zero external subresources, only `w3.org`/`localhost`:

```
$ <regex scan for @import|url(|href=|src=|from + https?:// over 66 files>
(no output)
```

**Commit authorship:**

```
$ git log -1 --format='%an <%ae> / %cn <%ce>'
Osher Elhadad <Osher.Elhadad@ibm.com> / Osher Elhadad <Osher.Elhadad@ibm.com>
$ git log -1 --format='%B' | grep -icE "co-authored|generated with"
0
```

---

**Files touched by this fix commit:** `core/tests/test_dashboard.py`, `dashboard/frontend/src/index.css`, `dashboard/frontend/tailwind.config.js`, `dashboard/frontend/dist/{index.html,assets/index-cVkeQUOc.css,assets/index-DXi969oT.js}` (rebuild), `skills/algorithms/evograph/dashboard/frontend/dist/assets/index-BGSNdw8n.js`, `examples/tau2_airline/run_full/PATCHES.md` (new).

Issue #117, opened 2026-07-25
Render the collected-but-discarded live event ticker in the SPA; fix blank algorithm badge


OsherElhadad:
## Problem

The dashboard SPA **collects the live event stream and then throws it away**, and the `algorithm` badge is blank everywhere because the reducer never sets it. Two small fixes turn "42 live events" from a bare counter into an actual play-by-play.

1. **Discarded event ticker.** `useRunStream` maintains a capped 200-entry `log` of streamed events (`dashboard/frontend/src/lib/useRunStream.ts:32,40-48`), but the only consumers of `stream` in the whole app are `stream.status` and `stream.count` (`RunDeepDive.tsx:60,63,75-76`). The UI shows "*N* live events" as a number — you cannot see *what* happened (no "cand_7 accepted +2.1%", "gate rejected", "optimizer_error"). The richest live signal is captured and unused.
2. **Blank `algorithm` badge.** `final.json` carries `algorithm` (`harness.py:1877`, `gepa.py:664`, `skillopt.py:371`), but `reduce_run`'s summary dict (`core/cap_evolve/dashboard.py:529-563`) never reads `final.get("algorithm")`. So the algorithm name is blank on every Hub row (`Hub.tsx:137-141`), the DeepDive chip (`RunDeepDive.tsx:72-74`), Compare, and the self-contained HTML "Optimize · <algo>" label.

## Why it matters

The plumbing for a live play-by-play already exists — rendering it is nearly free and is exactly what makes a live run feel alive. And a dashboard that can't tell you which algorithm produced a run is missing a basic label on its primary surface.

## What to change

1. Add a small scrolling **live event panel** in `RunDeepDive` that renders `stream.log` (event kind + candidate id + val + accept/reject reason), newest-first, capped. Reuse the existing 200-entry buffer.
2. Fix the reducer: add `"algorithm": final.get("algorithm")` to the summary dict at `core/cap_evolve/dashboard.py:529-563`. This fixes the badge in the SPA **and** the self-contained HTML dashboard at once.

## How to test

- Backend test: `reduce_run` on a fixture with `final.json` containing `algorithm: "gepa"` → `summary["algorithm"] == "gepa"`.
- Frontend test (Vitest): given a mocked `useRunStream` with a few events, the live panel renders their kinds/candidate ids.
- Manually: start a run, watch the ticker populate; Hub shows the algorithm name per run.

## Dependencies / parallelization

[...]

PR #194, opened 2026-07-29
perf(dashboard): memoize reduce_run on events mtime+size, drop dead SSE snapshot, paginate runs/rollouts

### LINKED ISSUE #119: Dashboard re-reduces the full event log on every request/SSE tick — cache by mtime; drop dead SSE snapshot

OsherElhadad:
Closes #119

The dashboard re-folded the **entire** event log (plus every rollout file, every candidate dir stat) on every HTTP request **and** every SSE tick (~2/s, 400 ms-debounced `invalidateQueries`), plus once more per candidate in `diff_candidate`. On a 200-candidate run that is **~275 ms of CPU per tick** spent re-deriving an answer that had not changed — the reason the "real-time" dashboard is laggiest exactly when the run is big.

## 1. `reduce_run` memoized on `events.jsonl` (mtime + size)

`core/cap_evolve/dashboard.py`: the fold moved to `_reduce_run_uncached`; `reduce_run` is now a thin memo over it. One `dict` entry per run root holding only its newest reduction — no cache class, no new dependency, stdlib `os.stat` only.

**Why the key is mtime *and* size, not mtime alone.** `st_mtime` granularity is 1 second on some filesystems (and the float `st_mtime` rounds), so an append that lands in the same second as the cached read can leave mtime looking unchanged — a cache keyed on mtime alone would then serve a *stale reduction during a live run*, which is strictly worse than no cache. `events.jsonl` is append-only, so **size always grows on a new event**: size is the strong signal, and mtime catches the (not-expected-here) same-length rewrite. There is a test that forces the mtime back to its cached value with `os.utime` and asserts the size change still invalidates.

**Why one key covers the reducer's other inputs.** The fold also reads `baseline.json`, `final.json`, `rollouts/`, `state.json` and the git store. Each of those is written *before* the event that announces it is appended — `harness` writes `baseline.json` then logs `baseline` (harness.py:376-378), persists rollout files then logs `evaluate` (harness.py:252 → 311), the iteration commit lands then `step` is logged, finalize writes `final.json` then logs `finalize`. So the stamp advances *after* those writes, never before.

**Race guard.** The stamp is re-read after the fold; if the log moved while we were folding, the result may already be stale, so it is not cached and the next call re-folds.

**Cache-hit results are shared objects** — documented on `reduce_run` as read-only. Every caller in the repo (`runs._reduce`, `trajectories.diff_candidate`, `export_static`, `render_html`, `report/scripts/{run,check}.py`) only reads.

## 2. Dead SSE snapshot payload deleted

`app.py` yielded `sse_format("snapshot", runs.load_run(...))` — a full reduced run serialized on every connect. The client (`useRunStream.ts:75`) **ignores the payload**: it dispatches `{type:'open'}` and then refetches `/api/runs/{id}` through its query cache. The frame now carries only `{"run_id": ...}` (kept as the open marker so no client change is needed). The now-unreachable `except runs.RunNotFound` around it went too — `_resolve_or_404` already 404s before the generator runs. `types.ts` `StreamEvent['snapshot']` narrowed from `RunDetail` to `{run_id}`.

## 3. Pagination contract

`?limit=` / `?offset=` on `GET /api/runs` and `GET /api/runs/{id}/rollouts`.

- **Unpaginated by default** — omit both and the response is the same full array as before (the SPA hub is unchanged).
- `limit` ≥ 1, `offset` ≥ 0, both validated by FastAPI (`422` otherwise). `offset` past the end returns `[]`.
- **Deterministic order** so consecutive pages tile the full list with no gaps or repeats: runs newest-first by dir mtime; rollouts by (split name, file name).
- **The window is applied before the expensive work** — to the discovered run dirs before any reduction (one reduction per row *on the page*, not per run in the dir), and to the matched rollout filenames before any file is opened (one JSON parse per row on the page, not per rollout in the split).
- Rows are best-effort: a half-written run or unparseable rollout is skipped, so a page can come back shorter than `limit`.

## Measured before/after

Real run dir from `bash examples/toy_calc/run.sh` (mock optimizer, zero-API: `baseline_val 0.0 → test_reward 1.0`, 3 iterations), grown to a long run: **215 events, 404 rollout files, 202 candidate dirs**. BEFORE is an actual `origin/main` worktree, not a flag.

| operation (x20) | BEFORE (origin/main) | AFTER (this branch) | speedup |
|---|---|---|---|
| `reduce_run` | 275.557 ms each | **0.009 ms** each | **~30,000x** |
| `runs.load_run` (HTTP + SSE path) | 254.873 ms each | **0.061 ms** each | **~4,200x** |
| `trajectories.diff_candidate` | 301.376 ms each | **53.482 ms** each | **5.6x** |
| `runs.list_runs` | 274.170 ms each | **0.791 ms** each | **347x** |

Cold pagination over 12 such runs: `/api/runs` full **2822.6 ms** → `limit=5` **1166.7 ms**. Rollouts on one run: full (404 rows) **18.42 ms** → `limit=50` **3.87 ms**.

The backend test suite also went **303.65 s → 2.06 s** (its own SSE test no longer sits through the 5-minute idle timeout).

## Overlap with sibling issues

- **#116** is factoring the events-tail helper out of this same SSE path. My only `app.py` change inside the SSE generator is the two lines that produced the snapshot frame; the `read_new_events` polling loop is untouched.
- **#117** is fixing the SPA render + a reducer field. My `dashboard.py` change adds a memo wrapper and renames the fold body to `_reduce_run_uncached` — a reducer *field* change lands inside that body and should merge cleanly; a merge conflict, if any, is the `def` line. My only frontend edit is one type annotation in `types.ts`.
- `dashboard/frontend/dist/` is checked in; I rebuilt it to verify and then **reverted the dist churn** so the hash-renamed bundle does not collide with #116/#117.

## Verification

```
$ PYTHONPATH=/tmp/wt-119/core python -m pytest core/tests -q -k "not test_maybe_launch_spawns_when_available"
185 passed, 1 deselected in 59.43s      # 179 baseline + 7 new = 186 collected

$ cd dashboard/backend && python -m pytest tests -q
49 passed, 1 warning in 2.06s           # 40 baseline + 9 new

compileall CLEAN

✓ built in 774ms
$ npx tsc --noEmit   → exit 0
$ npm test -- --run  → Test Files 13 passed (13) / Tests 45 passed (45)
```

> The one deselected test, `test_dashboard_launch.py::test_maybe_launch_spawns_when_available`, hard-codes port 7878 and fails in this environment because another process holds that port (the launcher falls through to 7882). It **fails identically on unmodified `origin/main`** — pre-existing and unrelated; it touches none of the files in this diff. See §3 of the Evidence comment for the reproduction.

**Fail-before / pass-after** (behavioral, no internals — reverting only `dashboard.py` to `origin/main`):

```
===== BEFORE (origin/main reducer) =====
>           assert dashboard.reduce_run(rd) is dashboard.reduce_run(rd)
E           AssertionError
FAILED core/tests/test_dashboard_cache.py::test_repeat_reduce_reuses_the_same_object_behavioral
1 failed, 1 passed

===== AFTER =====
7 passed in 0.06s
```

**Invalidation proven on the live-appending 202-candidate run dir:**

```
reduce #1 total nodes: 202 best_val: 1.0
reduce #2 is cached object: True
reduce #3 after append: fresh object: True total nodes: 203 (218.8 ms re-fold)
cand_9999 present in graph: True  -> NO stale serve during a live run
reduce #4 re-cached: True  (0.018 ms)
```

**Pagination bounds:** pages tile the full list exactly (`True`, 404 rollout rows / 12 runs); `offset` past the end → `[]`; `limit=0` and `offset=-1` → `422`.

Full commands + output in the 🔬 Evidence comment.

## Tests added

- `core/tests/test_dashboard_cache.py` (7) — hit on unchanged log; **invalidate on append**; **invalidate at identical mtime** (size busts the key); per-run isolation; missing event log never cached; 2 internals-free behavioral tests.
- `dashboard/backend/tests/test_pagination.py` (9) — default-unpaginated; pages tile; bounds `422`; offset past end; a paged rollout read parses only the page's files (5, not 20); SSE snapshot frame carries no reduced run.

Skipped: incremental SPA updates from stream events (#119 stretch item 4) — that is a frontend change and belongs with #117's SPA work; the cache already removes the cost of the per-tick refetch.

### DIFF
--- a/core/cap_evolve/dashboard.py
+++ b/core/cap_evolve/dashboard.py
@@ -229,8 +229,59 @@ def _step_candidate(ev: dict):
     return ev.get("candidate") or ev.get("candidate_id")
 
 
+# --- reduction cache -------------------------------------------------------
+# ``_reduce_run_uncached`` re-reads and re-folds the whole event log plus every
+# rollout file. The dashboard backend calls ``reduce_run`` on every request AND on
+# every SSE tick (~2/s), and once more per candidate in ``diff_candidate``, so a
+# long run spends most of its CPU re-deriving an answer that has not changed.
+#
+# Key: the run's ``events.jsonl`` ``(st_mtime_ns, st_size)``.
+#   * Both, not just mtime: mtime granularity is 1s on some filesystems (and plain
+#     float ``st_mtime`` rounds), so a same-second append can leave mtime looking
+#     unchanged. events.jsonl is append-only, so size always grows on a new event —
+#     size is the strong signal, mtime catches a same-length rewrite.
+#   * One key is enough for the reducer's *other* inputs (baseline.json, final.json,
+#     rollouts/, state.json, the git store) because each is written BEFORE the event
+#     that reports it is appended: harness writes baseline.json then logs "baseline",
+#     persists rollout files then logs "evaluate", finalize writes final.json then
+#     logs "finalize", the iteration commit lands then "step" is logged. So the stamp
+#     advances after those writes, never before.
+#
+# Bounded by run count: one entry per run root, holding only its newest reduction.
+_REDUCE_CACHE: dict[str, tuple[tuple[int, int], dict]] = {}
+
+
+def _events_stamp(root: Path):
+    """``(st_mtime_ns, st_size)`` of the run's event log, or ``None`` if unreadable."""
+    try:
+        st = (root / "events.jsonl").stat()
+    except OSError:
+        return None  # no event log → nothing stable to key on; never cache
+    return (st.st_mtime_ns, st.st_size)
+
+
 def reduce_run(run_dir) -> dict:
-    """Fold the run dir into ``{"graph": ..., "summary": ...}`` (redacted)."""
+    """Fold the run dir into ``{"graph": ..., "summary": ...}`` (redacted).
+
+    Memoized on the event log's mtime+size (see ``_REDUCE_CACHE``). The returned
+    object is SHARED between callers on a cache hit — treat it as read-only.
+    """
+    root = Path(run_dir.root)
+    stamp = _events_stamp(root)
+    if stamp is not None:
+        hit = _REDUCE_CACHE.get(str(root))
+        if hit is not None and hit[0] == stamp:
+            return hit[1]
+    reduced = _reduce_run_uncached(run_dir)
+    # Re-stat after the fold: if the log grew while we were folding, this result may
+    # already be stale, so don't cache it — the next call re-folds.
+    if stamp is not None and _events_stamp(root) == stamp:
+        _REDUCE_CACHE[str(root)] = (stamp, reduced)
+    return reduced
+
+
+def _reduce_run_uncached(run_dir) -> dict:
+    """The actual fold. Call ``reduce_run`` instead; this one always re-reads."""
     root = Path(run_dir.root)
     events = _read_jsonl(root / "events.jsonl")
     baseline = _read_json(root / "baseline.json")
--- a/core/cap_evolve/dashboard.py
+++ b/core/cap_evolve/dashboard.py
@@ -235,48 +235,102 @@ def _step_candidate(ev: dict):
 # every SSE tick (~2/s), and once more per candidate in ``diff_candidate``, so a
 # long run spends most of its CPU re-deriving an answer that has not changed.
 #
-# Key: the run's ``events.jsonl`` ``(st_mtime_ns, st_size)``.
-#   * Both, not just mtime: mtime granularity is 1s on some filesystems (and plain
+# Key: ``events.jsonl``'s ``(st_mtime_ns, st_size, st_ino)`` PLUS the git store's
+# ``.git`` + ``.git/logs/HEAD`` stamps. Keyed on ``root.resolve()`` so an absolute and
+# a relative path to the same run dir share one entry.
+#   * Size, not just mtime: mtime granularity is 1s on some filesystems (and plain
 #     float ``st_mtime`` rounds), so a same-second append can leave mtime looking
 #     unchanged. events.jsonl is append-only, so size always grows on a new event —
-#     size is the strong signal, mtime catches a same-length rewrite.
-#   * One key is enough for the reducer's *other* inputs (baseline.json, final.json,
-#     rollouts/, state.json, the git store) because each is written BEFORE the event
-#     that reports it is appended: harness writes baseline.json then logs "baseline",
-#     persists rollout files then logs "evaluate", finalize writes final.json then
-#     logs "finalize", the iteration commit lands then "step" is logged. So the stamp
-#     advances after those writes, never before.
+#     size is the strong signal, mtime catches a same-length rewrite, and ``st_ino``
+#     catches a same-size replace-via-rename (a new file gets a new inode).
+#   * INVARIANT the events part relies on: ``events.jsonl`` is only ever APPENDED to
+#     (``RunDir.log_event`` opens it "a"). A future writer that rewrites it in place
+#     to the same byte length within one mtime tick would serve a stale reduction —
+#     if you add such a writer, hash the file or bump a counter instead.
+#   * The event stamp alone covers baseline.json, final.json, rollouts/ and state.json,
+#     because each is written BEFORE the event reporting it: baseline.json then
+#     log_event("baseline"), rollout files (harness.py:252) then log_event("evaluate")
+#     (:311), final.json (:1964) then log_event("finalize") (:1966).
+#   * It does NOT cover the git store, whose ordering is the OPPOSITE: the harness logs
+#     "step" (harness.py:1323) and only commits 33 lines LATER (:1356); GEPA likewise
+#     logs (gepa.py:718) then commits (:635). So a commit lands with the event log
+#     untouched and ``git_log`` (:623) would stay stale for the rest of the iteration.
+#     Hence ``_git_stamp``. NOT ``.git/HEAD``: that is a constant symbolic ref
+#     ("ref: refs/heads/<branch>") and never changes on a commit — measured. ``.git``'s
+#     own mtime moves on every commit (index.lock create+rename) and ``.git/logs/HEAD``
+#     is the append-only reflog; both are branch-name agnostic, and either moving is
+#     enough to invalidate.
 #
-# Bounded by run count: one entry per run root, holding only its newest reduction.
-_REDUCE_CACHE: dict[str, tuple[tuple[int, int], dict]] = {}
+# Bounded: FIFO-capped at ``_REDUCE_MAX`` entries (the dashboard is a long-lived
+# process and ``list_runs`` installs one entry per discovered run).
+_REDUCE_CACHE: dict[str, tuple[tuple, dict]] = {}
+_REDUCE_MAX = 32
+
+
+def _stat_stamp(path: Path) -> tuple[int, int]:
+    """``(st_mtime_ns, st_size)``, or ``(0, 0)`` when the path is absent/unreadable."""
+    try:
+        st = path.stat()
+    except OSError:
+        return (0, 0)
+    return (st.st_mtime_ns, st.st_size)
+
+
+def _git_stamp(root: Path) -> tuple[int, int, int, int]:
+    """Stamp for the run's git store — moves on every ``VersionStore.commit``."""
+    return _stat_stamp(root / ".git") + _stat_stamp(root / ".git" / "logs" / "HEAD")
+
+
+def _resolve(root: Path) -> Path:
+    """``root.resolve()``, falling back to the path itself if resolution fails.
+
+    ``RunDir.open`` (rundir.py:201) does not resolve, and ``reduce_run`` is a public
+    API called with relative paths (report/scripts/run.py, check.py, export_static.py),
+    so without this an absolute and a relative path to one run dir get two entries.
+    """
+    try:
+        return root.resolve()
+    except OSError:
+        return root
 
 
 def _events_stamp(root: Path):
-    """``(st_mtime_ns, st_size)`` of the run's event log, or ``None`` if unreadable."""
+    """Full cache stamp for a run dir, or ``None`` when there is no event log.
+
+    ``(events mtime_ns, size, inode) + git stamp``. ~3 extra ``stat`` calls versus one
+    (~2 us each, measured) against a ~230 ms fold — noise.
+    """
     try:
         st = (root / "events.jsonl").stat()
     except OSError:
         return None  # no event log → nothing stable to key on; never cache
-    return (st.st_mtime_ns, st.st_size)
+    return (st.st_mtime_ns, st.st_size, st.st_ino) + _git_stamp(root)
 
 
 def reduce_run(run_dir) -> dict:
     """Fold the run dir into ``{"graph": ..., "summary": ...}`` (redacted).
 
-    Memoized on the event log's mtime+size (see ``_REDUCE_CACHE``). The returned
-    object is SHARED between callers on a cache hit — treat it as read-only.
+    Memoized on the event log's mtime+size+inode and the git store's stamp (see
+    ``_REDUCE_CACHE``). The returned object is SHARED between callers on a cache
+    hit — treat it as read-only.
     """
     root = Path(run_dir.root)
+    # ponytail: no lock around the miss path — N cold concurrent readers each fold
+    # (duplicate work, never a torn read: ``reduced`` is fully built before the
+    # single atomic dict store). Add a module lock if thundering herd shows up.
+    key = str(_resolve(root))
     stamp = _events_stamp(root)
     if stamp is not None:
-        hit = _REDUCE_CACHE.get(str(root))
+        hit = _REDUCE_CACHE.get(key)
         if hit is not None and hit[0] == stamp:
             return hit[1]
     reduced = _reduce_run_uncached(run_dir)
     # Re-stat after the fold: if the log grew while we were folding, this result may
     # already be stale, so don't cache it — the next call re-folds.
     if stamp is not None and _events_stamp(root) == stamp:
-        _REDUCE_CACHE[str(root)] = (stamp, reduced)
+        _REDUCE_CACHE[key] = (stamp, reduced)
+        while len(_REDUCE_CACHE) > _REDUCE_MAX:  # dicts are insertion-ordered → FIFO
+            _REDUCE_CACHE.pop(next(iter(_REDUCE_CACHE)))
     return reduced
 
 
--- a/core/tests/test_dashboard_cache.py
+++ b/core/tests/test_dashboard_cache.py
@@ -1,0 +1,151 @@
+"""``reduce_run`` is memoized on events.jsonl mtime+size (issue #119).
+
+The dashboard backend re-folds the whole event log on every request and every SSE
+tick, so the reduction is memoized. A stale reduction served during a LIVE run would
+be worse than no cache at all, so the invalidation cases are the ones that matter:
+an append must re-fold, and a same-second append (mtime unchanged at 1s resolution)
+must still re-fold because the size moved.
+"""
+
+import json
+import os
+import sys
+import tempfile
+from pathlib import Path
+
+REPO = Path(__file__).resolve().parents[2]
+sys.path.insert(0, str(REPO / "core"))
+
+_EVENTS = [
+    {"kind": "splits", "train": 4, "val": 2, "test": 2, "seed": 0},
+    {"kind": "baseline", "val": 0.25, "stderr": 0.0},
+    {"kind": "step", "candidate": "cand_0001", "accept": True, "reason": "up",
+     "val": 0.75, "parent": "seed", "parent_val": 0.25},
+]
+_BASELINE = {"val": {"reward": 0.25, "per_task": [
+    {"task_id": "t1", "reward": 0.0}, {"task_id": "t2", "reward": 0.5}]}, "best_id": "seed"}
+
+
+def _mk_run(tmp: Path, events=_EVENTS):
+    from cap_evolve import Budget, RunDir
+    rd = RunDir.create(tmp, ts="t", budget=Budget())
+    rd.events_path.write_text("\n".join(json.dumps(e) for e in events) + "\n",
+                              encoding="utf-8")
+    (rd.root / "baseline.json").write_text(json.dumps(_BASELINE), encoding="utf-8")
+    return rd
+
+
+def _spy(monkeypatch):
+    """Count real folds by wrapping the uncached reducer."""
+    from cap_evolve import dashboard
+    dashboard._REDUCE_CACHE.clear()
+    calls = []
+    real = dashboard._reduce_run_uncached
+
+    def counting(run_dir):
+        calls.append(1)
+        return real(run_dir)
+
+    monkeypatch.setattr(dashboard, "_reduce_run_uncached", counting)
+    return dashboard, calls
+
+
+def test_unchanged_event_log_hits_the_cache(monkeypatch):
+    dashboard, calls = _spy(monkeypatch)
+    with tempfile.TemporaryDirectory() as d:
+        rd = _mk_run(Path(d))
+        first = dashboard.reduce_run(rd)
+        second = dashboard.reduce_run(rd)
+        assert len(calls) == 1, f"expected one fold, got {len(calls)}"
+        assert second is first  # served from cache, not re-derived
+        assert first["summary"]["best_val"] == 0.75
+
+
+def test_appending_an_event_invalidates_the_cache(monkeypatch):
+    """The critical case: a live run appends, the next read must NOT be stale."""
+    dashboard, calls = _spy(monkeypatch)
+    with tempfile.TemporaryDirectory() as d:
+        rd = _mk_run(Path(d))
+        before = dashboard.reduce_run(rd)
+        assert before["summary"]["counts"]["total"] == 2  # seed + cand_0001
+        with rd.events_path.open("a", encoding="utf-8") as fh:
+            fh.write(json.dumps({"kind": "step", "candidate": "cand_0002",
+                                 "accept": True, "reason": "up", "val": 0.9,
+                                 "parent": "cand_0001", "parent_val": 0.75}) + "\n")
+        after = dashboard.reduce_run(rd)
+        assert len(calls) == 2, "append must re-fold"
+        assert after["summary"]["counts"]["total"] == 3
+        assert after["summary"]["best_val"] == 0.9  # fresh, not the cached 0.75
+
+
+def test_same_mtime_append_still_invalidates(monkeypatch):
+    """mtime alone is not enough: force the mtime back, size must still bust the key."""
+    dashboard, calls = _spy(monkeypatch)
+    with tempfile.TemporaryDirectory() as d:
+        rd = _mk_run(Path(d))
+        dashboard.reduce_run(rd)
+        st = rd.events_path.stat()
+        with rd.events_path.open("a", encoding="utf-8") as fh:
+            fh.write(json.dumps({"kind": "step", "candidate": "cand_0002",
+                                 "accept": True, "reason": "up", "val": 0.9,
+                                 "parent": "cand_0001"}) + "\n")
+        # Pretend the append happened within one mtime tick of the cached read.
+        os.utime(rd.events_path, ns=(st.st_atime_ns, st.st_mtime_ns))
+        assert rd.events_path.stat().st_mtime_ns == st.st_mtime_ns
+        after = dashboard.reduce_run(rd)
+        assert len(calls) == 2, "size change must invalidate even at identical mtime"
+        assert after["summary"]["best_val"] == 0.9
+
+
+def test_two_runs_do_not_share_a_cache_entry(monkeypatch):
+    dashboard, calls = _spy(monkeypatch)
+    with tempfile.TemporaryDirectory() as d1, tempfile.TemporaryDirectory() as d2:
+        a = _mk_run(Path(d1))
+        b = _mk_run(Path(d2), events=_EVENTS + [
+            {"kind": "step", "candidate": "cand_0002", "accept": True,
+             "reason": "up", "val": 0.95, "parent": "cand_0001"}])
+        ra, rb = dashboard.reduce_run(a), dashboard.reduce_run(b)
+        assert len(calls) == 2
+        assert ra["summary"]["best_val"] == 0.75
+        assert rb["summary"]["best_val"] == 0.95
+        # and each still hits its own entry
+        assert dashboard.reduce_run(a) is ra and dashboard.reduce_run(b) is rb
+        assert len(calls) == 2
+
+
+def test_repeat_reduce_reuses_the_same_object_behavioral():
+    """No internals: on an unchanged log the second reduce returns the SAME object.
+
+    Fails on the pre-#119 reducer (each call built a fresh structure) and is the
+    property the dashboard's per-request / per-SSE-tick path actually relies on.
+    """
+    from cap_evolve import dashboard
+    with tempfile.TemporaryDirectory() as d:
+        rd = _mk_run(Path(d))
+        assert dashboard.reduce_run(rd) is dashboard.reduce_run(rd)
+
+
+def test_reduce_after_append_reflects_the_new_event_behavioral():
+    """No internals: appending must be visible on the very next reduce."""
+    from cap_evolve import dashboard
+    with tempfile.TemporaryDirectory() as d:
+        rd = _mk_run(Path(d))
+        assert dashboard.reduce_run(rd)["summary"]["best_val"] == 0.75
+        with rd.events_path.open("a", encoding="utf-8") as fh:
+            fh.write(json.dumps({"kind": "step", "candidate": "cand_0002",
+                                 "accept": True, "reason": "up", "val": 0.9,
+                                 "parent": "cand_0001"}) + "\n")
+        r = dashboard.reduce_run(rd)
+        assert r["summary"]["best_val"] == 0.9
+        assert r["summary"]["counts"]["total"] == 3
+
+
+def test_missing_event_log_is_never_cached(monkeypatch):
+    """No events.jsonl → no stable stamp → always re-fold (correct over fast)."""
+    dashboard, calls = _spy(monkeypatch)
+    with tempfile.TemporaryDirectory() as d:
+        rd = _mk_run(Path(d))
+        rd.events_path.unlink()
+        dashboard.reduce_run(rd)
+        dashboard.reduce_run(rd)
+        assert len(calls) == 2
--- a/core/tests/test_dashboard_cache.py
+++ b/core/tests/test_dashboard_cache.py
@@ -1,10 +1,16 @@
-"""``reduce_run`` is memoized on events.jsonl mtime+size (issue #119).
+"""``reduce_run`` is memoized on events.jsonl mtime+size+inode and the git store's
+stamp (issue #119).
 
 The dashboard backend re-folds the whole event log on every request and every SSE
 tick, so the reduction is memoized. A stale reduction served during a LIVE run would
 be worse than no cache at all, so the invalidation cases are the ones that matter:
-an append must re-fold, and a same-second append (mtime unchanged at 1s resolution)
-must still re-fold because the size moved.
+an append must re-fold; a same-second append (mtime unchanged at 1s resolution) must
+still re-fold because the size moved; a same-size replace-via-rename must re-fold
+because the inode moved; and a git commit must re-fold even though the harness logs
+its event BEFORE committing, so the event log has not moved at all.
+
+Also pinned here: the FIFO cap (the dashboard is long-lived) and the resolved cache
+key (an absolute and a relative path to one run dir must share one entry).
 """
 
 import json
@@ -38,7 +44,9 @@ def _mk_run(tmp: Path, events=_EVENTS):
 def _spy(monkeypatch):
     """Count real folds by wrapping the uncached reducer."""
     from cap_evolve import dashboard
-    dashboard._REDUCE_CACHE.clear()
+    # A fresh dict rather than .clear() on the shared global, so monkeypatch restores
+    # whatever the rest of the suite had instead of leaving it permanently emptied.
+    monkeypatch.setattr(dashboard, "_REDUCE_CACHE", {})
     calls = []
     real = dashboard._reduce_run_uncached
 
@@ -149,3 +157,107 @@ def test_missing_event_log_is_never_cached(monkeypatch):
         dashboard.reduce_run(rd)
         dashboard.reduce_run(rd)
         assert len(calls) == 2
+
+
+def test_a_git_commit_with_no_new_event_invalidates(monkeypatch):
+    """B1: the harness logs "step" then commits 33 lines LATER (harness.py:1323/:1356).
+
+    So a commit lands with the event log untouched. Before the git stamp was folded
+    into the key, ``git_log`` stayed stale for the rest of the iteration.
+    """
+    from cap_evolve.store import make_store
+    dashboard, calls = _spy(monkeypatch)
+    with tempfile.TemporaryDirectory() as d:
+        rd = _mk_run(Path(d))
+        store = make_store({"kind": "git"}, rd.root)
+        store.init()
+        assert store.commit("iter 1: ACCEPT cand_0001")["ok"]
+        before = dashboard.reduce_run(rd)
+        assert len(before["summary"]["git_log"]) == 1
+        ev_stamp = rd.events_path.stat()
+        assert store.commit("iter 2: reject cand_0002")["ok"]
+        # the whole point: the event log did NOT move
+        assert rd.events_path.stat().st_mtime_ns == ev_stamp.st_mtime_ns
+        assert rd.events_path.stat().st_size == ev_stamp.st_size
+        after = dashboard.reduce_run(rd)
+        assert len(calls) == 2, "a new commit must re-fold"
+        assert len(after["summary"]["git_log"]) == 2, "git_log must show the new commit"
+
+
+def test_cache_is_capped_and_evicts_oldest(monkeypatch):
+    """B2: the dashboard is long-lived and list_runs installs one entry per run."""
+    dashboard, _ = _spy(monkeypatch)
+    monkeypatch.setattr(dashboard, "_REDUCE_MAX", 3)
+    with tempfile.TemporaryDirectory() as d:
+        roots = []
+        for i in range(6):
+            rd = _mk_run(Path(d) / f"r{i}")
+            dashboard.reduce_run(rd)
+            roots.append(str(Path(rd.root).resolve()))
+        assert len(dashboard._REDUCE_CACHE) == 3, "cap must hold"
+        # FIFO: the three newest survive, the three oldest are gone
+        assert list(dashboard._REDUCE_CACHE) == roots[-3:]
+
+
+def test_absolute_and_relative_paths_share_one_entry(monkeypatch):
+    """B3: RunDir.open does not resolve, so the key must (rundir.py:201)."""
+    from cap_evolve import RunDir
+    dashboard, calls = _spy(monkeypatch)
+    with tempfile.TemporaryDirectory() as d:
+        rd = _mk_run(Path(d) / "base")
+        abs_root = Path(rd.root).resolve()
+        cwd = os.getcwd()
+        try:
+            os.chdir(abs_root.parent)
+            a = dashboard.reduce_run(RunDir.open(abs_root))
+            b = dashboard.reduce_run(RunDir.open(Path(abs_root.name)))
+        finally:
+            os.chdir(cwd)
+        assert len(dashboard._REDUCE_CACHE) == 1, dict.keys(dashboard._REDUCE_CACHE)
+        assert len(calls) == 1 and b is a
+
+
+def test_truncating_the_event_log_invalidates(monkeypatch):
+    """N7: shrink was untested; size moving down must invalidate like moving up."""
+    dashboard, calls = _spy(monkeypatch)
+    with tempfile.TemporaryDirectory() as d:
+        rd = _mk_run(Path(d))
+        assert dashboard.reduce_run(rd)["summary"]["counts"]["total"] == 2
+        lines = rd.events_path.read_text(encoding="utf-8").splitlines()[:2]
+        rd.events_path.write_text("\n".join(lines) + "\n", encoding="utf-8")
+        after = dashboard.reduce_run(rd)
+        assert len(calls) == 2 and after["summary"]["counts"]["total"] == 1
+
+
+def test_same_size_replace_via_rename_invalidates(monkeypatch):
+    """N1: st_ino in the key catches a same-size rename-over at an identical mtime."""
+    dashboard, calls = _spy(monkeypatch)
+    with tempfile.TemporaryDirectory() as d:
+        rd = _mk_run(Path(d))
+        assert dashboard.reduce_run(rd)["summary"]["best_val"] == 0.75
+        st = rd.events_path.stat()
+        swapped = [dict(e) for e in _EVENTS]
+        swapped[2]["val"] = 0.95  # same JSON byte length as 0.75
+        tmp = rd.root / "events.new"
+        tmp.write_text("\n".join(json.dumps(e) for e in swapped) + "\n", encoding="utf-8")
+        assert tmp.stat().st_size == st.st_size, "probe must hold size constant"
+        os.replace(tmp, rd.events_path)
+        os.utime(rd.events_path, ns=(st.st_atime_ns, st.st_mtime_ns))
+        assert rd.events_path.stat().st_mtime_ns == st.st_mtime_ns
+        after = dashboard.reduce_run(rd)
+        assert len(calls) == 2, "a new inode must invalidate at identical mtime+size"
+        assert after["summary"]["best_val"] == 0.95
+
+
+def test_mutating_a_returned_reduction_is_visible_to_the_next_caller():
+    """N3 tripwire: the shared-object contract is read-only and NOT defended.
+
+    This pins the documented behaviour so a future deepcopy-on-hit (which would eat
+    the whole win) is a deliberate, test-breaking decision rather than an accident.
+    """
+    from cap_evolve import dashboard
+    with tempfile.TemporaryDirectory() as d:
+        rd = _mk_run(Path(d))
+        first = dashboard.reduce_run(rd)
+        first["summary"]["best_val"] = "MUTATED"
+        assert dashboard.reduce_run(rd)["summary"]["best_val"] == "MUTATED"
--- a/dashboard/backend/capevolve_dashboard/app.py
+++ b/dashboard/backend/capevolve_dashboard/app.py
@@ -25,8 +25,10 @@ def create_app(base_dir: Path, static_dir: Path | None = None) -> FastAPI:
         return {"ok": True, "base_dir": str(base)}
 
     @app.get("/api/runs")
-    def get_runs():
-        return runs.list_runs(base)
+    def get_runs(limit: int | None = Query(default=None, ge=1),
+                 offset: int = Query(default=0, ge=0)):
+        # Unpaginated by default (back-compat: the SPA hub expects a plain array).
+        return runs.list_runs(base, limit=limit, offset=offset)
 
     @app.get("/api/runs/{run_id}")
     def get_run(run_id: str):
@@ -42,8 +44,11 @@ def create_app(base_dir: Path, static_dir: Path | None = None) -> FastAPI:
             raise HTTPException(status_code=404, detail="run not found")
 
     @app.get("/api/runs/{run_id}/rollouts")
-    def get_rollouts(run_id: str, split: str | None = Query(default=None)):
-        return trajectories.list_rollouts(_resolve_or_404(run_id), split)
+    def get_rollouts(run_id: str, split: str | None = Query(default=None),
+                     limit: int | None = Query(default=None, ge=1),
+                     offset: int = Query(default=0, ge=0)):
+        return trajectories.list_rollouts(_resolve_or_404(run_id), split,
+                                          limit=limit, offset=offset)
 
     @app.get("/api/runs/{run_id}/rollout/{file_name}")
     def get_rollout(run_id: str, file_name: str):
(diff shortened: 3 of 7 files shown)


OsherElhadad:
## 🔬 Evidence

Worktree `/tmp/wt-119` (branch `perf/issue-119-reduce-cache`); BEFORE numbers from a real `origin/main` worktree at `/tmp/wt-119-main` (`git worktree add --detach /tmp/wt-119-main origin/main` → `HEAD is now at 6fca097`). Python 3.14, `/tmp/ce-venv`.

### 1. Baseline before any edit (origin/main)
```console
$ PYTHONPATH=/tmp/wt-119/core /tmp/ce-venv/bin/python -m pytest core/tests -q
179 passed in 60.10s (0:01:00)
```

### 2. Fail-before / pass-after (behavioral, no internals)

Only `core/cap_evolve/dashboard.py` is reverted to `origin/main`; the new test file stays.
```console
$ git stash push core/cap_evolve/dashboard.py
$ PYTHONPATH=/tmp/wt-119/core python -m pytest core/tests/test_dashboard_cache.py -q -k behavioral
    def test_repeat_reduce_reuses_the_same_object_behavioral():
        from cap_evolve import dashboard
        with tempfile.TemporaryDirectory() as d:
            rd = _mk_run(Path(d))
>           assert dashboard.reduce_run(rd) is dashboard.reduce_run(rd)
E           AssertionError: assert {'graph': {...}, 'summary': {...}} is {'graph': {...}, 'summary': {...}}
FAILED core/tests/test_dashboard_cache.py::test_repeat_reduce_reuses_the_same_object_behavioral
1 failed, 1 passed, 5 deselected in 0.06s

$ git stash pop
$ PYTHONPATH=/tmp/wt-119/core python -m pytest core/tests/test_dashboard_cache.py -q
7 passed in 0.06s
```

### 3. Core suite after the change
```console
$ cd /tmp/wt-119 && PYTHONPATH=/tmp/wt-119/core /tmp/ce-venv/bin/python -m pytest core/tests -q -k "not test_maybe_launch_spawns_when_available"
.........................................                                [100%]
185 passed, 1 deselected in 59.43s
```

**Note on the one deselected test.** `test_dashboard_launch.py::test_maybe_launch_spawns_when_available` hard-codes port 7878 and currently fails in this environment because another process holds that port, so the launcher falls through to 7882. It **fails identically on unmodified `origin/main`** — pre-existing, unrelated to this PR, and it does not touch any file I changed:
```console
$ lsof -nP -iTCP:7878 -sTCP:LISTEN
Python  20104 osherelhadad    6u  IPv4 ...  TCP 127.0.0.1:7878 (LISTEN)

$ cd /tmp/wt-119-main && PYTHONPATH=/tmp/wt-119-main/core python -m pytest core/tests/test_dashboard_launch.py -q
E       AssertionError: assert 'http://127.0.0.1:7882' == 'http://127.0.0.1:7878'
1 failed, 6 passed in 0.04s
```
With it included: `1 failed, 185 passed` (= 186 collected, 179 baseline + 7 new).

### 4. Backend suite
```console
$ cd /tmp/wt-119/dashboard/backend && /tmp/ce-venv/bin/python -m pytest tests -q
49 passed, 1 warning in 1.76s
```
Also **303.65 s → 2.06 s**: the new SSE test injects a `finalize` event so the generator returns immediately rather than holding the connection for the 5-minute idle timeout.

### 5. compileall
```console
$ cd /tmp/wt-119 && /tmp/ce-venv/bin/python -m compileall -q core skills && echo CLEAN
CLEAN
```

### 6. Measured before/after on a REAL run dir

Produced with the documented zero-API example:
```console
$ PATH=/tmp/ce-venv/bin:$PATH bash examples/toy_calc/run.sh
Working directory: /var/folders/.../T/toy_calc.XXXXXX.j3wnj8Cw8E
{
  "best_id": "cand_0001",
  "baseline_val": 0.0,
  "test_reward": 1.0,
  "test_delta": 1.0,
  "iterations": 3,
}
```

That real run dir (15 events / 12 rollouts / 4 candidate dirs) already shows the win — `18.829 ms` → `0.008 ms` per reduction (**2446x**). But #119 is about *long* runs, so it was grown to a realistic long run by appending synthetic `step` events + rollouts + candidate snapshots on top of the real one: **215 events, 404 rollout files, 202 candidate dirs.**

#### BEFORE — actual `origin/main` checkout
```console
$ /tmp/ce-venv/bin/python /tmp/measure_ab.py /tmp/wt-119-main/core /tmp/run_big 20
core=/tmp/wt-119-main/core  run=/tmp/run_big  n=20
  reduce_run                       x20:    4591.8 ms  ( 229.591 ms each)
  runs.load_run (HTTP/SSE)         x20:    4388.8 ms  ( 219.442 ms each)
  trajectories.diff_candidate      x20:    5550.4 ms  ( 277.522 ms each)
  runs.list_runs                   x20:    5105.7 ms  ( 255.285 ms each)
```

#### AFTER — this branch, identical inputs
```console
$ /tmp/ce-venv/bin/python /tmp/measure_ab.py /tmp/wt-119/core /tmp/run_big 20
core=/tmp/wt-119/core  run=/tmp/run_big  n=20
  reduce_run                       x20:       0.2 ms  (   0.009 ms each)
  runs.load_run (HTTP/SSE)         x20:       1.1 ms  (   0.056 ms each)
  trajectories.diff_candidate      x20:    1140.9 ms  (  57.044 ms each)
  runs.list_runs                   x20:      26.0 ms  (   1.299 ms each)
```

Per-operation, both independent A/B runs (run 1 / run 2 — the spread is machine noise; the ratio is stable):

| operation, ms each (x20) | BEFORE | AFTER | speedup |
|---|---|---|---|
| `reduce_run` | 275.557 / 229.591 | **0.009 / 0.009** | ~25,000-30,000x |
| `runs.load_run` (HTTP + every SSE tick) | 254.873 / 219.442 | **0.061 / 0.056** | ~4,000x |
| `trajectories.diff_candidate` | 301.376 / 277.522 | **53.482 / 57.044** | ~5x |
| `runs.list_runs` | 274.170 / 255.285 | **0.791 / 1.299** | ~200-350x |

The table in the PR description quotes run 1.

`diff_candidate` keeps 53 ms because `build_diffs` still walks 202 candidate dirs — that is real work, not a re-reduce; the re-reduce it did on top of that is gone. (The `export_static` note about doing it once for the whole graph is a bigger refactor and stays out of this diff.)

### 7. Live-run correctness — invalidation on the 202-candidate run dir

The critical property: a cache that serves stale data during a live run is worse than no cache. Reduce, confirm the cache hit, **append a real event to `events.jsonl`**, reduce again, assert the new candidate is visible.
```console
reduce #1                : total nodes: 203
reduce #2                : is the cached object: True
reduce #3 after append   : fresh object: True   total nodes: 204   (208.6 ms re-fold)
                           cand_8888 present in graph: True  -> NO stale serve during a live run
reduce #4                : re-cached at the new stamp: True  (0.019 ms)
```

And the same-mtime case, which is why the key is mtime **and** size (`core/tests/test_dashboard_cache.py::test_same_mtime_append_still_invalidates`): the test appends, then forces the mtime back with `os.utime(..., ns=(...))`, asserts `st_mtime_ns` is genuinely unchanged, and still requires a re-fold. Passing.

### 8. Pagination — bounds and tiling
```console
--- list_rollouts on one run (404 matching rollout files) ---
  full (no limit)        rows= 404    17.67 ms
  limit=50               rows=  50     4.15 ms
  limit=50&offset=50     rows=  50     5.04 ms
  pages tile the full list exactly: True (404 rows)
  offset past the end -> []

--- list_runs over 12 such long runs, COLD cache (every row = a real reduction) ---
  full (no limit)        rows= 12     3087.8 ms
  limit=5                rows=  5     1098.9 ms
  limit=5&offset=5       rows=  5     1155.7 ms
  pages tile the full list exactly: True (12 runs)
  offset past the end -> []
```

Bounds validation is a FastAPI `Query(ge=...)`, asserted in `test_runs_endpoint_limit_and_offset`: `?limit=0` → `422`, `?offset=-1` → `422`. And `test_rollouts_paging_only_opens_the_page` spies on `json.loads` to prove a `limit=5` page over 20 files parses exactly 5.

### 9. Frontend (one type annotation touched)
```console
$ cd /tmp/wt-119/dashboard/frontend && npm ci && npm run build
npm ci OK
dist/assets/index-b_9cC4ZS.css   16.36 kB │ gzip:   4.51 kB
dist/assets/index-CUh6vNsJ.js   830.26 kB │ gzip: 249.87 kB
✓ built in 774ms

$ npx tsc --noEmit ; echo exit=$?
exit=0

$ npm test -- --run
 Test Files  13 passed (13)
      Tests  45 passed (45)
```

`dashboard/frontend/dist/` is checked into the repo. The build above verified the change compiles, then the dist churn was **reverted** (`git checkout -- dashboard/frontend/dist`) so a hash-renamed bundle does not conflict with #116/#117.

### 10. Final diff — nothing stray
```console
$ cd /tmp/wt-119 && git status --short && git show --stat HEAD
perf(dashboard): memoize reduce_run on events mtime+size, drop dead SSE snapshot, paginate

 core/cap_evolve/dashboard.py                       |  53 +++++++-
 core/tests/test_dashboard_cache.py                 | 151 +++++++++++++++++++++
 dashboard/backend/capevolve_dashboard/app.py       |  24 ++--
 dashboard/backend/capevolve_dashboard/runs.py      |  17 ++-
 .../backend/capevolve_dashboard/trajectories.py    |  47 +++++--
 dashboard/backend/tests/test_pagination.py         | 134 ++++++++++++++++++
 dashboard/frontend/src/lib/types.ts                |   4 +-
 7 files changed, 402 insertions(+), 28 deletions(-)
```

OsherElhadad:
## 🔍 Review — PR #194

The core idea is right and the measured win is real and large — I reproduced it independently (below). What blocks merge is not the cache's happy path but its *scope*: the key covers `events.jsonl` only, and the reducer's output is derived from five other inputs, two of which (`git_log`, seed/rollout-derived per-task) provably change **without** the event log moving. Plus an unbounded cache in a long-lived server process and a cache key that is a raw un-normalized path string.

I want to be explicit that the two adversarial cases the PR description leads with — plain append and forced-same-mtime append — both **pass**. Size is the strong signal and it does its job. A same-second append does **not** serve stale data.

---

### Blocking

**B1. `core/cap_evolve/dashboard.py:623` + `:270` — `git_log` is in the cached payload but the git commit lands *after* the event that would invalidate it, so the graph's git panel serves a permanently stale commit list for the current iteration.**

The docstring at `dashboard.py:243-248` asserts the ordering guarantee: *"the iteration commit lands then `step` is logged. So the stamp advances after those writes, never before."* That is **backwards** in the actual harness:

- `core/cap_evolve/harness.py:1323` — `run_dir.log_event("step", ...)`
- `core/cap_evolve/harness.py:1356` — `store.commit(f"iter {...}: ...")`

The event is logged 33 lines **before** the commit. Same inversion in GEPA: `core/cap_evolve/gepa.py:718` logs `gepa_val_gate`, then `gepa.py:635`/`:639` commit. So the sequence on every iteration is: commit N-1 exists → `step` appended (stamp advances) → reduction cached with `git_log` = N-1 rows → commit N lands (stamp does **not** move) → every subsequent request for the rest of the iteration serves `git_log` missing commit N.

Reproduced:
```console
$ /tmp/ce-venv/bin/python /tmp/gitlog_stale.py
reduce #1 git_log: 1
reduce #2 after a commit lands with NO new event: 1 cached_object= True -> STALE git_log
```

Consequence: the self-contained `dashboard.html` "Iteration store (git)" panel (`dashboard.py:1320-1323`) and `summary.git_log` (`types.ts:127`) are wrong for exactly the window a user is watching a live run. Silent, and it never self-corrects until the *next* event.

Fix (pick one):
- Cheapest and in the spirit of the PR: **fold `.git/HEAD`'s stamp into the key** — `_events_stamp` returns `(ev_mtime_ns, ev_size, head_mtime_ns, head_size)`, stat'ing `root/".git"/"HEAD"` best-effort (missing → zeros). Two stats instead of one; the git ref moves on every commit.
- Or move `git_log` out of the cached reduction into its own `@lru_cache`-less call site (it is already a separate endpoint via `gitlog.py:33`, so the cached reduction arguably should not carry it at all).
- Either way, **correct the docstring at `dashboard.py:243-248`** — as written it documents a guarantee the code does not have, which is how the next person re-introduces this.

**B2. `core/cap_evolve/dashboard.py:270-279` — the cache is unbounded, in a process designed to be long-lived and to serve every run in a `.capevolve` dir.**

One entry per run root, never evicted. `list_runs` reduces *every* discovered run, so a single `GET /api/runs` on a dir with N runs installs N entries permanently. The comment at `:250` calls this "Bounded by run count" — run count is not a bound, it is monotonically increasing for the lifetime of the dashboard.

Measured retained size per entry, 209-candidate run:
```console
$ /tmp/ce-venv/bin/python /tmp/final_probes.py
  one 209-candidate reduction retained: 665 KiB
  100 such runs in one .capevolve dir -> ~65 MiB held forever (no eviction)
```
Confirmed no eviction path exists:
```console
$ /tmp/ce-venv/bin/python /tmp/memgrow.py
cache entries after 1 run: 1
cache entries after 8 more runs: 9
no eviction API present: True
```

Consequence: a dashboard left running over a research session with a few hundred archived runs pins tens to hundreds of MiB it will never release, and the *archived* runs are the ones whose entries are pure waste (their stamp never changes, so they are never even re-validated, just retained).

Fix: an insertion-ordered cap is 3 lines and stdlib —
```python
_REDUCE_MAX = 32
...
    if stamp is not None and _events_stamp(root) == stamp:
        _REDUCE_CACHE[str(root)] = (stamp, reduced)
        while len(_REDUCE_CACHE) > _REDUCE_MAX:
            _REDUCE_CACHE.pop(next(iter(_REDUCE_CACHE)))
```
(dicts are insertion-ordered; pop the oldest. Move-to-end on hit if you want true LRU, but FIFO is enough here.) Add one test asserting the cap holds.

**B3. `core/cap_evolve/dashboard.py:272,279` — the cache key is `str(root)`, un-normalized, so the same run dir reached by two different paths gets two independent entries.**

```console
$ /tmp/ce-venv/bin/python /tmp/keyprobe.py
keys after abs + relative open: ['/tmp/run_big', 'run_big']
same object: False -> duplicate entries
```

`RunDir.open` (`core/cap_evolve/rundir.py:201-205`) does **not** resolve `root`, and `reduce_run` is a public core API called from `skills/phases/report/scripts/run.py:40`, `check.py:81`, `export_static.py:76` and the CLI — relative paths are normal there. Symlinked run dirs give the same split.

Consequence: doubles the (unbounded, per B2) memory for the affected runs and defeats the cache for the second caller. Compounds B2.

Fix: `key = str(root.resolve())` — one call, and `root` is already a `Path` on line 269. Cheap; do it in the same edit as B2.

---

### Non-blocking

**N1. `core/cap_evolve/dashboard.py:258` — same-size, same-mtime rewrite serves stale data.** My probe:
```
3 same mtime + same size edit      size same=True mtime same=True folds=1 best_val=0.75 (want 0.95) -> STALE!
5 replace via rename, same size    size same=True folds=1 best_val=0.75 (want 0.25 seed) -> STALE!
```
I'm calling this non-blocking because `events.jsonl` is genuinely append-only in every writer (`rundir.log_event`), so this is not a real production path — but adding `st_ino` to the tuple closes case 5 (rename-over changes the inode) for one extra field on a stat call you already make. Case 3 (true in-place same-length edit) is only reachable by hand-editing and I would not defend against it.

**N2. `core/cap_evolve/dashboard.py:270-279` — no lock; N concurrent cold requests do N full folds.** The get/compute/set is not atomic across the compute step. Verified there is no *correctness* bug (no half-built entry is publishable — `reduced` is fully built before the dict store, and CPython's dict store is atomic), and no torn reads:
```console
$ /tmp/ce-venv/bin/python /tmp/concur.py
8 concurrent cold reduce_run: folds=8 (a lock would give 1) errors=[] wall=3765 ms
all results equal: True
node counts observed by concurrent readers: [204, 205, 207, 209] | folds during window: 13
```
So this is a thundering-herd *efficiency* miss, not a race: on a cold cache with several browser tabs open you pay 8× 230 ms instead of 1×. A module-level `threading.Lock` around the miss path (or accepting the duplicate work) — either is defensible; just say which in a comment, since the current code says nothing about threads at all and this is a threaded ASGI server.

**N3. `core/cap_evolve/dashboard.py:266-267` — the shared mutable return is documented but not enforced.** "The returned object is SHARED between callers on a cache hit — treat it as read-only" is a docstring-only contract on a *public core API* with out-of-tree callers (`skills/phases/report/scripts/`). I audited every in-tree consumer (`runs.py:74`, `trajectories.py:77`, `export_static.py:77`, `compare.py:28`, `dashboard.py:704,749`, `report/scripts/check.py:82`) — all read-only today, so no live bug. But one future `reduced["summary"]["x"] = ...` silently corrupts every other reader. Consider `copy.deepcopy` on hit (would eat most of the win — don't), or at minimum a test that mutating a returned reduction doesn't leak into the next call, so the contract has a tripwire.

**N4. `dashboard/backend/capevolve_dashboard/runs.py:63` — the newest-first sort key is `p.stat().st_mtime` alone, no tiebreaker.** With N runs created in the same mtime tick the order falls back to `iterdir()` order. It happened to be stable across 5 calls in my probe and pages still tiled correctly, but there is no *guarantee* — and pagination correctness depends on a total order. `key=lambda p: (-p.stat().st_mtime, p.name)` makes it deterministic for free.

**N5. `dashboard/backend/capevolve_dashboard/app.py:28-29,47-52` — no max page size.** `?limit=1000000` is accepted (200):
```
no max page size on /api/runs: 200
no max page size on rollouts : 200
limit=99999999999999999999 -> 200
```
Not a real DoS — the endpoint is `127.0.0.1`-bound and the *unpaginated* default already returns everything, so a huge `limit` is strictly no worse than the current default. But `Query(default=None, ge=1, le=1000)` costs nothing and makes the contract self-documenting. Traversal is clean: `?split=../../etc` → `[]` (the `sp_dir.is_dir()` guard at `trajectories.py:34`), and `resolve_run` (`runs.py:14-25`) already fences run ids.

**N6. `dashboard/backend/capevolve_dashboard/runs.py:57-61` + `trajectories.py:13-21` — the "a page can come back shorter than `limit`" behavior is documented but has no test, and it breaks the tiling invariant the other tests assert.** A corrupt rollout inside a page is skipped *after* windowing, so the client sees a short page and cannot distinguish "short page" from "last page". I verified tiling still reproduces the unpaged list (because the unpaged path skips the same file), so no records are *lost* relative to unpaged:
```
page with 1 corrupt file: len= 2 (limit 3)
unpaged rows: 7 tiled rows: 7 identical: True
```
Consistent, but a client paging until it gets a short page will stop early. Worth a sentence in the docstring saying "short page ≠ end of list; page until you get an empty page" — and a test pinning that.

**N7. No test for the truncation case.** It works (verified below), but `test_dashboard_cache.py` covers append, same-mtime append, two-runs, missing log — not shrink. One 4-line test.

---

### Nits

- `core/cap_evolve/dashboard.py:250` — "Bounded by run count" is misleading (see B2); reword once a real bound exists.
- `dashboard/backend/capevolve_dashboard/app.py:28` — the `# Unpaginated by default` comment is good; mirror it on the rollouts route (`:47`), which has none.
- `core/tests/test_dashboard_cache.py:39` — `_spy` mutates module state (`_REDUCE_CACHE.clear()`) without restoring it; `monkeypatch` restores the function but not the dict. Cross-file pollution didn't materialize (17 passed in both orders — checked), but a `monkeypatch.setattr(dashboard, "_REDUCE_CACHE", {})` is strictly cleaner than `.clear()` on the shared global.

---

### Cache correctness

| Scenario | Invalidates correctly? | How I tested it |
|---|---|---|
| plain append | ✅ | `/tmp/cachematrix.py` case 1 — `folds=2`, `best_val` 0.75→0.9 |
| **same-second append (mtime forced back with `os.utime`)** | ✅ | case 2 — asserted `st_mtime_ns` **identical**, still `folds=2`, `best_val=0.9`. Size is the strong signal and it works. **No stale serve.** |
| same-mtime **and** same-size in-place edit | ❌ | case 3 — `folds=1`, served cached 0.75 when disk said 0.95. Unreachable by any in-tree writer (append-only) → N1 |
| truncate / shrink | ✅ | case 4 — `folds=2`, `counts.total` 2→1 |
| replace via `os.replace` (same size, mtime forced back) | ❌ | case 5 — `folds=1`, stale. `st_ino` in the key would catch it → N1 |
| concurrent read (8 threads, cold) | ✅ correctness / ❌ efficiency | `/tmp/concur.py` — no errors, all results equal, no half-built entry; but 8 folds instead of 1 → N2 |
| concurrent read **while** appending | ✅ | 4 readers + writer: observed node counts `[204, 205, 207, 209]` monotone, final read 209 = max. No reader ever saw a count go backwards. |
| **git commit lands, no event appended** | ❌ | `/tmp/gitlog_stale.py` — cached object returned, `git_log` stuck at 1 row → **B1** |
| rollout file written, no event appended | ❌ (by design) | case 6 — `folds=1`. Acceptable: harness persists rollouts at `harness.py:252` *before* `log_event("evaluate")` at `:311`, so the docstring's claim **is** correct for this one. |
| `final.json` written, no event appended | ❌ (by design) | case 7 — `folds=1`, `test_reward=None`. Also acceptable: `harness.py:1964` writes `final.json` before `log_event("finalize")` at `:1966`. Verified. |
| missing `events.jsonl` | ✅ never cached | covered by the PR's own test; `_events_stamp` returns `None` |

The docstring's write-before-log argument holds for `baseline.json`, `final.json` and `rollouts/` (I checked each call site). It fails **only** for the git store — which is B1.

**Redaction: no regression.** The cache stores the output of `redact()` (`dashboard.py:626` — `return redact({...})` is inside `_reduce_run_uncached`, so redaction happens *before* the value reaches `_REDUCE_CACHE`). Verified with a planted secret:
```console
$ /tmp/ce-venv/bin/python /tmp/redactprobe.py
cache hit: True
secret leaked in first  : False
secret leaked in cached : False
redaction marker present: True
```

---

### Measured win — my numbers

Real run dir from `examples/toy_calc` + `mock` optimizer (15 events / 12 rollouts / 4 candidates), then grown to 215 events / 204 candidates / 412 rollouts. BEFORE = `origin/main` worktree (`/tmp/rv-194-main`, HEAD `6fca097`), core **and** backend both from main; AFTER = this branch. Identical inputs, same machine, warm-up call discarded.

**LARGE log (215 events, 204 candidates, 412 rollouts), x20:**

| op, ms each | BEFORE | AFTER | speedup |
|---|---|---|---|
| `reduce_run` | 230.316 | **0.014** | ~16,000x |
| `runs.load_run` (HTTP + every SSE tick) | 217.643 | **0.064** | ~3,400x |
| `trajectories.diff_candidate` | 268.168 | **63.527** | ~4.2x |
| `runs.list_runs` | 258.636 | **0.107** | ~2,400x |

**SMALL log (the unmodified real run — 15 events), x50:**

| op, ms each | BEFORE | AFTER | speedup |
|---|---|---|---|
| `reduce_run` | 17.955 | **0.008** | ~2,200x |
| `runs.load_run` | 18.276 | **0.056** | ~330x |
| `trajectories.diff_candidate` | 20.498 | **0.840** | ~24x |
| `runs.list_runs` | 19.572 | **0.064** | ~300x |

**Verdict: the claimed speedup reproduces, and it is far outside noise.** My absolute numbers land within ~10% of the PR's (230 vs 229.6 ms BEFORE; 0.014 vs 0.009 ms AFTER) — the ratio is stable across both. It is not a microbenchmark artifact: 217 ms → 0.064 ms on `runs.load_run` is the *actual* HTTP + per-SSE-tick path, at ~2 ticks/s, which is the thing #119 is about. The residual 63 ms on `diff_candidate` is `build_diffs` walking 204 candidate dirs — real work, correctly left out of scope (`export_static.py:73-75` already notes the batch fix).

The small-log case matters too and I'd flag it as *more* than the PR claims: even a 15-event run was paying 18 ms per tick, so this is not only a long-run fix.

---

### Pagination contract

**Shape:** `?limit=<int≥1>&offset=<int≥0>` on `GET /api/runs` and `GET /api/runs/{id}/rollouts`. Response envelope is **unchanged** — still a bare JSON array, no `{items, total, next}` wrapper. Both default to unpaginated (`limit=None`), so back-compat is preserved.

**Boundaries — verified, all correct:**
```console
$ /tmp/ce-venv/bin/python /tmp/pagin.py
default /api/runs rows: 6
limit=0 -> 422 | offset=-1 -> 422
limit=abc -> 422
offset past end -> []
runs tile exactly: True ['run_0','run_1','run_2','run_3','run_4','run_5']
rollouts default rows: 8
rollouts tile exactly: True n= 8
split traversal ../ -> []
```
No dropped or duplicated records at any seam, on either endpoint. Negative/zero/non-numeric rejected at 422 by FastAPI `Query(ge=...)`. Out-of-range offset → `[]`, not an error. Only gap is no upper bound (N5) and no tiebreaker on the runs sort (N4).

**Does the frontend consume it? — No, and that is FINE here.** `dashboard/frontend/src/lib/api.ts:94` is `getJSON<RunSummary[]>('/api/runs')` and `:99-101` builds only `?split=`; zero occurrences of `limit`/`offset` anywhere in `src/`. Because both endpoints default to unpaginated, the SPA keeps getting the full array and nothing silently truncates. I verified end to end (6 runs → 6 rows, 8 rollouts → 8 rows). So pagination is currently a **latent, unused capability** — correct, tested, and dead weight until a follow-up wires it up. Worth an issue so it doesn't rot; not a blocker.

`export_static.py:71` also calls `list_runs(self.base)` with no window → static export unaffected. Confirmed.

---

### Dead SSE snapshot code — genuinely dead ✅

Proved unreferenced, not assumed:
- `dashboard/frontend/src/lib/useRunStream.ts:75` — `es.addEventListener('snapshot', () => dispatch({ type: 'open' }))`. The handler takes **no** event arg; the payload is unreachable by construction. This is the whole proof.
- `grep -rn snapshot` across `dashboard/frontend/src/`: only `types.ts:264-266` (the type this PR narrows), `useRunStream.ts:75` above, and unrelated *candidate-snapshot* hits (`MemoryPanel`, `GitDiff`, `IterationsDiff`, `phases.ts`).
- No test on `origin/main` asserted the payload — checked `dashboard/backend/tests/` and `dashboard/frontend/src/test/` on main: the only `snapshot` hits are `test_memory_api.py:24` (candidate snapshots) and `RunDeepDive.test.tsx:29` (prose in a test name).
- The self-contained `dashboard.html` template does not use SSE at all — it consumes an inlined `__RUN_DATA__` blob (`dashboard.py:707-709`).
- Only mention outside code is a June design doc (`docs/superpowers/plans/2026-06-18-cap-evolve-dashboard-backend.md:738`), not a client.

The narrowed type is also load-bearing: `npx tsc --noEmit` → `exit=0`, so nothing was reading `StreamEvent`'s snapshot payload as `RunDetail`.

**SSE end to end with mid-connection appends: works.** Real run dir, `finalize` stripped so it looks live, two events appended from another thread while the connection was held:
```console
$ /tmp/ce-venv/bin/python /tmp/sse_live.py
GET /api/runs/run_live -> 200 total nodes: 4
stream status: 200
   event: snapshot
   data: {"run_id": "run_live"}
   event: event
   data: {"kind": "step", "candidate": "cand_9999", ...}
   event: event
   data: {"kind": "finalize", "test_reward": 1.0, "best_id": "cand_9999"}
   event: done
after appends, cand_9999 in graph: True | best_val: 1.0 | test_reward: 1.0
```
Frames arrive in order, the open marker carries only `run_id`, and the post-append refetch reflects the appended data — i.e. the cache invalidated on the live path.

---

### Merge-order note

**No `dist/` churn — confirmed:** `git diff --name-only origin/main...HEAD | grep -i dist` → empty. Epic #127's rebuild-never-hand-merge policy is not engaged by this PR. (The build does emit a new hashed bundle; the author reverted it, and I re-verified the tree is clean after my own `npm run build`.)

I ran **real trial merges**, not just file-overlap:
```console
$ for b in feat/issue-117-event-ticker feat/issue-142-protected-paths fix/issue-112-passk-na; do
    git merge --no-commit --no-ff origin/$b >/dev/null 2>&1; git diff --name-only --diff-filter=U; git merge --abort; done
(no output — zero conflicted files in all three)
```

| Branch | Overlap | Conflict |
|---|---|---|
| `feat/issue-116-follow-tail` (#191) | **none** | ✅ clean |
| `feat/issue-117-event-ticker` | `core/cap_evolve/dashboard.py` | ✅ clean (its hunk is at old `:535`, inside the reducer body; #194's is at `:229-283`, above it) |
| `feat/issue-142-protected-paths` (#197) | `core/cap_evolve/dashboard.py` | ✅ clean (hunk at old `:567`, the summary dict) |
| `fix/issue-112-passk-na` (#187) | `dashboard/frontend/src/lib/types.ts` | ✅ clean (different regions) |
| `fix/issue-120-cdn-font` (#192) | none | ✅ clean |

**vs #191 specifically — the flagged highest-risk overlap is a false alarm.** #191 moves the tail *body* out of `stream.py` into `core/cap_evolve/eventstream.py` and **re-exports `read_new_events` from `stream.py`** (`stream.py` diff: `from cap_evolve.eventstream import read_new_events  # noqa: F401 — re-export`). #194 touches `app.py`'s `gen()` but calls the tail only through `_stream.read_new_events(...)` (`app.py:121`) — the re-exported name #191 deliberately preserves. Zero file overlap, and the one API they share is API-stable across the refactor. **Either order works.**

**Recommended order: #194 → #191 → #197 → #117 → #187 → #192.** Put #194 early since it's the only one that is a pure win for the others (they all make the dashboard busier), and it's textually above every sibling's `dashboard.py` hunk. Rebuild `dist/` once after the #187/#192/#117 frontend cluster lands, per #127.

**One cross-PR caution:** if #191's `eventstream.py` ever gains its own read caching, the two caches must agree on invalidation — the byte-offset tail and the mtime+size reduction are keyed differently by design. Not a conflict today; worth a note in whichever lands second.

---

### Verification I re-ran

Core suite — **186 collected (179 baseline + 7 new), 1 pre-existing failure**:
```console
$ cd /tmp/rv-194 && PYTHONPATH=/tmp/rv-194/core /tmp/ce-venv/bin/python -m pytest core/tests -q
1 failed, 185 passed in 60.50s (0:01:00)
```
Pre-existing, port collision, unrelated — **independently confirmed on `origin/main`**:
```console
$ lsof -nP -iTCP:7878 -sTCP:LISTEN
Python  83469 osherelhadad ... TCP 127.0.0.1:7878 (LISTEN)

$ cd /tmp/rv-194-main && PYTHONPATH=/tmp/rv-194-main/core /tmp/ce-venv/bin/python -m pytest core/tests/test_dashboard_launch.py -q
1 failed, 6 passed in 0.05s
```
The author's claim on this stands.

Backend suite:
```console
$ cd /tmp/rv-194/dashboard/backend && /tmp/ce-venv/bin/python -m pytest tests -q
49 passed, 1 warning in 2.13s
```

Fail-before / pass-after — reverted **only** `dashboard.py` to main, kept the new test file:
```console
$ git checkout origin/main -- core/cap_evolve/dashboard.py
$ PYTHONPATH=/tmp/rv-194/core pytest core/tests/test_dashboard_cache.py -q
FAILED ...::test_appending_an_event_invalidates_the_cache
FAILED ...::test_same_mtime_append_still_invalidates
FAILED ...::test_two_runs_do_not_share_a_cache_entry
FAILED ...::test_repeat_reduce_reuses_the_same_object_behavioral
FAILED ...::test_missing_event_log_is_never_cached
6 failed, 1 passed in 0.06s
```
Reproduces (I get 6 failed vs the PR's narrower `-k behavioral` run — same conclusion, stronger).

Test-order pollution — clean both directions:
```console
$ pytest core/tests/test_dashboard_cache.py core/tests/test_dashboard.py -q   → 17 passed
$ pytest core/tests/test_dashboard.py core/tests/test_dashboard_cache.py -q   → 17 passed
```

compileall (added `dashboard/backend`, which the PR's run omitted):
```console
$ /tmp/ce-venv/bin/python -m compileall -q core skills dashboard/backend && echo CLEAN
CLEAN
```

Frontend:
```console
$ npm ci && npx tsc --noEmit; echo "tsc exit=$?"
npm ci OK
tsc exit=0
$ npm run build   → ✓ built in 765ms
$ npm test -- --run
 Test Files  13 passed (13)
      Tests  45 passed (45)
$ git status --short   → clean (dist churn removed)
```

Zero new runtime deps in core — `dashboard.py` imports remain `html, json, os, re, shutil, subprocess, pathlib` only. CONTRIBUTING satisfied.

---

**To unblock:** B1 (git stamp in the key + fix the docstring's ordering claim), B2 (cap the dict), B3 (`root.resolve()`). All three are small and land in the same function; B2+B3 are ~5 lines together. The perf work itself is sound and I'd merge it the moment the key covers what the payload actually contains.

OsherElhadad:
## 🔧 Review fixes

All 3 blocking + 7 non-blocking + 3 nits addressed in `a79280a`. Verification below is literal terminal output. Python 3.14, `/tmp/ce-venv`, worktree `/tmp/fx-194`; BEFORE numbers come from stashing only `core/cap_evolve/dashboard.py` back to the reviewed commit `05643ac`, or from a real `origin/main` worktree (`/tmp/fx-194-main`, HEAD `6fca097`) where the comparison is against pre-cache behaviour.

---

### Cache correctness

| Scenario | Invalidates? | Test pinning it |
|---|---|---|
| plain append | ✅ | `test_appending_an_event_invalidates_the_cache` |
| same-second append (`os.utime` forces mtime back) | ✅ | `test_same_mtime_append_still_invalidates` |
| truncate / shrink | ✅ | `test_truncating_the_event_log_invalidates` **(new)** |
| same-size replace via `os.replace`, mtime forced back | ✅ **now** (was ❌) | `test_same_size_replace_via_rename_invalidates` **(new)** |
| **git commit lands, event log untouched** | ✅ **now** (was ❌ — B1) | `test_a_git_commit_with_no_new_event_invalidates` **(new)** |
| nothing changed (must HIT) | ✅ hits | `test_unchanged_event_log_hits_the_cache` |
| two different runs | ✅ separate entries | `test_two_runs_do_not_share_a_cache_entry` |
| abs vs relative vs symlinked path to one run | ✅ **one** entry now (was 3 — B3) | `test_absolute_and_relative_paths_share_one_entry` **(new)** |
| >`_REDUCE_MAX` runs reduced | ✅ evicts FIFO (was unbounded — B2) | `test_cache_is_capped_and_evicts_oldest` **(new)** |
| missing `events.jsonl` | ✅ never cached | `test_missing_event_log_is_never_cached` |
| rollouts / `baseline.json` / `final.json` written before their event | ✅ by ordering | covered by the append tests; ordering re-verified at `harness.py:252`/`:311`, `:1964`/`:1966` |
| same-mtime **and** same-size **in-place** edit | ❌ **documented gap** | see N1 below |
| mutating a returned reduction leaks to the next caller | ⚠️ by design | `test_mutating_a_returned_reduction_is_visible_to_the_next_caller` **(new tripwire)** |

Full matrix re-run after the key change:

```console
$ /tmp/ce-venv/bin/python /tmp/matrix.py
plain append                                         folds=2  mtime_same=False size_same=False  invalidated=YES  best_val=0.9
same-second append (mtime forced back)               folds=2  mtime_same=True  size_same=False  invalidated=YES  best_val=0.9
truncate / shrink                                    folds=2  mtime_same=False size_same=False  invalidated=YES  counts.total=1
same-size replace via os.replace (mtime forced)      folds=2  mtime_same=True  size_same=True   invalidated=YES  best_val=0.95
same-mtime + same-size IN-PLACE edit (documented gap) folds=1  mtime_same=True  size_same=True   invalidated=no   best_val=0.75
nothing changed (must HIT)                           folds=1  mtime_same=True  size_same=True   invalidated=no   best_val=0.75
first git commit lands, no event                     folds=2  mtime_same=True  size_same=True   invalidated=YES  git_log=1
SECOND git commit lands, no event                    folds=2  mtime_same=True  size_same=True   invalidated=YES  git_log=2 (want 2)
missing events.jsonl (never cached)                  folds=2  -> always re-folds
```

Note row 2: **same-second append still invalidates** — no regression on what already worked.

---

### B1 proof — git-commit staleness fixed

Your scenario reproduced exactly: commit lands, `events.jsonl` verifiably untouched.

```console
############ BEFORE (reviewed commit 05643ac) ############
reduce #1 git_log rows: 1
events.jsonl moved?           False
reduce #2 git_log rows: 1  cached_object= True -> STALE git_log

############ AFTER (this fix) ############
reduce #1 git_log rows: 1
events.jsonl moved?           False
reduce #2 git_log rows: 2  cached_object= False -> FRESH git_log
```

**One correction to your suggested fix: `.git/HEAD` does not work.** It is a *constant symbolic ref* (`ref: refs/heads/<branch>`) — its bytes and mtime never change on a commit. Measured on a store built by the real `VersionStore.commit`:

```console
commit 2 (no tree change): moved -> ['logs/HEAD', 'index', 'refs/heads/master', 'refs/heads', '.']
commit 3 (no tree change): moved -> ['logs/HEAD', 'index', 'refs/heads/master', 'refs/heads', '.']
commit 6 (tree change):    moved -> ['logs/HEAD', 'index', 'refs/heads/master', 'refs/heads', '.']

$ # and directly:
HEAD               changed=False  before=(1785364438140548009, 23) after=(1785364438140548009, 23)
logs/HEAD          changed=True   before=(...,129) after=(...,248)
refs/heads/master  changed=True
```

`refs/heads/<branch>` moves but the branch name is not knowable without reading `HEAD`. So the stamp is `.git`'s own mtime (moves every commit via `index.lock` create+rename) **plus** `.git/logs/HEAD` (the append-only reflog) — both branch-agnostic, either moving is enough. Belt and braces because the reflog can be disabled (`core.logAllRefUpdates false`), and in that case `.git`'s mtime still moves — verified.

**Corrected docstring** (`core/cap_evolve/dashboard.py:232-262`), replacing the claim that was backwards:

```python
#   * The event stamp alone covers baseline.json, final.json, rollouts/ and state.json,
#     because each is written BEFORE the event reporting it: baseline.json then
#     log_event("baseline"), rollout files (harness.py:252) then log_event("evaluate")
#     (:311), final.json (:1964) then log_event("finalize") (:1966).
#   * It does NOT cover the git store, whose ordering is the OPPOSITE: the harness logs
#     "step" (harness.py:1323) and only commits 33 lines LATER (:1356); GEPA likewise
#     logs (gepa.py:718) then commits (:635). So a commit lands with the event log
#     untouched and ``git_log`` (:623) would stay stale for the rest of the iteration.
#     Hence ``_git_stamp``. NOT ``.git/HEAD``: that is a constant symbolic ref
#     ("ref: refs/heads/<branch>") and never changes on a commit — measured. ``.git``'s
#     own mtime moves on every commit (index.lock create+rename) and ``.git/logs/HEAD``
#     is the append-only reflog; both are branch-name agnostic, and either moving is
#     enough to invalidate.
```

And the append-only invariant is now stated where the next person will read it:

```python
#   * INVARIANT the events part relies on: ``events.jsonl`` is only ever APPENDED to
#     (``RunDir.log_event`` opens it "a"). A future writer that rewrites it in place
#     to the same byte length within one mtime tick would serve a stale reduction —
#     if you add such a writer, hash the file or bump a counter instead.
```

---

### B2 proof — the cache is now bounded (cap: 32 entries)

```console
###### BEFORE (reviewed commit) ######
cap (_REDUCE_MAX): NONE — unbounded
  after   1 runs reduced: cache entries=  1   retained ~=      431 KiB
  after   9 runs reduced: cache entries=  9   retained ~=     3883 KiB
  after  32 runs reduced: cache entries= 32   retained ~=    13807 KiB
  after  33 runs reduced: cache entries= 33   retained ~=    14239 KiB
  after  40 runs reduced: cache entries= 40   retained ~=    17259 KiB
100 runs in one .capevolve dir -> held: 42.1 MiB  (unbounded would be 42.1 MiB)

###### AFTER (this fix) ######
cap (_REDUCE_MAX): 32
  after   1 runs reduced: cache entries=  1   retained ~=      431 KiB
  after   9 runs reduced: cache entries=  9   retained ~=     3883 KiB
  after  32 runs reduced: cache entries= 32   retained ~=    13807 KiB
  after  33 runs reduced: cache entries= 32   retained ~=    13807 KiB
  after  40 runs reduced: cache entries= 32   retained ~=    13807 KiB
100 runs in one .capevolve dir -> held: 13.5 MiB  (unbounded would be 42.1 MiB)
```

Growth stops dead at 32. FIFO, 3 lines, exactly as you suggested:

```python
        _REDUCE_CACHE[key] = (stamp, reduced)
        while len(_REDUCE_CACHE) > _REDUCE_MAX:  # dicts are insertion-ordered → FIFO
            _REDUCE_CACHE.pop(next(iter(_REDUCE_CACHE)))
```

FIFO not LRU deliberately: the eviction victims are archived runs whose stamp never changes, so they are pure retention either way, and 32 is far above the working set of any real dashboard session. Pinned by `test_cache_is_capped_and_evicts_oldest`, which asserts both the count *and* that the surviving keys are the three newest.

---

### B3 proof — abs / relative / symlink collapse to ONE entry

```console
###### BEFORE (reviewed commit) ######
cache keys after abs + relative open: ['/tmp/big/run_big', 'run_big']
entries: 2  same object: False -> DUPLICATE entries, cache MISS
after symlinked path       : 3 entries, same object: False

###### AFTER (this fix) ######
cache keys after abs + relative open: ['/private/tmp/big/run_big']
entries: 1  same object: True -> ONE entry, cache HIT
after symlinked path       : 1 entries, same object: True
```

Symlinks collapse for free — `resolve()` follows them. Wrapped in `_resolve()` so an unresolvable path degrades to the raw string rather than raising inside a read-only view.

---

### Re-measured timings — the win survives the extra key work

Synthetic long run: **202 events / 200 candidates / 408 rollouts / 20 real git commits**. BEFORE = `origin/main` (no cache at all). Warm-up call discarded, x20.

**LARGE log:**

| op, ms each | BEFORE (`origin/main`) | AFTER (this fix) | speedup |
|---|---|---|---|
| `reduce_run` | 260.045 | **0.046** | **~5,700x** |
| `runs.load_run` (HTTP + every SSE tick) | 255.184 | **0.070** | **~3,650x** |
| `runs.list_runs` | 254.750 | **0.089** | **~2,860x** |
| `trajectories.diff_candidate` | 260.396 | **2.230** | **~117x** |

**SMALL log** (8 events / 6 candidates / 20 rollouts), x50:

| op, ms each | BEFORE | AFTER | speedup |
|---|---|---|---|
| `reduce_run` | 22.942 | **0.043** | ~530x |
| `runs.load_run` | 23.001 | **0.067** | ~343x |
| `runs.list_runs` | 22.779 | **0.087** | ~262x |
| `trajectories.diff_candidate` | 22.239 | **0.108** | ~206x |

**And the cost of the extra stats, measured rather than assumed** — old 1-stat key vs new 4-stat key, same inputs:

```console
###### key cost: ORIGINAL 1-stat key (PR as submitted) ######
  reduce_run                         x20:  0.019 ms each
  _events_stamp (key only)           x20:  0.005 ms each
###### key cost: NEW 4-stat key ######
  reduce_run                         x20:  0.045 ms each
  _events_stamp (key only)           x20:  0.018 ms each
```

So the extra stats cost **+0.026 ms per cache hit** — a 2.4x slower *hit*, which is **0.01% of the 260 ms fold** it avoids. Raw `os.stat` is ~2 us each on this machine (~1.85 us for a missing path, so a run dir without a git store pays even less):

```console
os.stat existing x200000: 437.3 ms total -> 2.186 us each
os.stat missing  x200000: 369.5 ms total -> 1.847 us each
```

Correctness bought for 26 microseconds. Note my BEFORE absolutes (260 ms) land close to yours (230 ms) and the PR's (229.6 ms) — same machine class, same ratio.

**Redaction still not bypassed** after the key change:

```console
$ /tmp/ce-venv/bin/python /tmp/redactprobe.py
cache hit (same object)  : True
secret leaked in first   : False
secret leaked in cached  : False
```

(The marker check printed `False` because `redact()` emits `«redacted»`, not `REDACTED` — verified directly: `redact({'x': 'tok sk-ant-...'})` → `{'x': 'tok «redacted»'}`. The secret is gone either way, before the value ever reaches `_REDUCE_CACHE`.)

---

### Response to all 13 findings

1. **B1 — `git_log` stale, docstring backwards.** ✅ Fixed. `_git_stamp` folded into the key; docstring rewritten to state the real (inverted) ordering with line refs. Your suggested `.git/HEAD` does not work — it is a constant symbolic ref; `.git` mtime + `.git/logs/HEAD` used instead, measured. Pinned by a test that asserts the event log did not move *and* that `git_log` grew.
2. **B2 — unbounded cache.** ✅ Fixed. `_REDUCE_MAX = 32`, FIFO, exactly your 3 lines. Growth measured flat past the cap; 42.1 MiB → 13.5 MiB at 100 runs. Test asserts count and which keys survive.
3. **B3 — un-normalized key.** ✅ Fixed. `root.resolve()`; abs + relative + symlink → 1 entry, 1 fold. Test pins it.
4. **N1 — same-size rewrite / rename-over.** ✅ Half-hardened, half-documented, matching your own de-escalation. `st_ino` added: rename-over is now **caught** (it was your case 5) for one extra field on a stat we already make. The true in-place same-length hand-edit (case 3) is **not** defended — no in-tree writer does it — but the append-only invariant it depends on is now written into the docstring with the instruction to hash instead if a non-append writer ever lands. That is the "document it so a future writer doesn't silently break the cache" option, and the cheap half of the hardening taken as well.
5. **N2 — no lock, thundering herd.** ✅ Documented, not fixed. Now says so explicitly at the call site: `# ponytail: no lock around the miss path — N cold concurrent readers each fold (duplicate work, never a torn read: reduced is fully built before the single atomic dict store). Add a module lock if thundering herd shows up.` Your own probe showed no correctness bug; a lock would serialize every miss across unrelated runs to save duplicate work in a case (several tabs, cold cache, same run, same instant) that costs one extra fold. Declining the lock, recording the decision.
6. **N3 — shared mutable return unenforced.** ✅ Tripwire added. `test_mutating_a_returned_reduction_is_visible_to_the_next_caller` asserts the leak, so a future `deepcopy`-on-hit (which would eat the whole win) has to break a test that names the tradeoff. Not deep-copying, as you advised.
7. **N4 — no sort tiebreaker.** ✅ Fixed. `key=lambda p: (-p.stat().st_mtime, p.name)`. New test forces three runs to an *identical* mtime with `os.utime` and asserts the order is total, stable across 5 calls, and tiles under paging.
8. **N5 — no max page size.** ✅ Fixed. `le=1000` on both `limit` params. Test asserts `1000`→200, `1001`→422, `99999999999999999999`→422, plus the existing `limit=0`/`offset=-1`→422.
9. **N6 — short page undocumented and untested.** ✅ Fixed. Both docstrings now say **"a SHORT PAGE IS NOT THE END OF THE LIST; page until you get an empty one."** New test puts a corrupt file mid-list, gets a 2-row page at `limit=3`, proves the *next* page still returns 3 (so short ≠ exhausted), proves `offset` past the end returns `[]`, and proves the pages still tile the unpaged list.
10. **N7 — no truncation test.** ✅ Added, `test_truncating_the_event_log_invalidates`.
11. **Nit — "Bounded by run count" misleading.** ✅ Reworded now that a real bound exists: *"Bounded: FIFO-capped at `_REDUCE_MAX` entries (the dashboard is a long-lived process and `list_runs` installs one entry per discovered run)."*
12. **Nit — mirror the `# Unpaginated by default` comment on the rollouts route.** ✅ Done, `app.py:50`.
13. **Nit — `_spy` clears the shared global.** ✅ Fixed: `monkeypatch.setattr(dashboard, "_REDUCE_CACHE", {})`, so the dict is restored rather than left emptied.

**On pagination staying in.** Keeping it. #119 asked for it, the measured bound is real (`limit=5` over 12 long runs: 3088 ms → 1099 ms cold, because the slice happens before any reduction), both endpoints default unpaginated so nothing truncates today, and the tests pin the tiling contract so whoever wires the SPA in **#117/#138** inherits a verified backend instead of writing one. Deliberately **not** wiring the frontend here — that is #117/#138 territory, and doing it in this PR would put UI churn in a perf PR and risk the `dist/` conflicts we are trying to avoid.

---

### Verification

```console
$ cd /tmp/fx-194 && PYTHONPATH=/tmp/fx-194/core /tmp/ce-venv/bin/python -m pytest core/tests -q
192 passed in 61.52s (0:01:01)

$ cd /tmp/fx-194/dashboard/backend && /tmp/ce-venv/bin/python -m pytest tests -q
52 passed, 1 warning in 2.48s

$ cd /tmp/fx-194 && /tmp/ce-venv/bin/python -m compileall -q core skills dashboard/backend && echo CLEAN
CLEAN

$ cd /tmp/fx-194/dashboard/frontend && npx tsc --noEmit ; echo exit=$?
exit=0

$ npm test -- --run
 Test Files  13 passed (13)
      Tests  45 passed (45)
```

Honest numbers: **192 core** (179 baseline + 7 from the original PR + 6 new here), **52 backend** (49 + 3 new), 45 frontend, `tsc` clean, `compileall` clean.

One honesty note on **#200**: `test_dashboard_launch.py::test_maybe_launch_spawns_when_available` **passed** in this run — all 7 launch tests pass in isolation (`7 passed in 0.02s`) and the full 192 includes it with nothing deselected. Port 7878 is currently held by a process on this box, so the contention that made it fail earlier is real but intermittent; it is pre-existing (fails identically on unmodified `origin/main`) and touches no file in this diff. I am reporting 192/192 because that is what actually ran, not masking it with `-k`.

No `dist/` churn — `git status --short` shows only the 6 source files, so the epic's rebuild policy is not engaged for this PR.

---

### Merge order

Confirming your trial-merge finding: **zero conflicted files** across all five siblings. Recommended order stands:

**#194 → #191 → #197 → #117 → #187 → #192**

The flagged #191 overlap is a **false alarm** — zero file overlap, and #191 deliberately re-exports `read_new_events` from `stream.py`, which is the only symbol `app.py:121` uses.

Issue #113, opened 2026-07-25
Guard tiny/empty val splits and add a small-sample correction to the paired gate


OsherElhadad:
## Problem

Two related small-sample hazards in the split + gate machinery:

1. **`make_splits` can produce an empty val set** for tiny task lists (`splits.py:113-119`) with **no warning**. The gate (which only runs on val) then has no data to decide on.
2. The paired significance gate uses a fixed `k·SE` z-style bar with **no t-correction / small-sample penalty and no minimum-val-size guard** (`gate.py:101-135`). On the default 25% val split of a small dataset, this is optimistic — a couple of lucky per-task deltas can clear the bar.

## Why it matters

cap-evolve's whole pitch is honest, variance-aware acceptance. On small benchmarks (common during onboarding and cheap iteration) an empty/tiny val silently degrades the gate into either "no data" or "too easy to pass," which is precisely the failure mode the gate exists to prevent.

## What to change

1. In `make_splits` / `ensure_splits`, **warn (and ideally refuse or auto-adjust)** when val ends up empty or below a minimum (e.g. < 5 tasks). Emit a `gate_warning`/`split_warning` event and surface it in the report.
2. In `gate.py`, add a **minimum-val-size guard**: below the threshold, either widen the bar, require more trials, or clearly flag the decision as low-confidence in the event log.
3. Consider a small-sample correction (t-based critical value instead of a fixed z-style `k`) for the paired test; at minimum document that the current bar is optimistic on tiny val sets.

## How to test

- New test: tiny task set (e.g. 6 tasks) → `make_splits` emits a warning / auto-adjusts rather than yielding empty val.
- New test: gate on a below-minimum val logs a low-confidence warning (or applies the widened bar).
- Existing split/gate tests still pass; default behavior on normal-sized datasets is unchanged.

## Dependencies / parallelization

- **Parallel-safe.** Independent core change. Pairs conceptually with the `pass^k` suppression issue (both harden honest-metrics on small runs) but touches different code.

PR #197, opened 2026-07-29
Protected-paths tamper guard: verify the optimizer never edited scoring/eval/task files

OsherElhadad:
Closes #142

## What this is

cap-evolve enforces honesty on the *evaluation* side — seeded frozen splits, a sealed test split, a val-only significance gate. But when the capability under optimization is **tool code or a skill package**, the optimizer is a coding agent with write tools running next to the project, and nothing structural stopped it from "improving" by editing the **scorer / eval harness / task data** instead of the target. A candidate that rewrites `score()` to return `1.0` is reward hacking, and every number after it is fiction.

This makes *"only edit the target, never the grader"* a structural guarantee, the same posture the test seal already has: a SHA-256 manifest of the protected paths, recorded before any optimizer runs and re-verified on the path every run takes.

## Threat model

**What this DOES protect against.** An optimizer (or anything else in the process) that modifies, deletes, or adds a file under the declared protected paths at any point between `baseline` and `finalize`. Concretely: monkeypatching `score()`, loosening the exact-match check, rewriting `tasks.jsonl` targets to match what the agent already outputs, editing a gold/answer-key file, dropping in a second scorer module, or changing `capevolve.yaml`'s split ratios / gate mode / `k_se` mid-run. Detection is byte-exact and pre-scoring: the run aborts before the tampered grader can produce a score, before the candidate is snapshotted or becomes `best`, and before the test seal is burned — so a tampered run cannot yield a headline number at all.

**What this does NOT protect against.** It is tamper *detection*, not prevention (see "why not read-only" below), and it is not a sandbox. It cannot stop: an adapter that reads its ground truth from somewhere outside the project dir (a network fetch, an env var, an absolute path elsewhere on disk — those bytes are not in the manifest); a grader that is *already* wrong or lenient at baseline (the manifest freezes whatever is there, it does not audit it); an optimizer that also rewrites `protected.json` itself or the `protected_paths` declaration and *starts a new run* (the manifest is only authoritative within one run — cross-run trust needs a signed/off-host manifest, out of scope); reward hacking that stays entirely inside the capability (e.g. a tool that special-cases the eval's inputs — that is a scorer-design problem, not a tamper problem); or a genuinely malicious process with the privileges to rewrite the run dir. It is one more structural rail, not a security boundary.

## Design decisions

[...]

PR #195, opened 2026-07-29
Guard tiny/empty val splits and add a Student-t small-sample correction to the paired gate

OsherElhadad:
Closes #113

## Problem

Two small-sample hazards made the acceptance gate — cap-evolve's entire honesty pitch — meaningless or optimistic on small benchmarks.

1. `make_splits` silently produced an **empty** (n=0) or **single-task** (n=1) val split for tiny task lists. With n=0 the gate has no data; with n=1 the paired delta has **0 degrees of freedom**, so `SE(Δ)` collapses to 0 and the gate falls through to "any Δ>0 wins" — unannounced.
2. The paired gate's bar was `k_se · SE`, a fixed **z** multiplier with no small-sample correction.

## Statistical justification for the correction

The bar `k_se · SE` is a z-style critical value. That is valid only when the standard error is **known** (or n is large enough that the estimate is effectively exact). In the paired gate it is not: `SE(Δ)` is estimated from the *same* n val deltas whose mean is being tested. For a paired mean difference with an estimated variance, the studentized statistic

```
T = Δ̄ / SE(Δ)      where SE(Δ) = s_Δ / √n
```

follows **Student's t with df = n−1**, not the standard normal. t has fatter tails, so `t_{1−α, n−1} > z_{1−α}` for every finite df. Using z therefore places the bar **too low** and accepts noise at exactly the sample sizes where noise dominates — the failure mode the gate exists to prevent. This is the textbook one-sample t-test on the paired differences, which *is* the correct test for a paired mean difference with unknown variance; the previous code was effectively using a z-test in its place.

The fix converts `k_se` to its one-sided normal significance level `α = P(Z > k_se)` and substitutes `t_{1−α, n−1}`. Same nominal α, correct reference distribution.

**Implementation (stdlib only, zero new deps).** `stats.py` gains:
- `betainc(a,b,x)` — regularized incomplete beta, via the continued fraction of *Numerical Recipes in C* (2nd ed.) §6.4 `betacf` with Lentz's modification.
- `t_cdf(t, df)` — Student's t CDF from `betainc`, per Abramowitz & Stegun, *Handbook of Mathematical Functions* (1964) **§26.7.1**.
- `t_critical(alpha, df)` — inverted by bisection on the monotone CDF (exact to 1e-10, works for non-integer df, no lookup table to typo).
- `t_multiplier_for_z(k_se, n)` — `max(k_se, t_critical(normal_sf(k_se), n-1))`.

[...]

PR #198, opened 2026-07-29
docs: document the real gate modes (paired is the default; add simplicity_tiebreak)

### LINKED ISSUE #104: Document the real gate modes (paired is the default; add simplicity_tiebreak; fix gate.py docstring)
## Problem

The acceptance gate's **real default mode is undocumented**, and `gate.py`'s own docstring disagrees with the shipped config default.

- `templates/project/capevolve.yaml:67` — real default: `gate_mode: paired` (with `significant | strict | threshold | simplicity_tiebreak` as the other modes).
- `core/cap_evolve/gate.py:3` — docstring says "The default is the *significance* gate…".
- Docs (`docs/ADAPTER_CONTRACT.md`, `docs/OPTIMIZE_YOUR_OWN.md`, `skills/phases/intake/SKILL.md`) describe modes as `significant | strict | threshold` and **never mention `paired` (the actual default) or `simplicity_tiebreak`.**

So the mode that actually governs every default run — `paired` (per-task paired SE over shared val tasks) — is documented nowhere outside a one-line yaml comment.

## Why it matters

The gate is the core honesty mechanism. Users tuning `gate_mode` or reasoning about why a candidate was accepted/rejected are reading docs that describe a different default than the code applies. `paired` also has a subtle, important property (it banks single-task gains because SE comes from between-task spread even at `num_trials=1`) that users should understand.

## What to change

1. Update `core/cap_evolve/gate.py:3` docstring to state `paired` is the default and briefly explain each mode.
2. Add a "Gate modes" section to `docs/ADAPTER_CONTRACT.md` (near "The gate"), `docs/OPTIMIZE_YOUR_OWN.md`, and `skills/phases/intake/SKILL.md`, listing all five modes with one line each, marking `paired` as default, and explaining the `Δ > k·SE` bar.
3. Cross-link to `docs/HONEST_EVAL.md`.

## How to test

- `grep -rn "paired\|simplicity_tiebreak" docs/ skills/phases/intake/SKILL.md` — both appear and `paired` is marked default.
- The documented default matches `templates/project/capevolve.yaml:67` and `gate.py`'s resolved default.

- **Parallel-safe.** Pure docs + one docstring. Independent of all other issues.

OsherElhadad:
Closes #104

## Ground truth (from the code, not the docs)

Every mode that exists in `core/cap_evolve/gate.py`, its real behavior, and which is really the default:

| `gate_mode` | Behavior | Default? | Proof (`file:line`) |
|---|---|---|---|
| **`paired`** | `accept ⟺ mean(per-task Δ) > k_se · SE(Δ)` where `Δ[t] = cand[t] − curr[t]` over shared val tasks; `SE = sqrt(var(Δ)/n)` (n−1 denominator). Falls back to `significant` when `paired_deltas` is empty. | **YES — the default for every real run** | `gate.py:117-151`; selected by `harness.py:1293` (`gate_kwargs["mode"] = "paired"`) and `gepa.py:703` (`gk["mode"] = "paired"`), both guarded by `if "mode" not in … and paired_deltas is not None`; shipped default `templates/project/capevolve.yaml:80` |
| `significant` | `accept ⟺ Δ(means) > k_se · sqrt(SE_cand² + SE_curr²)` — independent-samples test. | No — but it *is* the default of `decide()`'s own `mode=` parameter | `gate.py:153-176`; parameter default `gate.py:77` (`mode: str = "significant"`) |
| `threshold` | `accept ⟺ Δ > threshold` (flat margin). | No | `gate.py:178-180` |
| `strict` | `accept ⟺ Δ > 0`. | No | `gate.py:182-184` |
| `simplicity_tiebreak` | `Δ > 0` accepts; on a near-tie (`abs(Δ) ≤ 1e-9`) accept if `candidate_size < current_size`. | No | `gate.py:186-198` |

Anything else raises `ValueError: unknown gate mode` (`gate.py:200`), so this list is exhaustive.

**The asymmetry the old docs tripped over.** `decide()`'s `mode=` parameter really does default to `"significant"` — because a bare caller passing two means and two SEs has no per-task data to pair. But no real run goes through that path: the harness always builds `paired_deltas` and sets `mode="paired"`. Both facts are now documented together instead of one contradicting the other.

**Why `paired` matters (verified empirically, not asserted).** Its SE comes from the spread *between* tasks' deltas, not per-trial noise, so it is non-zero even at `num_trials=1`. Fixing one task of `n` and breaking nothing gives `mean(Δ) = SE = 1/n` exactly:

```
$ python -c "from cap_evolve.gate import decide; ..."
2 False paired Δ̄=+0.5000 <= 1.0·SE=0.5000 (SE=0.5000, n=2)
10 False paired Δ̄=+0.1000 <= 1.0·SE=0.1000 (SE=0.1000, n=10)
20 True  paired Δ̄=+0.0500 > 1.0·SE=0.0500 (SE=0.0500, n=20)   # float slack
```

So at the default `k_se=1.0` a 1-task gain just misses (the comparison is strict), and at the `gate_k_se: 0.2` the examples use it accepts. Under `significant` at `num_trials=1` the same 1-of-10 gain faces a `0.233·k` bar instead of `0.100·k` (`combined_stderr` carries the full between-task spread). The docs now state this precisely rather than the vaguer "~2-3x" claim that was in `INPUTS.md`.

## Bug found and fixed during the sweep

`docs/REPRODUCE_tau2.md` and `skills/algorithms/agent-optimize/SKILL.md` instructed agents to run the gate phase with `--mode paired`, but `skills/phases/gate/scripts/run.py` restricted `--mode` to `significant|strict|threshold|simplicity_tiebreak`:

```
$ python skills/phases/gate/scripts/run.py --current 0.5 --candidate 0.6 --mode paired
gate: error: argument --mode: invalid choice: 'paired' (choose from significant, strict, threshold, simplicity_tiebreak)
```

Every documented paired invocation exited 2. Added `paired` to the choices plus a `--paired-deltas` flag, so the default mode is actually reproducible standalone (it needs per-task deltas, which the old signature had no way to accept).

## Every surface fixed

**Code / docstrings**
- `core/cap_evolve/gate.py` — module docstring no longer claims `significant` is the default; states `paired` is the run default, cites the two selection sites, explains the parameter asymmetry, enumerates all five modes. `decide()`'s mode list marks `paired` as DEFAULT.
- `skills/phases/gate/scripts/run.py` — `paired` accepted; new `--paired-deltas`.
- `skills/algorithms/{hill-climb,gepa,skillopt}/scripts/run.py` — `--gate-mode` help said "recommended" and omitted `simplicity_tiebreak`; now says "the default" and lists all five.

**Canonical docs**
- `docs/HONEST_EVAL.md` — **new `## Gate modes` section** (the cross-link target #104 asked for): full table, the paired/single-task-gain explanation, and a small-sample caveat subsection. Guarantee 3 no longer says `mode="significant"`.
- `docs/ADAPTER_CONTRACT.md` — new `## Gate modes` next to "The gate", framed around what it means for your `score()`.
- `docs/OPTIMIZE_YOUR_OWN.md` — the spec block said `gate: <significant (k_se) | strict | threshold>`, which is not even the right key name; now `gate_mode` / `gate_k_se` with the real five, plus a `### Gate modes` table.
- `site/optimize-your-own.html` — the same block mirrored in HTML (would otherwise have drifted).
- `README.md`, `RUN.md`, `ci/benchmarks/README.md` — one-line gate descriptions now name `paired` and link the section.

**Skills**
- `skills/phases/intake/SKILL.md` — new `### Gate mode` intake question with all five modes (#104 item 2).
- `skills/phases/intake/inputs/INPUTS.md` — "**paired** recommended" → "the DEFAULT", exact wording, cross-link.
- `skills/phases/gate/SKILL.md` — frontmatter `description`/`argument-hint` said `--mode significant`; body's "significance rule (default)" documented only the unpaired formula and claimed SE=0 makes it "silently degrade" (it does not — it logs `gate_warning`). Both formulas now shown, `paired` marked default, working paired example in "How to run".
- `skills/phases/gate/references/concepts.md` — the significance-test section derived only the unpaired SE; now derives both and explains why pairing cancels between-task variance. Modes table gained `paired`. SE=0 and small-sample paragraphs corrected.
- `skills/phases/gate/meta.yaml` — manifest summary said "significance by default"; now "paired per-task significance by default" (manifest rebuilt).
- `skills/algorithms/skillopt/{SKILL.md,references/concepts.md}` — "`significant`/`paired`" (ambiguous about which) → `paired`.
- `skills/algorithms/agent-optimize/SKILL.md` — its `--mode paired` recipe passed `--current-stderr/--candidate-stderr`, which paired ignores; now passes `--paired-deltas` with the unpaired fallback noted.

**Config**
- `templates/project/capevolve.yaml`, `examples/tau2_airline/capevolve.yaml` — comments say "the default" rather than "recommended".
- `docs/REPRODUCE_tau2.md` — claimed `gate_mode significant` for a run whose `capevolve.yaml` is `paired`; corrected, and its gate-phase recipe updated.
- `docs/how-to/cap-evolve-with-exgentic-tau2.md` — keeps `significant`, but now documents *why* it is a deliberate override (its val is ~1 task, where the paired SE collapses to 0). Both mentions (line 77 summary and the yaml block) updated.

## Note for #113

#113 will change the small-sample behavior this PR documents. The paragraphs it must update:

- `docs/HONEST_EVAL.md` → `### Small-sample caveat (current behavior)` — explicitly states the bar is a fixed `k · SE` with **no t-correction and no minimum-val-size guard**, and is optimistic on tiny val splits. If #113 adds a t-based critical value or a min-val guard, this is the paragraph to rewrite.
- `skills/phases/gate/references/concepts.md` → the `**Small-sample caveat (current behavior)**` paragraph (same claim, skill-level).
- Both surfaces document the `SE = 0 → gate_warning + strict fallback` path (`gate.py:132-143`, `gate.py:155-165`). If #113 replaces that fallback with a widened bar or a low-confidence flag, the "Two degenerate cases are handled loudly" list and the SKILL.md "When the SE collapses to 0" paragraph both change.

I deliberately described current behavior only, and labelled it "(current behavior)" so #113's diff is obvious.

## Verification

`pytest core/tests -q` — 179 passed, 0 failed (baseline 179):
```
$ cd /tmp/wt-104 && PYTHONPATH=/tmp/wt-104/core python -m pytest core/tests -q
179 passed in 63.65s (0:01:03)
```

`compileall` clean:
```
$ python -m compileall -q core skills && echo "COMPILEALL_CLEAN"
COMPILEALL_CLEAN
```

Manifest rebuilt + every touched skill's `check.py` prints `"ok": true`:
```
  optimizer: run-optimizer

=== phases/gate
  "ok": true,
=== phases/intake
  "ok": true,
=== algorithms/skillopt
  "ok": true,
=== algorithms/hill-climb
  "ok": true,
=== algorithms/gepa
  "ok": true,
=== algorithms/agent-optimize
  "ok": true,
```

The newly documented CLI examples actually run (they did not before):
```
$ python scripts/run.py --current 0.50 --candidate 0.60 --mode paired --k-se 0.2 --paired-deltas 1,0,0,0,0,0,0,0,0,0
{"accept": true,  "reason": "paired Δ̄=+0.1000 > 0.2·SE=0.0200 (SE=0.1000, n=10)", "delta": 0.1, "threshold": 0.02}
$ python scripts/run.py --current 0.50 --candidate 0.60 --mode paired --k-se 1.0 --paired-deltas 1,0,0,0,0,0,0,0,0,0
{"accept": false, "reason": "paired Δ̄=+0.1000 <= 1.0·SE=0.1000 (SE=0.1000, n=10)", "delta": 0.1, "threshold": 0.1}
```

Every relative link + anchor in the changed markdown resolves on disk (67 checked, 0 broken).

Issue #104's own acceptance check:
```
$ grep -rln "paired" docs/ skills/phases/intake/SKILL.md
docs/ADAPTER_CONTRACT.md  docs/HONEST_EVAL.md  docs/OPTIMIZE_YOUR_OWN.md
docs/REPRODUCE_skillsbench.md  docs/REPRODUCE_tau2.md  skills/phases/intake/SKILL.md ...
$ grep -rl "simplicity_tiebreak" docs/ skills/phases/intake/SKILL.md
docs/ADAPTER_CONTRACT.md  docs/HONEST_EVAL.md  docs/OPTIMIZE_YOUR_OWN.md  skills/phases/intake/SKILL.md
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,8 +44,10 @@ significant margin → commit — and reports one honest number. It optimizes wh
 - **Learn from real agent failures.** Every iteration reads full trajectories and per-task
   causal feedback (which task ids a prior edit *broke* and *fixed*), so edits are large and
   don't regress the wins.
-- **Keep evaluation honest.** Acceptance is a val-only significance gate (Δ > k·SE); the
-  test split is sealed and scored exactly once. Both live in the core, not in editable docs.
+- **Keep evaluation honest.** Acceptance is a val-only significance gate (Δ > k·SE) — by
+  default the per-task [**`paired`** test](docs/HONEST_EVAL.md#gate-modes) over the same
+  val tasks; the test split is sealed and scored exactly once. Both live in the core, not
+  in editable docs.
 - **Inspect every change.** Each candidate is a git commit; the dashboard shows costs,
   timing, diffs, lineage, and a tasks × iterations pass/fail heatmap.
 
--- a/RUN.md
+++ b/RUN.md
@@ -75,6 +75,8 @@ artifact.
   default rule is `gate_mode: paired` — mean(per-task Δ) over the same val tasks must
   clear `gate_k_se · SE(Δ)`. Four other modes (`significant`, `threshold`, `strict`,
   `simplicity_tiebreak`): [docs/HONEST_EVAL.md § Gate modes](docs/HONEST_EVAL.md#gate-modes).
+  At the default `gate_k_se: 1.0`, `paired` needs **≥2 improved tasks** — a gain on
+  exactly one of `n` lands `mean(Δ) = SE(Δ)` and is rejected.
 - Only the **primary** metric (the scalar reward) gates; shown-only secondaries (e.g. `cost_usd`, `db_match`) are displayed but never affect accept/reject.
 - Multi-trial scoring reports mean + stderr; pass^k is reported when trials > 1.
 - Rejected approaches are remembered and never re-proposed.
--- a/RUN.md
+++ b/RUN.md
@@ -71,7 +71,10 @@ artifact.
 
 ## The rules (enforced by cap_evolve, restated here for bare hosts)
 - Train/val/test are split once with a seed; **test is scored only at finalize**.
-- Acceptance is gated on **val**, never on the data the optimizer edited against.
+- Acceptance is gated on **val**, never on the data the optimizer edited against. The
+  default rule is `gate_mode: paired` — mean(per-task Δ) over the same val tasks must
+  clear `gate_k_se · SE(Δ)`. Four other modes (`significant`, `threshold`, `strict`,
+  `simplicity_tiebreak`): [docs/HONEST_EVAL.md § Gate modes](docs/HONEST_EVAL.md#gate-modes).
 - Only the **primary** metric (the scalar reward) gates; shown-only secondaries (e.g. `cost_usd`, `db_match`) are displayed but never affect accept/reject.
 - Multi-trial scoring reports mean + stderr; pass^k is reported when trials > 1.
 - Rejected approaches are remembered and never re-proposed.
--- a/ci/benchmarks/README.md
+++ b/ci/benchmarks/README.md
@@ -93,7 +93,7 @@ has a **Type** column + filter.
   | `optimizer_model` | `claude-opus-4-8` | the optimization model (Claude Code's model) — dropdown, same gateway-alias list as `agent_model` |
   | `optimizer_usd_per_iter` | `0` (unlimited) | per-iteration $ cap on the optimizer — `0` disables Claude Code's native `--max-budget-usd` cap entirely; set e.g. `4.0` to bound it |
   | `optimizer_max_turns` | `80` | per-iteration turn cap on the optimizer |
-  | `gate_k_se` | `1.0` | acceptance-gate strictness (accept iff Δ > k_se·SE) |
+  | `gate_k_se` | `1.0` | acceptance-gate strictness — accept iff Δ > k_se·SE, on the default `gate_mode: paired` (per-task paired SE over the same val tasks; see [docs/HONEST_EVAL.md](../../docs/HONEST_EVAL.md#gate-modes)) |
   | `algorithm_focus` | `all` | hill-climb schedule: `all` \| `cyclic` \| `hardest-first` |
 
   Overriding `agent_model` takes precedence over any per-task `agent` a curated `tasks.json`
--- a/core/cap_evolve/gate.py
+++ b/core/cap_evolve/gate.py
@@ -1,9 +1,23 @@
 """Acceptance gate — the rule that decides whether a candidate edit is kept.
 
-The default is the *significance* gate (prior agent-optimization work's ``val_improvement_significant``):
-accept only when the val improvement exceeds k standard errors, so noise does
-not get mistaken for progress. All gates compare on VAL and never on TRAIN —
-``decide`` takes an explicit ``split`` and refuses anything but ``val``.
+The default mode for every real run is ``paired``. The harness scores candidate and
+current on the SAME val tasks, so it passes ``paired_deltas`` and sets
+``mode="paired"`` whenever the caller has not pinned a mode and per-task data
+aligns (``harness.py`` ``_gate_and_record``, ``gepa.py`` ``_full_val_gate``); the
+shipped ``gate_mode`` in ``templates/project/capevolve.yaml`` is ``paired`` too.
+Paired tests mean(per-task Δ) against the SE of those deltas, which cancels the
+cross-task variance and is far more powerful than the unpaired test.
+
+``decide``'s own ``mode`` parameter still defaults to ``significant`` (prior
+agent-optimization work's ``val_improvement_significant``) because that is the only
+mode a bare caller with no per-task data can honestly apply; ``paired`` also falls
+back to it when ``paired_deltas`` is empty.
+
+Modes: ``paired`` (default in runs) | ``significant`` | ``threshold`` | ``strict``
+| ``simplicity_tiebreak`` — see ``decide``'s docstring and the "Gate modes" table
+in ``docs/HONEST_EVAL.md``. Every mode's bar is a form of Δ > bar, every mode
+compares on VAL and never on TRAIN — ``decide`` takes an explicit ``split`` and
+refuses anything but ``val``.
 """
 
 from __future__ import annotations
@@ -72,8 +86,10 @@ def decide(
 ) -> GateDecision:
     """Decide whether to accept the candidate.
 
-    Modes:
-      - ``paired``: accept iff mean(per-task Δ) > k * SE(Δ), where Δ[t] =
+    Modes (``paired`` is the default for real runs — the harness selects it; the
+    ``mode`` parameter below defaults to ``significant`` only for bare callers that
+    have no per-task data to pair):
+      - ``paired`` (DEFAULT in runs): accept iff mean(per-task Δ) > k * SE(Δ), where Δ[t] =
         cand_reward[t] - curr_reward[t] over the SAME val tasks. This is the
         correct, far more powerful test: candidate and current are scored on the
         same tasks, so the cross-task variance cancels and only the *paired*
--- a/core/cap_evolve/gate.py
+++ b/core/cap_evolve/gate.py
@@ -94,17 +94,28 @@ def decide(
         correct, far more powerful test: candidate and current are scored on the
         same tasks, so the cross-task variance cancels and only the *paired*
         variance counts. Requires ``paired_deltas``; the loop uses it by default
-        when per-task data is available, else falls back to ``significant``.
+        when per-task data is available, else falls back to ``significant`` (which is
+        announced: the reason is prefixed ``paired→significant`` and a
+        ``gate_warning`` is logged). NOTE at ``k_se=1.0``: a gain on exactly ONE of
+        n tasks gives mean(Δ) = SE(Δ) *exactly*, so the strict ``>`` rejects it —
+        regardless of n and regardless of how large that one gain is. Banking needs
+        ≥2 improved tasks, or ``k_se < 1.0``.
       - ``significant``: accept iff delta > k * combined_SE (treats cand & current
         as INDEPENDENT samples — correct only when they were not scored on the
         same tasks; less powerful than ``paired``).
-      - ``threshold``:   accept iff delta > ``threshold`` (a flat margin).
+      - ``threshold``:   accept iff delta > ``threshold`` (a flat margin). Note
+        ``threshold`` defaults to ``0.0``, so leaving it unset makes this mode
+        identical to ``strict``.
       - ``strict``:      accept iff delta > 0 (any improvement).
       - ``simplicity_tiebreak``: like strict, but on a (near-)tie prefer the
-        smaller candidate (``candidate_size`` < ``current_size``).
+        smaller candidate (``candidate_size`` < ``current_size``). PRECONDITION:
+        needs both sizes; nothing in the harness or the algorithms populates them
+        today, so in a real run this mode behaves exactly like ``strict`` (issue
+        #206 tracks plumbing the sizes).
 
     ``run_dir`` (optional) is used only to log a ``gate_warning`` event when an SE
-    collapses to 0 (so the silent degeneration to strict is auditable).
+    collapses to 0 or ``paired`` falls back to ``significant`` (so neither
+    degradation is silent).
     """
     if split.lower() != "val":
         raise TrainGateError(
@@ -113,14 +124,24 @@ def decide(
         )
 
     delta = candidate_val - current_val
+    fell_back = ""
 
     if mode == "paired":
         deltas = list(paired_deltas or [])
         if not deltas:
             # No paired data — fall back to the independent significance test rather
             # than silently passing. (The loop should pass paired_deltas; this guards
-            # a direct caller.)
+            # a direct caller.) Announce it: asking for the strongest test and getting
+            # the weaker one must not be invisible, same posture as _warn_se_zero.
             mode = "significant"
+            fell_back = "paired→significant (no per-task deltas): "
+            log = getattr(run_dir, "log_event", None) if run_dir is not None else None
+            if callable(log):
+                log("gate_warning", mode="paired",
+                    reason=("mode='paired' was requested but no paired_deltas were "
+                            "supplied, so the weaker unpaired 'significant' test was "
+                            "applied instead. Pass per-task cand-curr deltas over the "
+                            "same val tasks to get the paired test."))
         else:
             n = len(deltas)
             mean_d = sum(deltas) / n
@@ -159,7 +180,7 @@ def decide(
             ok = delta > 0
             return GateDecision(
                 accept=ok,
-                reason=(f"Δ={delta:+.4f} {'>' if ok else '<='} 0 "
+                reason=(f"{fell_back}Δ={delta:+.4f} {'>' if ok else '<='} 0 "
                         f"(SE=0 → STRICT fallback, warned)"),
                 delta=delta, threshold=0.0,
             )
@@ -168,7 +189,7 @@ def decide(
         return GateDecision(
             accept=ok,
             reason=(
-                f"Δ={delta:+.4f} {'>' if ok else '<='} {k_se}·SE={bar:.4f} "
+                f"{fell_back}Δ={delta:+.4f} {'>' if ok else '<='} {k_se}·SE={bar:.4f} "
                 f"(SE={se:.4f})"
             ),
             delta=delta,
--- a/core/tests/test_documented_cli.py
+++ b/core/tests/test_documented_cli.py
@@ -1,0 +1,92 @@
+"""Every CLI invocation shown in the docs actually parses.
+
+Docs are the agent-facing API: `orchestrate` and the algorithm skills tell an agent
+to run these exact command lines. A flag that was renamed, or a `--mode` value the
+script's `choices` never had, turns a documented instruction into an exit-2 crash —
+or worse, silently does something else. Nothing else in CI covers the phase
+`scripts/run.py` surface, so this scans the docs and checks each flag (and each
+enumerated value) against the real argparse.
+
+Catches the class in #203 (docs instructing `cap-evolve finalize`, which does not
+exist) and the `--mode paired` breakage fixed in #198.
+"""
+
+import re
+import subprocess
+import sys
+from pathlib import Path
+
+REPO = Path(__file__).resolve().parents[2]
+# `$S` / `$SKILLS_DIR` etc. all stand in for the skills dir; a bare `scripts/run.py`
+# in a SKILL.md means that skill's own script.
+_SCRIPT = re.compile(r'(?:"?\$\{?\w+\}?"?/|\.\.\./)?([\w./-]*scripts/run\.py)')
+_FLAG = re.compile(r"(--[a-z][a-z0-9-]*)")
+# Placeholders (`<best_mean>`, `{workdir}`) and shell vars are not values to validate.
+_PLACEHOLDER = re.compile(r"^[<{$]|^'")
+
+
+def _spec(script: Path) -> tuple[set[str], dict[str, set[str]]]:
+    """(known flags, {flag: allowed choices}) from the script's own --help."""
+    out = subprocess.run([sys.executable, str(script), "--help"],
+                         capture_output=True, text=True, cwd=script.parent)
+    help_text = out.stdout + out.stderr
+    choices = {f"--{f}": set(v.split(","))
+               for f, v in re.findall(r"--([a-z][a-z0-9-]*) \{([^}]+)\}", help_text)}
+    return set(_FLAG.findall(help_text)), choices
+
+
+def _resolve(raw: str, doc: Path) -> Path | None:
+    for cand in (REPO / "skills" / raw, REPO / raw,
+                 doc.parent / raw):  # bare `scripts/run.py` inside a SKILL.md
+        if cand.is_file():
+            return cand
+    return None
+
+
+def test_documented_run_py_invocations_parse():
+    docs = [p for p in REPO.glob("**/*.md")
+            if "node_modules" not in p.parts and ".git" not in p.parts
+            # design specs/plans describe past or proposed states, not runnable commands
+            and "specs" not in p.parts and "plans" not in p.parts]
+    scanned, findings = 0, []
+    for doc in docs:
+        for line in doc.read_text(errors="replace").splitlines():
+            if "scripts/run.py" not in line or not line.lstrip().startswith(("python", "$")):
+                continue
+            m = _SCRIPT.search(line)
+            script = _resolve(m.group(1), doc) if m else None
+            if script is None:
+                continue
+            scanned += 1
+            known, choices = _spec(script)
+            # only the command itself runs; a trailing `# ...` comment is prose
+            tokens = line.split(" #")[0].split()
+            for i, tok in enumerate(tokens):
+                if not tok.startswith("--"):
+                    continue
+                flag, _, inline = tok.partition("=")
+                if flag not in known:
+                    findings.append(f"{doc.relative_to(REPO)}: {script.name} has no {flag}")
+                    continue
+                if flag in choices:
+                    val = inline or (tokens[i + 1] if i + 1 < len(tokens) else "")
+                    if val and not _PLACEHOLDER.match(val) and val not in choices[flag]:
+                        findings.append(
+                            f"{doc.relative_to(REPO)}: {flag} {val!r} not in "
+                            f"{sorted(choices[flag])}")
+    assert scanned > 20, f"scanner found only {scanned} invocations — regex likely broke"
+    assert not findings, "documented CLI invocations that do not parse:\n" + "\n".join(findings)
+
+
+def test_gate_check_py_passes():
+    """CI runs only `pytest core/tests`, so no skill `check.py` was ever exercised.
+
+    The gate's check.py is what covers the `--paired-deltas` parsing and the paired
+    k_se=1.0 rule, so run it here rather than leaving that coverage CI-invisible.
+    (A general runner for all 20 skills' check.py belongs in its own change; several
+    need network/optimizer setup. ponytail: one skill now, widen when CI can afford it.)
+    """
+    check = REPO / "skills" / "phases" / "gate" / "scripts" / "check.py"
+    out = subprocess.run([sys.executable, str(check)], capture_output=True, text=True,
+                         cwd=check.parent)
+    assert out.returncode == 0, f"gate check.py failed:\n{out.stdout}\n{out.stderr}"
--- a/docs/ADAPTER_CONTRACT.md
+++ b/docs/ADAPTER_CONTRACT.md
@@ -101,15 +101,22 @@ Which rule turns your `reward` into an accept/reject — `gate_mode` in
   variance cancels; this is the most powerful test and what the harness picks itself.
 - `significant` — accept iff `Δ(means) > k_se · sqrt(SE_cand² + SE_curr²)`. For
   unpaired comparisons; weaker, since it pays for cross-task variance.
-- `threshold` — accept iff `Δ > threshold` (flat domain margin).
+- `threshold` — accept iff `Δ > threshold` (flat domain margin). `threshold` defaults
+  to `0.0`, so leaving it unset makes this identical to `strict`.
 - `strict` — accept iff `Δ > 0`. Only for a near-zero-variance scorer.
 - `simplicity_tiebreak` — `strict`, plus on a near-tie prefer the smaller candidate.
+  ⚠️ Requires `candidate_size`/`current_size`, which nothing in the harness populates,
+  so today it behaves exactly like `strict`
+  ([#206](https://github.com/skillberry-ai/cap-evolve/issues/206)).
+
+At the default `gate_k_se: 1.0`, `paired` rejects a gain on exactly one of `n` tasks
+(`mean(Δ) = SE(Δ)` exactly, at any `n` and any magnitude) — ≥2 improved tasks, or
+`gate_k_se < 1.0`.
 
 The bar matters for your `score()`: a deterministic scorer with graded rewards gives
 the paired test real between-task spread, whereas an all-or-nothing scorer at
 `num_trials=1` can collapse the SE to 0, which triggers a logged `gate_warning` and a
-strict fallback. Full table, the single-task-gain property, and the small-sample
-caveat: [`HONEST_EVAL.md` § Gate modes](HONEST_EVAL.md#gate-modes).
+strict fallback. Full table, the `k_se = 1.0` rule, and the small-sample caveat: [`HONEST_EVAL.md` § Gate modes](HONEST_EVAL.md#gate-modes).
 
 ## Everything else is provided
 
--- a/docs/ADAPTER_CONTRACT.md
+++ b/docs/ADAPTER_CONTRACT.md
@@ -86,9 +86,30 @@ implemented, `tasks` is non-empty and stable, and `score` is deterministic. This
 mandatory before any budget is spent — a half-wired adapter can only produce a
 dishonest number.
 
-The gate reads only the **primary** metric (the scalar `reward`). Any shown-only
-secondaries a `score()` returns are carried through for display but can never move an
-accept/reject decision or the sealed number.
+The acceptance gate reads only the **primary** metric (the scalar `reward`). Any
+shown-only secondaries a `score()` returns are carried through for display but can
+never move an accept/reject decision or the sealed number.
+
+## Gate modes
+
+Which rule turns your `reward` into an accept/reject — `gate_mode` in
+`capevolve.yaml`, strictness via `gate_k_se` (default `1.0`). All val-only
+(`core/cap_evolve/gate.py`):
+
+- **`paired` — the default.** Accept iff `mean(per-task Δ) > k_se · SE(Δ)` over the
+  same val tasks. Candidate and current are scored on identical tasks, so cross-task
+  variance cancels; this is the most powerful test and what the harness picks itself.
+- `significant` — accept iff `Δ(means) > k_se · sqrt(SE_cand² + SE_curr²)`. For
+  unpaired comparisons; weaker, since it pays for cross-task variance.
+- `threshold` — accept iff `Δ > threshold` (flat domain margin).
+- `strict` — accept iff `Δ > 0`. Only for a near-zero-variance scorer.
+- `simplicity_tiebreak` — `strict`, plus on a near-tie prefer the smaller candidate.
+
+The bar matters for your `score()`: a deterministic scorer with graded rewards gives
+the paired test real between-task spread, whereas an all-or-nothing scorer at
+`num_trials=1` can collapse the SE to 0, which triggers a logged `gate_warning` and a
+strict fallback. Full table, the single-task-gain property, and the small-sample
+caveat: [`HONEST_EVAL.md` § Gate modes](HONEST_EVAL.md#gate-modes).
 
 ## Everything else is provided
 
--- a/docs/AGENT_ORCHESTRATION.md
+++ b/docs/AGENT_ORCHESTRATION.md
@@ -86,8 +86,10 @@ The agent then drives the loop against `run_dir`, using the phase scripts direct
 S="$CAPEVOLVE_SKILLS_DIR"; R=.capevolve/run_myrun; P=.capevolve/project
 # honest full-val eval of a candidate dir/id (writes rollouts+results):
 python "$S/phases/evaluate/scripts/run.py" --run-dir "$R" --project "$P" --candidate "$R/work/cand_1" --split val --n-trials 1
-# the significance gate:
-python "$S/phases/gate/scripts/run.py" --mode paired --k-se 1.0 --current <best_mean> --candidate <cand_mean> --current-stderr <s> --candidate-stderr <s>
+# the significance gate (paired is the default mode; it NEEDS the per-task deltas —
+# without them it falls back to the weaker unpaired `significant`, which the reason says):
+python "$S/phases/gate/scripts/run.py" --mode paired --k-se 1.0 --current <best_mean> --candidate <cand_mean> --paired-deltas <d1,d2,... per-task cand-curr over the shared val tasks>
+# no per-task data? then: --mode significant --current-stderr <s> --candidate-stderr <s>
 # spend snapshot (the "see your constraints" step — no new command needed):
 python -c "from cap_evolve import RunDir; import json; print(json.dumps(RunDir.open('$R').spent.to_dict(), indent=2))"
 # seal test ONCE + report:
--- a/docs/HONEST_EVAL.md
+++ b/docs/HONEST_EVAL.md
@@ -18,8 +18,8 @@ construction*, and the rules below are enforced in code, not just documented.
 3. **Acceptance is gated on val, with significance.** `gate.decide(...)` refuses
    any split but `val` (`TrainGateError`) and, by default, accepts a candidate
    only when the improvement exceeds `k · SE` — so noise is not mistaken for
-   progress (`mode="significant"`). Other modes (`strict`, `threshold`,
-   `simplicity_tiebreak`) exist but never relax the val-only rule. The gate reads
+   progress. The default mode in a real run is **`paired`** (see [Gate
+   modes](#gate-modes)); no mode relaxes the val-only rule. The gate reads
    only the **primary** metric (the scalar `reward`); any shown-only secondary
    metrics a scorer emits (`Score.metrics`) are for display and cannot move the
    decision.
@@ -27,6 +27,56 @@ construction*, and the rules below are enforced in code, not just documented.
 4. **Variance is measured, not assumed.** With `num_trials > 1`, each task gets a
    mean and stderr; `combined_stderr` mixes between-task and within-task error;
    `pass_k` reports the probability all k i.i.d. trials succeed (tau-bench style).
+
+## Gate modes
+
+Set with `gate_mode` in `capevolve.yaml` (strictness with `gate_k_se`, default
+`1.0`). All five live in `core/cap_evolve/gate.py` `decide()`; all are val-only.
+
+| `gate_mode` | rule | when to use |
+|---|---|---|
+| **`paired`** — **the default** | accept iff `mean(per-task Δ) > k_se · SE(Δ)`, where `Δ[t] = cand[t] − curr[t]` over the same val tasks | default and recommended: candidate and current are scored on the *same* tasks, so cross-task variance cancels and the test is far more powerful than `significant` |
+| `significant` | accept iff `Δ(means) > k_se · sqrt(SE_cand² + SE_curr²)` | when the two sides were *not* scored on the same tasks. Less powerful — it pays for cross-task variance that the paired test cancels |
+| `threshold` | accept iff `Δ > threshold` (flat margin) | you have a domain minimum worthwhile gain ("don't bother unless +2pp") |
+| `strict` | accept iff `Δ > 0` | only with a near-zero-variance scorer (deterministic, single correct answer) |
+| `simplicity_tiebreak` | `Δ > 0` accepts; on a near-tie (`abs(Δ) ≤ 1e-9`) accept the *smaller* candidate (`candidate_size < current_size`) | Occam bias against edits that bloat without earning it |
+
+**`paired` is what actually runs by default.** `templates/project/capevolve.yaml`
+ships `gate_mode: paired`, and the harness sets `mode="paired"` itself whenever the
+caller pinned no mode and per-task val data aligns (`harness.py`
+`_gate_and_record`, `gepa.py` `_full_val_gate`). The algorithm skills'
+`--gate-mode auto` means exactly "let the engine pick paired".
+
+Note the one asymmetry: `gate.decide()`'s own `mode=` parameter still defaults to
+`"significant"`, because a bare caller passing only two means and two SEs has no
+per-task data to pair. `paired` also falls back to `significant` when
+`paired_deltas` is empty.
+
+### Why `paired` banks single-task gains
+
+Under `paired`, the SE comes from the *spread between tasks' deltas*, not from
+per-trial noise — so it is non-zero even at `num_trials=1`. A candidate that fixes
+one task out of `n` and breaks nothing gets `Δ̄ = 1/n` and `SE = 1/n` exactly, so at
+the default `k_se=1.0` it just misses the strict `Δ̄ > k·SE` bar, while at the
+`gate_k_se: 0.2` the examples use it accepts. Under `significant` at
+`num_trials=1`, both SEs come from the between-task spread of the *absolute* scores
+(`combined_stderr`), which is much larger — a one-task gain on 10 tasks faces a
+`0.233·k` bar instead of `0.100·k`.
+
+### Small-sample caveat (current behavior)
+
+The bar is a fixed `k · SE`, with no t-correction and no minimum-val-size guard, so
+it is optimistic on tiny val splits. Two degenerate cases are handled loudly, never
+silently:
+
+- **`SE = 0`** (`paired` with `n=1`, or every task moved identically; `significant`
+  with a zero combined SE, typically `num_trials=1`) — the gate logs a
+  `gate_warning` event and applies a documented **strict fallback** (accept any
+  `Δ > 0`). The decision `reason` says `SE=0 → STRICT fallback, warned`.
+- **no paired data** — `paired` falls back to `significant` rather than passing.
+
+Fix: raise `num_trials` and use a val split big enough that between-task spread is
+meaningful.
 
 ## Why no central engine?
 
--- a/docs/HONEST_EVAL.md
+++ b/docs/HONEST_EVAL.md
@@ -37,9 +37,9 @@ Set with `gate_mode` in `capevolve.yaml` (strictness with `gate_k_se`, default
 |---|---|---|
 | **`paired`** — **the default** | accept iff `mean(per-task Δ) > k_se · SE(Δ)`, where `Δ[t] = cand[t] − curr[t]` over the same val tasks | default and recommended: candidate and current are scored on the *same* tasks, so cross-task variance cancels and the test is far more powerful than `significant` |
 | `significant` | accept iff `Δ(means) > k_se · sqrt(SE_cand² + SE_curr²)` | when the two sides were *not* scored on the same tasks. Less powerful — it pays for cross-task variance that the paired test cancels |
-| `threshold` | accept iff `Δ > threshold` (flat margin) | you have a domain minimum worthwhile gain ("don't bother unless +2pp") |
+| `threshold` | accept iff `Δ > threshold` (flat margin) | you have a domain minimum worthwhile gain ("don't bother unless +2pp"). Note `threshold` defaults to `0.0`, so leaving it unset makes this identical to `strict` |
 | `strict` | accept iff `Δ > 0` | only with a near-zero-variance scorer (deterministic, single correct answer) |
-| `simplicity_tiebreak` | `Δ > 0` accepts; on a near-tie (`abs(Δ) ≤ 1e-9`) accept the *smaller* candidate (`candidate_size < current_size`) | Occam bias against edits that bloat without earning it |
+| `simplicity_tiebreak` | `Δ > 0` accepts; on a near-tie (`abs(Δ) ≤ 1e-9`) accept the *smaller* candidate (`candidate_size < current_size`) | Occam bias against edits that bloat without earning it. ⚠️ **Requires `candidate_size`/`current_size`, which nothing in the harness currently populates — so today this mode behaves exactly like `strict`** ([#206](https://github.com/skillberry-ai/cap-evolve/issues/206)) |
 
 **`paired` is what actually runs by default.** `templates/project/capevolve.yaml`
 ships `gate_mode: paired`, and the harness sets `mode="paired"` itself whenever the
@@ -52,16 +52,34 @@ Note the one asymmetry: `gate.decide()`'s own `mode=` parameter still defaults t
 per-task data to pair. `paired` also falls back to `significant` when
 `paired_deltas` is empty.
 
-### Why `paired` banks single-task gains
+### The `k_se = 1.0` rule: improve ≥2 tasks, or bank nothing
 
 Under `paired`, the SE comes from the *spread between tasks' deltas*, not from
-per-trial noise — so it is non-zero even at `num_trials=1`. A candidate that fixes
-one task out of `n` and breaks nothing gets `Δ̄ = 1/n` and `SE = 1/n` exactly, so at
-the default `k_se=1.0` it just misses the strict `Δ̄ > k·SE` bar, while at the
-`gate_k_se: 0.2` the examples use it accepts. Under `significant` at
-`num_trials=1`, both SEs come from the between-task spread of the *absolute* scores
-(`combined_stderr`), which is much larger — a one-task gain on 10 tasks faces a
-`0.233·k` bar instead of `0.100·k`.
+per-trial noise — so it is non-zero even at `num_trials=1`, where `significant`
+cannot form a bar at all. But that same spread has one sharp consequence at the
+shipped default `gate_k_se: 1.0`:
+
+> **A candidate that improves exactly ONE of `n` val tasks can never be accepted at
+> `k_se = 1.0`.** Fix one task by `m`, break nothing: `Δ̄ = m/n` and `SE(Δ) = m/n` —
+> *exactly equal* — so the strict `Δ̄ > k·SE` comparison rejects it.
+
+Both parameters cancel out of that identity, which is what makes it a rule rather
+than an edge case:
+
+- **Magnitude-independent.** The `m` cancels, so fixing one task **perfectly**
+  (`m = 1.0`) is rejected identically to nudging it by `0.01`.
+- **`n`-independent.** The `n` cancels too, so a larger val split does *not* help.
+  (`n = 20` happens to accept only because `0.05` is not exactly representable in
+  IEEE-754 binary — floating-point slack, not a rule. `n = 50` rejects again.)
+
+**So: improve ≥2 tasks at `k_se = 1.0`, or lower `gate_k_se`** — the examples use
+`0.2`, where a 1-of-n gain banks. Two improved tasks of `n` clear the bar at every
+`n` (`n=10`: `Δ̄ = 0.2000 > SE = 0.1333`).
+
+This is not an argument against `paired`. For every other shape of gain it is
+strictly more powerful than `significant`, whose SEs come from the between-task
+spread of the *absolute* scores (`combined_stderr`) and are much larger — a
+one-task gain on 10 tasks faces a `0.233·k` bar there instead of `0.100·k`.
 
 ### Small-sample caveat (current behavior)
 
--- a/docs/OPTIMIZE_YOUR_OWN.md
+++ b/docs/OPTIMIZE_YOUR_OWN.md
@@ -118,11 +118,15 @@ you want intake to ask):
 |---|---|
 | **`paired`** (**default**) | `mean(per-task Δ) > k_se · SE(Δ)` over the same val tasks — cross-task variance cancels, so this is the most powerful test and the one the harness picks itself (`--gate-mode auto`) |
 | `significant` | `Δ(means) > k_se · sqrt(SE_cand² + SE_curr²)` — unpaired, weaker |
-| `threshold` | `Δ > threshold` — flat domain margin |
+| `threshold` | `Δ > threshold` — flat domain margin. `threshold` defaults to `0.0`, so unset = `strict` |
 | `strict` | `Δ > 0` — only for a near-zero-variance scorer |
-| `simplicity_tiebreak` | `strict`, plus on a near-tie prefer the smaller candidate |
+| `simplicity_tiebreak` | `strict`, plus on a near-tie prefer the smaller candidate. ⚠️ needs `candidate_size`/`current_size`, which the harness does not populate — **behaves as `strict` today** ([#206](https://github.com/skillberry-ai/cap-evolve/issues/206)) |
 
-Details, why `paired` banks a single-task gain, and the small-sample caveat:
+**At the default `gate_k_se: 1.0`, a gain on exactly one of `n` tasks gives
+`mean(Δ) = SE(Δ)` exactly and is rejected** (at any `n`, at any magnitude) — improve
+≥2 tasks, or set `gate_k_se` below 1.0 as the examples do.
+
+That rule in full, and the small-sample caveat:
 [`HONEST_EVAL.md` § Gate modes](HONEST_EVAL.md#gate-modes).
 
 ## B — drive the `cap-evolve` CLI yourself
--- a/docs/OPTIMIZE_YOUR_OWN.md
+++ b/docs/OPTIMIZE_YOUR_OWN.md
@@ -100,12 +100,30 @@ you want intake to ask):
 - max_optimizer_usd:    <cumulative optimizer-only $ cap; 0 = unlimited>
 - optimizer_usd_per_iter: <PER-ITERATION $ cap enforced by the optimizer CLI itself, e.g. claude `--max-budget-usd N`>
 - optimizer_max_turns:  <per-iteration WORK cap passed to the agent CLI, e.g. claude `--max-turns N`>
-- gate:                 <significant (k_se) | strict | threshold>
-                        # significant: accept only if Δ > k_se · SE — k_se is how many standard errors
-                        # the val gain must clear (e.g. 0.2 = lenient, 1.0 = strict) so noise isn't mistaken for progress
+- gate_mode:            <paired (DEFAULT) | significant | threshold | strict | simplicity_tiebreak>
+- gate_k_se:            <how many standard errors the val gain must clear; default 1.0
+                        # (e.g. 0.2 = lenient, 1.0 = strict) so noise isn't mistaken for progress>
+                        # paired (default): Δ = per-task cand−curr on the SAME val tasks; accept if mean(Δ) > k_se·SE(Δ)
+                        # see docs/HONEST_EVAL.md#gate-modes for all five modes
 - stall:                <stop after N consecutive rejects; 0 = run all max_iterations>
 - store:                git          # versions every iteration as a commit for an inspectable process
 ```
+
+### Gate modes
+
+`gate_mode` picks the acceptance rule; all five are val-only
+(`core/cap_evolve/gate.py`), and `gate_k_se` sets strictness (default `1.0`):
+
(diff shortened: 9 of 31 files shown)


OsherElhadad:
## 🔬 Evidence

Every command below was run in the worktree at `/tmp/wt-104` (branch `docs/issue-104-gate-modes`), Python `/tmp/ce-venv/bin/python`. All output is literal.

---

### 1. Reasoning: establish ground truth from the code before touching a doc

The issue asserts the default is `paired` and points at `templates/project/capevolve.yaml:67`. I did not take that on faith — a yaml default can be overridden by the caller, so "what the yaml says" and "what actually runs" are different questions. I read `gate.py` and `stats.py` end to end, then traced every caller of `gate.decide` to find where `mode` is really resolved.

```
$ grep -rn "gate_kwargs\|from .gate\|import gate\|decide(" core/cap_evolve/*.py | sort
core/cap_evolve/__init__.py:17:from .gate import GateDecision, TrainGateError, decide
core/cap_evolve/gepa.py:53:from . import gate as gate_mod
core/cap_evolve/gepa.py:704:    decision = gate_mod.decide(
core/cap_evolve/harness.py:1292:    if "mode" not in gate_kwargs and paired_deltas is not None:
core/cap_evolve/harness.py:1293:        gate_kwargs["mode"] = "paired"
core/cap_evolve/harness.py:1294:    decision = gate_mod.decide(
core/cap_evolve/skillopt.py:210:    gate_kwargs: dict | None = None,
...
```

Two resolution sites, both the same shape:

```
$ sed -n '1291,1299p' core/cap_evolve/harness.py
    paired_deltas = _paired_deltas(current_val, cand_val)
    if "mode" not in gate_kwargs and paired_deltas is not None:
        gate_kwargs["mode"] = "paired"
    decision = gate_mod.decide(
        current_val.reward, cand_val.reward, split="val",
        candidate_stderr=cand_val.stderr, current_stderr=current_val.stderr,
        paired_deltas=paired_deltas, run_dir=run_dir,
        **gate_kwargs,
    )

$ sed -n '700,708p' core/cap_evolve/gepa.py
    gk = dict(gate_kwargs or {})
    paired = _paired_deltas(parent_result, cand_val)
    if "mode" not in gk and paired is not None:
        gk["mode"] = "paired"
    decision = gate_mod.decide(
        parent_result.reward, cand_val.reward, split="val",
        candidate_stderr=cand_val.stderr, current_stderr=parent_result.stderr,
        paired_deltas=paired, run_dir=run_dir, **gk,
    )
```

`skillopt.py` and `hill-climb` route through `harness.run_step`, so they inherit the same resolution. And the algorithm skills pass `--gate-mode auto` → **no** `mode` key at all, which is precisely the branch that selects `paired`:

```
$ sed -n '105,106p' skills/algorithms/hill-climb/scripts/run.py
        gate_kwargs=({"k_se": args.k_se} if args.gate_mode == "auto"
                     else {"mode": args.gate_mode, "k_se": args.k_se}),
```

**Conclusion:** `paired` is the default in every real run, and `decide()`'s `mode="significant"` parameter default only governs bare direct callers. This is the asymmetry the old docs collapsed into a flat (and wrong) "the default is significance".

### 2. Enumerate every mode that ACTUALLY exists (exhaustive, not from the issue text)

```
$ grep -n "mode == \"paired\"\|mode == \"significant\"\|mode == \"threshold\"\|mode == \"strict\"\|mode == \"simplicity_tiebreak\"\|mode: str = \|raise ValueError(f\"unknown gate mode" core/cap_evolve/gate.py
77:    mode: str = "significant",
117:    if mode == "paired":
153:    if mode == "significant":
178:    if mode == "threshold":
182:    if mode == "strict":
186:    if mode == "simplicity_tiebreak":
200:    raise ValueError(f"unknown gate mode: {mode!r}")
```

Five modes, and the terminal `raise` proves the list is exhaustive (any sixth name is an error, not a silent default). Confirmed at runtime in §4.

```
$ grep -n "gate_mode: paired" templates/project/capevolve.yaml
80:gate_mode: paired                        # DEFAULT: per-task paired SE, same tasks both sides — banks real 1-task gains.
```

(Line 80 post-edit; it was line 67 when #104 was filed — the file has grown since.)

### 3. Runtime proof of every behavioral claim I wrote into the docs

```
$ python - <<'PY'   # (paired SE at num_trials=1, one task fixed of n)
2 False paired Δ̄=+0.5000 <= 1.0·SE=0.5000 (SE=0.5000, n=2)
3 False paired Δ̄=+0.3333 <= 1.0·SE=0.3333 (SE=0.3333, n=3)
5 False paired Δ̄=+0.2000 <= 1.0·SE=0.2000 (SE=0.2000, n=5)
7 False paired Δ̄=+0.1429 <= 1.0·SE=0.1429 (SE=0.1429, n=7)
10 False paired Δ̄=+0.1000 <= 1.0·SE=0.1000 (SE=0.1000, n=10)
20 True paired Δ̄=+0.0500 > 1.0·SE=0.0500 (SE=0.0500, n=20)
--- n=1 (SE collapses) ---
paired Δ̄=+1.0000 > 0 (SE=0 → STRICT fallback, warned; n=1)
--- paired with no deltas falls back to significant ---
Δ=+0.1000 > 1.0·SE=0.0283 (SE=0.0283)
--- paired vs significant on the SAME data (10 tasks, num_trials=1, +1 task) ---
cand combined_stderr=0.1633 cur combined_stderr=0.1667
k_se=1.0  paired: False | paired Δ̄=+0.1000 <= 1.0·SE=0.1000 (SE=0.1000, n=10)
k_se=1.0  signif: False | Δ=+0.1000 <= 1.0·SE=0.2333 (SE=0.2333)
k_se=0.2  paired: True | paired Δ̄=+0.1000 > 0.2·SE=0.0200 (SE=0.1000, n=10)
k_se=0.2  signif: True | Δ=+0.1000 > 0.2·SE=0.0467 (SE=0.2333)
--- unknown mode ---
ValueError: unknown gate mode: 'nope'
--- train refused ---
TrainGateError: Acceptance must be gated on VAL, got split='train'. Gating on train overfits the
```

What each line substantiates:

- **`mean(Δ) = SE = 1/n` exactly** for a 1-of-n gain — so at the default `k_se=1.0` the strict `>` comparison rejects it, and at `gate_k_se: 0.2` it accepts. The `n=20 True` row is float slack at `0.05`, not a different rule. The docs now say "just misses the strict bar at k=1.0, accepts at 0.2" rather than the older, vaguer "banks 1-task gains" claim (true only for `k < 1`).
- **paired SE is non-zero at `num_trials=1`** (`SE=0.1000` with all per-trial SEs zero) — this is why paired works on single-trial runs, and it comes from between-task delta spread.
- **The old `INPUTS.md` "~2-3x smaller" figure**: measured `0.2333 / 0.1000 = 2.33x` on this case. Directionally right, but I removed the specific multiplier from the docs since it is data-dependent, and stated the mechanism (combined_stderr carries absolute between-task spread) instead.
- **`SE=0` does not silently act strict** — it emits `SE=0 → STRICT fallback, warned` and logs `gate_warning` (`gate.py:35-55`). `skills/phases/gate/SKILL.md` previously said `significant` "silently degrades to strict", which was factually wrong; corrected.
- **`paired` with no deltas falls back to `significant`**, not to a free pass (`gate.py:117-123`).
- **Exhaustiveness and the val-only rule** confirmed by the `ValueError` / `TrainGateError`.

### 4. The bug the sweep exposed

Docs told agents to run `--mode paired`; the CLI rejected it. Before my change:

```
$ python skills/phases/gate/scripts/run.py --current 0.5 --candidate 0.6 --mode paired
gate: error: argument --mode: invalid choice: 'paired' (choose from significant, strict, threshold, simplicity_tiebreak)
```

Callers of that broken instruction: `docs/REPRODUCE_tau2.md:212` and `skills/algorithms/agent-optimize/SKILL.md:70`. Per the house rule that a fix belongs where all callers route, I fixed the shared script (added `paired` to `choices` + a `--paired-deltas` argument, since paired genuinely needs per-task data the old signature could not express) rather than deleting `paired` from the two docs. After:

```
$ python scripts/run.py --current 0.50 --candidate 0.60 --mode paired --k-se 0.2 --paired-deltas 1,0,0,0,0,0,0,0,0,0
{
  "accept": true,
  "reason": "paired Δ̄=+0.1000 > 0.2·SE=0.0200 (SE=0.1000, n=10)",
  "delta": 0.1,
  "threshold": 0.020000000000000004
}

$ python scripts/run.py --current 0.50 --candidate 0.62 --mode significant --k-se 1.0 --candidate-stderr 0.03 --current-stderr 0.03
{
  "accept": true,
  "reason": "Δ=+0.1200 > 1.0·SE=0.0424 (SE=0.0424)",
  "delta": 0.12,
  "threshold": 0.042426406871192854
}

$ python scripts/run.py --current 0.50 --candidate 0.60 --mode paired --k-se 1.0 --paired-deltas 1,0,0,0,0,0,0,0,0,0
{
  "accept": false,
  "reason": "paired Δ̄=+0.1000 <= 1.0·SE=0.1000 (SE=0.1000, n=10)",
  "delta": 0.1,
  "threshold": 0.1
}
```

Both examples now in `skills/phases/gate/SKILL.md` are copy-pasteable and produce the output above.

### 5. Whole-repo sweep (a partial fix is a failed fix)

```
$ grep -rn "gate_mode" . --include="*.py" --include="*.yaml" --include="*.yml" --include="*.md" --include="*.json" --include="*.html"
```
returned 27 hits across `core/`, `docs/`, `examples/`, `skills/`, `templates/`, `ci/`. I then swept the mode *names* independently (`grep -rn "significant"` over `*.md *.html *.yaml`) to catch surfaces that describe modes without using the key, which is how I found `docs/OPTIMIZE_YOUR_OWN.md:103` (`- gate: <significant (k_se) | strict | threshold>` — wrong key name *and* wrong mode list) and its HTML mirror `site/optimize-your-own.html:200`. That HTML mirror is the kind of surface a docs-only sweep usually misses.

Two inconsistencies found that the issue did not mention:
- `docs/REPRODUCE_tau2.md:141` claimed `gate_mode significant` for a run whose `examples/tau2_airline/capevolve.yaml:42` is `paired` — the reproduction recipe contradicted the config it reproduces. Corrected to `paired`.
- `docs/how-to/cap-evolve-with-exgentic-tau2.md` uses `significant` in two places. I did **not** flip these: that walkthrough has ~1 val task, where the paired SE collapses to 0 and falls back to strict, so `significant` + `num_trials: 5` is the correct choice there. I documented it as a deliberate override with the reason, in both the summary (line 77) and the yaml block (line 208).

Post-change verification of #104's own stated test:

```
$ grep -rl "simplicity_tiebreak" docs/ skills/phases/intake/SKILL.md
docs/OPTIMIZE_YOUR_OWN.md
docs/ADAPTER_CONTRACT.md
docs/HONEST_EVAL.md
skills/phases/intake/SKILL.md

$ grep -rn "the default" docs/HONEST_EVAL.md | head -3
33:Set with `gate_mode` in `capevolve.yaml` (strictness with `gate_k_se`, default
37:| **`paired`** — **the default** | accept iff `mean(per-task Δ) > k_se · SE(Δ)`, where `Δ[t] = cand[t] − curr[t]` over the same val tasks | default and recommended: candidate and current are scored on the *same* tasks, so cross-task variance cancels and the test is far more powerful than `significant` |
```

### 6. Tests — 179 passed, 0 failed (baseline 179)

```
$ cd /tmp/wt-104 && PYTHONPATH=/tmp/wt-104/core /tmp/ce-venv/bin/python -m pytest core/tests -q
179 passed in 63.65s (0:01:03)
```

No test was added: per CONTRIBUTING ("Add a test for any core change"), the only core change is a docstring — zero behavioral surface. The one behavioral change is the gate skill's argparse `choices`, and `skills/phases/gate/scripts/check.py` already covers the script's contract and passes (§8). The paired-mode behavior itself is already covered by `core/tests/test_w1_engine.py`.

### 7. compileall

```
$ /tmp/ce-venv/bin/python -m compileall -q core skills && echo "COMPILEALL_CLEAN"
COMPILEALL_CLEAN
```

### 8. Manifest + every touched skill's check.py

`skills/phases/gate/meta.yaml` carried the same stale claim in its manifest `summary` ("significance by default") — a surface that is neither a doc nor a docstring, and that `build_manifest.py` bakes into `manifest.json`. Fixed and rebuilt:

```
$ /tmp/ce-venv/bin/python skills/_registry/build_manifest.py skills
  optimizer: run-optimizer

$ git diff --stat skills/_registry/manifest.json
 skills/_registry/manifest.json | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

$ for s in phases/gate phases/intake algorithms/skillopt algorithms/hill-climb algorithms/gepa algorithms/agent-optimize; do echo "=== $s"; python skills/$s/scripts/check.py | grep '"ok"'; done
=== phases/gate
  "ok": true,
=== phases/intake
  "ok": true,
=== algorithms/skillopt
  "ok": true,
=== algorithms/hill-climb
  "ok": true,
=== algorithms/gepa
  "ok": true,
=== algorithms/agent-optimize
  "ok": true,
```

### 9. Relative links + anchors resolve on disk

Checked every non-http markdown link in all changed `.md` files, resolving the path relative to the containing file and, where an `#anchor` is present, confirming a matching heading exists in the target:

```
$ python - <<'PY'   # resolves each link + slugifies target headings
checked=67 broken=0
```

That covers the depth-sensitive ones I hand-wrote, e.g. `skills/phases/intake/inputs/INPUTS.md` → `../../../../docs/HONEST_EVAL.md#gate-modes` (four levels up, easy to get wrong) and `ci/benchmarks/README.md` → `../../docs/HONEST_EVAL.md#gate-modes`.

### 10. Commit authorship

```
$ git log -1 --format='%H%n%an <%ae>%n%s'
c4d51bee4d7b3632c1ed3c4ffe90e0fd1a76d57e
docs: document the real gate modes (paired is the default)

$ git log -1 --format='%B' | grep -ci "co-authored-by\|generated with"
0
```

OsherElhadad:
## 🔍 Review — PR #198

**Verdict: CHANGES REQUESTED** (2 blocking, 6 non-blocking, 3 nits)

The ground-truth research here is genuinely good — the `paired`-is-the-default asymmetry is correctly traced, all five modes are exhaustively enumerated, and the quantitative claims reproduce to the digit (I re-derived them independently). But the PR ships **new executable CLI functionality with zero tests** while explicitly arguing none is needed, and it leaves the very config surface it was fixing carrying a claim its own arithmetic disproves.

---

### Blocking

**B1 — `skills/phases/gate/scripts/run.py:38`: new executable parsing, zero tests, and it crashes with a raw traceback on malformed input.**

```python
deltas = [float(x) for x in args.paired_deltas.split(",") if x.strip()] or None
```

This is the only new *logic* in the PR and it is untested. Three concrete defects:

1. **Unhandled `ValueError` → raw traceback, exit 1.** A user or agent typo produces a Python stack trace instead of an argparse error:
   ```
   $ python run.py --current 0.5 --candidate 0.6 --mode paired --paired-deltas "1,abc,0"
     File ".../skills/phases/gate/scripts/run.py", line 38, in main
       deltas = [float(x) for x in args.paired_deltas.split(",") if x.strip()] or None
   ValueError: could not convert string to float: 'abc'
   exit=1
   ```
   Same for the very plausible `--paired-deltas "1;0;0"` (semicolons). Consequence: an agent driving the phase chain (which is the documented use, `agent-optimize/SKILL.md:72`, `REPRODUCE_tau2.md:212`) gets an unparseable crash rather than "invalid value for --paired-deltas".
   Fix: `type=` converter or a `try/except ValueError: p.error(...)`. Argparse then exits 2 with a clean message, matching every other flag.

2. **No length validation against the val split.** `--paired-deltas 1,0` on a 50-task val split is silently accepted as `n=2`:
   ```
   $ python run.py --current 0.5 --candidate 0.6 --mode paired --paired-deltas "1,0"
     "reason": "paired Δ̄=+0.5000 <= 1.0·SE=0.5000 (SE=0.5000, n=2)"
   ```
   Note `Δ̄=+0.5000` while `--current 0.5 --candidate 0.6` implies `Δ=+0.10`. The script prints `delta: 0.5` — the *paired* mean — and never cross-checks it against `candidate − current`. Consequence: a truncated paste (an easy failure mode when copying 50 deltas by hand) silently produces a different accept/reject than the run made, which is exactly the "wrong accept/reject" class this repo exists to prevent. `gate.decide` can't catch this — it never sees the val size.
   Fix: warn (or error) when `abs(mean(deltas) − (candidate − current)) > 1e-6`. That single consistency check catches truncation, wrong-order pasting, and wrong-sign deltas at once, using data the script already has.

3. **No test, and `check.py` does not cover it.** `skills/phases/gate/scripts/check.py` exercises only `mode="significant"` (lines 27/29/38) — `grep -n "mode\|paired\|deltas" check.py` shows zero `paired` and zero `--paired-deltas`. Nothing in `core/tests/` invokes any phase `scripts/run.py` (`grep -rln "phases/gate/scripts" core/tests/ ci/ .github/` → empty), and `ci.yml:42` only runs `pytest core/tests`. So this argument is covered by **nothing** in CI. The PR's justification ("`check.py` already covers the script's contract") does not hold for the code it added.
   Fix: extend `check.py` with a `paired` case + a malformed-input case, or add a `core/tests/test_gate_cli.py` that shells the script. Either satisfies `CONTRIBUTING.md:33`.

> On `gate.py` specifically the PR's claim **is** correct — `git diff -U0 origin/main...HEAD -- core/cap_evolve/gate.py` shows exactly two hunks, `@@ -3,4 +3,18 @@` (module docstring) and `@@ -75,2 +89,4 @@` (`decide` docstring). Zero executable lines. That part of the "no test needed" argument stands; it does not extend to the CLI.

**B2 — `templates/project/capevolve.yaml:80`: the shipped default config asserts a property the PR's own arithmetic disproves.**

```yaml
gate_mode: paired    # DEFAULT: per-task paired SE, same tasks both sides — banks real 1-task gains.
gate_k_se: 1.0
```

Those two lines are adjacent and contradictory. The PR itself establishes (and I verified exactly, below) that at `k_se=1.0` a 1-of-n gain gives `mean(Δ) = SE` exactly, so the strict `>` **rejects** it. The template therefore promises "banks real 1-task gains" beside the one `k_se` value at which it provably cannot. `docs/HONEST_EVAL.md:60` gets this right ("just misses"); the config a user actually edits does not.

Consequence: the highest-traffic surface — the scaffolded yaml every new project starts from, generated by intake — teaches the wrong thing. This is the identical failure mode the PR was opened to fix (config comment vs. reality), reintroduced. It also survives unchanged in `examples/skillsbench/capevolve.yaml:41` and `examples/tau2_airline/PROMPT.md:180` / `examples/skillsbench/PROMPT.md:258` — though those pair it with `gate_k_se: 0.2`, where the claim is true, so only the template is actually wrong.

Fix (one line):
```yaml
gate_mode: paired    # DEFAULT: per-task paired SE, same tasks both sides. NOTE: a gain on
                     # exactly 1 of n tasks gives mean(Δ)=SE, so it does NOT clear k_se=1.0 —
                     # lower gate_k_se (the examples use 0.2) to bank single-task gains.
```

---

### Non-blocking

**N1 — `docs/AGENT_ORCHESTRATION.md:90` was missed by the sweep, and it is the same bug the PR fixed.** The PR names two callers of the broken `--mode paired` instruction. There were **three**. This line is unchanged by the PR and still reads:
```
python "$S/phases/gate/scripts/run.py" --mode paired --k-se 1.0 --current <best_mean> --candidate <cand_mean> --current-stderr <s> --candidate-stderr <s>
```
It no longer exits 2 (the `choices` fix helps it), but it now silently does the **wrong thing** — `paired` with no `--paired-deltas` falls back to `significant` and consumes the stderr flags:
```
$ python run.py --mode paired --k-se 1.0 --current 0.50 --candidate 0.62 --current-stderr 0.03 --candidate-stderr 0.03
  "reason": "Δ=+0.1200 > 1.0·SE=0.0424 (SE=0.0424)"     ← unpaired, despite --mode paired
```
Consequence: an agent following AGENT_ORCHESTRATION.md believes it applied the paired gate and applied the weaker unpaired one. Fix: apply the same edit the PR made to `REPRODUCE_tau2.md:212` — add `--paired-deltas`, or change to `--mode significant`. Note `site/agent-orchestration.html:108` mirrors this page and says "paired significance gate" — check whether it needs the same treatment.

**N2 — the `paired`→`significant` fallback is silent to the user.** `gate.py:117-123` and `run.py:38` mean that asking for `paired` with no deltas yields a `significant` decision whose `reason` string contains **no trace** of the substitution:
```
$ python run.py --current 0.5 --candidate 0.6 --mode paired            # no deltas at all
  "reason": "Δ=+0.1000 > 0 (SE=0 → STRICT fallback, warned)"
```
The user asked for paired, got significant, which then degraded to strict — a two-step slide to the weakest possible gate, with the output only disclosing the second step. Compare `_warn_se_zero` (`gate.py:49-68`), which the PR rightly praises for being loud: SE=0 logs a `gate_warning` and stamps the reason. The paired→significant fallback does neither. Given this repo's stated posture ("we do NOT silently behave strict"), the asymmetry is hard to defend on honesty grounds. It is non-blocking only because it predates this PR — but this PR is what documents and promotes the fallback, so it is the right place to fix it. Fix: prefix the reason with `paired→significant (no per-task deltas): ` and log a `gate_warning`, reusing the existing mechanism. In the CLI, `run_dir` is `None` so no event is possible at all — worth noting in the docs.

**N3 — `simplicity_tiebreak` is documented as a usable mode but is unreachable as one in every real run.** The tiebreak branch (`gate.py:186-199`) fires only when `candidate_size` **and** `current_size` are both non-`None`. `grep -rn "candidate_size" core/cap_evolve/harness.py core/cap_evolve/gepa.py skills/algorithms/*/scripts/run.py` → **nothing**; the only producer in the repo is `core/tests/test_core.py:83`. So `gate_mode: simplicity_tiebreak` in a real run is bit-identical to `strict`:
```
mode=simplicity_tiebreak, no sizes:  tie → "Δ=+0.0000 <= 0 (no simpler tie)"
mode=strict,              same tie:  tie → "Δ=+0.0000 <= 0"
```
Five documentation surfaces now advertise it (`HONEST_EVAL.md:42`, `ADAPTER_CONTRACT.md:106`, `OPTIMIZE_YOUR_OWN.md:123`, `intake/SKILL.md:57`, `site/optimize-your-own.html`) with no hint that the size plumbing does not exist. A user who sets it gets `strict` and never learns. Fix: add "requires `candidate_size`/`current_size`, which the harness does not currently populate — effectively `strict` in a normal run" to the table row, or open a follow-up to plumb sizes. Documenting a mode into prominence is exactly when this should be caught.

**N4 — the "1-of-n misses the default bar" finding is under-promoted for how consequential it is.** It appears in full only at `docs/HONEST_EVAL.md:55-64` (a `###` subsection) and `gate/references/concepts.md:36` (a parenthetical: "the default `k=1` just misses, since the comparison is strict"). Every other surface still says the opposite in effect: `intake/SKILL.md:50` — "can bank a genuine single-task gain"; `inputs/INPUTS.md:165` — "so real 1-task gains bank"; `gate/SKILL.md:60` — "can bank a real single-task gain". None of the three mentions `k_se`. Since intake's default is `1.0`, the practical reading of those lines is false. Fix: append "(needs `gate_k_se < 1.0`; at the default `1.0` a 1-of-n gain lands exactly on the bar and is rejected)" to each. This is the single most actionable thing the PR discovered and it should not require reading HONEST_EVAL to find.

**N5 — `examples/tau2_airline/capevolve.yaml:43` keeps the "~2-3x smaller" figure the PR says it removed.** The PR states it "removed the specific multiplier from the docs since it is data-dependent" — and it did, from `INPUTS.md`. But the same figure survives in the committed example:
```yaml
#   paired tests mean(per-task Δ) on the SAME tasks both sides → SE ~2-3x smaller than
```
It happens to be right (I measured `2.333×` on the 10-task/1-gain case), but if it was worth removing from one surface for being data-dependent, it is worth removing from the other. Inconsistent application of the PR's own editorial decision.

**N6 — link count is 66, not the claimed 67.** My independent resolver over all changed `.md` files reports `resolved=66 broken=0`. Zero broken is what matters and that reproduces; the count does not. Likely a different file set (the PR may have counted a `site/` or non-changed file). Cosmetic, but it is a stated verification number.

---

### Nits

**n1 — `docs/HONEST_EVAL.md:60`, wording.** "while at the `gate_k_se: 0.2` the examples use it accepts" is ungrammatical. Suggest: "while at the `gate_k_se: 0.2` used by the examples, it accepts."

**n2 — `skills/phases/gate/scripts/run.py:38`, the `or None` idiom.** `[] or None` → `None` is correct but non-obvious; a reader has to reason about falsy-list truthiness to see that empty input reaches the documented fallback rather than passing `[]`. `deltas = [...] ; deltas = deltas or None` on two lines, or an explicit `if not deltas: deltas = None`, reads plainly at zero cost.

**n3 — `run.py` has no `--run-dir`, so no `gate_warning` can ever be logged from the CLI.** `decide(..., run_dir=)` is never passed, so every SE=0 / fallback warning the docs promise is invisible in standalone use — the JSON `reason` is the only channel. Worth one sentence in `gate/SKILL.md` so nobody hunts for a `gate_warning` event that cannot exist. (Not worth adding the flag.)

---

### Gate-mode table verification

Read `core/cap_evolve/gate.py` line-for-line against the PR's `docs/HONEST_EVAL.md:36-42` table. **Exhaustiveness confirmed:** the five `if mode ==` branches at `gate.py:117, 153, 178, 182, 186` are followed by `raise ValueError(f"unknown gate mode: {mode!r}")` at `gate.py:200` with no `else`/default path — any sixth name raises, so the table is complete.

| Row | Documented | Code | Verdict |
|---|---|---|---|
| `paired` | `mean(per-task Δ) > k_se · SE(Δ)`, Δ over same val tasks | `gate.py:124-151` — `mean_d = sum(d)/n`; `var = Σ(d−m)²/(n−1)`; `se = sqrt(var/n)`; `ok = mean_d > k_se*se` | ✅ exact. Sample variance with `n−1` denominator, as `concepts.md:29` states |
| `paired`, n=1 | (in caveat) SE→0, strict fallback, warned | `gate.py:131-133` — `if n >= 2 … else se = 0.0`, then `gate.py:135-144` warns + `ok = mean_d > 0` | ✅ exact. Verified: `paired Δ̄=+1.0000 > 0 (SE=0 → STRICT fallback, warned; n=1)` |
| `paired`, no deltas | falls back to `significant` | `gate.py:118-122` — `if not deltas: mode = "significant"` | ✅ behavior correct — but silent, see **N2** |
| `significant` | `Δ(means) > k_se · sqrt(SE_c² + SE_p²)` | `gate.py:153-176` — `se = sqrt(cand_se² + cur_se²)`; SE=0 → warn + strict | ✅ exact |
| `threshold` | `Δ > threshold` (flat margin) | `gate.py:178-180` — `ok = delta > threshold` | ✅ exact. Note `threshold` defaults to `0.0`, so an unset `threshold` mode ≡ `strict`; not documented anywhere, minor |
| `strict` | `Δ > 0` | `gate.py:182-184` — `ok = delta > 0` | ✅ exact |
| `simplicity_tiebreak` | `Δ > 0` accepts; near-tie `abs(Δ) ≤ 1e-9` → accept smaller | `gate.py:186-199` — `if delta > 0: accept`; `tie = abs(delta) <= 1e-9`; requires both sizes non-`None` **and** `candidate_size < current_size` | ✅ tolerance `1e-9` and `<` (strict, so equal sizes reject) both correct. ⚠️ the *both-sizes-required* precondition is undocumented and never met in practice — **N3** |
| val-only | every mode | `gate.py:112-116` — `TrainGateError` before any mode dispatch | ✅ exact, and it is genuinely before the branch |

**Verdict: 8/8 rows behaviorally accurate.** Two undocumented preconditions (`simplicity_tiebreak` sizes; `threshold` defaulting to 0).

---

### The k_se=1.0 arithmetic

Derived from scratch, not taken from the PR. Let `n` val tasks, `j` of them improve by `m`, the rest unchanged. Deltas: `j` copies of `m`, `n−j` zeros.

Take `j = 1` (the case in question) — one task improves by `m`, `n−1` unchanged:

```
mean(Δ) = m/n

Σ(Δ − mean)²  =  (m − m/n)²  +  (n−1)·(m/n)²          one improved task, n−1 zeros
              =  m²(n−1)²/n²  +  (n−1)m²/n²
              =  (m²(n−1)/n²)·[(n−1) + 1]
              =  m²(n−1)/n

var(Δ) = Σ/(n−1) = m²/n                                sample variance, n−1 denominator
SE(Δ)  = sqrt(var/n) = sqrt(m²/n²) = m/n
```

So **`SE(Δ) = m/n = mean(Δ)` exactly**, for every `n ≥ 2` and every `m`. The gate's comparison is strict `>` (`gate.py:145`), so `mean_d > k_se·se` with `k_se = 1.0` is `m/n > m/n` → **False**. Rejected.

Verified two ways — exact `fractions.Fraction` arithmetic, and the real `decide()` float path:

```
n=  2 j=1  mean=0.500000 SE=0.500000  k=1 -> m==SE      (exact rationals)
n=  5 j=1  mean=0.200000 SE=0.200000  k=1 -> m==SE
n= 10 j=1  mean=0.100000 SE=0.100000  k=1 -> m==SE
n=100 j=1  mean=0.010000 SE=0.010000  k=1 -> m==SE
n= 10 j=2  mean=0.200000 SE=0.133333  k=1 -> m>SE       (2 tasks DOES clear it)

n=  5 accept=False  paired Δ̄=+0.2000 <= 1.0·SE=0.2000 (SE=0.2000, n=5)     (decide())
n= 10 accept=False  paired Δ̄=+0.1000 <= 1.0·SE=0.1000 (SE=0.1000, n=10)
n= 50 accept=False  paired Δ̄=+0.0200 <= 1.0·SE=0.0200 (SE=0.0200, n=50)
n=100 accept=False  paired Δ̄=+0.0100 <= 1.0·SE=0.0100 (SE=0.0100, n=100)
```

**Claim (b) VERIFIED, and it is stronger than the PR states.** Three points the docs do not make:

1. **It is magnitude-independent.** `m` cancels — a 1-of-10 gain of `+1.0` (full task fixed) and of `+0.05` (marginal) both land exactly on the bar. Fixing one task *perfectly* is rejected identically to barely nudging it.
2. **It is n-independent.** The PR's evidence table shows `n=20 accept=True`, which the PR correctly calls "float slack at 0.05". My exact-rational run confirms: `n=20 j=1 → m==SE`. The `True` is IEEE-754 noise, not a rule. **At every n, a 1-of-n gain is rejected at `k_se=1.0`.** A reader of the PR's table could reasonably conclude "large val splits fix this" — they do not. Worth stating explicitly, since the arithmetic is knife-edge and a reader may see either verdict depending on float luck.
3. **The cliff is at exactly j=2.** `j=2, n=10` gives `mean=0.200 > SE=0.133` → accepts. So the default gate's real rule is "improve ≥2 tasks or bank nothing."

**Claim (a) VERIFIED:** paired SE is non-zero at `num_trials=1` — every case above has all per-trial SEs zero yet `SE=0.1000` at n=10, sourced from between-task delta spread (`gate.py:127-129` uses only the deltas, never `candidate_stderr`).

**The `0.233` vs `0.100` figure VERIFIED** (`HONEST_EVAL.md:63`): 10 tasks, `num_trials=1`, current 5/10 → candidate 6/10:
```
cand combined_stderr 0.1633  cur 0.1667
significant SE_diff = 0.2333   paired SE = 0.1000   ratio = 2.333
```

**Are the docs clear enough? No — see N4.** `HONEST_EVAL.md:55-64` states it correctly but is the only surface that does; three other surfaces still say "banks 1-task gains" unqualified, and `templates/project/capevolve.yaml:80` states it *beside `gate_k_se: 1.0`* (**B2**). "The default configuration cannot bank a single-task improvement of any size, on any val size" deserves a callout box, not a `###` subsection — it is a finding about the shipped default, not a footnote about `paired`.

---

### Untested code changes

| File:line | Executable change | Test | Status |
|---|---|---|---|
| `core/cap_evolve/gate.py:1-20, 89-92` | **none** — docstrings only, verified via `git diff -U0` showing 2 hunks both inside `"""..."""` | n/a | ✅ PR's claim holds |
| `skills/phases/gate/scripts/run.py:24` | `choices` gains `"paired"` | none | ❌ |
| `skills/phases/gate/scripts/run.py:29-31` | **new `--paired-deltas` argument** | none | ❌ |
| `skills/phases/gate/scripts/run.py:38` | **new parsing: split/filter/float/`or None`** | none | ❌ **B1** |
| `skills/phases/gate/scripts/run.py:42` | new `paired_deltas=deltas` kwarg to `decide` | none | ❌ |
| `skills/algorithms/{gepa,hill-climb,skillopt}/scripts/run.py` | `--gate-mode` help string only | n/a | ✅ |

Missing tests, minimum set: (1) `--mode paired --paired-deltas` produces a paired reason; (2) `--mode paired` with no deltas hits the documented fallback; (3) malformed deltas exit cleanly, not by traceback. `check.py` covers **zero** of these — it tests only `significant` (lines 27/29/38). `CONTRIBUTING.md:33` requires a test for core changes; this is a skill script, so the letter of the rule is arguably satisfied — but the spirit ("new accept/reject-affecting code gets a test") plainly is not, and `ci.yml:42` runs only `pytest core/tests`, so nothing in CI touches it.

---

### Other documented-but-broken CLI invocations

I built a repo-wide checker: extract every fenced `python .../scripts/{run,check}.py ...` from all `.md` (excluding `docs/superpowers/`), resolve the script, run its real `--help`, and diff the documented flags and literal choice-values against argparse's actual surface.

**On `origin/main` — 38 invocations, 2 failures, both the exact bug this PR fixes:**
```
MAIN scanned=38 findings=2
  docs/AGENT_ORCHESTRATION.md:90       --mode 'paired' NOT IN ['significant','strict','threshold','simplicity_tiebreak']
  skills/algorithms/agent-optimize/SKILL.md:70   --mode 'paired' NOT IN [...]
```
Reproduced live:
```
$ cd /tmp/rv-198-main/skills/phases/gate/scripts && python run.py --current 0.5 --candidate 0.6 --mode paired
gate: error: argument --mode: invalid choice: 'paired' (choose from significant, strict, threshold, simplicity_tiebreak)
exit=2
```
**Bug confirmed. But the PR names only two callers and my checker found the same two plus `REPRODUCE_tau2.md:212` had already been `significant` on main — so the third broken caller, `AGENT_ORCHESTRATION.md:90`, is unfixed (N1).**

**On this branch — 39 invocations, 0 failures:**
```
scanned=39 findings=0
```
Every documented flag and choice-value now parses. (This is flag-and-choice level, not semantic — which is precisely why N1 slips through: `AGENT_ORCHESTRATION.md:90` *parses* fine and silently does the wrong thing.)

**How long was it broken?** `git log -S` dates it:
```
--mode paired added to docs:                    5d6d913  2026-07-22  (PR #50, dual-mode orchestration)
gate run.py 4-choice list:                      59b72de  2026-06-15  (original gate skill)
```
So **8 days** (2026-07-22 → 2026-07-30), introduced by #50 writing `--mode paired` into two docs against a script that had rejected it for five weeks. Nothing caught it because no CI job executes a documented command.

**Is a repo-wide check warranted? Yes — and it is ~40 lines.** The class of drift is structural: 39 invocations across 24 files, argparse surfaces in 20 skills, and zero coupling between them. My checker took one pass to write and runs in seconds. **Recommend a follow-up issue** (out of scope here): a `ci.yml` step that greps fenced `scripts/run.py` invocations and asserts each flag/choice against `--help`. Two caveats for whoever picks it up: (a) `--help` needs `PYTHONPATH=core` or every script exits 2 and the check silently passes everything — I hit exactly this and had to fix my own harness; (b) flag-level checking cannot catch N1-style semantic drift, so it is a floor, not a ceiling.

---

### #113 contradiction

Real and bidirectional. PR #195 (`origin/fix/issue-113-small-samples`) **already implements** both mechanisms this PR documents as absent.

**#195 ships** (`gate.py:133-142` on that branch):
```python
from .stats import t_multiplier_for_z
k_eff = t_multiplier_for_z(k_se, n)      # Student-t at df=n-1, always >= k_se
bar = k_eff * se
if n < LOW_CONFIDENCE_VAL_TASKS:         # = 5, splits.py:27
    low_conf = f" [LOW CONFIDENCE: only {n} val tasks]"
```
plus `check_val_size` raising `TinyValSplitError` on val of 0–1 tasks.

**This PR asserts the opposite, in two places:**

1. `docs/HONEST_EVAL.md:66-68` — heading `### Small-sample caveat (current behavior)`, body: *"The bar is a fixed `k · SE`, with no t-correction and no minimum-val-size guard, so it is optimistic on tiny val splits."*
2. `skills/phases/gate/references/concepts.md:71-74` — same heading, body: *"The bar is a fixed `k · SE` — a z-style critical value with no t-correction and no minimum-val-size guard."*

Both statements become **false** the moment #195 lands. And #195 edits `HONEST_EVAL.md` guarantee 3 to say the opposite — so a merged repo asserts both. I confirmed the merge is clean but the content is not:

```
$ T=$(git merge-tree --write-tree HEAD origin/fix/issue-113-small-samples)
$ git show $T:docs/HONEST_EVAL.md | grep -n "Student-t\|no t-correction"
33:   uses the equivalent **Student-t** multiplier at df = n−1 (`stats.t_critical`).
76:   The bar is a fixed `k · SE`, with no t-correction and no minimum-val-size guard, so
```

**Line 33 and line 76 of the same file, directly contradicting each other. Git will not flag this** — different hunks, no textual overlap, `merge-tree` reports CLEAN. It merges green and ships a self-contradicting honesty doc.

**Paragraphs that must change when #195 lands** (both authored by this PR, both pre-labelled by this author for exactly this — good foresight, but the labelling does not prevent the contradiction, it only documents it):
- `docs/HONEST_EVAL.md:66-81` — whole `### Small-sample caveat (current behavior)` subsection. Delete "no t-correction and no minimum-val-size guard"; replace with the t-multiplier + `LOW_CONFIDENCE_VAL_TASKS=5` + `TinyValSplitError` behavior. The two `SE=0` / `no paired data` bullets stay accurate.
- `skills/phases/gate/references/concepts.md:71-74` — same paragraph, same fix. **This one is not in #195's diff** (#195 touches only `gate.py` + `HONEST_EVAL.md`), so if #195 merges second it will not even see this file, and a stale "no t-correction" claim survives in the skill reference indefinitely. **This is the paragraph most likely to rot.**
- `skills/phases/gate/SKILL.md:96` cross-link to `#gate-modes` — verify the anchor survives; harmless if the heading is untouched.

Third-order interaction worth flagging: #195's `check_val_size` **refuses** a val split of 1 task. `docs/how-to/cap-evolve-with-exgentic-tau2.md:227` documents `{"train":["9"],"val":["9"],"test":["9"]}`. Post-#195 that walkthrough **fails at split freeze** unless `CAPEVOLVE_ALLOW_TINY_VAL=1`. Neither PR mentions it. Flag on #195, not here.

**Recommended merge order: #198 first, then #195.** Rationale: #198 is 25 files of contended docs and grows stale fastest; #195 is 2 files and its author is already reasoning about these paragraphs. Merging #198 first means #195's rebase surfaces both stale paragraphs as content it must rewrite. The reverse order leaves #198 (the larger, docs-heavy PR) to remember to hunt down a caveat paragraph in a skill reference file that #195 never touched. **Whichever order: #195 must not merge without editing `skills/phases/gate/references/concepts.md:71-74`.** Add it as an explicit checklist item on #195 — this is the actionable coordination step.

---

### Merge-order note

Trial-merged all six sibling branches against this HEAD with `git merge-tree --write-tree`:

```
fix/issue-103-adapter-methods     CLEAN
refactor/issue-106-deleak-intake  CLEAN
fix/issue-113-small-samples       CLEAN     ← but semantically contradictory, above
fix/issue-102-counts              CLEAN
feat/issue-142-protected-paths    CONFLICT (content): docs/HONEST_EVAL.md
feat/issue-134-provider-creds     CLEAN
```

File-level overlap:

| Branch (PR) | Shared files | Risk |
|---|---|---|
| #197 `feat/issue-142-protected-paths` | `docs/HONEST_EVAL.md`, `templates/project/capevolve.yaml` | 🔴 **only real textual conflict.** #197 renames `## The four guarantees` → `## The five guarantees` and inserts guarantee 5 right where this PR inserts `## Gate modes`. Both edit the same region. Hand-resolvable (~10 lines) — keep #197's "five guarantees" heading + new item 5, keep this PR's `## Gate modes` section after item 4. The `capevolve.yaml` halves don't collide (#197 adds `protected_paths`, this PR edits the gate block) |
| #195 `fix/issue-113-small-samples` | `core/cap_evolve/gate.py`, `docs/HONEST_EVAL.md` | 🟡 merges clean, ships contradictions. See above. `gate.py` doesn't collide: this PR touches only docstrings, #195 only the `paired` branch body |
| #181 `fix/issue-103-adapter-methods` | `RUN.md`, `docs/ADAPTER_CONTRACT.md`, `docs/how-to/cap-evolve-with-exgentic-tau2.md`, `skills/phases/intake/SKILL.md` | 🟡 4 shared files, all clean. Highest re-conflict risk if either rebases — this PR appends a `## Gate modes` section to `ADAPTER_CONTRACT.md` while #181 rewrites the adapter-method text above it. Merge close together |
| #185 `refactor/issue-106-deleak-intake` | `skills/phases/intake/SKILL.md` | 🟢 clean; this PR inserts a `### Gate mode` block, #185 removes leaked content elsewhere |
| #189 `fix/issue-102-counts` | `README.md` | 🟢 clean; different lines |
| #190 `feat/issue-134-provider-creds` | `templates/project/capevolve.yaml` | 🟢 clean; different section |

**`skills/_registry/manifest.json` — mechanical but guaranteed.** This PR changes it (1 line, from `meta.yaml`'s `summary`). Any other PR touching any `meta.yaml` regenerates the same file. It is a build artifact: **regenerate with `python skills/_registry/build_manifest.py skills` after every merge; never hand-merge or accept a `<<<<<<<` resolution.** A hand-resolved manifest silently desyncs from `meta.yaml` and there is no CI check comparing them (worth its own follow-up alongside the CLI checker). #189 is listed as touching the manifest — but its diff vs `main` shares only `README.md` with this PR, so no manifest collision from that pair.

**Recommended order:** #189 (README, trivial) → #185 (intake) → #181 (intake + ADAPTER_CONTRACT) → **#198 (this)** → #197 (resolve the one HONEST_EVAL conflict) → #195 (rewrite both caveat paragraphs) → #190. Regenerate `manifest.json` after any merge that touched a `meta.yaml`.

---

### Verification I re-ran

**pytest — 178 passed, 1 FAILED. The PR claims 179 passed.**
```
$ cd /tmp/rv-198 && PYTHONPATH=/tmp/rv-198/core /tmp/ce-venv/bin/python -m pytest core/tests -q
1 failed, 178 passed in 63.71s (0:01:03)

E       AssertionError: assert 'http://127.0.0.1:7887' == 'http://127.0.0.1:7878'
```
**Not this PR's fault — it fails identically on `origin/main`:**
```
$ cd /tmp/rv-198-main && PYTHONPATH=.../core python -m pytest core/tests/test_dashboard_launch.py -q
1 failed, 6 passed in 0.05s
```
Cause: `dashboard_launch.py:78` calls `_free_port(port)`, which walks up from 7878; port 7878 is occupied on this machine so it returns 7887. **Environment-dependent test, pre-existing, not blocking this PR** — but it means the PR's "179 passed" is not reproducible on a machine with 7878 in use, and the baseline is really 178+1-flaky. Worth a separate issue: the test should monkeypatch `_free_port`.

**compileall — clean:**
```
$ /tmp/ce-venv/bin/python -m compileall -q core skills && echo COMPILEALL_CLEAN
COMPILEALL_CLEAN
```

**manifest.json — reproduces byte-exactly (not hand-edited):**
```
$ /tmp/ce-venv/bin/python skills/_registry/build_manifest.py skills
wrote skills/_registry/manifest.json (20 skill(s))
$ git diff --stat skills/_registry/manifest.json
(empty — committed manifest == fresh build)
```

**All 6 touched skills' check.py — green:**
```
phases/gate                      "ok": true,
phases/intake                    "ok": true,
algorithms/skillopt              "ok": true,
algorithms/hill-climb            "ok": true,
algorithms/gepa                  "ok": true,
algorithms/agent-optimize        "ok": true,
```

**Both newly documented CLI examples (`gate/SKILL.md:92,96`) — copy-pasteable, output matches:**
```
$ python run.py --current 0.50 --candidate 0.62 --mode significant --k-se 1.0 --candidate-stderr 0.03 --current-stderr 0.03
  "accept": true,  "reason": "Δ=+0.1200 > 1.0·SE=0.0424 (SE=0.0424)"
$ python run.py --current 0.50 --candidate 0.60 --mode paired --k-se 0.2 --paired-deltas 1,0,0,0,0,0,0,0,0,0
  "accept": true,  "reason": "paired Δ̄=+0.1000 > 0.2·SE=0.0200 (SE=0.1000, n=10)"
```

**`--paired-deltas` edge cases** (the untested surface, B1):
```
empty string / flag absent → falls back to significant → SE=0 → strict:  "Δ=+0.1000 > 0 (SE=0 → STRICT fallback, warned)"
"1,abc,0"                  → ValueError traceback, exit=1                  ← B1.1
"1;0;0"                    → ValueError traceback, exit=1                  ← B1.1
"1,0"  (wrong length)      → silently n=2, Δ̄=+0.5000 ≠ candidate−current   ← B1.2
"1, 0, 0,"                 → tolerated, n=3 (whitespace + trailing comma OK)
"1"    (n=1)               → "paired Δ̄=+1.0000 > 0 (SE=0 → STRICT fallback, warned; n=1)"
--mode significant + --paired-deltas → deltas correctly ignored
```

**exit-2 reproduction on main — confirmed** (see "Other documented-but-broken CLI invocations").

**`REPRODUCE_tau2.md:141` `significant`→`paired` correction — factually right.** Verified against the *actual committed run artifacts*, not the config:
```
$ grep -rhoE 'paired Δ̄=[^"\\]{0,70}' examples/tau2_airline/run_full/ui/ | sort -u
paired Δ̄=+0.0520 > 0.2·SE=0.0078 (SE=0.0391, n=50)
paired Δ̄=+0.0280 > 0.2·SE=0.0047 (SE=0.0234, n=50)
paired Δ̄=-0.0700 <= 0.2·SE=0.0078 (SE=0.0392, n=50)
  … 10 distinct paired decisions, n=50, k=0.2; zero unpaired "Δ=" reasons
```
The published run demonstrably used `paired`/`k=0.2`. The doc said `significant`. Good catch by the PR — and this is the strongest evidence in it.

**The declined `exgentic` change — the author is right; SE→0 verified.** `docs/how-to/cap-evolve-with-exgentic-tau2.md:227` documents `{"train":["9"],"val":["9"],"test":["9"]}` — literally n=1 val:
```
paired,      n=1 val task:                  "paired Δ̄=+0.1000 > 0 (SE=0 → STRICT fallback, warned; n=1)"
significant, n=1 task, num_trials=5, se=.2: "Δ=+0.1000 <= 0.5·SE=0.1414 (SE=0.2828)"   ← real bar, correctly REJECTS
combined_stderr([0.6],[0.2]) = 0.2   (nonzero: within-task trial variance survives at n_tasks=1)
combined_stderr([0.6],[0.0]) = 0.0   (…but only if trials actually vary)
```
`paired` at n=1 degrades to strict and accepts a `+0.10` blip; `significant` + `num_trials: 5` holds a `0.1414` bar and rejects it. **Keeping `significant` there is the honest choice and the PR's stated reasoning is sound** — this is the right call, well argued, and the added yaml comment (`:209-212`) explains it for the next reader. One caveat: it depends on trials genuinely varying (second line above), which `num_trials: 5` makes likely but does not guarantee.

**Is "paired silently degrades to strict at n=1" an honesty hazard needing a code warning? It already has one** — `gate.py:135` calls `_warn_se_zero`, logs `gate_warning`, and stamps `SE=0 → STRICT fallback, warned` into the reason. The PR is right that the old `gate/SKILL.md` claim of "silently degrades" was factually wrong, and the correction is an improvement. Two residual gaps: (a) the CLI passes no `run_dir` so the event can never fire there (n3); (b) the *paired→significant* fallback has no equivalent warning at all (N2). #195's `TinyValSplitError` closes the n=1 case properly by refusing the split upfront — the better fix.

**Docs cross-consistency — `HONEST_EVAL.md`, `gate/references/concepts.md`, `site/optimize-your-own.html` agree** on all five modes, `paired` as default, the `decide()` asymmetry, and the fallback. `site/` already said "paired significance gate" before this PR (`index.html:118,225`, `architecture.html:205`, `agent-orchestration.html:108`) so no stale HTML survives. `llms.txt:21` says "paired significance gate" — consistent, no edit needed.

**Repo-wide staleness sweep — clean.** `grep -rn "gate_mode"` across `*.py *.yaml *.yml *.md *.json *.html *.txt` → 40 hits, every one consistent with `paired`-as-default. The two pre-existing `significant` values (`how-to/cap-evolve-with-exgentic-tau2.md:77,209`) are now explicitly annotated as deliberate overrides. No surviving instance of "the default is significance" or `significant | strict | threshold` as an exhaustive list. `templates/adapters/*/capevolve.yaml` (5 files) all already ship `paired`. `ci/`, `examples/`, `presentation/` clean.

**Links in changed `.md` — 66 resolved, 0 broken** (PR claimed 67; see N6). Includes the depth-sensitive `INPUTS.md` → `../../../../docs/HONEST_EVAL.md#gate-modes`, verified by slugifying the target's headings.

**Security — clean.** No secrets, keys, tokens, absolute local paths, or internal hostnames in the diff.

OsherElhadad:
## 🔧 Review fixes

Pushed as `fa11f9c`. Every finding verified before acting; two of them were stronger than reported and I say so below. All output literal, from `/tmp/fx-198`.

---

### ⚠️ #195 merge note — read before merging either PR

**Order: #198 (this) → #195.** They merge clean and git will flag nothing, but #195 *implements* what two paragraphs here describe as absent. Both must be rewritten by #195's author:

| File | Line | Text that becomes false |
|---|---|---|
| `docs/HONEST_EVAL.md` | **86** | "The bar is a fixed `k · SE`, with no t-correction and no minimum-val-size guard" |
| `skills/phases/gate/references/concepts.md` | **78** | "a z-style critical value with no t-correction and no minimum-val-size guard" |

`concepts.md` is **not in #195's diff** (`git diff --name-only origin/main...origin/fix/issue-113-small-samples` → 8 files, `concepts.md` absent), so it will rot silently. `HONEST_EVAL.md` is in the diff, but post-merge its line 33 region and line 86 would contradict each other in the same file.

---

### Blocking

**1. B1 — `--paired-deltas` untested, crashed on malformed input, no length validation. FIXED.**

Both defects reproduced. `parse_paired_deltas()` now routes through `p.error` (argparse exit 2) with an actionable message, and cross-checks `mean(deltas)` against `candidate − current` — the only length check possible here, since the script never sees the val split, and it catches truncation, wrong-order pastes and flipped signs at once:

```
$ python run.py --current 0.5 --candidate 0.6 --mode paired --paired-deltas "1,abc,0"
gate: error: --paired-deltas: could not convert string to float: 'abc'. Expected
comma-separated numbers (e.g. 1,0,0,0,0), one per val task; got '1,abc,0'.
exit=2

$ python run.py --current 0.5 --candidate 0.6 --mode paired --paired-deltas "1;0;0"
gate: error: --paired-deltas: could not convert string to float: '1;0;0'. Expected
comma-separated numbers (e.g. 1,0,0,0,0), one per val task; got '1;0;0'.
exit=2

$ python run.py --current 0.5 --candidate 0.6 --mode paired --paired-deltas "1,0"
gate: error: --paired-deltas has 2 value(s) whose mean is +0.5000, but --candidate
minus --current is +0.1000. The deltas must be the per-task cand-curr over ALL val
tasks used for those means (a truncated paste is the usual cause) — otherwise the
decision reports a gain and an n the run never made.
exit=2

$ python run.py --current 0.5 --candidate 0.6 --mode paired --paired-deltas "1,0,0,0,0,0,0,0,0,0"
  "reason": "paired Δ̄=+0.1000 <= 1.0·SE=0.1000 (SE=0.1000, n=10)"
exit=0
```

Tolerance is `1e-4`, not `1e-6`, so means copied at the 4dp the dashboard prints still pass; a truncation error is orders of magnitude larger than that.

On coverage: you were right that `check.py` tested only `significant` and that **nothing in CI touched this script at all**. `check.py` now covers `paired` (the mode every real run uses), the malformed / wrong-length / empty / happy CLI cases, and the fallback disclosure — 11 notes, up from 4:

```
"paired at k_se=1.0 rejects a 1-of-n gain (mean(Δ)=SE(Δ))"
"paired banks >=2 improved tasks at the default k_se"
"paired→significant fallback is disclosed, not silent"
"--paired-deltas refuses non-numeric"
"--paired-deltas refuses semicolons"
"--paired-deltas refuses wrong length (mean 0.5 != candidate-current 0.1)"
"happy path parses; mean(Δ) is cross-checked against candidate-current"
```

And `core/tests/test_documented_cli.py::test_gate_check_py_passes` runs `check.py` under pytest, so CI now reaches it.

**2. B2 — the template contradicted itself. FIXED, and I made the `gate_k_se: 1.0` call explicitly.**

```yaml
gate_mode: paired                        # DEFAULT: per-task paired SE, same tasks both sides — cross-task
                                         # variance cancels, so it is the most powerful test and works at num_trials=1.
# gate_k_se: how many SEs the val gain must clear. 1.0 is deliberately conservative:
# improving exactly ONE of n tasks gives mean(Δ)=SE(Δ) exactly, so it does NOT clear
# 1.0 — at any n, and no matter how large that single gain is. You need >=2 improved
# tasks to bank at 1.0. That is the intended default (one task moving is weak evidence
# it wasn't luck); set 0.2, as the examples do, to bank single-task gains.
gate_k_se: 1.0
```

**Decision on the default: keep `1.0`.** One task out of `n` moving is genuinely weak evidence the move wasn't luck — that is what "mean(Δ) lands exactly on the bar" is telling us, not a bug in the bar. Lowering the shipped default to `0.2` to make single-task gains bankable would make the *scaffolded* gate more permissive than the honest one, which is the wrong direction for a repo whose whole point is refusing to bank noise. The examples that want the lenient bar set it explicitly and now say why. Stated either way, as asked.

---

### The corrected `k_se = 1.0` rule

> **A candidate that improves exactly ONE of `n` val tasks can never be accepted at `k_se = 1.0`.** Fix one task by `m`, break nothing: `Δ̄ = m/n` and `SE(Δ) = m/n` — *exactly equal* — so the strict `>` rejects it.
>
> - **Magnitude-independent.** `m` cancels: fixing one task **perfectly** is rejected identically to nudging it by `0.01`.
> - **`n`-independent.** `n` cancels: a bigger val split does **not** help.
>
> **Rule: improve ≥2 tasks at `k_se = 1.0`, or bank nothing.**

You were right that I understated it twice, and right about the `n=20` row. Confirmed with `Fraction`:

```
n=2  mean=1/2  var/n=1/4    mean²=1/4    equal=True
n=3  mean=1/3  var/n=1/9    mean²=1/9    equal=True
n=10 mean=1/10 var/n=1/100  mean²=1/100  equal=True
n=20 mean=1/20 var/n=1/400  mean²=1/400  equal=True
n=50 mean=1/50 var/n=1/2500 mean²=1/2500 equal=True
```

and magnitude-independence + the float slack, through the live `decide()`:

```
n=10 m=1.0:  accept=False  paired Δ̄=+0.1000 <= 1.0·SE=0.1000
n=10 m=0.01: accept=False  paired Δ̄=+0.0010 <= 1.0·SE=0.0010     ← m cancels
n=20 m=1.0:  accept=True   paired Δ̄=+0.0500 >  1.0·SE=0.0500     ← IEEE-754 slack only
n=50 m=1.0:  accept=False  paired Δ̄=+0.0200 <= 1.0·SE=0.0200     ← n=50 rejects again
2-of-10:     accept=True   paired Δ̄=+0.2000 >  1.0·SE=0.1333     ← the real threshold
```

The misleading `n=20 accept=True` row is **removed** and replaced with an explicit annotation that it is floating-point slack, not a rule. Propagated to all **four** surfaces that said "banks 1-task gains" unqualified (you named three; the template was the fourth): `templates/project/capevolve.yaml`, `examples/skillsbench/{capevolve.yaml,PROMPT.md}`, `examples/tau2_airline/{capevolve.yaml,PROMPT.md}`, `skills/phases/intake/SKILL.md`, `skills/phases/intake/inputs/INPUTS.md`, `skills/phases/gate/SKILL.md`, `skills/phases/gate/references/concepts.md`, `docs/HONEST_EVAL.md`, `docs/OPTIMIZE_YOUR_OWN.md`, `docs/ADAPTER_CONTRACT.md`, `RUN.md`, `site/optimize-your-own.html`.

---

### Non-blocking

**3. N1 — third broken caller in `AGENT_ORCHESTRATION.md:90`. CONFIRMED and FIXED.** It parsed but silently ran `significant`. Reproduced the exact documented line, then fixed it to pass `--paired-deltas` with the fallback spelled out. Broken 8 days, as you noted. `site/agent-orchestration.html:108` says "the paired significance gate (Δ > k·SE)" in prose with no CLI line — accurate, no edit needed.

**4. N2 — silent `paired→significant` fallback. FIXED in `gate.py`.** Now announced in the reason *and* logged as a `gate_warning`, matching `_warn_se_zero`'s posture. The two-step slide you described is now fully visible:

```
$ python run.py --mode paired --k-se 1.0 --current 0.50 --candidate 0.62 --current-stderr 0.03 --candidate-stderr 0.03
  "reason": "paired→significant (no per-task deltas): Δ=+0.1200 > 1.0·SE=0.0424 (SE=0.0424)"

$ python run.py --mode paired --current 0.5 --candidate 0.6
  "reason": "paired→significant (no per-task deltas): Δ=+0.1000 > 0 (SE=0 → STRICT fallback, warned)"
```

Both steps of the slide now appear in one string.

**5. N3 — `simplicity_tiebreak` is dead in practice. DECISION: BOTH — filed as a real bug AND documented the precondition.** Filing alone would leave five surfaces lying until someone picks the issue up; documenting alone would leave a broken mode permanently broken. So: **#206** tracks plumbing `candidate_size`/`current_size` through `harness._gate_and_record` and `gepa._full_val_gate`, and every one of the five surfaces now carries the warning ("⚠️ requires sizes the harness does not populate — **behaves as `strict` today**"), plus `gate.py`'s own docstring. `gate/SKILL.md` also notes that only `core/tests/test_core.py` exercises the branch.

**6. N4 — the 1-of-n finding was under-promoted.** Fixed on all three surfaces you named plus the rest; `HONEST_EVAL`'s `### Why paired banks single-task gains` is renamed **`### The k_se = 1.0 rule: improve ≥2 tasks, or bank nothing`** with the rule in a blockquote.

**7. N5 — "~2-3x smaller" survived in `examples/tau2_airline/capevolve.yaml`. FIXED.** You're right that applying my own editorial decision to one surface and not the other was inconsistent. `grep -rn "2-3x"` → empty.

**8. N6 — link count. CONCEDED.** My resolver over the 15 changed `.md` files now reports `resolved=74 broken=0`. I can't reproduce either 66 or 67; the file set and link-extraction regex clearly differ. Zero broken is the claim that matters and it reproduces on both our tooling.

**9. `threshold` defaults to `0.0`. DOCUMENTED** on every surface that lists the mode (`HONEST_EVAL`, `ADAPTER_CONTRACT`, `OPTIMIZE_YOUR_OWN`, `gate/SKILL.md`, `gate/concepts.md`, `intake/SKILL.md`, the site mirror) and in `gate.py`'s docstring — unset `threshold` is identical to `strict`.

### Nits

**10. n1** — the ungrammatical "while at the `gate_k_se: 0.2` the examples use it accepts" is gone; that whole paragraph was rewritten. **n2** — the `[] or None` idiom is replaced by an explicit `parse_paired_deltas()` with an early `if not fields: return None`. **n3** — documented: `run.py` takes no `--run-dir`, so `decide(run_dir=None)` always holds there and **no `gate_warning` can ever fire from the standalone CLI** — the `reason` string is the only channel. Added as a blockquote in `gate/SKILL.md` so nobody hunts for an event that cannot exist. Agreed the flag isn't worth adding.

---

### 11. The documented-CLI checker — ADDED as `core/tests/test_documented_cli.py`

~60 lines including docstrings, so within the "genuinely ~40 lines" bar. It scans every `python .../scripts/run.py` line in the repo's `.md` files, resolves the script, and validates each flag **and each enumerated value** against the real argparse `--help`. References **#203**.

**Fails on `main` with exactly your 2 findings:**
```
$ cd /tmp/fx198-main && pytest core/tests/test_documented_cli.py -q
E   AssertionError: documented CLI invocations that do not parse:
E     docs/AGENT_ORCHESTRATION.md: --mode 'paired' not in ['significant', 'simplicity_tiebreak', 'strict', 'threshold']
E     skills/algorithms/agent-optimize/SKILL.md: --mode 'paired' not in ['significant', 'simplicity_tiebreak', 'strict', 'threshold']
1 failed
```
**Passes on this branch:**
```
$ cd /tmp/fx-198 && pytest core/tests/test_documented_cli.py -q
2 passed
```
It found one further real hit while I built it (a stale `--candidate-stderr/--current-stderr` in a trailing `#` comment in `REPRODUCE_tau2.md`); since a trailing comment isn't executed, the scanner truncates at ` #` rather than me contorting the prose. `assert scanned > 20` guards the regex silently matching nothing.

---

### Verification

**pytest — 181 passed, 0 failed.** You were right that my "179" wasn't reproducible; the honest baseline is `main` = 179, this branch = 181 (+2 from the new file).
```
$ cd /tmp/fx-198 && PYTHONPATH=/tmp/fx-198/core pytest core/tests -q
181 passed in 63.56s (0:01:03)

$ cd /tmp/fx198-main && pytest core/tests -q      # origin/main, without my new test
179 passed in 61.31s (0:01:01)
```
On **#200**: I could *not* reproduce the `test_dashboard_launch.py` failure even with 7878 genuinely occupied —
```
$ lsof -iTCP:7878 -sTCP:LISTEN
Python  15375 osherelhadad   6u  IPv4 ...  TCP localhost:7878 (LISTEN)     ← occupied
$ pytest core/tests/test_dashboard_launch.py -q
7 passed in 0.02s
```
So it is flakier than "port 7878 is in use" — timing-dependent on top of that. Either way pre-existing and orthogonal; **#200**'s fix (monkeypatch `_free_port`) is still the right one, and its diagnosis should note the port check alone doesn't determine the outcome.

**compileall** → `COMPILEALL_CLEAN`. **manifest** → byte-exact from a fresh `build_manifest.py` (empty `git diff --stat`). **check.py — 12 green** (I ran all phases + algorithms, not just the 6 touched):
```
phases/gate  intake  baseline  evaluate  diagnose  finalize  report  implement-and-check
algorithms/skillopt  hill-climb  gepa  agent-optimize          → all "ok": true
```
**Links** → `files=15 resolved=74 broken=0`.

Thanks — B1's length check and the two ways the `k_se` claim was understated were the most valuable parts of this review, and the `n=20` row would have taught readers the opposite of the truth.

OsherElhadad:
Superseded by #347, which implements the reviewed scope as a fresh minimal change against current main rather than rebasing this branch. Scoped down: the documented gate modes are corrected to match `gate.py`, and `simplicity_tiebreak` is **deleted** as unreachable rather than documented. Closing this in favour of #347.

Issue #143, opened 2026-07-25
Cross-host skill parity: per-host metadata + stdlib-only fallback


OsherElhadad:
## Problem / opportunity

cap-evolve advertises broad host support (Claude Code plugin + ~14 optimizer backends), but the skill packaging is **Claude-Code-centric**: `install.sh` admits several host skill-dir mappings are "best-guess" (`install.sh:38-40`), there's no per-host display metadata, and no explicit **stdlib-only fallback** for environments where the host provides no native tools. So "works on host X" is uneven in practice (also see #126, #121).

## Why it matters

Prior art in this space (Arbor, Apache-2.0 — inspiration only) ships each skill with a cross-platform body **plus per-host metadata** (e.g. a small manifest giving a display name / short description / default prompt for a specific coding-agent UI) **and a stdlib-only tool fallback script** so the suite runs even where native tooling is absent. The result is genuine multi-host parity rather than one first-class host and several "best-guess" ones. cap-evolve's honesty hooks are already core-owned stdlib scripts (good foundation); this extends that discipline to skill *packaging* across hosts.

## What to change

1. Add optional **per-host metadata** to skills (display name / short description / default invocation) for the hosts cap-evolve claims, so each host's UI presents the skill correctly.
2. Provide a **stdlib-only fallback** for any skill step that assumes native host tools, so the suite degrades gracefully in a bare environment (mirror the core-owned hook approach).
3. Turn the "best-guess" host dirs into **verified** placements (or clearly mark unverified, ties into #126) and cover them with an install smoke test per host where feasible.

## How to test

- Installing into each claimed host places skills where that host discovers them (smoke test / documented manual check), with metadata rendered.
- A skill step runs via the stdlib fallback in an environment with no native tools.
- `install.sh` no longer silently "best-guesses" without marking it (consistent with #126).

## Dependencies / parallelization

- Coordinates with **#126** (verified-vs-best-guess host labels) and **#121** (`doctor` checks placement). Packaging/skills work; **parallel-safe** with algorithm/UX issues.

OsherElhadad:
Closing — ask 1 (per-host display metadata) has no consumer anywhere in core or skills, so it is YAGNI; ask 2 (stdlib-only fallback) is already true by construction, since every core import is stdlib or first-party and the single yaml import is function-local and optional. Ask 3 is real and belongs to #208 (CI never runs ./install.sh), where it is ~25 lines.

Full evidence in the close comment on #245.

Issue #100, opened 2026-07-25
Reconcile results across README, docs/RESULTS.md, and the site (Qwen runs live only on the site)


OsherElhadad:
## Problem

Reported results have diverged across three surfaces, and the source-of-truth relationship is now **inverted**.

- `site/results.html:231-307` contains two **Qwen 2.5 14B** runs (`#qwen-tools` +41.2% test, `#qwen-all` +125% test) that **do not exist** in `docs/RESULTS.md` or `README.md`.
- `site/README.md:65-69` explicitly states the site "summarize[s] the source markdown docs, they do not replace them" and instructs readers to "cross-check against `docs/RESULTS.md`."

So the canonical doc (`RESULTS.md`) is now a **subset** of the site, and the site is publishing numbers that have no committed backing in the source-of-truth doc. The "cross-check the canonical doc" instruction is currently false.

## Why it matters

Same brand-critical honesty theme as the README headline issue: the tool sells "every number is artifact-backed and cross-checkable." Right now that guarantee does not hold.

## What to change

1. Make `docs/RESULTS.md` the **single source of truth** for every published number (baseline, optimized, split, model, task/trial counts, commit, cost).
2. For each figure on `site/results.html`, confirm it is present in `RESULTS.md` **and** artifact-backed. Move the two Qwen 14B runs into `RESULTS.md` (with artifacts) or remove them from the site.
3. Add a short CI check or a `scripts/check_results_consistency.py` that greps the headline figures out of `README.md`, `docs/RESULTS.md`, and `site/results.html` and fails if a number appears on one surface but not another.

## How to test

- New consistency script (or a manual `grep`) shows every published percentage/score appears identically on all three surfaces.
- Every `site/results.html` run anchor (`#qwen-tools`, `#qwen-all`, etc.) maps to a `RESULTS.md` section with a committed artifact reference.

## Dependencies / parallelization

- Sibling of **README headline** and **CHANGELOG** honesty issues. **Parallel-safe** with all of Group B (skills/docs drift). Best done alongside the README headline fix.

OsherElhadad:
Closing — the reported bug is gone. Both Qwen sections are in the canonical `docs/RESULTS.md` and all 8 `site/results.html` run anchors map to a heading there, so this issue's acceptance criterion passes on main unchanged. The site is a curated summary of the source docs by design, not a mirror of them.

Full evidence in the close comment on #241.

Issue #200, opened 2026-07-29
test_maybe_launch_spawns_when_available asserts a hard-coded port and fails whenever 7878 is in use


OsherElhadad:
`core/tests/test_dashboard_launch.py::test_maybe_launch_spawns_when_available` fails on any machine where TCP **7878** is already in use — including a developer box running a cap-evolve dashboard from an earlier session.

## The test
`core/tests/test_dashboard_launch.py:55-56`
```python
out = dl.maybe_launch("/runs", mode="auto", port=7878)
assert out["dashboard"] == "http://127.0.0.1:7878"
```

It asserts the **requested** port. But `maybe_launch` treats `port` as a *starting preference* and picks the next free port when it's taken, so on an occupied box it correctly returns `http://127.0.0.1:7882` and the assertion fails. The test is asserting the opposite of the behavior the code is designed to have: **the port-picking fallback is the feature, and this test fails when it works.**

## Impact
Not hypothetical — it produced false-negative signal repeatedly during epic #127. Three independent agents hit it on PRs whose diffs were pure Markdown, each had to stash their changes and re-run on a clean tree to prove innocence. `Popen` is already monkeypatched here, so nothing actually binds a port during the test; the collision comes purely from the real free-port probe inspecting the host.

## Fix options
1. **Assert the contract, not the constant** (preferred) — assert the URL matches `http://127.0.0.1:<int>` and that the chosen port appears in the spawned argv, so the picker's actual guarantee is what's tested:
   ```python
   host, _, port = out["dashboard"].removeprefix("http://").partition(":")
   assert host == "127.0.0.1" and port.isdigit()
   assert port in calls["cmd"]          # the spawn got the port that was reported
   ```
2. **Monkeypatch the free-port probe** to return a deterministic value, making the test hermetic and letting it keep an exact assertion.
3. Add a **separate** test that explicitly covers "requested port busy → a different free port is returned and reported", since that path is currently untested — the bug is that the happy path accidentally depends on it.

(1) + (3) together give better coverage than today with no host dependence.

## Acceptance
- The test passes with something listening on 7878 (verify: bind 7878, run the test).
- The port-fallback path has its own explicit test.
- No test in the suite asserts a hard-coded port number it does not itself control. Worth grepping for other `127.0.0.1:<literal>` assertions while here.

## Provenance
[...]

PR #202, opened 2026-07-29
Onboarding honesty: real toy_calc screenshot (+ fix the dashboard charts that never rendered) + verified/best-guess host labels

### LINKED ISSUE #126: Onboarding polish: fix "not the toy" getting-started screenshot + label verified vs best-guess hosts
## Problem

Two smaller onboarding-credibility papercuts flagged in the docs/site review:

1. **"Getting started" shows a screenshot that isn't what you'll see.** `site/getting-started.html:105` caption admits: "Shown here from a real benchmark run, not the toy." The literal getting-started walkthrough shows a dashboard the `toy_calc` run does not produce.
2. **Advertised host support outruns verification.** README/site advertise ~14 optimizer backends / hosts ("Codex, Gemini, opencode, Cursor, Droid, Copilot, Kimi, Pi, Antigravity…") as first-class paths, but `install.sh:38-40` comments that several host skill-dir mappings are **"best-guess"** and can't be reliably placed without `--dest`. The confidence of the marketing outruns the verification status.

## Why it matters

On a project selling honesty, a "not the toy" screenshot on the getting-started page and unqualified host claims are exactly the small credibility dings a skeptical reader notices. Both are cheap to fix.

## What to change

1. Either generate a **real `toy_calc` dashboard screenshot** for `site/getting-started.html:105`, or move that image to a clearly-labeled "what a real run looks like" aside so the getting-started shot matches what the toy actually produces.
2. On every surface that lists hosts/backends (README "Choose your path", site, INSTALL), **label verified vs best-guess** hosts (e.g. a ✅/➖ column), matching the honesty already shown in `install.sh:38-40`. Verified = we've confirmed the skill-dir placement + a headless command; best-guess = needs `--dest`.

## How to test

- The getting-started screenshot is produced by (or clearly labeled as not from) the toy run.
- Host lists on all surfaces carry a verified/best-guess distinction consistent with `install.sh`.

- **Parallel-safe.** Docs/site only. Small; good first issue. Pairs with the `doctor` and PyPI issues (install-friendliness cluster).

OsherElhadad:
Closes #126.

Two onboarding-credibility papercuts from the docs/site review. Both are honesty fixes: in each case the artifact was making a stronger claim than the evidence supported.

## 1. The getting-started screenshot showed something you won't see

The old caption admitted it in writing: *"Shown here from a real benchmark run, not the toy."* The page tells you to run `bash examples/toy_calc/run.sh`, then shows you a τ²-bench dashboard.

**Why nobody had ever screenshotted the toy: the dashboard's charts have never rendered — for any run, since the dashboard was introduced (4c87ed1).** `ParentNode.append()` returns `undefined`, so every

```js
el.append(svg('text', {...})).textContent = v      // TypeError
```

threw and aborted the whole chart IIFE. Chrome, on the real toy `dashboard.html`:

```
[INFO:CONSOLE:180] "Uncaught TypeError: Cannot set properties of undefined (setting 'textContent')"
```

It went unnoticed because the panel `<h2>` titles are emitted *before* the throw, so the existing test's `assert "Score over iterations" in text` passed against a permanently empty panel.

Fixed once in the shared `svg()` helper (a `text:` pseudo-attribute) rather than at each of the seven call sites — same root cause, smaller diff, and no sibling call site left broken. The regression test now asserts on the *chaining shape* that throws, so this cannot come back:

```python
assert not re.search(r"\.append\(svg\(.*\)\)\s*\.\s*\w", text), \
    "chaining off .append(svg(...)) throws: append() returns undefined"
```

### Old vs new

Both images are in this commit, so this comparison is checkable.

**Before** — `site/assets/dash-overview.png`, a τ²-bench run presented as the outcome of the toy walkthrough:

**After** — `site/assets/toy-calc-dashboard.png`, a headless screenshot of the `dashboard.html` that `bash examples/toy_calc/run.sh` actually wrote:

This is a **real screenshot**, not a transcript fallback. It is deliberately modest, and that is the point: `$0.0000` and `0` tokens because the toy calls no model; exactly 2 tasks (`a1`, `a4`) and 4 candidates because that is all `toy_calc` has; `0.6s` wall clock. The cumulative-best stair and the per-task heatmap are visible here **for the first time**.

`docs/GETTING_STARTED.md` additionally now pastes the **literal, unedited** run output in place of the old hand-written one-liner summary, and explains the two lines that look like failures but aren't — `"dashboard": "skipped"` (that's the *optional live server*, not the dashboard file) and `test_pass_k {"2": 0.0}` (with two test tasks, pass^2 needs both to pass on both trials).

The README hero shot **stays**: it is a genuine τ² run and was already labelled as one. It now also points at Getting started for what the quickstart really produces.

## 2. Host support advertised more than we've verified

README/site/INSTALL/RUN listed ~14 hosts as first-class, while `install.sh` privately conceded several skill-dir mappings were "best-guess". New **`docs/HOST_SUPPORT.md`** is the single source of truth, with three honest tiers:

| Badge | Means | Bar |
|---|---|---|
| ✅ verified | Executed in CI | A CI job or committed run record that **actually invokes it** |
| 🟡 docs-checked | Flags/dirs from vendor docs; **never run here** | A `references/<host>.md` |
| ➖ best-guess | Skill dir inferred from dotdir convention | Nothing — pass `--dest` |

**Result: 2 verified, 4 docs-checked, 7 best-guess** (+ `generic`, n/a).

| Optimizer | Status | Provenance |
|---|---|---|
| `claude-code` | ✅ verified | `optimizer_skill: claude-code` at [`ci/benchmarks/lib/run_suite.sh:135`](ci/benchmarks/lib/run_suite.sh), executed by [`integration-tests.yml`](.github/workflows/integration-tests.yml) (tau2 task 9, real models) and [`benchmarks.yml`](.github/workflows/benchmarks.yml); also the optimizer behind committed run records [`examples/skillsbench/run_full/`](examples/skillsbench/run_full/) + [`examples/tau2_airline/run_full/`](examples/tau2_airline/run_full/) |
| `mock` | ✅ verified | `toy-example` job in [`ci.yml`](.github/workflows/ci.yml) runs `examples/toy_calc/run.sh` and asserts `baseline_val 0.0 → test_reward 1.0` on every push |
| `codex` | 🟡 docs-checked | `references/codex.md` only — no CI job, no run record |
| `gemini-cli` | 🟡 docs-checked | `references/gemini-cli.md` |
| `opencode` | 🟡 docs-checked | `references/opencode.md` |
| `ibm-bob` | 🟡 docs-checked | `references/ibm-bob.md` calls `--chat-mode code` "likely-but-unverified". Bob automating this repo's issue/PR workflows is **not** evidence it works as a cap-evolve optimizer |
| `cursor`, `droid`, `copilot`, `kimi`, `pi`, `antigravity`, `openclaw` | ➖ best-guess | Named best-guess in `install.sh`; `kimi`'s API-key env var and `antigravity`'s headless flags are self-described unverified; `openclaw`'s skill mechanism is "UNVERIFIED" and its `skills_dir` deliberately blank |

Deliberately **not** upgraded: `ibm-bob` (CI presence is for labelling/PR-fixing, not optimizing) and `codex`/`gemini` (their appearances in `benchmarks.yml` are *model* names like `azure/gpt-5.3-codex`, not host invocations). No row moved to ✅ without a linkable artifact.

README, RUN.md, `docs/INSTALL.md`, `install.sh` and the site now **link** to that page instead of restating the list, so the claim can only drift in one file.

## Coordination

- **#143 (cross-host skill parity)** — `docs/HOST_SUPPORT.md` is intended as exactly the base #143 builds on: one canonical table, one badge vocabulary, a documented "Promoting a host to ✅" bar, and a per-host capability-parity row (today only `claude-code` has `parallel: "true"`, and the registry says to keep it that way until verified). #143 should extend this file, not fork a second list.
- **#123 (site chrome)** — touches `site/getting-started.html` only in the `<figure>` and one `<p>` in `index.html`; no CSS/nav/layout, so it should not conflict.
- **#100 (results provenance)** — same instinct one layer down: this PR labels *screenshot* and *host* provenance, #100 labels *results* provenance.

## Verification

```
$ PYTHONPATH=/tmp/wt-126/core python -m pytest core/tests -q
178 passed, 1 failed
```

The single failure is **pre-existing and environment-only** — `test_dashboard_launch.py::test_maybe_launch_spawns_when_available` hardcodes `assert out["dashboard"] == "http://127.0.0.1:7878"`, but `maybe_launch` probes for a free port, and this box had 7878 occupied by leftover dashboard servers. Proof it is not from this diff — same test, same box, with `core/` reset to unmodified `origin/main`:

```
$ git checkout origin/main -- core/ && python -m pytest core/tests/test_dashboard_launch.py::test_maybe_launch_spawns_when_available -q
E       AssertionError: assert 'http://127.0.0.1:7892' == 'http://127.0.0.1:7878'
1 failed
```

On a clean port it is the documented **179 passed**. (Worth its own issue: the test should assert on the returned port, not a literal.)

The regression test genuinely catches the bug — it fails on the old code and passes on the fix:

```
$ git stash push core/cap_evolve/dashboard.py     # revert only the fix
$ python -m pytest core/tests/test_dashboard.py::test_render_html_self_contained_and_parseable -q
1 failed
$ git stash pop
$ python -m pytest core/tests/test_dashboard.py -q
10 passed
```

Charts render, zero console errors (before: 1 SVG + TypeError; after: 2 SVGs, clean):

```
$ chrome --headless --dump-dom file://$D/.capevolve/run_demo/dashboard.html 2>&1 | grep INFO:CONSOLE
(no output)
$ grep -c '<svg' dom.html
2
```

Links — the exact command from the `docs-links` CI job:

```
$ lychee --offline --no-progress --include-fragments \
    --exclude-path examples/tau2_airline/run_full --exclude-path examples/skillsbench/run_full \
    "README.md" "docs/**/*.md" "examples/**/*.md" "*.md"
🔍 264 Total 🔗 115 Unique ✅ 221 OK 🚫 0 Errors 👻 43 Excluded
```

Every local `src`/`href` in `site/*.html` resolves on disk, and `compileall` + `bash -n install.sh` are clean. Full commands and output in the 🔬 Evidence comment below.

## Files touched

| File | Change |
|---|---|
| `core/cap_evolve/dashboard.py` | Root-cause fix: `text:` pseudo-attribute in `svg()`; 7 call sites unchained |
| `core/tests/test_dashboard.py` | Regression test for the throwing chain shape |
| `docs/HOST_SUPPORT.md` | **New** — single source of truth for host verification |
| `site/assets/toy-calc-dashboard.png` | **New** — real toy_calc dashboard screenshot |
| `site/getting-started.html` | Real screenshot + honest caption |
| `docs/GETTING_STARTED.md` | Literal run output + real screenshot |
| `README.md` | Host-verification note; hero-shot caption clarified |
| `RUN.md`, `docs/INSTALL.md`, `install.sh` | Link to HOST_SUPPORT.md |
| `site/index.html`, `site/optimize-your-own.html` | Host-verification note |


OsherElhadad:
## 🔬 Evidence

All commands run in a clean worktree off `origin/main` (`6fca097`), branch `fix/issue-126-onboarding-polish`. Literal output, unedited.

### 1. Baseline before any change — 179 passed

```
$ cd /tmp/wt-126 && PYTHONPATH=/tmp/wt-126/core python -m pytest core/tests -q
179 passed in 63.13s (0:01:03)
```

### 2. The run that produced the screenshot (reproducible, zero API cost)

```
$ cd /tmp/wt-126 && bash examples/toy_calc/run.sh
Working directory: /var/folders/zh/srgnbq_97qvb6002zsr40tgc0000gn/T/toy_calc.XXXXXX.9t5qIAzQje
{
  "best_id": "cand_0001",
  "baseline_val": 0.0,
  "test_reward": 1.0,
  "test_baseline_reward": 0.0,
  "test_delta": 1.0,
  "test_pass_k": {
    "1": 1.0,
    "2": 0.0
  },
  "iterations": 3,
  "dashboard_server": "skipped"
}
```

Re-run after all edits — byte-identical metrics (the server line differs only because the optional dashboard backend got installed on the box mid-session):

```
Working directory: /var/folders/.../toy_calc.XXXXXX.kqo0daF52i
{
  "best_id": "cand_0001",
  "baseline_val": 0.0,
  "test_reward": 1.0,
  "test_baseline_reward": 0.0,
  "test_delta": 1.0,
  "iterations": 3,
  "dashboard_server": "http://127.0.0.1:7884"
}
```

The screenshot was taken from that run's own `dashboard.html`:

```
$ D=/var/folders/.../toy_calc.XXXXXX.9t5qIAzQje
$ "/Applications/Google Chrome.app/Contents/MacOS/Google Chrome" --headless --disable-gpu \
    --hide-scrollbars --window-size=1280,1560 --force-device-scale-factor=2 \
    --screenshot=/tmp/toy-final.png --virtual-time-budget=12000 \
    "file://$D/.capevolve/run_demo/dashboard.html"
264656 bytes written to file /tmp/toy-final.png
$ python -c "import struct;d=open('/tmp/toy-final.png','rb').read();print(struct.unpack('>II',d[16:24]))"
(2560, 3120)
$ cp /tmp/toy-final.png site/assets/toy-calc-dashboard.png
```

### 3. The chart bug — before / after

The run data was always present; only the JS was broken:

```
$ python -c "... json.loads(re.search(r'id=\"run-data\">(.*?)</script>', html, re.S).group(1))"
[{"id":"seed","iteration":0,"val":0.0,"best_so_far":0.0,"status":"seed"},
 {"id":"cand_0001","iteration":1,"val":1.0,"best_so_far":1.0,"status":"accepted"},
 {"id":"cand_0002","iteration":2,"val":1.0,"best_so_far":1.0,"status":"rejected"},
 {"id":"cand_0003","iteration":3,"val":1.0,"best_so_far":1.0,"status":"rejected"}]
```

**BEFORE** (`core/cap_evolve/dashboard.py` reverted to `origin/main`):

```
$ chrome --headless --enable-logging=stderr --v=1 --dump-dom "file://$OD/.../dashboard.html" \
    2>&1 >/tmp/dom_old.html | grep 'INFO:CONSOLE'
[INFO:CONSOLE:180] "Uncaught TypeError: Cannot set properties of undefined (setting 'textContent')", source: .../dashboard.html (180)
OLD svg count: 1
```

**AFTER** (the fix):

```
$ chrome --headless --enable-logging=stderr --v=1 --dump-dom "file://$D/.../dashboard.html" \
    2>&1 >/tmp/dom3.html | grep 'INFO:CONSOLE' || echo "ZERO console errors"
ZERO console errors
$ grep -c '<svg' /tmp/dom3.html
2
$ grep -o 'Per-task pass/fail' /tmp/dom3.html | head -1
Per-task pass/fail
```

1 SVG (the logo only) + a TypeError → 2 SVGs and a clean console. The stair chart and heatmap in the PR screenshot are rendering for the first time.

Regression age — broken since the dashboard landed, never in a release that worked:

```
$ git log --oneline -L 985,985:core/cap_evolve/dashboard.py | head -3
4c87ed1 v2 W4+W5: rich observability dashboard + Claude Code plugin
+    el.append(svg('text',{...,'text-content':''})).textContent=v.toFixed(2);}
```

### 4. The regression test actually catches it

Fails on old code, passes on the fix:

```
$ git stash push core/cap_evolve/dashboard.py -q      # revert ONLY the fix
$ python -m pytest core/tests/test_dashboard.py::test_render_html_self_contained_and_parseable -q
core/tests/test_dashboard.py:187: AssertionError
FAILED core/tests/test_dashboard.py::test_render_html_self_contained_and_parseable
1 failed in 0.04s

$ git stash pop -q
$ python -m pytest core/tests/test_dashboard.py -q
..........                                                               [100%]
10 passed in 0.07s
```

### 5. Full suite

```
$ PYTHONPATH=/tmp/wt-126/core python -m pytest core/tests -q
179 passed in 60.82s (0:01:00)
```

A later re-run on the same box showed `178 passed, 1 failed`. That failure is **pre-existing and environment-only**, not from this diff — `test_maybe_launch_spawns_when_available` hardcodes port 7878 while `maybe_launch` probes for a free one, and leftover dashboard servers had taken it:

```
$ lsof -nP -iTCP -sTCP:LISTEN | grep -E '78[7-9][0-9]'
Python ... 127.0.0.1:7879 (LISTEN)
Python ... 127.0.0.1:7880 (LISTEN)
...   (7879-7891 occupied)

$ git checkout origin/main -- core/     # my changes fully absent
$ python -m pytest core/tests/test_dashboard_launch.py::test_maybe_launch_spawns_when_available -q
E       AssertionError: assert 'http://127.0.0.1:7892' == 'http://127.0.0.1:7878'
1 failed in 0.03s
$ git checkout HEAD -- core/
```

Same failure with the diff reverted ⇒ unrelated. Worth its own issue: assert on the returned port, not a literal.

```
$ python -m compileall -q core skills && echo OK
compileall OK
$ bash -n install.sh && echo OK
install.sh syntax OK
```

### 6. Links resolve (the CI `docs-links` job's exact command)

Caught a real break mid-work — I had removed `dash-overview.png` as orphaned, but README uses it:

```
$ lychee --offline --no-progress --include-fragments \
    --exclude-path examples/tau2_airline/run_full --exclude-path examples/skillsbench/run_full \
    "README.md" "docs/**/*.md" "examples/**/*.md" "*.md"
[README.md]:
[ERROR] file:///private/tmp/wt-126/site/assets/dash-overview.png (at 27:13) | File not found.
🔍 262 Total 🔗 115 Unique ✅ 217 OK 🚫 2 Errors 👻 43 Excluded
```

Restored it (it is a genuine τ² run, already labelled as such) and re-ran — clean:

```
$ lychee --offline --no-progress --include-fragments \
    --exclude-path examples/tau2_airline/run_full --exclude-path examples/skillsbench/run_full \
    "README.md" "docs/**/*.md" "examples/**/*.md" "*.md"
🔍 264 Total (in 31ms) 🔗 115 Unique ✅ 221 OK 🚫 0 Errors 👻 43 Excluded
```

Site HTML isn't covered by lychee (it globs `*.md`), so every local `src`/`href` was checked directly:

```
$ cd site && for f in *.html; do for r in $(grep -oE '(src|href)="[^"#]+"' "$f" \
    | sed -E 's/.*="([^"]*)"/\1/;s/\?.*$//' | grep -vE '^(https?:|mailto:|//)'); do
    [ -e "$r" ] || echo "MISSING in $f: $r"; done; done
ALL site local refs resolve on disk (query strings stripped)
```

Both PR-embedded images resolve from the pushed commit:

```
$ curl -s -o /dev/null -w "%{http_code}\n" .../site/assets/toy-calc-dashboard.png
200
$ curl -s -o /dev/null -w "%{http_code}\n" .../site/assets/dash-overview.png
200
```

### 7. Host verification — the evidence behind each ✅

Only two hosts clear the bar. `claude-code`:

```
$ grep -n "optimizer_skill" ci/benchmarks/lib/run_suite.sh
135:optimizer_skill:    claude-code
```

...invoked by `integration-tests.yml` (real models, real $):

```
$ sed -n '46,58p' .github/workflows/integration-tests.yml
      - name: Run tau2 task-9 end-to-end (agent gpt-oss-120b, optimizer claude-opus-4-8)
        run: |
          out="$(bash ci/benchmarks/lib/run_suite.sh tau2 ...)"
          python3 ci/benchmarks/lib/assert_run.py "$run_dir" --min-iterations 1
```

...and behind the committed run records:

```
$ grep -n "optimizer_skill" examples/*/capevolve*.yaml
examples/skillsbench/capevolve.yaml:13:optimizer_skill: claude-code
examples/tau2_airline/capevolve.yaml:16:optimizer_skill: claude-code
examples/tau2_airline/capevolve.itest.yaml:11:optimizer_skill: claude-code
$ head -8 examples/skillsbench/run_full/report.md
- Best candidate: `cand_0004`
- Baseline val: 0.33333333333333337
- **Held-out test (optimized skills): 0.6666666666666666**
```

`mock`:

```
$ sed -n '81,89p' .github/workflows/ci.yml
      - name: Run toy_calc and assert honest improvement
        run: |
          out="$(bash examples/toy_calc/run.sh)"
          echo "$out" | grep -q '"baseline_val": 0.0' \
            && echo "$out" | grep -q '"test_reward": 1.0' \
            || { echo "::error::toy_calc did not reach baseline_val 0.0 -> test_reward 1.0"; exit 1; }
$ grep -n "optimizer_skill" templates/project/capevolve.yaml
13:optimizer_skill: mock
```

**Everything else has no such artifact.** No CI workflow invokes any other host as an optimizer:

```
$ for h in opencode openclaw cursor droid copilot kimi antigravity; do
    echo "$h -> $(grep -rilE "\b$h" .github/workflows/ | tr '\n' ' ')"; done
opencode ->
openclaw ->
cursor ->
droid ->
copilot ->
kimi ->
antigravity ->
```

`codex` and `gemini` *do* appear in `benchmarks.yml`, but as **model ids**, not hosts — so they stay 🟡:

```
$ grep -n -iE 'codex|gemini' .github/workflows/benchmarks.yml | head -4
66:          - "Azure/gpt-5.1-codex-2025-11-13"
67:          - "azure/gpt-5.3-codex"
74:          - "GCP/gemini-2.0-flash"
75:          - "gcp/gemini-2.5-flash"
```

No test shells out to any host either (the one `cursor` hit is an unrelated pagination variable):

```
$ grep -rn "codex\|gemini-cli\|opencode\|cursor\|droid\|copilot\|kimi\|antigravity\|openclaw" core/tests/
core/tests/test_resume.py:138:    pool, lineage, accepts, merges, cursor, offset = gepa._reconstruct_gepa(
```

`ibm-bob` appears in CI only for issue/PR automation, which is **not** evidence of optimizer use:

```
$ grep -n "bob" .github/workflows/fix-issues-with-bob.yml
14:    uses: skillberry-ai/skillberry-common/.github/workflows/fix-issues-with-bob.yml@main
```

The ➖/🟡 tiers also match what the repo already admits about itself:

```
$ grep -n -iE 'unverified|best.guess' skills/optimizers/run-optimizer/references/*.md install.sh
install.sh:38:    # cursor/droid/copilot/kimi/pi/antigravity dirs follow each tool's dotdir
install.sh:39:    # convention (best-guess) — override with --dest or $CAPEVOLVE_SKILLS_DIR if
references/antigravity.md:9:# best-guess for current builds — verify with `agy --help` before relying on it:
references/ibm-bob.md:23:  Advanced-mode only — treat as likely-but-unverified under `--chat-mode code`.)
references/kimi.md:19:  registry as a convenience but is **unverified** against current docs — prefer
references/openclaw.md:16:OpenClaw's native skill/instruction mechanism is **UNVERIFIED**. Web evidence is thin
```

**Tally: 2 verified (`claude-code`, `mock`) · 4 docs-checked (`codex`, `gemini-cli`, `opencode`, `ibm-bob`) · 7 best-guess (`cursor`, `droid`, `copilot`, `kimi`, `pi`, `antigravity`, `openclaw`) · 1 n/a (`generic`).**

### 8. Commit provenance

```
$ git log --format='%H%n%an <%ae>%n%s' -1
1c43f4639e98b3e5ae5c6ce4c762236f4fbe81ea
Onboarding honesty: real toy_calc screenshot + verified/best-guess host labels
$ git log -1 --format='%B' | grep -ci 'co-authored-by\|generated with' || echo "0 — clean"
0 — clean
```

OsherElhadad:
## 🔍 Review — PR #202

I re-ran every claim in the evidence comment. All of them reproduce, including the headline one — and the chart bug is **worse** than the PR states. Two documentation inaccuracies to fix; nothing structural.

---

### Blocking

None.

---

### Non-blocking

**1. `docs/GETTING_STARTED.md:60-61` — the `test_pass_k` explanation is factually wrong.**

The new copy says:

> `test_pass_k` `"2": 0.0` is honest too: with two test tasks, pass^2 requires both to pass on both trials.

`pass^k` has nothing to do with the number of *tasks*. `core/cap_evolve/loop.py:75` computes it **per task over that task's trials**, then means across tasks:

```python
rel = [stats.pass_k(s.trial_rewards or [s.reward], k) for s in scores if (s.trial_rewards or [s.reward])]
```

and `core/cap_evolve/stats.py:65-66` returns `0.0` whenever `k > n`:

```python
if k <= 0 or k > n:
    return 0.0
```

The real reason is `num_trials: 1` (`templates/project/capevolve.yaml:66`) — with one trial per task, `k=2` exceeds `n=1`, so every task scores 0 and the mean is 0. Verified on the actual artifact and directly:

```
$ /tmp/ce-venv/bin/python -c "from cap_evolve.stats import pass_k; print(pass_k([1.0],2), pass_k([1.0,1.0],2))"
0.0 1.0
```

Two passing trials → `1.0`. The number of tasks is irrelevant. This is a "getting started" page on an honesty-first project explaining a statistic incorrectly — the one place it's least affordable.

**Fix:** replace with the true cause, e.g. "`test_pass_k` `\"2\": 0.0` is expected: the toy runs `num_trials: 1`, and pass^2 needs two trials per task to estimate — with `k > n` the estimator honestly returns 0 rather than guessing. A run with `num_trials: 2+` populates it."

**2. `docs/GETTING_STARTED.md:37-38` — "literal, unedited output" is conditionally false, and the pasted block is the *degraded* branch.**

The pasted output hard-codes:

```
...
  "dashboard_server": "skipped"
```

My reproduction on a box where the optional backend *is* installed:

```
$ cd /tmp/rv-202 && bash examples/toy_calc/run.sh
Working directory: /var/folders/zh/.../toy_calc.XXXXXX.9cKxRQ7oBb
{
  ...
  "dashboard_server": "http://127.0.0.1:7899"
}
```

Every metric line matches; only the two dashboard-server lines differ. The PR's own evidence comment shows exactly this divergence in its second run ("the server line differs only because the optional dashboard backend got installed on the box mid-session") — but the doc still presents one branch as *the* literal output and then spends a sentence explaining `"skipped"` as though it were universal. A reader who did `pip install -e dashboard/backend` (which `docs/INSTALL.md` recommends) sees different text under a heading promising it's literal.

**Fix:** either add one clause — "`\"dashboard\": \"skipped\"` when the optional live-server backend isn't installed; you'll see a `http://127.0.0.1:...` URL instead if it is" — or drop the word "unedited" and say "output from a run without the optional dashboard backend".

**3. `llms.txt:39-41` — the machine-readable surface still carries an unqualified 14-backend list and never links `HOST_SUPPORT.md`.**

```
- optimizers: run-optimizer + optimizers/registry.yaml (claude-code, codex, gemini-cli,
  opencode, cursor, droid, copilot, kimi, pi, antigravity, openclaw, ibm-bob, generic,
  mock — one runner resolves the name; the agent set matches obra/superpowers)
```

The doc index at `llms.txt:26-32` lists ADAPTER_CONTRACT / HONEST_EVAL / EXTENDING / ROADMAP but not the new page. Grep of every host-listing surface:

```
$ grep -rln -iE 'Codex, Gemini|opencode.*Cursor|Kimi.*Pi.*Antigravity|codex \| gemini' \
    README.md RUN.md docs/ site/ llms.txt install.sh CHANGELOG.md
README.md  RUN.md  docs/OPTIMIZE_YOUR_OWN.md  docs/INSTALL.md  docs/ROADMAP.md
docs/specs/2026-06-24-multi-agent-support-design.md  docs/how-to/cap-evolve-with-exgentic-tau2.md
site/index.html  site/optimize-your-own.html  install.sh  CHANGELOG.md  llms.txt

$ grep -rln 'HOST_SUPPORT' README.md RUN.md docs/ site/ llms.txt install.sh
README.md  RUN.md  docs/INSTALL.md  site/index.html  site/optimize-your-own.html  install.sh
```

`llms.txt` is what an LLM reads to describe this project. It's the surface most likely to reproduce the exact overclaim #126 was filed about, and it's the cheapest possible fix (one em-dash clause + one index line). `docs/OPTIMIZE_YOUR_OWN.md:90,144` has the same bare enumeration; ROADMAP/specs/how-to are historical and fine to leave.

**Fix:** add `HOST_SUPPORT.md` to the `llms.txt` doc index and one clause to line 41 ("only `claude-code` and `mock` are CI-executed — see `docs/HOST_SUPPORT.md`"). Same clause in `OPTIMIZE_YOUR_OWN.md:144`, which already has the ← swap comment where it would fit.

**4. `docs/HOST_SUPPORT.md:62` — the `claude` / `claude-code` install-*destination* row is graded ✅ verified, but nothing executes `install.sh`.**

The table's own bar (line 17) is "A CI job or committed run record that **actually invokes it**." For the *optimizer* table that bar is met. For the *destination* table it isn't — nothing runs `install.sh`:

```
$ grep -rn 'install\.sh' .github/ ci/ 2>/dev/null
(no output)
$ grep -rln 'install.sh' core/tests/
(no output)
$ grep -rn 'claude/skills\|CAPEVOLVE_SKILLS_DIR' .github/workflows/ ci/benchmarks/lib/*.sh
ci/benchmarks/lib/run_suite.sh:37:export CAPEVOLVE_SKILLS_DIR="$REPO/skills"
```

CI sets `CAPEVOLVE_SKILLS_DIR` to the repo tree, i.e. it deliberately takes the *first* precedence branch and never exercises the `--host` mapping at all. So `~/.claude/skills` is verified in the ordinary sense (it's the documented Claude Code path, and `core/tests/test_native_skills.py:75` asserts the string) but not by the bar this page sets for itself, one screen earlier. That's the exact species of overclaim the page exists to eliminate — mild, but it's the page whose whole value is that its ✅ means one specific thing.

**Fix:** either footnote the destination table ("✅ here means the vendor-documented canonical path, not a CI-executed install — `install.sh` is not exercised in CI") or downgrade that one row to 🟡 with the same note. Prefer the footnote; it's honest and keeps the useful signal.

---

### Nits

**5. `core/tests/test_dashboard.py:186` — the regression guard is narrower than its own comment.** Detail in *The chart bug* below.

**6. No CHANGELOG entry for a shipped user-visible regression.** Detail below.

**7. `docs/HOST_SUPPORT.md:57` vs `skills/optimizers/registry.yaml:109`** — the page says gemini's global path is `~/.gemini/extensions/cap-evolve/skills` while the registry's `skills_dir` is `.gemini/skills`. The page *does* explain this ("skills live inside an extension", "the global install path differs from the per-workdir one"), so it's disclosed, not wrong — noting only because a reader diffing the two files will trip on it and #143 will have to encode the distinction.

---

### The chart bug

**Did it verify? Yes — TRUE, and understated.**

Built a dashboard from `origin/main` (`6fca097`) via a real toy run, loaded it in headless Chrome:

```
$ OD=/var/folders/zh/.../toy_calc.XXXXXX.xpvXtWC0OT
$ chrome --headless --disable-gpu --enable-logging=stderr --v=1 --dump-dom \
    "file://$OD/.capevolve/run_demo/dashboard.html" 2>/tmp/err_old.txt >/tmp/dom_old.html
$ grep 'INFO:CONSOLE' /tmp/err_old.txt
[INFO:CONSOLE:180] "Uncaught TypeError: Cannot set properties of undefined (setting 'textContent')",
  source: .../dashboard.html (180)
```

Same procedure on this branch:

```
$ grep 'INFO:CONSOLE' /tmp/err_new.txt || echo "ZERO console errors"
ZERO console errors
```

**The PR undersells the blast radius.** It says the charts died and the `<h2>` titles survived. In fact the throw is at top level of a single inline `<script>` (only 2 script tags in the file; the first is the JSON island), so it aborts **every remaining IIFE**, not just the chart ones. Rendered panel titles:

```
$ grep -o '<h2>[^<]*</h2>' /tmp/dom_old.html        # origin/main
Summary · Narrative · Pipeline phases · What not to try — dead ends
Score over iterations — cumulative best                              (5 panels)

$ grep -o '<h2>[^<]*</h2>' /tmp/dom_new.html        # this PR
… the same 5, plus:
Per-task pass/fail across iterations · Diff vs parent · Lineage
Cost · tokens · latency (optimizer vs runner) · Cost & time per iteration
Evaluations · Annotations & diagnoses · Candidates                   (13 panels)
```

Eight panels were missing — heatmap, lineage, diffs, cost, evaluations, candidates, annotations. Element counts:

| | origin/main | this PR |
|---|---|---|
| `<svg>` | 1 (logo only) | 5 |
| `<path>` | 1 | 6 |
| `<rect>` | 2 | 19 |
| `<text>` | **0** | 16 |
| `<circle>` | 4 | 14 |

Zero SVG `<text>` on main is the signature. Age confirmed: introduced in `4c87ed1` (2026-06-17), i.e. the dashboard has never worked past panel 5 in any release. That makes the existing test's "core panels present" assertion — which checks `Per-task pass/fail`, `Lineage`, `Candidates`, `Annotations` — pass on the *source string* while all four panels were absent from the *rendered page*. Good catch, and the diagnosis is exactly right.

**Is the fix complete? Yes — no sibling instances remain.**

```
$ grep -rnE '\.append\([^;]*\)\s*\.\s*[a-zA-Z]' --include='*.html' --include='*.js' \
    --include='*.jsx' --include='*.ts' --include='*.tsx' . | grep -vE 'dist/|node_modules|presentation/reveal|run_full/ui'
(no output)

$ grep -rnE '\.append\(.*\)\.' dashboard/backend/capevolve_dashboard/*.py \
    skills/algorithms/evograph/dashboard/backend/app.py
(no output)

$ grep -rn '\.append(' dashboard/frontend/src skills/algorithms/evograph/dashboard/frontend/src
(no output — the SPA is React; evograph ships dist only, no src)
```

All 7 sites in `dashboard.py` converted. Notably `dashboard.py:1132` (main) had a **second** latent throw — `el.append(svg('rect',…)).addEventListener('mousemove',()=>{})`, a dead no-op — and the PR **deletes** it rather than converting it. Correct call: it did nothing, the real handler is on `bar` two lines down.

The `text:` pseudo-attribute is the right shape: it mirrors the existing `$()` helper at `dashboard.py:884-885`, which has handled `text`/`html` this way since the file was written. Reuse over invention, one shared fix rather than seven call-site guards, and no real SVG attribute is named `text` so there's no collision.

**Would a future recurrence be caught? Only this one call pattern — no.**

`core/tests/test_dashboard.py:186`:

```python
assert not re.search(r"\.append\(svg\(.*\)\)\s*\.\s*\w", text), \
    "chaining off .append(svg(...)) throws: append() returns undefined"
```

I checked it against the historical string and four plausible future forms:

```
catches old svg pattern:                      True
chained on $() helper                      -> caught: False
chained on a variable                      -> caught: False
multiline svg( arg                         -> caught: False
appendChild chain (works — returns node)   -> caught: False   [correctly ignored]
```

It fires on the exact regression and nothing else. It is genuinely a *fails-on-old-code* test (the PR's stash-and-run demonstration reproduces), which beats most regression tests — but it's a string grep over generated JS, not a rendering assertion, and the comment above it promises more than the regex delivers.

The deeper point the PR itself raises is unaddressed: **the pre-existing test passed while 8 of 13 panels were absent**, because it asserts panel titles against the HTML *source*, where they always appear. Nothing in this PR changes that. A future breakage in the heatmap IIFE — a typo'd variable, a bad property, a helper rename — leaves both the old assertions and the new regex green.

The lazy strong version, given a browser is already a hard dependency of nothing here: assert on the *reduced-data → source* invariant instead. One assertion that would have caught this class:

```python
# every panel the run's data warrants must emit a chart-bearing IIFE that
# survives to the end of the script — assert the script parses as a unit
assert text.count("(function(){") == text.count("})();"), "unbalanced IIFE"
```

That's weak. The assertion with teeth needs a DOM. If `pytest` gaining a headless browser is unacceptable (reasonable), the honest alternative is a `# ponytail:`-style note on the test saying "source-level only; a rendering failure inside a panel IIFE is NOT covered — see #126" so the next person doesn't trust it further than it goes. **Recommend at minimum: soften the comment to match the regex's actual reach, and file the follow-up.** Not blocking — this PR leaves testing strictly better than it found it.

**Does it deserve its own issue/CHANGELOG entry? Yes — CHANGELOG, firmly.**

Separate issue: no. The fix is 20 lines, it's already reviewed here with evidence, and splitting it costs a rebase for zero gain. Merge it in this PR.

CHANGELOG: yes, and its absence is the one process gap. This is a user-visible regression that shipped in `4c87ed1` and silently broke the majority of the flagship dashboard for six weeks. `CHANGELOG.md` has an active `## [Unreleased] / ### Fixed` section carrying entries considerably less user-facing than this one (CI runner robustness, stdin draining). A user who noticed their dashboard was half-empty and gave up has no way to learn it's fixed. `git diff --stat origin/main...HEAD -- CHANGELOG.md` is empty.

**Fix:** one bullet under `### Fixed` — "Dashboard charts never rendered. `ParentNode.append()` returns `undefined`, so `el.append(svg(…)).textContent = x` threw at top level of the dashboard's inline script, aborting it and leaving 8 of 13 panels (heatmap, lineage, diffs, cost, candidates, annotations) absent from every generated `dashboard.html` since `4c87ed1`. The `svg()` helper now takes a `text:` pseudo-attribute."

---

### Host grades audit

Optimizer backends — every ✅ traced to its cited artifact:

| Host | Claimed | Cited artifact | Proves it? |
|---|---|---|---|
| `claude-code` | ✅ verified | `ci/benchmarks/lib/run_suite.sh:135` → `optimizer_skill: claude-code`, driven by `integration-tests.yml:48-58` (tau2 task-9, real models, `assert_run.py --min-iterations 1`); plus `examples/{tau2_airline,skillsbench}/capevolve.yaml` + committed `run_full/` | ✅ Line 135 confirmed verbatim. Real-model CI job with a run assertion. |
| `mock` | ✅ verified | `templates/project/capevolve.yaml:13` → `optimizer_skill: mock`; `ci.yml:81-88` runs `examples/toy_calc/run.sh` and greps `baseline_val 0.0` + `test_reward 1.0` | ✅ `run.sh:17` copies exactly that template as the spec, so the toy-example job *is* a mock-optimizer execution. Chain closes. |
| `codex` | 🟡 docs-checked | — | ✅ correctly withheld |
| `gemini-cli` | 🟡 docs-checked | — | ✅ correctly withheld |
| `opencode` | 🟡 docs-checked | — | ✅ correctly withheld |
| `ibm-bob` | 🟡 docs-checked | — | ✅ correctly withheld |
| `cursor` / `droid` / `copilot` / `kimi` / `pi` / `antigravity` / `openclaw` | ➖ best-guess | — | ✅ matches `install.sh:36-40` |
| `generic` | n/a | — | ✅ |

**Demotions — all three judgments correct.**

`codex` / `gemini` appear in `benchmarks.yml` only as model ids, never as `optimizer_skill`:

```
$ grep -n -iE 'codex|gemini' .github/workflows/benchmarks.yml | head -6
66:          - "Azure/gpt-5.1-codex-2025-11-13"
67:          - "azure/gpt-5.3-codex"
74:          - "GCP/gemini-2.0-flash"
75:          - "gcp/gemini-2.5-flash"

$ grep -rn 'optimizer_skill' .github/ ci/ examples/ templates/
ci/benchmarks/lib/run_suite.sh:135:optimizer_skill:    claude-code
examples/*: claude-code (×5)   templates/adapters/*: claude-code (×5)
templates/project/capevolve.yaml:13:optimizer_skill: mock
```

`claude-code` and `mock` are the *only* two values anywhere in the repo. No wrongly-demoted host.

`ibm-bob`: 6 of 14 workflows are Bob-driven but all are issue/PR automation delegating to `skillberry-ai/skillberry-common`. Not optimizer evidence. Correct.

The `parallel: "true"` parity claim (`HOST_SUPPORT.md` capability table) checks out — `grep -c 'parallel: "true"' skills/optimizers/registry.yaml` → `1`, on `claude-code` (`registry.yaml:88`).

**Verdict: the optimizer table is accurate. One destination row overclaims (finding 4).**

**As a single source of truth for #143:** structurally sound. Two tables split along the right seam (does-it-run vs where-do-files-go, which genuinely have different evidence), a stated bar with a per-row citation, and a "Promoting a host to ✅" section that names the artifact requirement. The capability-parity table is the natural extension point for #143's cross-host skill matrix. The one gap for #143's purposes: parity is currently one row (parallel subagents) inferred from `registry.yaml`, with no mechanism keeping the table honest if the registry changes — cf. #189's `test_advertised_counts.py`, which is the pattern that would pin it. Worth #143 adding, not this PR.

---

### Screenshot integrity

**Reproduced. Effectively pixel-identical.** Ran the toy myself, captured with the PR's documented command:

```
$ chrome --headless --disable-gpu --hide-scrollbars --window-size=1280,1560 \
    --force-device-scale-factor=2 --screenshot=/tmp/my-toy.png --virtual-time-budget=12000 \
    "file://$D/.capevolve/run_demo/dashboard.html"
264419 bytes written to file /tmp/my-toy.png

$ python3 -c "import struct; [print(p, struct.unpack('>II',open(p,'rb').read()[16:24]), len(open(p,'rb').read())) for p in ['/tmp/my-toy.png','site/assets/toy-calc-dashboard.png']]"
/tmp/my-toy.png                       (2560, 3120) 264419
site/assets/toy-calc-dashboard.png    (2560, 3120) 264656
```

Same dimensions, 237 bytes apart. Compared visually: every KPI matches (best val 1.000, baseline 0.000, Δ +1.000, held-out test 1.000, 4 candidates, frontier 1, $0.0000, 0 tokens), narrative identical, all 6 pipeline phases DONE, dead-ends panel identical. Sole difference: WALL CLOCK `0.6s` (committed) vs `0.5s` (mine) — wall-clock noise, and the only field that *can't* be deterministic. This is a genuine capture of a genuine run.

**Charts genuinely visible:** yes, and they're the charts that never rendered before. The cumulative-best stair is there with its 0.00 baseline dot, the vertical jump at iteration 1, the blue record-holder rings, the amber rejected markers, the gold star at 1, and the full 0.00/0.25/0.50/0.75/1.00 axis labels — those axis labels *are* the `<text>` elements the TypeError was killing. The heatmap shows both tasks (`a1`, `a4`) × 4 iterations, red at iteration 0 → green after, with its legend.

**Leaks: none.** No metadata chunks whatsoever:

```
$ python3 -c "<PNG chunk walk>"
chunk IHDR len=13
chunk IDAT ... (×65)
chunk IEND len=0
```

No `tEXt`/`iTXt`/`zTXt`/`eXIf`/`tIME`/`pHYs` — nothing to strip. And no leaked strings:

```
$ strings -a site/assets/toy-calc-dashboard.png | grep -iE '/Users|/home/|/var/folders|osherelhadad|api[_-]?key|token|localhost|127\.0\.0\.1|ibm\.com|http'
(no output)
```

Visually clean too: the dashboard header renders `run_demo · 4 candidates · test sealed` — no filesystem path, no hostname, no server URL. The `/var/folders/...` temp path exists in the run output but not in the dashboard chrome, so it never entered the frame. Given #190's base-URL leak, I checked this three ways; it's clean.

**Honest for onboarding? Yes.** The all-zeros risk is real but the copy defuses it in both places, pre-emptively and specifically. `docs/GETTING_STARTED.md:68-71`: "`$0.0000` / `0` tokens because no model is called, and exactly 2 tasks × 4 candidates because that is all `toy_calc` has" — plus a pointer to `RESULTS.md` for scale. `site/getting-started.html:104-111` says the same and links the real-run walkthrough. Both reference the correct direction of travel instead of hiding the smallness. The image `alt` text is unusually good — it describes the actual numbers, so it survives as information when the image doesn't load.

**README hero τ² screenshot left in place: correct call.** #126 asked to fix the *getting-started* shot specifically ("`site/getting-started.html:105` caption admits: 'Shown here from a real benchmark run, not the toy'"), and offered "or move that image to a clearly-labeled 'what a real run looks like' aside" as an accepted resolution. The README hero was already labelled "a real tau2-bench airline run"; `README.md:29` now additionally says "**Not** the zero-API toy" with a link to the toy's real output. That's the aside #126 asked for. Removing a genuine, correctly-labelled hero shot would have been worse onboarding, not better.

**Consistency with #182:** clean, and I checked rather than assumed. `git diff origin/main...origin/fix/issue-99-sealed-headline -- README.md` touches only the Results table (adding ✅ artifact-backed / ⚠️ reported) — it does not touch the hero `` or its caption. #182 grades the τ² val row (which is what the hero visualizes, baseline 0.536 → best 0.712) as **✅ artifact-backed, `run_full/` committed**. So #202's "a real τ²-bench airline run" and #182's ✅ agree. The row #182 downgrades to ⚠️ is the *held-out test* 30.0 → 47.5, which the hero doesn't show. No divergence.

**Image hygiene:** 264,656 bytes — the *third smallest* of 9 dashboard PNGs in `site/assets/` (`dash-gitdiffs.png` is 559 KB, `dash-cost.png` 373 KB, `logo.png` 226 KB). Reasonable, and 2560×3120 @2× is justified for a screenshot whose value is legible small text. No old file needed removing: `dash-overview.png` is still used by 3 referrers, so restoring it after lychee flagged the break was correct, not a patch-up:

```
$ grep -rn 'dash-overview' --include='*.md' --include='*.html' .
README.md:27 · site/index.html:115 · site/run-end-to-end.html:212
```

Asset orphan sweep found no newly-dangling files; `toy-calc-dashboard.png` has 2 referrers.

---

### Merge-order note

**All seven branches merge clean — no textual conflicts, verified rather than eyeballed:**

```
$ for b in feat/issue-123-site-chrome fix/issue-99-sealed-headline fix/issue-102-counts \
           fix/issue-120-cdn-font perf/issue-119-reduce-cache feat/issue-121-doctor \
           docs/issue-104-gate-modes; do
    git merge-tree --write-tree --messages origin/fix/issue-126-onboarding-polish origin/$b >/dev/null 2>&1
    echo "$b -> exit=$?"
  done
feat/issue-123-site-chrome -> exit=0     fix/issue-99-sealed-headline -> exit=0
fix/issue-102-counts       -> exit=0     fix/issue-120-cdn-font       -> exit=0
perf/issue-119-reduce-cache-> exit=0     feat/issue-121-doctor        -> exit=0
docs/issue-104-gate-modes  -> exit=0
```

**vs #196 (`feat/issue-123-site-chrome`) — the flagged high risk does not materialize, and I confirmed the drift guard specifically.** #196's `scripts/sync-site-chrome.py` rewrites only `<head>`, nav, and footer. Every #202 site edit is in a page *body*: `getting-started.html:104-111` (the `<figure>`), `index.html:299-306` (a `<p class="section-intro">` in the Choose-your-path section), `optimize-your-own.html:246` (a step paragraph). Disjoint regions. Merged them for real and ran the guard:

```
$ cd /tmp/mt196   # worktree at origin/feat/issue-123-site-chrome
$ git merge --no-edit origin/fix/issue-126-onboarding-polish
 12 files changed, 185 insertions(+), 21 deletions(-)
 create mode 100644 docs/HOST_SUPPORT.md
 create mode 100644 site/assets/toy-calc-dashboard.png
$ python3 scripts/sync-site-chrome.py --check
site chrome in sync (9 pages)
GUARD EXIT=0
```

**Neither order breaks.** #202's site edits are not clobbered by #196's generator and do not trip its drift guard. #196's new `site-chrome` CI job also adds a lychee pass over `site/*.html` with `--root-dir site`, which will now cover `assets/toy-calc-dashboard.png` — a free extra check on this PR's new asset, and it resolves.

**#194 (`perf/issue-119-reduce-cache`) and #192 (`fix/issue-120-cdn-font`) — same-file, non-overlapping, and I ran the merged tests:**

```
# hunk locations don't intersect
#194 dashboard.py:  @@ -229,8 +229,59 @@                (reducer)
#202 dashboard.py:  @@ 885, 982, 1003, 1026, 1108, 1129, 1157   (inline JS)
#192 test_dashboard.py: @@ -243,3 +243,34 @@
#202 test_dashboard.py: @@ 5, 177

$ cd /tmp/mt119 && git merge -q origin/fix/issue-126-onboarding-polish   # Auto-merging dashboard.py
$ pytest core/tests/test_dashboard.py core/tests/test_dashboard_cache.py -q
17 passed in 0.13s

$ cd /tmp/mt192 && git merge -q origin/fix/issue-126-onboarding-polish   # Auto-merging test_dashboard.py
$ pytest core/tests/test_dashboard.py -q
11 passed in 0.10s
```

Worth noting for #192: its new test adds font/CSS assertions to the same file, and #202's new regex scans the whole generated HTML. They coexist — but if #192 ever adds JS to `dashboard.py` it must not use `.append(svg(…)).prop`, or #202's guard will (correctly) fail it. That's the guard working.

**#189 (`fix/issue-102-counts`) — the one coordination item.** It edits `llms.txt` and adds `core/tests/test_advertised_counts.py`, which pins advertised counts against reality. Finding 3 asks #202 to edit `llms.txt:39-41`. Same file, different lines, merges clean today — but whichever lands second should re-run `test_advertised_counts.py`, and #189 is the natural owner of "does `HOST_SUPPORT.md` agree with `registry.yaml`" as a test.

**Recommended order:**

1. **#202 now** — it carries the chart fix, which is a live user-facing bug and shouldn't wait behind a 526-line site refactor.
2. **#194, #192** — same-file, verified green against #202.
3. **#189** — then re-run `test_advertised_counts.py` over the merged `llms.txt` + `HOST_SUPPORT.md`.
4. **#196 last** of the site work — biggest diff, and its drift guard is a clean gate to land against once the body edits are settled.
5. #182, #193, #198 — independent, any time.

If #196 lands first instead, #202 needs no rework: the guard passes as shown.

---

### Verification I re-ran

```
$ cd /tmp/rv-202 && PYTHONPATH=/tmp/rv-202/core /tmp/ce-venv/bin/python -m pytest core/tests -q
1 failed, 178 passed in 60.35s (0:01:00)
```

The one failure is #200 (port 7878 contention — the test hardcodes a literal while `maybe_launch` probes for a free port), pre-existing and unrelated; the PR's evidence comment demonstrates it reproducing with `core/` reverted to `origin/main`. Effective baseline: **179/179**.

```
$ /tmp/ce-venv/bin/python -m compileall -q core skills && echo OK
compileall OK

$ bash -n install.sh && echo OK
install.sh syntax OK
```

`install.sh` is behaviorally inert beyond `bash -n` — the diff is two comment lines inside `detect_dest()`, before the `case`. No branch, no variable, no precedence change; `git diff origin/main -- install.sh` is 2 added `#` lines. Documented install path unaffected.

```
$ lychee --offline --no-progress --include-fragments \
    --exclude-path examples/tau2_airline/run_full --exclude-path examples/skillsbench/run_full \
    "README.md" "docs/**/*.md" "examples/**/*.md" "*.md"
🔍 264 Total (in 6ms) 🔗 115 Unique ✅ 221 OK 🚫 0 Errors 👻 43 Excluded
```

Matches the PR's claim exactly (264/115/221/0/43). Independent Python walk of every relative Markdown link and image, including the new PNG and `HOST_SUPPORT.md` from all 6 referrers: `broken local md links: 0`. Site HTML (which lychee's `*.md` globs miss):

```
$ cd site && for f in *.html; do for r in $(grep -oE '(src|href)="[^"#]+"' "$f" \
    | sed -E 's/.*="([^"]*)"/\1/;s/\?.*$//' | grep -vE '^(https?:|mailto:|//|#)'); do
    [ -e "$r" ] || echo "MISSING in $f: $r"; done; done
(no output — all site local refs resolve)
```

Real toy run reproduced (metrics byte-identical to the PR's, dashboard-server line differs per finding 2), screenshot reproduced at 264,419 vs 264,656 bytes and same dimensions, browser console captured before (TypeError, 1 svg, 5 panels) and after (zero errors, 5 svgs, 13 panels), PNG chunk-walked (no metadata), served the run dir over HTTP (`dashboard.html -> 200 size=35406`), and merge-tested all 7 concurrent branches with #196's drift guard run against the merged tree.

Every verification claim in the PR description reproduced.

OsherElhadad:
## 🔧 Review fixes

All 7 findings addressed; 1 partially declined with reasoning (#5 — full DOM coverage deferred, but the assertion with teeth is **in this PR**). Pushed as `22ad618`.

---

### 1. `docs/GETTING_STARTED.md` — the `test_pass_k` explanation was factually wrong ✅ fixed

You're right and I was wrong: pass^k has nothing to do with the number of tasks. `loop.py:75` computes it **per task over that task's trials**, then means across tasks; `stats.py:65` returns `0.0` whenever `k > n`. Re-verified directly:

```
$ PYTHONPATH=core /tmp/ce-venv/bin/python -c "
from cap_evolve.stats import pass_k
print('pass_k([1.0],2)     =', pass_k([1.0],2))
print('pass_k([1.0,1.0],2) =', pass_k([1.0,1.0],2))
print('pass_k([1.0,0.0],2) =', pass_k([1.0,0.0],2))"
pass_k([1.0],2)     = 0.0
pass_k([1.0,1.0],2) = 1.0
pass_k([1.0,0.0],2) = 0.0
```

Two trials → `1.0`. Task count irrelevant. Real cause is `templates/project/capevolve.yaml:66` → `num_trials: 1`.

**On the #187 interaction — I did read `origin/fix/issue-112-passk-na` and ran the toy on it, rather than reasoning about it.** #187 omits any `k > min(trials)` instead of emitting `0.0`, so the `"2"` key genuinely disappears. Both runs, same box, same command:

```
$ cd /tmp/fx-202 && bash examples/toy_calc/run.sh      # this branch
  "test_pass_k": {
    "1": 1.0,
    "2": 0.0
  },

$ cd /tmp/pk187 && bash examples/toy_calc/run.sh       # origin/fix/issue-112-passk-na
  "test_pass_k": {
    "1": 1.0
  },
```

**How I handled it:** I did *not* drop the line. The transcript block in this doc is a real capture of *today's* code, and pre-emptively deleting the `"2": 0.0` line would make the block wrong until #187 lands — the exact failure mode this PR exists to end. So the explanation is correct for both states and names the transition explicitly:

> `test_pass_k` `"2": 0.0` is honest too, and it is **not** about the number of tasks: pass^k is computed per task over that task's *trials*, then averaged. The toy runs `num_trials: 1`, so `k=2` exceeds the one trial available and the estimator returns 0 rather than guessing. Re-run with `num_trials: 2` or more to populate it. (Once **PR #187** — issue #112 — lands, an unmeasurable `k` is **omitted** instead of reported as `0.0`, so the `"2"` key simply disappears from this block. Same fact, better rendering.)

**Dependency to note: whoever merges #187 should delete the `"2": 0.0` line from the pasted block in `docs/GETTING_STARTED.md`** (the parenthetical then becomes redundant and can go too). No ordering constraint — the doc is honest either way — but it's a one-line follow-up, not zero.

---

### 2. `docs/GETTING_STARTED.md` — "literal, unedited output" was conditionally false ✅ fixed

Correct, and the sharper point stands: the doc pasted the *degraded* branch and then explained `"skipped"` as if universal, under a heading promising literalness. I took the annotate-the-variance route and stopped claiming unqualified literalness:

> Below is real output of the command above, captured on a box **without** the optional live-server backend installed. Every metric line is deterministic — you will get those bytes exactly. The only additions are the three `# ← varies` markers, flagging the lines that legitimately differ per environment

Three lines now carry `# ← varies` (working dir, the `{"dashboard": …}` line, `"dashboard_server"`), followed by:

> If you *did* install the optional backend (`pip install -e dashboard/backend`, as `INSTALL.md` suggests), those two lines instead read `{"dashboard": "http://127.0.0.1:7878"}` and `"dashboard_server": "http://127.0.0.1:78xx"`.

Note I also dropped the word **unedited** entirely — with `# ← varies` markers added the block *is* edited, and claiming otherwise would reintroduce the same species of small dishonesty one layer down. "Every metric line is deterministic — you will get those bytes exactly" is the claim that's actually true and actually useful.

---

### 3. `llms.txt` + `docs/OPTIMIZE_YOUR_OWN.md` — unqualified list on the LLM-facing surface ✅ fixed (plus one you didn't catch)

`llms.txt` doc index gained the page:

```
- [docs/HOST_SUPPORT.md](docs/HOST_SUPPORT.md): which hosts/optimizers are CI-verified
  vs docs-checked vs best-guess. Single source of truth — do not restate the list.
```

and `llms.txt:41`:

> mock — one runner resolves the name; the agent set matches obra/superpowers). **Host-agnostic by design, but only `claude-code` and `mock` are CI-executed** — the rest are docs-checked or best-guess; see docs/HOST_SUPPORT.md.

`OPTIMIZE_YOUR_OWN.md` both sites (`:90` intake spec, `:144` the yaml `← swap` comment) got the same clause.

**Re-grepped, and found a surface neither of us listed:** `skills/optimizers/run-optimizer/SKILL.md:42-45` — the optimizer skill's own body, i.e. what an agent reads when it's *actually choosing a backend*. Arguably worse than `llms.txt`. Qualified:

> Only **`claude-code`** and **`mock`** are CI-executed; every other name is docs-checked or a best-guess dir mapping. Grades + per-row evidence live in `docs/HOST_SUPPORT.md` (single source of truth — do not restate the list elsewhere).

Post-fix sweep — every surface with a host list, and whether it links the table:

```
$ grep -rln -iE 'codex.*gemini-cli.*opencode|Codex, Gemini|kimi.*pi.*antigravity' \
    README.md RUN.md docs/ site/ llms.txt install.sh CHANGELOG.md skills/ templates/
README.md  RUN.md  docs/OPTIMIZE_YOUR_OWN.md  docs/INSTALL.md  docs/ROADMAP.md
docs/specs/2026-06-24-multi-agent-support-design.md  docs/how-to/cap-evolve-with-exgentic-tau2.md
site/index.html  site/optimize-your-own.html  llms.txt  install.sh  CHANGELOG.md
skills/optimizers/run-optimizer/SKILL.md  templates/adapters/{swe_bench,tau2_bench,skillsbench,huggingface_litellm,jsonl_litellm}/capevolve.yaml

$ grep -rln 'HOST_SUPPORT' README.md RUN.md docs/ site/ llms.txt install.sh skills/ templates/
README.md  RUN.md  docs/INSTALL.md  docs/OPTIMIZE_YOUR_OWN.md  site/index.html
llms.txt  site/optimize-your-own.html  install.sh  skills/optimizers/run-optimizer/SKILL.md
```

**Deliberately left:** `docs/ROADMAP.md`, `docs/specs/`, `docs/how-to/` (historical, as you said), `CHANGELOG.md` (append-only history), and the five `templates/adapters/*/capevolve.yaml` — those are `# or: codex | gemini-cli | opencode | mock` on a config line, four names not fourteen, and adding a doc pointer to every generated project's yaml is noise. Say the word if you want them.

---

### 4. `docs/HOST_SUPPORT.md:62` — the `claude-code` install-destination ✅ that nothing proves ✅ **downgraded**

You said be strict, so I went past the footnote you preferred: I **downgraded the row to 🟡** *and* added the note, because the footnote alone leaves a ✅ on the page meaning something different from the ✅ one screen above it — and a badge that means two things is the failure this page exists to prevent.

Re-verified your evidence first:

```
$ grep -rn 'install\.sh' .github/ ci/ core/tests/
(no output)
$ grep -rn 'CAPEVOLVE_SKILLS_DIR' .github/workflows/ ci/benchmarks/lib/*.sh examples/*/run.sh
ci/benchmarks/lib/run_suite.sh:37:export CAPEVOLVE_SKILLS_DIR="$REPO/skills"
examples/toy_calc/run.sh:9:export CAPEVOLVE_SKILLS_DIR="$REPO/skills"
examples/tau2_airline/run.sh:10:export CAPEVOLVE_SKILLS_DIR="$REPO/skills"
examples/skillsbench/run.sh:12:export CAPEVOLVE_SKILLS_DIR="$REPO/skills"
```

Confirmed — and it's broader than CI: **all four** run scripts set it, so *nothing anywhere* exercises the `--host` mapping. The row is now:

| `--host` | Destination | Status |
|---|---|---|
| `claude` / `claude-code` | `~/.claude/skills` | 🟡 docs-checked (vendor-canonical path; `install.sh` is **not** CI-executed) |

with a blockquote above the table stating the scope, so no future reader promotes a row on the wrong bar:

> **No row here can be ✅ by this page's own bar.** Nothing in CI executes `install.sh`; CI sets `$CAPEVOLVE_SKILLS_DIR` to the repo tree, which takes the *first* precedence branch and bypasses the `--host` mapping entirely. So the badges below grade **how well-attested the path is**, not whether an install was executed. 🟡 on `claude` is the strongest available claim: it is the vendor-documented canonical path and `core/tests/test_native_skills.py` asserts the `.claude/skills` string — but no test runs the installer.

The destination table now has **zero ✅ rows**, which is the honest state. Genuinely proving artifact would be a CI job that runs `install.sh --host claude --dest $tmp` and asserts the tree — cheap, and worth filing, but out of scope here.

---

### 5. `core/tests/test_dashboard.py` — the regex overpromised ✅ **both** done, and the assertion with teeth is in this PR

Your measurement was right: it fires on one shape and misses `$()` chains, variable chains, and multiline. Comment softened to say exactly that:

```python
# Cheap source-level guard for the *exact* historical regression only:
# ...
# This regex catches nothing else — not a chain off `$()`, off a variable, or
# split across lines. `test_dashboard_js_renders_all_panels` is the assertion
# with teeth; this one is a fast tripwire on the known shape.
```

**And the real fix.** You framed this as "the assertion with teeth needs a DOM" and offered the `# ponytail:`-note escape hatch if a headless browser in pytest is unacceptable. It is unacceptable — but a **browser** isn't what's needed, only `append()` returning `undefined`. That's ~20 lines of shim and `node`, which is already a hard CI dependency (`ci.yml:53` `setup-node@v4`, preinstalled on `ubuntu-latest`). So: `test_dashboard_js_renders_all_panels` extracts the generated inline script, runs it under a minimal DOM shim whose `append()` returns `undefined` exactly like `ParentNode.append()`, and counts what the script *constructs*:

```python
assert p.returncode == 0, f"dashboard JS threw:\n{p.stderr}"
assert counts.get("section", 0) >= 13, f"panels not built: {counts}"
assert counts.get("h2", 0) == counts.get("section", 0), f"panel/title mismatch: {counts}"
# SVG <text> is the signature: 0 on the broken code, 13+ once the charts render.
assert counts.get("text", 0) >= 10, f"charts emitted no labels: {counts}"
assert counts.get("svg", 0) >= 4, f"charts missing: {counts}"
```

`pytest.skip` when `node` is absent, so no new hard dependency for local runs.

#### Would it have caught the original bug? Yes — reintroduce-and-fail, literal

**A. Exact historical bug** (`core/cap_evolve/dashboard.py` reverted to `origin/main`):

```
$ git checkout origin/main -- core/cap_evolve/dashboard.py
$ PYTHONPATH=core /tmp/ce-venv/bin/python -m pytest core/tests/test_dashboard.py -q
E             TypeError: Cannot set properties of undefined (setting 'textContent')
FAILED core/tests/test_dashboard.py::test_render_html_self_contained_and_parseable
FAILED core/tests/test_dashboard.py::test_dashboard_js_renders_all_panels - A...
2 failed, 9 passed in 0.14s
```

The full node stack trace is in the failure message, pointing at the offending line:

```
probe.js:125
    el.append(svg('text',{...,'text-content':''})).textContent=v.toFixed(2);}
                                                              ^
TypeError: Cannot set properties of undefined (setting 'textContent')
```

**B. The variant your audit proved the regex is blind to** — chained off a variable:

```
$ # rewrote one site to:  const lbl=svg('text',{…}); el.append(lbl).textContent=v.toFixed(2);
existing regex matches this variant? False
E             TypeError: Cannot set properties of undefined (setting 'textContent')
FAILED core/tests/test_dashboard.py::test_dashboard_js_renders_all_panels - A...
1 failed, 10 passed in 0.12s
```

**The regex stayed green. The new test failed.** That's the whole point — it catches the *class*, not the shape.

**C. Restored:**

```
$ git checkout HEAD -- core/cap_evolve/dashboard.py
$ PYTHONPATH=core /tmp/ce-venv/bin/python -m pytest core/tests/test_dashboard.py -q
11 passed in 0.10s
```

Element counts on the fixed code — your browser measured 16 `<text>`; the shim measures 13 on the *test* fixture (smaller run: 4 candidates, 2 tasks), same signature, **0 → 13**:

```
{"main":1,"div":84,"section":13,"h2":13,"p":2,"svg":5,"line":5,"text":13,
 "path":6,"circle":8,"rect":8,"table":3,"th":20,"tr":12,"td":62,"code":6,"span":11}
```

`section: 13` / `h2: 13` — the 13 panels, asserted as *constructed*, not as strings in the source.

**Deferred (declined here):** no layout/geometry/visual assertions, and no real CSS or event dispatch — the shim has no box model. A panel that builds all its elements at wrong coordinates still passes. That needs a real headless browser and belongs in its own issue; the throw-anywhere-in-a-panel class (the one that actually bit us) is covered now.

---

### 6. CHANGELOG ✅ added

Firmly agreed — six weeks of a half-empty flagship dashboard with no way for an affected user to learn it's fixed. Under `## [Unreleased] / ### Fixed`, first entry: names the mechanism, the blast radius (8 of 13 panels, `<text>` count 0), the `4c87ed1` regression point, the `text:` pseudo-attribute fix, the deleted second latent throw, and the new test.

---

### 7. `HOST_SUPPORT.md:57` vs `registry.yaml:109` gemini path ✅ cross-referenced

Disclosed but easy to trip on, as you said. The destination row now points at the divergence and its owner:

> `~/.gemini/extensions/cap-evolve/skills` | 🟡 docs-checked (skills live inside an extension — deliberately **not** the registry's per-workdir `skills_dir: .gemini/skills`; see #143)

---

### Verification — literal

**Full suite. 180 passed, 0 failed** (179 + the new test). Ran twice; **#200 did not reproduce this session** (port 7878 was free), so this is a clean number rather than the 179 + 1-known-failure shape:

```
$ cd /tmp/fx-202 && PYTHONPATH=/tmp/fx-202/core /tmp/ce-venv/bin/python -m pytest core/tests -q
....................................                                     [100%]
180 passed in 62.47s (0:01:02)
```

`test_dashboard_launch.py::test_maybe_launch_spawns_when_available` still hardcodes port 7878 and will fail again on a box with servers on that port — **#200, pre-existing, unrelated**, unchanged by this diff.

```
$ /tmp/ce-venv/bin/python -m compileall -q core skills && echo "compileall OK"
compileall OK
$ bash -n install.sh && echo "install.sh syntax OK"
install.sh syntax OK
```

Skip path when `node` is missing (no new hard dependency):

```
$ PATH=/usr/bin:/bin pytest core/tests/test_dashboard.py::test_dashboard_js_renders_all_panels -q -rs
SKIPPED [1] core/tests/test_dashboard.py:234: node not available; JS execution check skipped
1 skipped in 0.01s
```

Links — CI's exact `docs-links` command, after all doc edits:

```
$ lychee --offline --no-progress --include-fragments \
    --exclude-path examples/tau2_airline/run_full --exclude-path examples/skillsbench/run_full \
    "README.md" "docs/**/*.md" "examples/**/*.md" "*.md"
🔍 267 Total (in 11ms) 🔗 117 Unique ✅ 222 OK 🚫 0 Errors 👻 45 Excluded
```

267/117/222/**0**/45 — up from 264/115/221/0/43 (the new `HOST_SUPPORT.md` + `#187` links). `HOST_SUPPORT.md` resolves from **all 9** referrers now:

```
README.md (×2)  RUN.md  docs/INSTALL.md (×2)  docs/OPTIMIZE_YOUR_OWN.md (×2)
install.sh  llms.txt (×2)  site/index.html  site/optimize-your-own.html
skills/optimizers/run-optimizer/SKILL.md
```

Site HTML (lychee's `*.md` globs miss it):

```
$ cd site && for f in *.html; do for r in $(grep -oE '(src|href)="[^"#]+"' "$f" \
    | sed -E 's/.*="([^"]*)"/\1/;s/\?.*$//' | grep -vE '^(https?:|mailto:|//|#)'); do
    [ -e "$r" ] || echo "MISSING in $f: $r"; done; done
(no output — all site local refs resolve)
```

`core/cap_evolve/dashboard.py` is **byte-identical to `1c43f46`** after the reintroduce-and-restore cycle (`git diff HEAD -- core/cap_evolve/dashboard.py` → empty). This commit touches no production code — only tests and docs.

```
$ git log --format='%H%n%an <%ae>' -1
22ad618f17129c8d44b00d4ff8bfbcfc0aabc4a6
$ git log -1 --format='%B' | grep -ci 'co-authored-by\|generated with'
0
```

Files touched: `CHANGELOG.md`, `core/tests/test_dashboard.py`, `docs/GETTING_STARTED.md`, `docs/HOST_SUPPORT.md`, `docs/OPTIMIZE_YOUR_OWN.md`, `llms.txt`, `skills/optimizers/run-optimizer/SKILL.md` — 7 files, +140/−23.

---

### Merge order (restated)

Per your verified merge testing — **no conflict with #196**, its generator rewrites only `<head>`/nav/footer while every #202 site edit is page-body, and `sync-site-chrome.py --check` passed on the real merged tree:

1. **#202 first** — it carries the live chart fix (8 of 13 dashboard panels dead since `4c87ed1`); shouldn't wait behind a site refactor.
2. **#194, #192** — same-file, non-overlapping hunks, verified green against #202.
3. **#189** — then re-run `core/tests/test_advertised_counts.py` over the merged `llms.txt` + `HOST_SUPPORT.md`. **This one now matters more:** finding 3 added a `llms.txt` doc-index line and a clause to the optimizer line, so #189's count assertions meet a changed file. Same file, different lines, still merges clean.
4. **#196 last** of the site work — biggest diff, and its drift guard is a clean gate to land against.
5. #182, #193, #198 — independent, any time.

**#187** is orthogonal to merge order, but see finding 1: after it lands, drop the `"2": 0.0` line from the `docs/GETTING_STARTED.md` transcript block.

Follow-ups worth filing, none blocking: a CI job that actually executes `install.sh --host` (would let the destination table earn a real ✅), and browser-level layout assertions for the dashboard.

Thanks for the audit — finding 1 was a real error on my part, and the "your test wouldn't catch the next one" push produced the only assertion here that would have caught the original.

PR #199, opened 2026-07-29
fix(algorithm): give GEPA & SkillOpt the same optimizer context as hill-climb, un-gate the CLI flags

OsherElhadad:
Closes #109

The two flagship algorithms ran with a **thinner** optimizer context than the basic climber. `_inject_optimizer_context` was only reachable from `harness.run_step` (the hill-climb path) — GEPA bypasses `run_step` entirely, and `skillopt_loop` had no capability/optimizer parameters at all. On the CLI side, seven optimizer-context flags were gated behind `if algorithm_name == "hill-climb"` and **silently dropped** for gepa/skillopt.

## Before / after — what the optimizer actually receives

Measured with a real run (`OptimizerContext(capabilities="system-prompt", optimizer_name="claude-code", bench_repo=…, capability_sources=…, target_model="weak")`), same toy_calc adapter, same mock optimizer:

| | GEPA before | GEPA after | SkillOpt before | SkillOpt after |
|---|---|---|---|---|
| `./trajectories/` | ❌ absent | ✅ 3 files, scoped to parent minibatch | ✅ | ✅ |
| `./guidance/<cap>/` | ❌ absent | ✅ `system-prompt` | ✅ | ✅ |
| `./guidance/diagnose/` | ❌ absent | ✅ | ✅ | ✅ |
| `./guidance/sources/` | ❌ absent | ✅ | ❌ (flag dropped) | ✅ |
| `./guidance/optimizer/<name>.md` | ❌ absent | ✅ | ❌ | ✅ |
| native skills (`.claude/skills/`) + `CLAUDE.md` | ❌ absent | ✅ | ❌ | ✅ |
| `CAP_BRIEF` (allowed edit space) | ❌ **empty** | ✅ | ❌ **empty** | ✅ |
| `BENCH_REPO` pointer | ❌ | ✅ | ❌ | ✅ |
| `PARALLEL_NOTE` fan-out | ❌ | ✅ | ❌ sequential (wrong) | ✅ |
| `{{TARGET_READER}}` consuming-LLM brief | ❌ | ✅ | ❌ | ✅ |
| intake-authored `--instructions-file` template | ❌ ignored | ✅ honored | ❌ ignored | ✅ honored |
| `target_profile` run event | ❌ never logged | ✅ | ❌ | ✅ |
| **rendered prompt size** (built-in default template) | **1,460 chars** | **22,770 chars** | 20,473 | 21,391 |
[...]

PR #204, opened 2026-07-29
fix(dashboard): render the live event ticker in the SPA; populate the algorithm label

### LINKED ISSUE #117: Render the collected-but-discarded live event ticker in the SPA; fix blank algorithm badge

OsherElhadad:
Closes #117

## What

Two fixes, both on the render path plus one shared reducer/producer fix.

**1. The live event ticker is rendered instead of discarded.** `useRunStream` already
maintained a capped 200-entry `log` of streamed events, but the only consumers of
`stream` in the whole app were `stream.status` and `stream.count` — so a live run
showed "42 live events" as a bare number and you could not see *what* happened. New
`EventTicker` component + an **Events** tab in `RunDeepDive` renders that existing
buffer newest-first: event kind, candidate id, val, and the accept/reject reason,
with accept/reject colouring and an `aria-live="polite"` region. No new fetching, no
new buffer — it reads the 200-entry log that was already there.

**2. The algorithm label is populated everywhere.** This fixes the Hub rows, the
DeepDive chip, Compare, and the self-contained HTML `Optimize · <algo>` phase label
in one change, because all four read `summary.algorithm`.

## Badge root cause (differs from the issue's diagnosis)

The issue proposed a one-liner: add `"algorithm": final.get("algorithm")` to the
reducer's summary dict, citing `harness.py:1877`, `gepa.py:664`, `skillopt.py:371`
as producers that put `algorithm` into `final.json`.

**That one-liner would still render blank.** Those three lines are the *return dicts*
of `hill_climb_loop` / `gepa_loop` / `skillopt_loop`, and that dict is only
`print(json.dumps(result))`-ed to stdout by the algorithm skill's `run.py`. Nothing
persists it. `final.json` is written by `harness.finalize()`, whose payload is:

```python
payload = {"test": ..., "best_id": ..., "test_baseline": ..., "baseline_id": ..., "test_delta": ...}
```

No `algorithm` key. Verified on a real run dir before the fix:

```
final.json keys: ['baseline_id', 'best_id', 'test', 'test_baseline', 'test_delta']
```

**Root cause in one sentence:** the algorithm name was never persisted into the run
dir at all — the loops return it only to stdout and `harness.finalize()` omits it
from `final.json` — so there was no producer for the reducer to read.

### Where it is fixed, and every producer checked

Fixed at the single choke point all three deterministic loops already route through,
`harness._init_memory_store`, which now logs one `algorithm` event. That is one
argument per loop rather than three separate `log_event` calls:

| Producer | Path | Covered by |
|---|---|---|
| `hill-climb` (3 focus schedules) | `hill_climb_loop` | `_init_memory_store(..., algorithm=algorithm)` |
| `gepa` | `gepa_loop` | `_init_memory_store(..., algorithm="gepa")` |
| `skillopt` | `skillopt_loop` | `_init_memory_store(..., algorithm=algorithm)` |
| `evograph` | agent-mode only, no deterministic loop | `cli.py` agent-mode handoff |
| `agent-optimize` | agent-mode only, no deterministic loop | `cli.py` agent-mode handoff |

Agent-mode runs (`evograph`, `agent-optimize`) short-circuit in `cap-evolve run`
right after baseline and never enter a loop, so they get the same event stamped at the
handoff — otherwise the label would stay blank for every agent-driven run.

Using an **event** rather than `final.json` has a second benefit the issue's approach
could not give: the label is correct for **live** runs from iteration 1, because
`final.json` only exists after finalize. `reduce_run` still falls back to
`final.get("algorithm")` for run dirs written before this event existed.

## Before / after

Same run dir, same SPA build; only the reducer differs.

```
BEFORE  summary.algorithm = None                    header: "run_demo\ndone\n0 live events"
AFTER   summary.algorithm = 'hill-climb:all'        header: "run_demo\ndone\nhill-climb:all\n0 live events"

BEFORE  GET /api/runs  -> "algorithm": null
AFTER   GET /api/runs  -> "algorithm": "hill-climb:all"
```

## Coordination with concurrent work

- **#116** (factoring the `events.jsonl` tail helper out of the dashboard SSE) — no
  overlap: `stream.py` / `read_new_events` and the SSE route in `app.py` are
  untouched. This PR only *consumes* the events the stream already delivers.
- **#119** (caching the reducer) — overlap is one hunk: the `algorithm` lookup is 5
  lines inside `reduce_run` just above the `summary` dict. It is a pure read over the
  already-loaded `events` list, so it is cache-safe; if #119 lands first, this is a
  trivial re-apply inside whatever the cached function becomes.
- `dashboard/frontend/dist/` is intentionally **not** rebuilt in this PR, matching the
  convention of recent frontend-src PRs (e.g. #178), so the bundle hashes don't
  conflict with #116/#119/#138/#139.

## Verification

`npm ci && npm run build`:

```
> tsc -b && vite build
transforming...✓ 2810 modules transformed.
dist/assets/index-DrRhMHxe.css   16.70 kB │ gzip:   4.58 kB
dist/assets/index-0-YIP72o.js   831.44 kB │ gzip: 250.20 kB
✓ built in 634ms
```

Backend suite (179 baseline → 182; +3 reducer/producer tests):

```
182 passed in 59.27s
```

Frontend suite (45 baseline → 48; +3 EventTicker tests):

```
$ npx vitest run
 Test Files  14 passed (14)
      Tests  48 passed (48)
```

Real run, zero API cost — `examples/toy_calc` via `cap-evolve run` with the `mock`
optimizer, then the dashboard served against the resulting run dir. Full commands and
output in the 🔬 Evidence comment below.

### DIFF
--- a/core/cap_evolve/cli.py
+++ b/core/cap_evolve/cli.py
@@ -325,6 +325,10 @@ def _cmd_run(argv):
     # primitives, and sealing with `cap-evolve finalize`. cap-evolve run does
     # setup+baseline, then hands off here — no algorithm subprocess, no auto-finalize.
     if orchestration_mode == "agent":
+        # Agent-mode runs never enter a deterministic loop, so stamp the algorithm
+        # event here — otherwise the dashboard label would be blank for every
+        # agent-driven run (evograph, agent-optimize, ...).
+        _RunDir.open(workdir / run_dir).log_event("algorithm", name=algorithm_name)
         print(json.dumps({"mode": "agent", "run_dir": run_dir, "algorithm": algorithm_name,
                           "stop_condition": str(spec.get("stop_condition", "")),
                           "next": "drive via the orchestrate Agent-mode loop; "
--- a/core/cap_evolve/dashboard.py
+++ b/core/cap_evolve/dashboard.py
@@ -535,8 +535,16 @@ def reduce_run(run_dir) -> dict:
     elif baseline_val == 0 and best_val:
         delta_pct = None  # undefined %Δ off a zero baseline; show absolute Δ instead
 
+    # Which algorithm produced this run. Every loop logs one `algorithm` event at its
+    # start, so this is populated for LIVE runs too — final.json only exists after
+    # finalize. `final.get` is the fallback for run dirs written before that event.
+    algorithm = next((ev.get("name") for ev in reversed(events)
+                      if ev.get("kind") == "algorithm" and ev.get("name")),
+                     None) or final.get("algorithm")
+
     summary = {
         "run_id": root.name,
+        "algorithm": algorithm,
         "baseline_val": baseline_val,
         "best_val": best_val,
         "best_id": best_id,
--- a/core/cap_evolve/gepa.py
+++ b/core/cap_evolve/gepa.py
@@ -482,7 +482,7 @@ def gepa_loop(
     highest-val pool member; the test split is never touched.
     """
     gate_kwargs = dict(gate_kwargs or {})
-    rejected, history, store = _init_memory_store(run_dir, store)
+    rejected, history, store = _init_memory_store(run_dir, store, algorithm="gepa")
     cache = EvalCache(run_dir.root / "eval_cache.json")
     rng = random.Random(seed)
     run_dir.log_event("gepa_start", seed=seed, minibatch_size=minibatch_size,
--- a/core/cap_evolve/harness.py
+++ b/core/cap_evolve/harness.py
@@ -1374,11 +1374,19 @@ def run_step(
 
 # ---- memory + version store wiring ----------------------------------------
 
-def _init_memory_store(run_dir: RunDir, store):
+def _init_memory_store(run_dir: RunDir, store, algorithm: str | None = None):
     """Create the optimizer memory (rejected + accepted history) and ensure a
-    version store (default git) is initialized + holds an initial 'seed' commit."""
+    version store (default git) is initialized + holds an initial 'seed' commit.
+
+    Also stamps the run's ``algorithm`` event. Every deterministic loop (hill-climb,
+    gepa, skillopt) routes through here, so logging it once here is what makes the
+    dashboard's algorithm label work for all of them — and from iteration 1, not only
+    once ``final.json`` exists.
+    """
     from .memory import History, RejectedMemory
     from .store import VersionStore
+    if algorithm:
+        run_dir.log_event("algorithm", name=algorithm)
     rejected = RejectedMemory(run_dir.rejected_path)
     history = History(run_dir.history_path)
     if store is None:
@@ -1866,7 +1874,7 @@ def hill_climb_loop(
     per-instance frontier and parent selection (see ``cap_evolve.gepa``).
     """
     gate_kwargs = dict(gate_kwargs or {})
-    rejected, history, store = _init_memory_store(run_dir, store)
+    rejected, history, store = _init_memory_store(run_dir, store, algorithm=algorithm)
 
     # Resolve the consuming-LLM profile once; its brief steers every iteration's prompt.
     from . import target_profile as _tp
--- a/core/cap_evolve/harness.py
+++ b/core/cap_evolve/harness.py
@@ -1374,6 +1374,28 @@ def run_step(
 
 # ---- memory + version store wiring ----------------------------------------
 
+def last_algorithm_event(run_dir: RunDir) -> str | None:
+    """The name of the run's most recent ``algorithm`` event, or None.
+
+    Same last-write-wins read the dashboard reducer does (``dashboard.py``), so a
+    re-stamp with an unchanged name is detectable and can be skipped.
+    """
+    if not run_dir.events_path.exists():
+        return None
+    name = None
+    for line in run_dir.events_path.read_text(encoding="utf-8").splitlines():
+        line = line.strip()
+        if not line or '"algorithm"' not in line:
+            continue
+        try:
+            ev = json.loads(line)
+        except json.JSONDecodeError:
+            continue
+        if ev.get("kind") == "algorithm" and ev.get("name"):
+            name = ev["name"]
+    return name
+
+
 def _init_memory_store(run_dir: RunDir, store, algorithm: str | None = None):
     """Create the optimizer memory (rejected + accepted history) and ensure a
     version store (default git) is initialized + holds an initial 'seed' commit.
@@ -1382,10 +1404,18 @@ def _init_memory_store(run_dir: RunDir, store, algorithm: str | None = None):
     gepa, skillopt) routes through here, so logging it once here is what makes the
     dashboard's algorithm label work for all of them — and from iteration 1, not only
     once ``final.json`` exists.
+
+    NOTE for #114 (removing write-only optimizer memory): if this function is renamed,
+    split or inlined, the ``algorithm`` stamp must move with it — it is the single
+    choke point the dashboard label depends on. ``test_dashboard.py``'s
+    ``test_every_deterministic_loop_stamps_the_algorithm_event`` guards the drop.
     """
     from .memory import History, RejectedMemory
     from .store import VersionStore
-    if algorithm:
+    # Idempotent: --resume re-enters here, and re-stamping the same name would add a
+    # dead duplicate row to the dashboard's event ticker. A resume that *changes*
+    # algorithm/--focus still stamps, so the label stays last-write-wins and correct.
+    if algorithm and last_algorithm_event(run_dir) != algorithm:
         run_dir.log_event("algorithm", name=algorithm)
     rejected = RejectedMemory(run_dir.rejected_path)
     history = History(run_dir.history_path)
--- a/core/cap_evolve/skillopt.py
+++ b/core/cap_evolve/skillopt.py
@@ -224,7 +224,7 @@ def skillopt_loop(
     ``edit_budget_schedule`` / ``slow_updates`` / per-epoch ``epoch_stats``.
     """
     gate_kwargs = dict(gate_kwargs or {})
-    rejected, history, store = harness._init_memory_store(run_dir, store)
+    rejected, history, store = harness._init_memory_store(run_dir, store, algorithm=algorithm)
 
     train_ids = list(run_dir.read_splits().train)
     n_train = len(train_ids)
--- a/core/tests/test_dashboard.py
+++ b/core/tests/test_dashboard.py
@@ -291,3 +291,23 @@ class _NullStore:
 
     def commit(self, *a, **k):
         pass
+
+
+def test_resume_does_not_double_stamp_the_algorithm_event():
+    """--resume re-enters _init_memory_store; re-stamping the same name would add a
+    dead duplicate row to the dashboard event ticker (review finding 4)."""
+    from cap_evolve import Budget, RunDir, dashboard, harness
+
+    def algorithm_events(rd):
+        return [json.loads(x) for x in rd.events_path.read_text(encoding="utf-8").splitlines()
+                if x.strip() and json.loads(x).get("kind") == "algorithm"]
+
+    with tempfile.TemporaryDirectory() as d:
+        rd = RunDir.create(Path(d), ts="t", budget=Budget())
+        harness._init_memory_store(rd, _NullStore(), algorithm="hill-climb:all")
+        harness._init_memory_store(rd, _NullStore(), algorithm="hill-climb:all")  # --resume
+        assert len(algorithm_events(rd)) == 1
+        # A resume that CHANGES the algorithm/--focus still corrects the label.
+        harness._init_memory_store(rd, _NullStore(), algorithm="hill-climb:cyclic")
+        assert len(algorithm_events(rd)) == 2
+        assert dashboard.reduce_run(rd)["summary"]["algorithm"] == "hill-climb:cyclic"
--- a/core/tests/test_dashboard.py
+++ b/core/tests/test_dashboard.py
@@ -243,3 +243,51 @@ def test_no_target_profile_event_leaves_summary_none():
         r = dashboard.reduce_run(rd)
         assert r["summary"]["target_profile"] is None
         assert "consuming model" not in dashboard.render_ansi(r, color=False)
+
+
+# ---- algorithm label (issue #117) ----------------------------------------
+
+def test_algorithm_event_surfaces_in_summary_without_final():
+    """The label must work for a LIVE run — i.e. before final.json exists."""
+    from cap_evolve import dashboard
+    events = [{"kind": "algorithm", "name": "hill-climb:all"}] + _BASE_EVENTS
+    with tempfile.TemporaryDirectory() as d:
+        rd = _mk_run(Path(d), events=events, baseline=_BASELINE)
+        assert not (rd.root / "final.json").exists()
+        assert dashboard.reduce_run(rd)["summary"]["algorithm"] == "hill-climb:all"
+
+
+def test_algorithm_falls_back_to_final_json_then_none():
+    from cap_evolve import dashboard
+    with tempfile.TemporaryDirectory() as d:
+        rd = _mk_run(Path(d), events=_BASE_EVENTS, baseline=_BASELINE,
+                     final={"test": {"reward": 0.8}, "best_id": "cand_0001",
+                            "algorithm": "gepa"})
+        assert dashboard.reduce_run(rd)["summary"]["algorithm"] == "gepa"
+    with tempfile.TemporaryDirectory() as d:
+        rd = _mk_run(Path(d), events=_BASE_EVENTS, baseline=_BASELINE)
+        assert dashboard.reduce_run(rd)["summary"]["algorithm"] is None
+
+
+def test_every_deterministic_loop_stamps_the_algorithm_event(tmp_path=None):
+    """Each loop's _init_memory_store call must log it — otherwise the badge is
+    blank for that algorithm (the actual root cause of issue #117)."""
+    from cap_evolve import Budget, RunDir, dashboard, harness
+    for algo in ("hill-climb:all", "skillopt", "gepa"):
+        with tempfile.TemporaryDirectory() as d:
+            rd = RunDir.create(Path(d), ts="t", budget=Budget())
+            harness._init_memory_store(rd, _NullStore(), algorithm=algo)
+            assert dashboard.reduce_run(rd)["summary"]["algorithm"] == algo
+
+
+class _NullStore:
+    """Minimal VersionStore stand-in: no git, no commits."""
+
+    def init(self):
+        pass
+
+    def log(self):
+        return ["seed"]  # non-empty → _init_memory_store skips the seed commit
+
+    def commit(self, *a, **k):
+        pass
--- a/core/tests/test_orchestration_mode.py
+++ b/core/tests/test_orchestration_mode.py
@@ -107,6 +107,13 @@ def test_agent_mode_stops_after_baseline_with_handoff(tmp_path, capsys):
     splits = json.loads((run_dir / "splits.json").read_text(encoding="utf-8"))
     assert splits["test_used"] is False, "agent mode must not score/seal the test"
 
+    # Agent mode is the one producer NOT behind harness._init_memory_store, so the
+    # dashboard's algorithm badge depends on cli.py stamping the event here. Without
+    # this assert, dropping that stamp goes unnoticed (issue #117).
+    from cap_evolve import dashboard
+    from cap_evolve.rundir import RunDir
+    assert dashboard.reduce_run(RunDir.open(run_dir))["summary"]["algorithm"] == "hill-climb"
+
 
 def test_deterministic_mode_still_runs_full_pipeline(tmp_path, capsys):
     """The same setup in deterministic mode runs the algorithm + finalize (test sealed)."""
--- a/dashboard/backend/capevolve_dashboard/app.py
+++ b/dashboard/backend/capevolve_dashboard/app.py
@@ -109,7 +109,12 @@ def create_app(base_dir: Path, static_dir: Path | None = None) -> FastAPI:
                 yield _stream.sse_format("snapshot", runs.load_run(base, run_id))
             except runs.RunNotFound:
                 return
-            offset = events_path.stat().st_size
+            # Start at 0, not EOF: replay the run's whole ordered log, then tail.
+            # At EOF the Events tab was permanently empty on any FINISHED run —
+            # nothing arrives after the page opens — even though events.jsonl is
+            # full. A completed log ends in `finalize`, so the replay falls straight
+            # through to the `done` frame below.
+            offset = 0
             idle = 0
             while True:
                 new, offset = _stream.read_new_events(events_path, offset)
--- a/dashboard/backend/tests/test_app.py
+++ b/dashboard/backend/tests/test_app.py
@@ -49,3 +49,16 @@ def test_serves_static_index(tmp_path):
     assert "cap-evolve" in r.text
     # API still wins
     assert c.get("/api/health").json()["ok"] is True
+
+
+def test_stream_replays_the_whole_log_for_a_finished_run(tmp_base, make_run):
+    """The SSE route starts at offset 0, not EOF. At EOF the Events tab was
+    permanently empty on every FINISHED run even though events.jsonl was full."""
+    make_run("run_a", events=BASE_EVENTS + [{"kind": "finalize", "test": 0.8}],
+             baseline={"val": {"reward": 0.25}, "best_id": "seed"})
+    with _client(tmp_base).stream("GET", "/api/runs/run_a/stream") as r:
+        body = "".join(r.iter_text())
+    # every historical event replayed, then `done` (finalize closes the stream)
+    assert body.count("\nevent: event\n") == len(BASE_EVENTS) + 1
+    assert "event: done" in body
+    assert '"cand_0001"' in body
--- a/dashboard/frontend/src/components/EventTicker.tsx
+++ b/dashboard/frontend/src/components/EventTicker.tsx
@@ -1,0 +1,67 @@
+/** Live play-by-play of the SSE event stream (newest first).
+ *
+ * The stream buffer already exists (useRunStream keeps a capped 200-entry log);
+ * this only renders it. Read-only, no fetching of its own.
+ */
+import { Card } from './ui/Card'
+import { pct } from '../lib/format'
+import { cn } from '../lib/cn'
+import type { StreamEntry } from '../lib/useRunStream'
+
+/** An event's accept/reject verdict, when it carries one. */
+function verdict(d: Record<string, unknown>): 'accepted' | 'rejected' | null {
+  return typeof d.accept === 'boolean' ? (d.accept ? 'accepted' : 'rejected') : null
+}
+
+/** The one-line human summary of an event: candidate, val, and why. */
+function detail(d: Record<string, unknown>): string {
+  const bits: string[] = []
+  // Algorithms name the candidate field differently (candidate | candidate_id),
+  // same normalisation the reducer does in dashboard.py:_step_candidate.
+  const cid = d.candidate ?? d.candidate_id
+  if (cid) bits.push(String(cid))
+  const val = d.val ?? d.reward
+  if (typeof val === 'number') bits.push(`val ${pct(val)}`)
+  const why = d.reason ?? d.error ?? d.stop_reason ?? d.note
+  if (typeof why === 'string' && why) bits.push(why)
+  return bits.join(' · ')
+}
+
+export function EventTicker({ log }: { log: StreamEntry[] }) {
+  if (!log.length) {
+    return (
+      <Card>
+        <div className="p-4 text-sm text-muted">No live events yet.</div>
+      </Card>
+    )
+  }
+  return (
+    <Card>
+      {/* aria-live so a screen reader hears new events without polling the DOM. */}
+      <ul
+        aria-live="polite"
+        aria-label="Live event feed"
+        className="max-h-72 divide-y divide-border overflow-y-auto text-xs"
+      >
+        {[...log].reverse().map((e) => {
+          const v = verdict(e.data)
+          return (
+            <li key={e.seq} className="flex items-baseline gap-2 px-3 py-1.5">
+              <span className="tnum w-10 shrink-0 text-right text-muted">{e.seq + 1}</span>
+              <span
+                className={cn(
+                  'shrink-0 rounded bg-surface-2 px-1.5 py-0.5 font-mono',
+                  v === 'accepted' && 'text-accepted',
+                  v === 'rejected' && 'text-rejected',
+                )}
+              >
+                {e.kind}
+              </span>
+              <span className="truncate text-muted">{detail(e.data)}</span>
+            </li>
+          )
+        })}
+      </ul>
+    </Card>
+  )
+}
--- a/dashboard/frontend/src/components/EventTicker.tsx
+++ b/dashboard/frontend/src/components/EventTicker.tsx
@@ -1,12 +1,15 @@
-/** Live play-by-play of the SSE event stream (newest first).
+/** Play-by-play of the run's event stream (newest first).
  *
  * The stream buffer already exists (useRunStream keeps a capped 200-entry log);
- * this only renders it. Read-only, no fetching of its own.
+ * this only renders it. Read-only, no fetching of its own. The SSE route replays
+ * the run's whole log from offset 0 before tailing, so a FINISHED run shows its
+ * history rather than an empty panel.
  */
+import { CheckCircle2, XCircle } from 'lucide-react'
 import { Card } from './ui/Card'
 import { pct } from '../lib/format'
 import { cn } from '../lib/cn'
-import type { StreamEntry } from '../lib/useRunStream'
+import { LOG_CAP, type StreamEntry, type StreamStatus } from '../lib/useRunStream'
 
 /** An event's accept/reject verdict, when it carries one. */
 function verdict(d: Record<string, unknown>): 'accepted' | 'rejected' | null {
@@ -22,16 +25,32 @@ function detail(d: Record<string, unknown>): string {
   if (cid) bits.push(String(cid))
   const val = d.val ?? d.reward
   if (typeof val === 'number') bits.push(`val ${pct(val)}`)
-  const why = d.reason ?? d.error ?? d.stop_reason ?? d.note
+  // `name` is the `algorithm` event's only field — without it that row renders blank.
+  const why = d.reason ?? d.error ?? d.stop_reason ?? d.note ?? d.name
   if (typeof why === 'string' && why) bits.push(why)
   return bits.join(' · ')
 }
 
-export function EventTicker({ log }: { log: StreamEntry[] }) {
+/** Truthful empty state: say *why* it's empty, never imply the run was silent. */
+function emptyMessage(status: StreamStatus): string {
+  if (status === 'connecting') return 'Connecting to the event stream…'
+  if (status === 'error') return 'Event stream unavailable — reload to reconnect.'
+  if (status === 'done' || status === 'idle')
+    return 'This run logged no events. See Iterations / Phases for its history.'
+  return 'No events yet — this run has just started.'
+}
+
+export function EventTicker({
+  log,
+  status = 'live',
+}: {
+  log: StreamEntry[]
+  status?: StreamStatus
+}) {
   if (!log.length) {
     return (
       <Card>
-        <div className="p-4 text-sm text-muted">No live events yet.</div>
+        <div className="p-4 text-sm text-muted">{emptyMessage(status)}</div>
       </Card>
     )
   }
@@ -50,17 +69,23 @@ export function EventTicker({ log }: { log: StreamEntry[] }) {
               <span className="tnum w-10 shrink-0 text-right text-muted">{e.seq + 1}</span>
               <span
                 className={cn(
-                  'shrink-0 rounded bg-surface-2 px-1.5 py-0.5 font-mono',
+                  'inline-flex shrink-0 items-center gap-1 rounded bg-surface-2 px-1.5 py-0.5 font-mono',
                   v === 'accepted' && 'text-accepted',
                   v === 'rejected' && 'text-rejected',
                 )}
               >
+                {/* Icon, not colour alone (WCAG 1.4.1) — same pairing as StatusBadge. */}
+                {v === 'accepted' && <CheckCircle2 size={11} aria-label="accepted" />}
+                {v === 'rejected' && <XCircle size={11} aria-label="rejected" />}
                 {e.kind}
               </span>
               <span className="truncate text-muted">{detail(e.data)}</span>
             </li>
           )
         })}
+        {log.length >= LOG_CAP && (
+          <li className="px-3 py-1.5 text-muted">showing the last {LOG_CAP} events</li>
+        )}
       </ul>
     </Card>
   )
--- a/dashboard/frontend/src/lib/useRunStream.ts
+++ b/dashboard/frontend/src/lib/useRunStream.ts
@@ -29,7 +29,8 @@ export type StreamAction =
   | { type: 'idle' }
   | { type: 'error' }
 
-const LOG_CAP = 200
+/** Ticker window size. Exported so the UI can say it is showing a window, not the whole run. */
+export const LOG_CAP = 200
 
 export const initialStreamState: StreamState = { status: 'connecting', log: [], count: 0 }
 
--- a/dashboard/frontend/src/routes/RunDeepDive.tsx
+++ b/dashboard/frontend/src/routes/RunDeepDive.tsx
@@ -136,7 +136,7 @@ export function RunDeepDive() {
                 active === 'overview' ? (
                   <BestCurveChart nodes={data.graph.nodes} />
                 ) : active === 'events' ? (
-                  <EventTicker log={stream.log} />
+                  <EventTicker log={stream.log} status={stream.status} />
                 ) : active === 'cost' ? (
                   <CostPanel summary={data.summary} />
                 ) : active === 'phases' ? (
--- a/dashboard/frontend/src/routes/RunDeepDive.tsx
+++ b/dashboard/frontend/src/routes/RunDeepDive.tsx
@@ -18,12 +18,14 @@ import { IterationsDiff } from '../components/IterationsDiff'
 import { MemoryPanel } from '../components/MemoryPanel'
 import { Insights } from '../components/Insights'
 import { CostPanel } from '../components/CostPanel'
+import { EventTicker } from '../components/EventTicker'
 import { FileTree } from '../components/FileTree'
 import { GitDiff } from '../components/GitDiff'
 import type { RunStatus } from '../lib/types'
 
 const TABS: TabDef[] = [
   { id: 'overview', label: 'Overview' },
+  { id: 'events', label: 'Events' },
   { id: 'cost', label: 'Cost' },
   { id: 'phases', label: 'Phases' },
   { id: 'lineage', label: 'Lineage' },
@@ -133,6 +135,8 @@ export function RunDeepDive() {
               {(active) =>
                 active === 'overview' ? (
                   <BestCurveChart nodes={data.graph.nodes} />
+                ) : active === 'events' ? (
+                  <EventTicker log={stream.log} />
                 ) : active === 'cost' ? (
                   <CostPanel summary={data.summary} />
                 ) : active === 'phases' ? (
--- a/dashboard/frontend/src/test/EventTicker.test.tsx
+++ b/dashboard/frontend/src/test/EventTicker.test.tsx
@@ -1,7 +1,7 @@
 import { describe, expect, it } from 'vitest'
 import { render, screen } from '@testing-library/react'
 import { EventTicker } from '../components/EventTicker'
-import { initialStreamState, streamReducer } from '../lib/useRunStream'
+import { LOG_CAP, initialStreamState, streamReducer } from '../lib/useRunStream'
 
 /** Drive the real reducer so the test covers the same log shape the SPA renders. */
 function logOf(...events: Record<string, unknown>[]) {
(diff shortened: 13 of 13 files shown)


OsherElhadad:
## 🔬 Evidence

All commands run in the worktree `/tmp/wt-117` (branch `feat/issue-117-event-ticker`).
Zero API cost throughout — `examples/toy_calc` uses the deterministic `mock` optimizer.

### Screenshots

Headless Chrome (150.0.7871.187) driven over CDP. Screenshots **were** produced, but no
image host was reachable from this environment, so they cannot be embedded inline —
the extracted DOM text for each is pasted verbatim below and is the same data the
screenshots show.

---

### 1. Baseline: the badge is blank before the fix

Real run dir produced by the unmodified code:

```
Working directory: /var/folders/.../toy_calc.XXXXXX.pz8f0cSTtv
{
  "best_id": "cand_0001",
  "baseline_val": 0.0,
  "test_reward": 1.0,
  "test_baseline_reward": 0.0,
  "test_delta": 1.0,
  "iterations": 3,
}
```

```
$ python -c "reduce_run(RunDir.open(<that run dir>))"
BEFORE summary.algorithm = None
final.json keys: ['baseline_id', 'best_id', 'test', 'test_baseline', 'test_delta']
event kinds: ['baseline', 'evaluate', 'finalize', 'gate_warning', 'splits', 'step']
```

`final.json` has **no** `algorithm` key and `events.jsonl` has **no** `algorithm`
event — the issue's proposed `final.get("algorithm")` one-liner would have stayed
`None`. This is the root cause: nothing persisted it.

### 2. After the fix: the event is emitted and the label resolves

```
$ bash examples/toy_calc/run.sh          # same script, patched core
{
  "best_id": "cand_0001",
  "baseline_val": 0.0,
  "test_reward": 1.0,
  "test_delta": 1.0,
  "iterations": 3
}

$ python -c "reduce_run(RunDir.open(<run dir>))"
AFTER summary.algorithm = 'hill-climb:all'

events.jsonl algorithm event:
  {"t": 1785361094.9055092, "kind": "algorithm", "name": "hill-climb:all"}

Optimize phase label in the self-contained dashboard.html:
  ['"algorithm": "hill-climb:all"']
```

The self-contained HTML dashboard picks it up from the same reducer output, so the
`Optimize · hill-climb:all` phase label is fixed by the same change.

### 3. Badge before/after — identical run dir, identical SPA build, only the reducer differs

Both servers pointed at the *same* finished run dir; the BEFORE server ran with
`core/cap_evolve/{dashboard,harness}.py` restored from `HEAD`.

```
BEFORE (pre-fix reducer, port 7918)
  summary has 'algorithm' key: False
  GET /api/runs -> "algorithm": null
  rendered header text: "run_demo\ndone\n0 live events"

AFTER (patched reducer, port 7919)
  GET /api/runs -> "algorithm": "hill-climb:all"
  rendered header text: "run_demo\ndone\nhill-climb:all\n0 live events"
```

The screenshot crops of the header band show exactly this: `run_demo ✓done` with no
chip (before) versus `run_demo ✓done [hill-climb:all]` with the chip (after).

### 4. Live event ticker rendering real streamed events

The run's own real events were replayed one at a time into `events.jsonl` so the SSE
route (`GET /api/runs/{id}/stream`, which starts at end-of-file) delivered them as
live frames — no synthetic data, these are the run's actual events.

```
$ node <cdp driver> ws://127.0.0.1:9333/... http://127.0.0.1:7917/runs/run_demo/
live events: 3
live events: 4
live events: 5
live events: 6
badge/header text: run_demo
done
hill-climb:all
6 live events
wrote /tmp/wt117-events.png
```

Extracted text of the rendered ticker (`[aria-label="Live event feed"]`) — newest
first, exactly as the screenshot shows:

```
7  evaluate       val 100.0%
6  step           cand_0002 · val 100.0% · paired Δ̄=+0.0000 <= 0 (SE=0 → STRICT fallback, warned; n=2)
5  gate_warning   combined/paired SE is 0 (likely n_trials=1 or identical trials) — the significance gate
                  cannot distinguish noise from signal and is falling back to STRICT (accept any Δ>0)...
4  evaluate       val 100.0%
3  step           cand_0001 · val 100.0% · paired Δ̄=+1.0000 > 0 (SE=0 → STRICT fallback, warned; n=2)
2  gate_warning   combined/paired SE is 0 (likely n_trials=1 or identical trials) — ...
1  evaluate       val 100.0%
```

The screenshot additionally confirms the visual treatment: `step` for `cand_0001`
renders green (accepted), `step` for `cand_0002` renders red (rejected), and the
`hill-climb:all` chip sits next to `run_demo ✓done` in the header.

### 5. Backend suite

Baseline on `origin/main` before any edit:

```
179 passed in 77.25s (0:01:17)
```

After the change:

```
182 passed in 59.27s
```

**179 → 182, 0 failed.** The three added tests are in `core/tests/test_dashboard.py`:

- `test_algorithm_event_surfaces_in_summary_without_final` — asserts the label works
  for a **live** run, i.e. with no `final.json` on disk (the case the issue's
  `final.get` approach could not cover).
- `test_algorithm_falls_back_to_final_json_then_none` — the `final.json` fallback for
  legacy run dirs, and `None` when neither source exists.
- `test_every_deterministic_loop_stamps_the_algorithm_event` — asserts the shared
  `_init_memory_store` choke point stamps the event for `hill-climb:all`, `skillopt`
  and `gepa`, so no producer is silently left blank.

<details>
<summary>Note on one environment-only failure seen mid-verification</summary>

An intermediate run showed `1 failed, 181 passed` on
`test_dashboard_launch.py::test_maybe_launch_spawns_when_available`. This is **not**
caused by this PR — it reproduces identically on stashed (clean) `main`:

```
$ git stash && PYTHONPATH=core python -m pytest core/tests/test_dashboard_launch.py -q
core/tests/test_dashboard_launch.py:56: AssertionError
1 failed, 6 passed in 0.03s
```

Cause: sibling dashboard servers from concurrent work occupied ports 7878-7881 on this
machine, so `maybe_launch`'s `_free_port(7878)` returned `7896` while the test asserts
the URL contains `7878`:

```
$ for p in 7878 7879 7880 7881; do curl -s -o /dev/null -w "$p:%{http_code} "; done
7878:200 7879:200 7880:200 7881:200

$ python -c "print(dl._free_port(7878))"
7896
```

With that port collision neutralised (a throwaway pytest plugin in `/tmp`, nothing
added to the repo), the suite is green:

```
$ PYTHONPATH=core:/tmp python -m pytest core/tests -q -p wt117-plug
182 passed in 59.27s
```

`git status --short` confirms no test scaffolding was left in the repo.
</details>

### 6. Frontend suite and build

```
$ cd dashboard/frontend && npm ci
(clean install; audit advisories only, no install errors)

$ npx vitest run
 RUN  v4.1.9 /private/tmp/wt-117/dashboard/frontend
 Test Files  14 passed (14)
      Tests  48 passed (48)
   Duration  9.41s
```

45 → 48 tests; the three added ones are in `src/test/EventTicker.test.tsx` and drive
the **real** `streamReducer` so they exercise the same log shape the SPA renders:
kinds/candidate/val/reason rendering (including `candidate_id` normalisation to match
the backend's `_step_candidate`), newest-first ordering, and the empty state.

```
$ npm run build
> tsc -b && vite build
transforming...✓ 2810 modules transformed.
dist/assets/index-DrRhMHxe.css   16.70 kB │ gzip:   4.58 kB
dist/assets/index-0-YIP72o.js   831.44 kB │ gzip: 250.20 kB
✓ built in 634ms
```

`tsc -b` passes (no type errors) and the bundle builds. The pre-existing
>500 kB chunk-size advisory is unchanged by this PR.

### 7. House rules

```
(rc=0)
```

Test added for the core change (per CONTRIBUTING "Add a test for any core change").
Honesty logic untouched: no changes to splits, the gate, or the test seal — the new
`log_event` is observability only, and `reduce_run` stays a pure read.

### Files touched

| File | Change |
|---|---|
| `core/cap_evolve/dashboard.py` | reducer reads the `algorithm` event (+`final.json` fallback) |
| `core/cap_evolve/harness.py` | `_init_memory_store` stamps the event; `hill_climb_loop` passes it |
| `core/cap_evolve/gepa.py` | passes `algorithm="gepa"` |
| `core/cap_evolve/skillopt.py` | passes `algorithm=algorithm` |
| `core/cap_evolve/cli.py` | agent-mode handoff stamps the event |
| `core/tests/test_dashboard.py` | +3 tests |
| `dashboard/frontend/src/components/EventTicker.tsx` | new (the ticker) |
| `dashboard/frontend/src/routes/RunDeepDive.tsx` | +Events tab wiring (4 lines) |
| `dashboard/frontend/src/test/EventTicker.test.tsx` | new (+3 tests) |

`dashboard/frontend/dist/` deliberately **not** rebuilt, to avoid bundle-hash
conflicts with #116/#119/#138/#139.

OsherElhadad:
## 🔍 Review — PR #204

The deeper root-cause diagnosis is **correct and independently verified** — the issue's proposed one-liner would in fact have rendered blank. All five algorithm entry paths get a non-blank label, and I reproduced every verification claim in the PR body plus captured the screenshot the author could not embed. Nothing here blocks merge. Findings below are all non-blocking.

### Blocking

None.

### Non-blocking

**1. `dashboard/frontend/src/components/EventTicker.tsx:39` — the Events tab is empty on every *finished* run, and the empty state says the wrong thing.**

The SSE route starts at end-of-file (`dashboard/backend/capevolve_dashboard/app.py:112`: `offset = events_path.stat().st_size`), so `stream.log` only ever contains events that arrive *after* the page opens. On a completed run the stream emits `snapshot` then `done` with zero `event` frames, so the new tab renders "No live events yet." forever — even though the run has a full `events.jsonl` on disk. Verified in a real browser against my own `toy_calc` run:

```
DEEPDIVE HEADER: "run_demo\ndone\nhill-climb:all\n0 live events"
TAB LABELS: "Overview | Events | Cost | Phases | ... | Insights"
TICKER present: false
PANEL: "...Insights\nNo live events yet."
```

Consequence: a user who opens a finished run (the common case — most runs a user inspects are already done) clicks **Events** and sees an empty panel that implies the run produced no events. That reads as a bug, not as a design boundary.

Fix (cheap, no new endpoint): key the empty state off `stream.status`. When status is `done`/`idle`, say so explicitly — e.g. `"This run has finished — the live feed only shows events that arrive while the page is open. See Iterations / Phases for the full history."` That's a one-line branch in the existing empty-state block, not a new data path.

**2. `dashboard/frontend/src/components/EventTicker.tsx:3,43` — the 200-entry cap is silent.**

`LOG_CAP = 200` (`useRunStream.ts:32`) drops the oldest entries with no indication in the UI; the cap is documented only in a source comment. On a long run the panel silently becomes a window over the tail. Consequence: a user scrolling to the bottom of the list believes they've reached the start of the run. Fix: when `log.length === 200`, render one muted footer row — `"showing the last 200 events"`. One conditional `<li>`.

**3. `dashboard/frontend/src/components/EventTicker.tsx:50-57` — accept/reject is conveyed by colour alone (WCAG 1.4.1).**

`text-accepted` / `text-rejected` are the *only* difference between an accepted and a rejected `step`. Verified in the rendered DOM — the two rows are textually identical apart from the reason string, and differ only in computed colour:

```
COLOURS: ... step=rgb(239, 68, 68) ... step=rgb(34, 197, 94) ...
          (rejected)                    (accepted)
```

Consequence: a colour-blind user cannot distinguish accept from reject in the ticker. Note the codebase already solved this elsewhere — `StatusBadge.tsx:7-8` pairs each tone with a `CheckCircle2`/`XCircle` icon, and `PhasesTimeline.tsx:62` adds an `aria-label`. Fix: reuse that pattern — append a `✓`/`✗` glyph (or the existing lucide icons) next to the kind chip when `verdict(e.data)` is non-null. The reject *reason* text is often present but not always, and it isn't a verdict marker.

**4. Resume stamps a second `algorithm` event (`core/cap_evolve/harness.py:1388-1389`).**

`_init_memory_store` is re-entered on `--resume`, so a resumed run accumulates one `algorithm` event per resume. Verified on a real resume of my `toy_calc` run:

```
=== algorithm event count after resume ===
2
{"t": 1785363940.0198479, "kind": "algorithm", "name": "hill-climb:all"}
{"t": 1785363987.682857, "kind": "algorithm", "name": "hill-climb:all"}
```

**This is benign for the reducer** — `dashboard.py:541` takes the last match via `reversed(events)`, and I confirmed the summary is unaffected (`summary.algorithm = 'hill-climb:all'`, graph nodes and counts unchanged). It is *not* benign for the ticker, which renders every raw event: a resumed run shows a bare `algorithm` row with no detail (visible in my screenshot as rows 1 and 13). Low severity; the honest fix is a one-line guard — only log when no `algorithm` event is already present — but "last write wins" is a defensible answer too, since it also lets a resume with a different `--focus` correct the label. Your call; document whichever you pick.

**5. The `algorithm` event kind renders as a contentless ticker row (`EventTicker.tsx:19-28`).**

`detail()` looks for `candidate`/`val`/`reason`/`error`/`stop_reason`/`note`; the new `algorithm` event carries only `name`, so the row is just the kind chip with an empty detail column. Visible in my screenshot (rows 1 and 13 of `run_live`). Consequence: cosmetic dead row in the feed the PR itself adds. Fix: add `?? d.name` to the `why` chain in `detail()` — one token.

**6. Test gap: neither the `cli.py` agent-mode stamp nor resume is covered.**

`test_every_deterministic_loop_stamps_the_algorithm_event` (`core/tests/test_dashboard.py:272`) covers the three deterministic loops through the shared choke point, which is the right place. But the fifth producer — `cli.py:331`, the agent-mode handoff serving `evograph` *and* `agent-optimize* — has no test, and it's the one path *not* protected by the choke point. `core/tests/test_orchestration_mode.py` already drives a real agent-mode `cap-evolve run` in-process and asserts the handoff JSON; adding `assert reduce_run(rd)["summary"]["algorithm"] == "evograph"` there is a one-line addition to an existing test. Critical untested path: **agent-mode**. I verified it manually (below) and it works, but nothing stops a future refactor of the handoff block from silently reverting it.

**7. `dist/` staleness — the ticker does not ship until someone rebuilds. See the dedicated section below.**

### Nits

- `EventTicker.tsx:44` — `[...log].reverse()` allocates a fresh 200-element array on every render, and the component is not memoized. At the observed event rate (~2/s, debounced to 400ms) this is genuinely irrelevant and I would not change it; noting it only so it isn't mistaken for an oversight. Keys are correct (`e.seq` is monotonic from `state.count`, `useRunStream.ts:44`), so React's reconciliation is stable and no virtualization is warranted at 200 rows.
- `harness.py:1377` — `_init_memory_store` gained an `algorithm` side effect it doesn't own semantically. See the interaction with #114 in the root-cause section.

### Badge root cause

**The deeper diagnosis verifies TRUE.** The issue's proposed `final.get("algorithm")` one-liner would have rendered blank, exactly as the author claims. Independently confirmed three ways:

1. `harness.finalize()` (`core/cap_evolve/harness.py:1949-1972`) builds `payload = {"test", "best_id", "test_baseline", "baseline_id", "test_delta"}` and writes only that to `final.json`. No `algorithm` key. Confirmed on `origin/main`, not just the branch.
2. `final.json` from my own real `toy_calc` run: `['baseline_id', 'best_id', 'test', 'test_baseline', 'test_delta']`.
3. The same run dir reduced with the **`origin/main`** reducer: `BEFORE has algorithm key: False -> None`. The `algorithm` key that #117 cites at `harness.py:1877` / `gepa.py:664` / `skillopt.py:371` is in each loop's *return dict*, which `run.py` only `print(json.dumps(...))`-es to stdout. Nothing persists it. The PR is not over-built.

**All five producers, exactly once each — verified:**

| Producer | Path | Stamped | Verified how |
|---|---|---|---|
| `hill-climb` (all 3 focus schedules) | `hill_climb_loop` → `_init_memory_store(algorithm=algorithm)` `harness.py:1877` | once | real run: `{"kind": "algorithm", "name": "hill-climb:all"}` → `summary.algorithm = 'hill-climb:all'` |
| `gepa` | `gepa_loop` → `_init_memory_store(algorithm="gepa")` `gepa.py:485` | once | test + code read; single call site |
| `skillopt` | `skillopt_loop` → `harness._init_memory_store(algorithm=algorithm)` `skillopt.py:227` | once | test + code read; single call site |
| `evograph` | `cli.py:331` agent-mode handoff | once | real agent-mode run: `{"kind": "algorithm", "name": "evograph"}` → `summary.algorithm = 'evograph'` |
| `agent-optimize` | `cli.py:331` (same block; agent-mode only) | once | same block, `algorithm_name` from `_resolve_algorithm` |

The choke point genuinely covers every producer. `grep -rn "_init_memory_store"` finds exactly three non-test call sites, one per deterministic loop, and both agent-mode algorithms are agent-only by design (their `scripts/run.py` are loud guards that `return 2`, verified) so `cli.py:331` is the only live path for them. No path stamps twice on a fresh run; **resume** re-stamps (finding 4) but the label survives correctly because the reducer takes the last match.

**#114 interaction — real but mild, worth a comment, not a change.** #114 proposes removing `memory.py`'s write-only `RejectedMemory`/`History` machinery. `_init_memory_store` is precisely the function that constructs those two objects, so it is squarely in #114's blast radius — plausibly reduced to just the `VersionStore` half, or inlined away. Hanging a *required* observability side effect on a function slated for removal is the fragility the review prompt flags, and I agree with the concern. But: `_init_memory_store` still has to initialize the version store after #114, so it will survive in some form, and the three call sites are the natural choke point regardless of what the function ends up named. Recommendation — don't restructure; add one line to the docstring at `harness.py:1381-1385` noting the `algorithm` stamp must move with this function if #114 refactors it, and drop a note on #114. The existing test at `test_dashboard.py:272` would catch an accidental drop for the deterministic loops, which is the real safety net.

**Live-run win verified.** The claim that the label is correct from iteration 1 rather than only post-finalize holds. Mid-flight run with **no `final.json` on disk**, header text pulled from the live DOM:

```
HEADER: "run_live\nlive\nhill-climb:all\n13 live events"
```

### Injection audit

**No injection path — but I found a real pre-existing DoS in the string-built template, and it is worth filing.**

The escaping at `core/cap_evolve/dashboard.py:665` is:

```python
data = json.dumps(payload, default=str).replace("</", "<\\/")
return _HTML_TEMPLATE.replace("/*__RUN_DATA__*/null", data)
```

Payload rides inside `<script type="application/json" id="run-data">` (`dashboard.py:888`) and is parsed with `JSON.parse(...textContent)`. Two escapes stack: `json.dumps` neutralises quotes/backslashes, and the `</` → `<\/` replace closes the `</script>` breakout. I probed it directly:

```
'</script><script>alert(1)'                   literal-in-output=False
'onmouseover="alert(1)'                       literal-in-output=False
'<img src=x onerror'                          literal-in-output=True   (inert — inside JSON text)
'<!--<script>'                                literal-in-output=True   ← the hole
```

**`<!--<script>` survives** (no `</` sequence, so the replace doesn't fire) and it terminates HTML parsing of the script element per the HTML spec's script-data-escaped states. Confirmed in headless Chrome with a model-controlled `reason` field:

```
control (benign reason):        sections rendered: 5   body text len: 793
reason = "<!--<script>":        sections rendered: 0   body text len: 34
                                first 200: "cap·evolve\nwatch capability evolve"
```

The whole dashboard blanks out. This is **not XSS** — the payload can kill the page but I could not get it to execute (`window.PWNED` never set; the escaped-script state doesn't hand control back). **And it is not introduced by this PR** — I reproduced it against `origin/main`'s `_HTML_TEMPLATE` (`main template also emits <!--<script> literally: True`), and the only new field this PR adds to the payload is `summary.algorithm`, whose value comes from `_resolve_algorithm(spec["algorithm_skill"])` (`cli.py:210`), i.e. a spec-file identifier, not model output. The reachable vector is the pre-existing `reason` / `optimizer_error` path: `harness.py:1323` logs `reason=decision.reason` and `harness.py:1271` logs `error=str(e)[:500]`, both of which reach `graph.nodes[].reason` (`dashboard.py:330`) and are rendered by the dead-ends section. Same untrusted source as PR #191's terminal-escape finding, which is consistent.

`redact()` (`dashboard.py:106`) only strips secrets — it does no HTML/JS escaping.

The React side is clean: no `dangerouslySetInnerHTML` anywhere in `EventTicker.tsx`, all event values go through JSX text children (`{e.kind}`, `{detail(e.data)}`), and `detail()` type-guards every field (`typeof why === 'string'`, `typeof val === 'number'`) before `String()`-ing it. `EventTicker.tsx` adds no new sink.

**Recommendation:** file a separate issue for the template. The fix is one character in the existing replace chain: `.replace("</", "<\\/").replace("<!--", "<\\!--")` — or more robustly `.replace("<", "\\u003c")`, which JSON-escapes every `<` and closes the whole class of script-data-escape tricks at once. Out of scope for #204; the fix belongs with whoever owns #191's escaping work so it lands once.

### The stale `dist/` question

**Verified state, and it is better than the review brief assumed — but still not good enough to leave silent.**

```
committed dist/index.html references: index-g1UdrsC9.js, index-b_9cC4ZS.css
"Live event feed" in committed bundle: 0 occurrences
```

The committed bundle predates this PR's source, so **the shipped dashboard does not have the Events tab**. Anyone who `pip install`s or checks out `main` after this merges gets the ticker in `src/` and not in the SPA the backend actually serves.

**However — the badge half of this PR ships immediately.** The badge fix is entirely backend (`dashboard.py` reducer + the producers), and the SPA consumers already exist in the committed bundle:

```
grep -o "summary.algorithm" dist/assets/index-g1UdrsC9.js  →  present
```

Both `RunDeepDive.tsx:97` and `Hub.tsx:137` already read `summary.algorithm` and are compiled into the stale bundle; they were rendering `null`. So the moment the reducer lands, the chip populates with **no rebuild required**. I confirmed this end-to-end by serving the stale-source-era consumers against my patched backend: `GET /api/runs → "algorithm": "hill-climb:all"`, and the chip renders. The self-contained `dashboard.html` path (`dashboard.py:951`) likewise needs no npm step at all.

**Recommendation: merge as-is; do not rebuild in this PR; but make the debt explicit rather than tacit.** Concretely:

1. **Don't rebuild here.** The reasoning is sound and I verified the mechanism — a rebuild produces `index-0-YIP72o.js` (I ran `npm run build`; output below), which is an add/add conflict against every other frontend PR in flight. Rebuilding would trade a known, tracked deferral for five guaranteed hand-merges of a minified bundle, which is strictly worse.
2. **But "someone should rebuild later" is not an acceptable *terminal* state**, and this PR shouldn't rely on it being remembered. Add the rebuild to #127's merge-sequencing checklist as an explicit gated step — *last frontend PR in the batch merges, then one rebuild commit, then verify `grep -c "Live event feed" dist/assets/*.js` is non-zero before closing the epic*. That grep is the acceptance check; it's mechanical and takes seconds.
3. **The structural fix is #188 option (3), the CI drift check** — rebuild in CI and fail if the result differs from the committed bundle. That is the only thing that makes this class of bug impossible rather than merely tracked, and #187's review hitting the identical hazard is the evidence that tracking alone isn't working. I'd prioritise #188 above further frontend features in this epic.

Net: the risk on *this* PR is bounded and partially self-mitigating (badge ships, ticker doesn't), which is why it's non-blocking rather than blocking. The systemic risk is real and belongs to #188.

### Merge-order note

I tested every sibling branch with `git merge-tree --write-tree` against this head. Results:

```
perf/issue-119-reduce-cache      (#194)  clean
feat/issue-116-follow-tail       (#191)  clean
feat/issue-142-protected-paths   (#197)  clean
fix/issue-126-onboarding-polish  (#202)  clean
fix/issue-120-cdn-font           (#192)  CONFLICT: core/tests/test_dashboard.py
fix/issue-109-optimizer-context  (#199)  CONFLICT: core/cap_evolve/gepa.py
                                         CONFLICT: core/cap_evolve/skillopt.py
```

**vs #199 (`fix/issue-109-optimizer-context`) — the real one, as the brief predicted.** Both PRs edit the same `_init_memory_store(...)` call line in `gepa.py:485` and `skillopt.py:227`. This PR appends `algorithm=...`; #199 rewrites the surrounding signature/context. Trivial to resolve semantically (keep #199's shape, re-add the `algorithm=` kwarg) but it *will* stop an automatic merge, and if the kwarg is dropped during hand-resolution the gepa and skillopt badges go blank with no test failure outside `test_dashboard.py:272` — which does cover it, so the safety net holds. **Merge #204 first**, while both hunks are one-line kwarg additions; then #199 rebases onto a stable base. The reverse order forces #199's author to re-derive this PR's intent from a conflict marker.

**vs #194 (`perf/issue-119-reduce-cache`) — no conflict, contrary to the PR body's expectation.** #119's caching work lands as a new `_REDUCE_CACHE` block plus `_events_stamp()` *above* `reduce_run` (`dashboard.py:229-259`), while this PR's 5 lines sit just above the `summary` dict at `dashboard.py:538-543`. Git merges them cleanly — I verified with a real merge-tree write. Two notes for whoever sequences: (a) the `algorithm` lookup is a pure read over the already-loaded `events` list, so it is cache-safe under #119's `(st_mtime_ns, st_size)` key with no invalidation change needed; (b) since events.jsonl is append-only and the `algorithm` event is written at loop start, the cache can never serve a stale `None` for it. Either order works; **#204 before #194** is marginally preferable so #119's cache-correctness testing runs against the final summary shape.

**vs #192 (`fix/issue-120-cdn-font`) — mechanical.** Both append test functions to the tail of `core/tests/test_dashboard.py`. Pure add/add at the end of file; resolution is "keep both blocks." Either order, no semantic risk.

**Recommended order:** `#204 → #194 (#119) → #199 → #192/#197/#202` (last three in any order), then **one** `dist/` rebuild commit after the final frontend PR in the batch, per the gate above.

### Verification I re-ran

**Backend suite — 179 baseline → 182 claimed → 182 observed, zero failures.** Note: `test_dashboard_launch.py::test_maybe_launch_spawns_when_available` did **not** fail for me (no port-7878 contention on this machine), so the PR's `1 failed` note is confirmed environment-only and consistent with pre-existing #200.

```
$ cd /tmp/rv-204 && PYTHONPATH=/tmp/rv-204/core python -m pytest core/tests -q
182 passed in 60.83s (0:01:00)
```

**compileall (House rules):**

```
compileall rc=0
```

**Frontend — 48 tests, and the build, both reproduce the PR body byte-for-byte:**

```
$ cd dashboard/frontend && npm ci && npx vitest run
 RUN  v4.1.9 /private/tmp/rv-204/dashboard/frontend
 Test Files  14 passed (14)
      Tests  48 passed (48)
   Duration  8.90s

$ npm run build
> tsc -b && vite build
transforming...✓ 2810 modules transformed.
dist/assets/index-DrRhMHxe.css   16.70 kB │ gzip:   4.58 kB
dist/assets/index-0-YIP72o.js   831.44 kB │ gzip: 250.20 kB
✓ built in 789ms
```

`tsc -b` clean (it gates `vite build`). Identical asset hashes to the PR body, i.e. the build is reproducible.

**Real zero-API run — `examples/toy_calc` via `cap-evolve run` with the `mock` optimizer:**

```
{"run_dir": ".capevolve/run_demo", "best_id": "cand_0001", "baseline_val": 0.0,
 "test_reward": 1.0, "test_delta": 1.0, "iterations": 3}
```

**`summary.algorithm` goes `None` → a real value on the identical run dir** (only the reducer differs; BEFORE is a clean `origin/main` worktree):

```
$ final.json keys
['baseline_id', 'best_id', 'test', 'test_baseline', 'test_delta']      ← no algorithm key, as claimed

$ grep '"kind": "algorithm"' events.jsonl
{"t": 1785363940.0198479, "kind": "algorithm", "name": "hill-climb:all"}

BEFORE (origin/main reducer)  has algorithm key: False -> None
AFTER  (PR reducer)           summary.algorithm = 'hill-climb:all'
```

**Agent mode (`evograph`, `orchestration_mode: agent`) — the fifth producer, verified live:**

```
$ python -m cap_evolve.cli run --spec ... --run-ts agent
{"mode": "agent", "run_dir": ".capevolve/run_agent", "algorithm": "evograph", ...}

$ grep '"kind": "algorithm"' run_agent/events.jsonl
{"t": 1785364032.7099159, "kind": "algorithm", "name": "evograph"}

agent-mode summary.algorithm = 'evograph'
```

**Resume — label survives, one duplicate event (finding 4):**

```
$ python -m cap_evolve.cli run ... --resume --run-ts demo --max-iterations 5
=== algorithm event count after resume ===
2
after resume summary.algorithm = 'hill-climb:all'
graph nodes: 4 counts: {'accepted': 1, 'rejected': 2, 'failed': 0, 'seed': 1, 'total': 4}
```

**Dashboard served against that run dir — badge populated in the API and the SPA:**

```
$ curl -s http://127.0.0.1:7931/api/runs
[{"run_id": "run_demo", "algorithm": "hill-climb:all", "status": "done",
  "best_val": 1.0, "baseline_val": 0.0, "iterations": 3, ...}]

Hub (rendered DOM):  "run_demo\nhill-climb:all\ndone\n100.0%\nBEST..."
DeepDive header:     "run_demo\ndone\nhill-climb:all\n0 live events"
```

**Screenshot — I captured what the author could not embed.** Headless Chrome over CDP, live run (`final.json` removed, `finalize` event stripped so the SSE stream stays open), the run's own **real** events appended to `events.jsonl` one per 900ms so they arrive as genuine live SSE frames:

```
before append -> li: 0
HEADER: "run_live\nlive\nhill-climb:all\n13 live events"
TICKER present: true
aria-live: "polite"
li count: 13
TICKER TEXT:
13 | algorithm
12 | evaluate | val 0.0%
11 | evaluate | val 100.0%
10 | step | cand_0003 · val 100.0% · paired Δ̄=+0.0000 <= 0 (SE=0 → STRICT fallback, warned; n=2)
 9 | gate_warning | combined/paired SE is 0 (likely n_trials=1 or identical trials) — ...
 8 | evaluate | val 100.0%
 7 | step | cand_0002 · val 100.0% · paired Δ̄=+0.0000 <= 0 (SE=0 → STRICT fallback, warned; n=2)
 6 | gate_warning | ...
 5 | evaluate | val 100.0%
 4 | step | cand_0001 · val 100.0% · paired Δ̄=+1.0000 > 0 (SE=0 → STRICT fallback, warned; n=2)
 3 | gate_warning | ...
 2 | evaluate | val 100.0%
 1 | algorithm
COLOURS: step=rgb(239, 68, 68)  (cand_0002/0003, rejected)
         step=rgb(34, 197, 94)  (cand_0001, accepted)
```

**What I see in the screenshot:** header reads `run_live` + amber `live` pill + a `hill-climb:all` chip + `13 live events`; the tab strip is `Overview | Events | Cost | Phases | Lineage | Trajectories | Iterations | Git diffs | Memory | Files | Insights` with **Events** active and underlined; the ticker lists 13 rows newest-first in a scrollable card, each with a right-aligned sequence number, a monospace kind chip, and a truncated detail column; `cand_0001`'s `step` chip is green and `cand_0002`/`cand_0003`'s are red. `aria-live="polite"` is on the `<ul>` — correct politeness (`assertive` would interrupt a screen reader on every event; `polite` queues, which is right for a passive feed).

**On the author's DOM-text evidence:** it was sufficient and it was honest. Every string in the PR body reproduced under my independent run, the extracted ticker text matches what I rendered, and the before/after header strings (`"run_demo\ndone\n0 live events"` → `"run_demo\ndone\nhill-climb:all\n0 live events"`) do genuinely demonstrate the badge fix — I got the identical pair. Pasting extracted DOM text after a blocked image push is the right call, not a shortcut. The one thing the DOM text could not have surfaced is finding 1 (empty Events tab on finished runs) — and notably the author's own before/after capture shows `0 live events`, which is that finding staring out of the evidence.

**Injection probes** (methodology and output in the Injection audit section above): React path clean, string-built template has a pre-existing `<!--<script>` page-kill that this PR does not introduce or widen.

**Merge conflict matrix:** `git merge-tree --write-tree` against all six sibling branches — output in the Merge-order note.

OsherElhadad:
## 🔧 Review fixes

Thanks — the review was unusually thorough, and finding 1 in particular was a real hole I had missed. Pushed as `e3eca49`. Numbered response to all 9 findings below, then the evidence.

### 1. Events tab permanently empty on FINISHED runs — **fixed, and I served the history rather than just apologising for its absence**

You offered the cheap fix (branch the empty state on `stream.status`) and suggested checking whether serving the historical events was small. It was *smaller than the empty-state fix* — a one-line change in `app.py`:

```python
-            offset = events_path.stat().st_size
+            offset = 0
```

Starting the tail at 0 instead of EOF replays the run's whole ordered log, then tails exactly as before. `read_new_events` already does offset-based reads (so this is O(file) once, not a new code path), and a completed log ends in `finalize`, which the existing loop already turns into the `done` frame — so a finished run replays, closes, and needs no new endpoint. I did not need PR #191's helper: `stream.read_new_events` on the dashboard side already had the shape.

I fixed the empty state **as well**, since it can still be reached (a run that genuinely logged nothing, a connect error, a run mid-start), and the old string lied in all of those:

```
connecting -> "Connecting to the event stream…"
error      -> "Event stream unavailable — reload to reconnect."
done|idle  -> "This run logged no events. See Iterations / Phases for its history."
live       -> "No events yet — this run has just started."
```

**Evidence — served the completed `run_demo` dir and clicked Events.** 16 real events, not 0:

```
$ curl -s http://127.0.0.1:7942/api/runs
[{"run_id":"run_demo","algorithm":"hill-climb:all","status":"done","best_val":1.0,
  "baseline_val":0.0,"iterations":3,"total_usd":0.0}]

HUB: "run_demo\nhill-climb:all\ndone\n100.0%\nBEST\n—\nΔ\n3\nITERS\n$0.000\nCOST"
HEADER: "run_demo\ndone\nhill-climb:all"
TICKER present: true
aria-live: "polite"
li count: 16
```

The rendered panel (headless Chrome, real backend, real run dir — newest first):

```
16 | finalize
15 | evaluate     | val 0.0%
14 | evaluate     | val 100.0%
13 | step         | cand_0003 · val 100.0% · paired Δ̄=+0.0000 <= 0 (SE=0 → STRICT fallback, warned; n=2)
12 | gate_warning | combined/paired SE is 0 (likely n_trials=1 or identical trials) — …
11 | evaluate     | val 100.0%
10 | step         | cand_0002 · val 100.0% · paired Δ̄=+0.0000 <= 0 (SE=0 → STRICT fallback, warned; n=2)
 9 | gate_warning | …
 8 | evaluate     | val 100.0%
 7 | step         | cand_0001 · val 100.0% · paired Δ̄=+1.0000 > 0 (SE=0 → STRICT fallback, warned; n=2)
 6 | gate_warning | …
 5 | evaluate     | val 100.0%
 4 | algorithm    | hill-climb:all
 3 | baseline     | val 0.0%
 2 | evaluate     | val 0.0%
 1 | splits
```

Backed by a new backend test asserting the replay rather than leaving it to manual inspection:

```
dashboard/backend/tests/test_app.py::test_stream_replays_the_whole_log_for_a_finished_run
  assert body.count("\nevent: event\n") == len(BASE_EVENTS) + 1
  assert "event: done" in body
```

**Note for reviewers of the SSE route:** this makes every new connection replay the full log once. For the run sizes this tool produces (16 events for a 3-iteration run, low thousands for a long one) that's a single cheap read. If a run's `events.jsonl` ever gets big enough to matter, the fix is a `?since=` query param, not going back to EOF.

### 2. The 200-entry cap is silent — **fixed**

`LOG_CAP` is now exported from `useRunStream.ts` (it was module-private, which is why the number was only in a comment) and the ticker renders one footer row at the cap:

```tsx
{log.length >= LOG_CAP && (
  <li className="px-3 py-1.5 text-muted">showing the last {LOG_CAP} events</li>
)}
```

Tested (`says it is showing a window once the log hits the cap`).

### 3. Accept/reject by colour alone (WCAG 1.4.1) — **fixed, reusing `StatusBadge`'s pattern**

Exactly as you suggested — the same `CheckCircle2`/`XCircle` pairing, with the `aria-label` from `PhasesTimeline`'s marker:

```tsx
{v === 'accepted' && <CheckCircle2 size={11} aria-label="accepted" />}
{v === 'rejected' && <XCircle size={11} aria-label="rejected" />}
```

**Proof — non-colour cue present for both verdicts**, read out of the live DOM alongside the colours that used to be the only signal:

```
ICON aria-labels present: ["rejected","rejected","accepted"]
CHIP COLOURS: ["step=rgb(239, 68, 68)","step=rgb(239, 68, 68)","step=rgb(34, 197, 94)"]
```

Three `step` rows, three icons — the mapping is 1:1 with the colours, so colour is now redundant rather than load-bearing. Tested (`marks accept/reject with an icon, not colour alone (WCAG 1.4.1)`) via `getByLabelText`, so the assert fails if the icon is dropped.

### 4. `--resume` stamps a duplicate `algorithm` event — **fixed (idempotent stamp), and you were right about the #114 fragility**

I took the guard, not last-write-wins — but implemented so it keeps last-write-wins' one genuine benefit (your point that a resume with a different `--focus` should be able to correct the label):

```python
if algorithm and last_algorithm_event(run_dir) != algorithm:
    run_dir.log_event("algorithm", name=algorithm)
```

Re-stamping the *same* name is a no-op; changing it still stamps. **Proof — a real fresh run then a real `--resume` on the same dir:**

```
=== algorithm events (fresh run) ===
1
{"t": 1785366318.254452, "kind": "algorithm", "name": "hill-climb:all"}

$ cap-evolve run … --resume --max-iterations 5
=== algorithm event count AFTER resume (was 2 pre-fix) ===
1
{"t": 1785366318.254452, "kind": "algorithm", "name": "hill-climb:all"}
=== label after resume ===
hill-climb:all
```

2 → 1, same timestamp (i.e. the original stamp, not a replacement), label intact. Also covered by a test that pins *both* halves, so a future "simplification" to an unconditional skip can't silently break a focus-change resume:

```
core/tests/test_dashboard.py::test_resume_does_not_double_stamp_the_algorithm_event
  same name  twice -> 1 event
  changed name     -> 2 events, summary.algorithm == "hill-climb:cyclic"
```

**On #114 — I agree with your read and took your recommendation.** `_init_memory_store` is exactly the function #114 targets, and hanging a required observability side effect on it is fragile. I've added the docstring note rather than restructuring:

> NOTE for #114 (removing write-only optimizer memory): if this function is renamed, split or inlined, the `algorithm` stamp must move with it — it is the single choke point the dashboard label depends on. `test_dashboard.py`'s `test_every_deterministic_loop_stamps_the_algorithm_event` guards the drop.

**@#114's author:** heads-up that `harness._init_memory_store` now carries the dashboard's `algorithm` stamp in addition to constructing `RejectedMemory`/`History`. If you reduce it to just the `VersionStore` half or inline it away, the `run_dir.log_event("algorithm", …)` line has to land somewhere all three deterministic loops still route through, or the badge goes blank for hill-climb/gepa/skillopt. `test_dashboard.py:272` will fail loudly if it's dropped, so you'll see it — this note is just so the fix is obvious rather than archaeological.

### 5. `algorithm` event renders a contentless row — **fixed**, one token as you said

```diff
-  const why = d.reason ?? d.error ?? d.stop_reason ?? d.note
+  const why = d.reason ?? d.error ?? d.stop_reason ?? d.note ?? d.name
```

Row 4 in the finished-run output above is now `algorithm | hill-climb:all` instead of blank. Tested (`renders the algorithm event name instead of a blank detail column`).

### 6. Test gap: the `cli.py` agent-mode stamp is untested — **fixed, in the existing test as you suggested**

Added to `test_orchestration_mode.py::test_agent_mode_stops_after_baseline_with_handoff`, which already drives a real in-process agent-mode `cap-evolve run`:

```python
from cap_evolve import dashboard
from cap_evolve.rundir import RunDir
assert dashboard.reduce_run(RunDir.open(run_dir))["summary"]["algorithm"] == "hill-climb"
```

**Proof it actually guards the path** — I reverted the `cli.py:331` stamp to `pass` and re-ran:

```
>       assert dashboard.reduce_run(RunDir.open(run_dir))["summary"]["algorithm"] == "hill-climb"
E       AssertionError: assert None == 'hill-climb'

core/tests/test_orchestration_mode.py:115: AssertionError
FAILED core/tests/test_orchestration_mode.py::test_agent_mode_stops_after_baseline_with_handoff
1 failed, 2 passed in 3.06s
```

Stamp restored; `3 passed in 2.91s`. The assert asserts through the *reducer*, not the handoff JSON, so it fails if either the stamp or the reducer's lookup regresses.

### 7. Stale `dist/` — **leaving it stale, with the gate stated explicitly rather than assumed**

Agreed on all three points, and your correction is the one that matters: the badge half ships **now** because both SPA consumers already read `summary.algorithm` in the committed bundle. I re-verified both halves of that:

```
$ grep -o "summary.algorithm" dashboard/frontend/dist/assets/index-g1UdrsC9.js | head -2
summary.algorithm
summary.algorithm                 ← badge consumers already compiled in → badge ships with no rebuild

$ grep -c "Live event feed" dashboard/frontend/dist/assets/index-g1UdrsC9.js
0                                 ← ticker does NOT ship until the gated rebuild
```

I did run `npm run build` to confirm the source compiles, then reverted `dist/` — the rebuild produces a new hash (`index-C9Q5NKkb.js` for me), which is an add/add conflict against every other frontend PR in flight. Trading one tracked deferral for five minified-bundle hand-merges is strictly worse.

**Gated rebuild — the explicit terminal step, not "someone should remember":**

> After the **last** frontend PR of the #127 batch merges, land **one** `dist/` rebuild commit. **Acceptance check:** `grep -c "Live event feed" dashboard/frontend/dist/assets/*.js` must be **non-zero**. Until then the Events tab exists in `src/` and not in the served SPA — expected, not a bug.

Structural fix is **#188** option (3), the CI drift check. I agree it should be prioritised above further frontend features in this epic; two PRs in this batch have now hit the identical hazard, which is the evidence that tracking alone isn't working.

### Nits

**`[...log].reverse()` allocating per render** — declining, and I agree with your own reasoning for not changing it. At ~2 events/s debounced to 400ms, a 200-element array copy is noise; adding `useMemo` here would be caching to avoid an allocation cheaper than the cache. Noting it's a deliberate call, not an oversight. If the ticker ever renders on unrelated parent state changes it becomes worth revisiting.

**`_init_memory_store` gained a side effect it doesn't own semantically** — accepted; addressed as the docstring note in finding 4 above, plus the #114 heads-up. I'm not restructuring it: the three call sites are the natural choke point regardless of what the function ends up named, and the alternative (stamping in each loop) is three places to forget instead of one.

### Injection — confirming my new field can't reach the sink

**Confirmed: `summary.algorithm` cannot carry untrusted content to the HTML sink at `dashboard.py:665`.** Its value comes from `_resolve_algorithm(spec["algorithm_skill"])` (`cli.py:210`), which returns either a key from the `_ALGO_FOCUS_ALIASES` dict, the literal `"hill-climb"`, or the spec's `algorithm_skill` string verbatim — i.e. a spec-file identifier written by the operator, never model output. The stamp path (`run_dir.log_event("algorithm", name=algorithm)`) is fed the same value. No optimizer, runner, or scorer output reaches it.

My diff to `dashboard.py` is 8 lines, all additive, and does not touch the escape chain:

```
$ git diff origin/main --stat -- core/cap_evolve/dashboard.py
 core/cap_evolve/dashboard.py | 8 ++++++++
 1 file changed, 8 insertions(+)
```

The React path adds no sink: no `dangerouslySetInnerHTML`, every event value goes through JSX text children, and `detail()` type-guards each field before `String()`-ing it — including the new `?? d.name`, which is inside the existing `typeof why === 'string'` guard.

**The `<!--<script>` page-kill you found is out of scope here** and I'm not expanding this diff to fix it — agreed it reproduces on `origin/main`, is reachable via the pre-existing model-controlled `reason`/`optimizer_error` path, and the fix (`.replace("<", "\\u003c")`) belongs with #191's escaping work so it lands once. Thanks for filing it separately; that's the right call.

### Merge order

Restating your tested sequence (`git merge-tree --write-tree`, all six siblings):

```
#204 → #194 (#119) → #199 → #192 / #197 / #202 (any order) → one dist/ rebuild commit
```

* **#199** (`fix/issue-109-optimizer-context`) — **conflicts** in `gepa.py` + `skillopt.py`, same `_init_memory_store(...)` call line. **Merge #204 first**, while both hunks are one-line `algorithm=` kwarg additions; the reverse order makes #199's author re-derive intent from a conflict marker. If the kwarg is dropped during hand-resolution, `test_dashboard.py:272` fails — the net holds.
* **#194** (`perf/issue-119-reduce-cache`) — **merges clean**, correcting my PR body's expectation. Thanks for testing it; the `algorithm` lookup is a pure read over the already-loaded `events` list, so it's cache-safe under #119's `(st_mtime_ns, st_size)` key with no invalidation change.
* **#192** — trivial, both append to the tail of `test_dashboard.py`; resolution is "keep both".
* **#194 / #197 / #202** — clean.

### Verification

**Core suite — 182 → 183 (one new test), zero failures:**

```
$ cd /tmp/fx-204 && PYTHONPATH=/tmp/fx-204/core python -m pytest core/tests -q
183 passed in 63.13s (0:01:03)
```

`test_dashboard_launch.py::test_maybe_launch_spawns_when_available` **passed** — no port-7878 contention on this machine either, consistent with your finding that the earlier `1 failed` was environment-only (#200).

**Dashboard backend suite — 42 → 43 (the new SSE replay test):**

```
$ cd dashboard/backend && PYTHONPATH=/tmp/fx-204/core python -m pytest tests -q
43 passed, 1 warning in 1.99s
```

**compileall (House rules):**

```
compileall rc=0
```

**`npm test` — 48 → 52 (CI never runs vitest, tracked as #207, so run locally):**

```
$ npm test
> vitest run
 RUN  v4.1.9 /private/tmp/fx-204/dashboard/frontend
 Test Files  14 passed (14)
      Tests  52 passed (52)
   Duration  9.12s
```

**`tsc -b --noEmit`:**

```
$ npx tsc -b --noEmit
tsc rc=0
```

**`npm run build`:**

```
$ npm run build
> tsc -b && vite build
dist/assets/index-DrRhMHxe.css   16.70 kB │ gzip:   4.58 kB
dist/assets/index-C9Q5NKkb.js   832.02 kB │ gzip: 250.41 kB
✓ built in 728ms
```

(Reverted afterwards — `dist/` stays stale per finding 7. The new hash vs your `index-0-YIP72o.js` is this commit's source changes.)

**Real zero-API run — `examples/toy_calc` with the `mock` optimizer, fresh then resumed:**

```
$ python -m cap_evolve.cli run --spec … --run-ts demo --dashboard off
{"run_dir": ".capevolve/run_demo", …, "iterations": 3}

fresh:         1 algorithm event, summary.algorithm = hill-climb:all
after --resume: 1 algorithm event, summary.algorithm = hill-climb:all
total events:  16   (all 16 now replay into the Events tab)
```

**Screenshot** — the finished-run Events tab, headless Chrome against the real backend serving the real completed run dir: header `run_demo` + green `done` badge + `hill-climb:all` chip, **Events** tab active, 16 rows newest-first, `cand_0001`'s `step` chip green with a check icon and `cand_0002`/`cand_0003`'s red with an x icon, `aria-live="polite"` on the `<ul>`. This is the case that previously showed "No live events yet."

### Files touched

```
core/cap_evolve/harness.py                              idempotent stamp + last_algorithm_event() + #114 docstring note
core/tests/test_dashboard.py                            + test_resume_does_not_double_stamp_the_algorithm_event
core/tests/test_orchestration_mode.py                   + agent-mode reduced-label assert
dashboard/backend/capevolve_dashboard/app.py            SSE starts at offset 0 (replay then tail)
dashboard/backend/tests/test_app.py                     + test_stream_replays_the_whole_log_for_a_finished_run
dashboard/frontend/src/components/EventTicker.tsx       verdict icons, cap footer, ?? d.name, status-aware empty state
dashboard/frontend/src/lib/useRunStream.ts              export LOG_CAP
dashboard/frontend/src/routes/RunDeepDive.tsx           pass stream.status to EventTicker
dashboard/frontend/src/test/EventTicker.test.tsx        + 5 tests (a11y, cap, algorithm name, 2 empty states)
```

`dist/` deliberately unchanged.

Issue #129, opened 2026-07-25
Active failure-memory: re-inject rejected approaches as optimizer constraints


OsherElhadad:
## Problem / opportunity

Rejected candidates are recorded (`rejected.jsonl`) but the current optimizer-memory that would feed them back is **write-only/dead** (see #114). So the optimizer can — and does — re-propose variations of approaches that already failed the gate, wasting expensive rollouts on known dead ends.

## Why it matters

Prior art in this space (Arbor, Apache-2.0 — inspiration only) treats **pruned/failed directions as first-class hard constraints**: each proposal round re-injects the failed approaches *and their reasons*, with an explicit instruction not to re-propose anything sharing the same hidden assumption unless it explains how it counters the earlier lesson. This converts wasted rollouts into permanent search pruning — a direct sample-efficiency win, which is exactly cap-evolve's stated priority for GEPA/SkillOpt.

## What to change

1. Build a compact **"already tried & rejected (do not repeat)"** block from `rejected.jsonl` — normalized approach signature + the rejection reason (gate delta, what it broke) — deduped.
2. Inject it into every optimizer proposal prompt (via the shared injection path from #109), with an explicit "don't re-propose the same idea; if you revisit a rejected direction, state what's different" instruction.
3. Bound its size (most-recent / most-relevant N) so it doesn't crowd the prompt.
4. Surface the dead-end list in the dashboard Insights tab (it already derives a "what not to try" section — wire the same data).

## How to test

- New test: seed a run's `rejected.jsonl` with a known-failed edit; assert the next proposal prompt contains that approach in the constraints block with its reason.
- Behavior check on the synthetic adapter: the mock optimizer, given a rejected edit in-context, doesn't re-emit it (or the harness flags a repeat).
- Bounded: with many rejections, the block stays within a size cap.

## Dependencies / parallelization

- **Depends on / coordinates with #114** (which removes the dead write-only memory) and **#109** (shared injection). Do #114 first (clean the dead path), then add this as the real replacement. **Parallel-safe** with UX/docs.

OsherElhadad:
Closing — re-injection of rejected approaches already happens on main by a richer path: `prior_iterations/<candidate>/diff.patch` carries the exact capability edit of every prior iteration including rejected ones, LEDGER.md records each outcome, and the prompt body carries the explicit do-not-repeat and never-re-introduce-a-regression instructions plus JOURNAL.md for refuted ideas.

Full evidence in the close comment on #222.

Issue #128, opened 2026-07-25
Persist synthesized insight/priors across iterations (a reflective signal that survives context)


OsherElhadad:
## Problem / opportunity

cap-evolve's optimizer learns from the *current* candidate's failed traces (via `diagnose` / `REFLECTION.md`) plus the LEDGER/JOURNAL/RUNMAP handover, but there is **no durable, synthesized "what we've learned so far" prior** that survives across iterations independent of the transcript. Each proposal re-derives understanding from scratch-ish context; hard-won insight from iteration 3 isn't guaranteed to reach iteration 30 in compact form.

## Why it matters

Prior art in this space (Arbor, Apache-2.0 — technique inspiration only, no code copied) demonstrates the value of **bottom-up insight synthesis**: after each experiment, an LLM distills the learning into a short insight and propagates it upward into a compact "global priors" block that is re-fed into the next proposal. It's a cheap, durable analogue of a gradient that doesn't depend on the full transcript surviving context limits, and it measurably steers later proposals toward what's working and away from what isn't.

cap-evolve already has the raw material (per-candidate diagnosis, per-task impact, lineage). What's missing is a **compact, persisted, continually-updated priors summary** fed to every proposal.

## What to change

1. Maintain a small, run-scoped **priors/insight artifact** (e.g. `INSIGHTS.md`, bounded to ~200 words) that is updated after each accepted/informative candidate by summarizing "what helped, what hurt, what's still open" from the candidate's diagnosis + per-task impact.
2. Inject it into every optimizer proposal prompt (all algorithms — see #109 for the shared injection path), alongside the existing focused failures.
3. Keep it honest: priors are *candidates for re-testing*, never asserted truth; every prior remains subject to the val gate. Persist it in the run dir and show it in the dashboard Insights tab.
4. Cross-run (stretch): optionally seed a new run's priors from a prior run on the same benchmark, clearly labeled as unverified and re-filtered by the val/test gate.

## How to test

- New test: after N iterations on the synthetic adapter, `INSIGHTS.md` exists, is bounded in length, and contains references to the accepted edits' effects; the proposal prompt contains it.
- Ablation-style check: a run with insights injected proposes edits that reference prior learnings (assert marker presence), vs a control without.
- Reduced/empty case: first iteration has an empty-but-valid priors block (no crash).

## Dependencies / parallelization

[...]

Issue #109, opened 2026-07-25
GEPA & SkillOpt run blind: wire optimizer-context injection + capability brief; un-gate CLI flags from hill-climb-only


OsherElhadad:
## Problem (flagship algorithm issue)

The two "advanced" algorithms — **GEPA** and **SkillOpt** — run with a *thinner* optimizer context than the basic `hill-climb`, because the optimizer-context injection and the capability/instructions wiring are effectively **hill-climb-only**. The flagship sample-efficient algorithms are second-class citizens.

Evidence:
- `_inject_optimizer_context` is only called from `harness.py:1228` inside `run_step` (the hill-climb path). **GEPA bypasses `run_step`**, so its optimizer gets **no `./trajectories/`, no `./guidance/<cap>/`, no native-skill injection** — only `REFLECTION.md` (800-char-truncated traces) + LEDGER/JOURNAL/RUNMAP.
- `skillopt_loop` has **no `capabilities` / `optimizer_name` params** and calls `harness._focus_instructions(current_val, minibatch_ids, label)` with no capability/optimizer args (`skillopt.py:290`) → empty `CAP_BRIEF`, sequential `PARALLEL` note, no bench repo, no custom template.
- **CLI side:** `--capabilities`, `--instructions-file`, `--bench-repo`, `--optimizer-name`, `--capability-sources` are all gated behind `if algorithm_name == "hill-climb"` (`cli.py:354-384`). So even when a user passes these flags, GEPA/SkillOpt silently ignore them.

Net: the benchmark-authored optimizer brief (edit-space, existing-tool-code mandate, non-overfitting guardrail, parallel-fan-out pattern) — the thing that makes hill-climb produce good edits — never reaches the algorithms that are supposed to be *better*.

## Why it matters

GEPA and SkillOpt are marketed as the flagship, most sample-efficient algorithms. Running them with an impoverished prompt undercuts their whole value and likely explains any underperformance vs hill-climb. This is the single highest-impact algorithm fix.

## What to change

1. Factor optimizer-context injection so **GEPA and SkillOpt call `_inject_optimizer_context`** (trajectories, guidance, native-skill injection) around each optimizer invocation — not just `run_step`.
[...]

PR #205, opened 2026-07-29
feat: model tiering — cheap model for auxiliary steps, strong model for proposals

### LINKED ISSUE #132: Model tiering: cheap model for auxiliary steps, strong model for proposals
## Problem / opportunity

cap-evolve uses coding-agent optimizers via the registry, but there's no first-class notion of **spending compute in proportion to task difficulty** — using a cheaper model for auxiliary/mechanical steps (diagnosis summarization, reflection distillation, insight synthesis) and reserving the strong model for the actual edit proposal.

## Why it matters

Prior art in this space (Arbor, Apache-2.0 — inspiration only) runs its coordinator, executors, and auxiliary agents on **different model tiers** (e.g. a cheap model for the isolated related-work/summarization agent, a strong model for the strategist), and gates expensive auxiliary work behind demonstrated promise (only spend on novelty checks for candidates that already beat best). This is a disciplined, measurable cost lever. cap-evolve already tracks per-role cost (intake/optimizer/runner) in the dashboard — the missing piece is letting those roles use different models and gating the expensive ones.

## What to change

1. Allow per-role model configuration in `capevolve.yaml` (e.g. `optimizer_model`, `diagnose_model`, `runner_model`), defaulting to a single model when only one is set (no behavior change for existing configs).
2. Route mechanical/auxiliary steps (reflection distillation, insight synthesis from #128, rejected-summary from #129) to the cheaper tier.
3. Gate expensive auxiliary work behind promise where applicable (e.g. only do deep diagnosis on candidates within reach of best).
4. Surface per-tier spend in the existing CostPanel.

## How to test

- New test: a spec with distinct `optimizer_model`/`diagnose_model` routes calls to the right model (assert via the mock optimizer/registry).
- A single-model spec behaves exactly as today.
- CostPanel shows per-tier breakdown.

- Independent of the algorithm-correctness cluster. Coordinates with #128/#129 (their auxiliary steps are the cheap-tier candidates). **Parallel-safe.** Medium effort.

OsherElhadad:
Closes #132

Lets a spec declare a **strong proposer model** and a **cheap auxiliary model**, and routes each model call site to the right tier. Two new `capevolve.yaml` keys, zero new runtime deps, no new abstraction layer — tier resolution is one function in `specfile.py` that `cli.py` (argv) and `pricing.py` (cost) both read.

## Call site → tier

I enumerated the actual LLM call sites rather than assuming. **The only LLM-shelling callable in core is the edit proposer**, invoked at exactly two places:

| Call site | What it does | Tier | Why |
|---|---|---|---|
| `harness.py:1262` `optimizer(workdir, instructions)` (via `run_step`) | generates the capability edit | **proposer (strong)** | this call determines result quality — never downgraded |
| `gepa.py:575` `optimizer(workdir, instructions)` (minibatch step) | generates the capability edit | **proposer (strong)** | same call, same reason |
| `gepa._write_reflection` / `_write_focus` / `_build_merge` | reflection distillation, component focus, component-wise merge | *aux (n/a today)* | **pure Python — zero model calls** |
| `harness._build_ledger` / `_build_runmap` / `_reconcile_journal` | cross-iteration bookkeeping files | *aux (n/a today)* | **pure Python — zero model calls** |
| `skillopt._failure_patterns` / `_categorize` | failure patterns, longitudinal categorization | *aux (n/a today)* | **pure Python — zero model calls** |
| `skills/phases/diagnose` | reflective dataset + failure clustering | *aux (n/a today)* | **pure Python — zero model calls** |
| runner rollouts | the agent under test | runner (`runner_model`, unchanged) | separate pre-existing role |

**So the honest scope:** the aux tier costs **$0 today** because every auxiliary step is already pure Python. The LLM-backed aux steps the issue names come from #128 (insight synthesis) and #129 (rejected-summary) — both still open. The tier resolves, routes, and prices correctly now, so those steps adopt it with a one-line change (`aux_usd=` on their `update_spent`, `CAPEVOLVE_AUX_MODEL` for the model). I did **not** invent an LLM aux step just to have something to route.

Per issue point 3 (gating expensive aux work behind promise): gepa already does exactly this — the expensive full-val eval only runs after the cheap minibatch local gate passes (`gepa.py:596`). No change needed.

## Backward-compatibility guarantee

Both tiers fall back to `optimizer_model`. An existing single-model spec produces:
- the **identical optimizer argv** (`--model <optimizer_model>`, once),
- the **identical estimate** (proposal priced off the same model, same assumed tokens),
- an **identical event stream** (the `model_tiers` event is logged *only* when the spec is actually tiered),
- `aux_usd == 0.0`, so `total_usd` is unchanged.

Proven by these tests (all pass):
- `test_single_model_optimizer_argv_is_unchanged_by_tiering` — asserts on the **real command built by `cap-evolve run --plan-only`** that reaches `run-optimizer`, not on parsed config: `--model claude-opus-4-8` present, `--model` count == 1.
- `test_spec_with_no_model_emits_no_model_flag` — a model-less spec still emits **no** `--model` token.
- `test_estimate_single_model_is_byte_identical_to_pre_tiering` — the estimated `optimizer_usd` equals `iters × pricing.call_cost(model, "optimizer")`, the pre-tiering formula.
- `test_spent_roundtrips_aux_and_stays_readable_without_it` — an old `state.json` with no aux keys loads, reads back `aux_usd == 0.0`, and yields the **unchanged** `total_usd`.
- End-to-end (below): single-model and tiered runs both reach `test_reward 1.0`; the only diff in `final.json`/`baseline.json` is wall-clock timing, and the only event-stream diff is the one added `model_tiers` line.

## Cost accounting argument (honesty-adjacent — treated as such)

The failure mode is a cheap aux call being reported at the strong model's rate, or vice-versa. Three separate guards, each tested:

1. **Separate buckets.** Aux spend lands in a new `Spent.aux_usd` / `aux_tokens`, never in `optimizer_usd`. If it shared the bucket it would be (a) capped by `max_optimizer_usd` at the wrong rate and (b) calibrated into `$/optimizer-call`, corrupting every future estimate. `test_aux_spend_has_its_own_bucket_and_never_inflates_optimizer_usd`, `test_max_optimizer_usd_caps_only_the_proposer_tier`, `test_calibration_keeps_the_tiers_separate` (asserts `usd_per_optimizer_call == 0.4/2`, **not** `(0.4+0.04)/2`).
2. **Separate pricing profile.** `pricing.TIER_ROLE` pins each tier to a pricing role, and `ASSUMED_TOKENS["aux"] = (4_000, 500)` — a mechanical digest is smaller on both sides than a 10k/2k proposal, so aux would be overstated even on the *correct* model. `test_aux_priced_at_its_own_model_and_token_profile` asserts the same model priced under both tiers gives **different** numbers, so a tier label can't be a no-op hiding a mis-priced call.
3. **No unbudgeted tier.** `Spent.total_usd` includes `aux_usd`, so a new tier is not a way to spend past `max_usd`. `test_max_usd_counts_aux_spend`.

The proposal path is pinned to the strong tier — `test_tiered_spec_routes_the_PROPOSAL_to_the_strong_model` asserts the cheap model string appears **nowhere** on the proposal command line.

## New config keys

Documented in `templates/project/capevolve.yaml`, `skills/phases/intake/inputs/INPUTS.md` (where `optimizer_model`/`target_model` are already documented), `skills/phases/intake/SKILL.md`, and `CHANGELOG.md`.

| Key | Default | Meaning |
|---|---|---|
| `proposer_model` | `""` → `optimizer_model` | STRONG model that proposes the edit. Never downgraded. |
| `aux_model` | `""` → `optimizer_model` | CHEAP model for auxiliary/mechanical steps. |

Typical use: keep `optimizer_model` as the strong proposer and set only `aux_model`. The aux model reaches subprocesses as `CAPEVOLVE_AUX_MODEL` (env, not argv — no core call site spends it yet, so a flag nobody reads would be dead plumbing).

Also surfaced: `--plan-only` gains `model_tiers`, `estimate` gains `proposer_model`/`aux_model`/`tiered`, and the dashboard CostPanel shows the per-tier split (hidden on an untiered run).

## Merge sequencing (#109 / #134)

- **`harness.py` is UNTOUCHED** — zero conflict surface with #109's optimizer-context rework. The proposer tier is resolved where the model already becomes argv (`cli.py`), not inside the harness.
- **Credential resolution is untouched** — no conflict with #134. I touch model *ids* only; nothing reads or moves an auth env var.
- Regions touched, for sequencing:
  - `core/cap_evolve/specfile.py` — **additive only** (new `TIERS`, `model_for_tier`, `is_tiered`; the YAML reader is unchanged).
  - `core/cap_evolve/cli.py` — `_cmd_run` optimizer-argv construction (~line 194), `--plan-only` dump, the `model_tiers` log after baseline, `_calibrate`, `_estimate_core`.
  - `core/cap_evolve/pricing.py` — **additive only** (`ASSUMED_TOKENS["aux"]`, `TIER_ROLE`, `tier_call_cost`; `call_cost` body unchanged).
  - `core/cap_evolve/rundir.py` — `Spent` gains two fields **appended last** (`from_dict` passes positionally, so order is stable), `update_spent` gains two kwargs, `total_usd` includes aux.
  - `core/cap_evolve/dashboard.py` — cost/token split + `model_tiers` in the summary.

## Verification

Full suite — 179 baseline + 20 new, zero failures:

```
$ PYTHONPATH=/tmp/wt-132/core python -m pytest core/tests -q
........................................................................ [ 36%]
........................................................................ [ 72%]
.......................................................                  [100%]
199 passed in 62.14s (0:01:02)
```

```
compileall rc=0
```

The 20 new tests:

```
$ python -m pytest core/tests/test_model_tiering.py -v
test_single_model_spec_resolves_both_tiers_to_optimizer_model PASSED
test_no_model_at_all_stays_blank_for_both_tiers PASSED
test_tiered_spec_splits_the_tiers PASSED
test_proposer_model_overrides_optimizer_model_for_the_proposal_only PASSED
test_unknown_tier_is_a_loud_error PASSED
test_single_model_optimizer_argv_is_unchanged_by_tiering PASSED
test_spec_with_no_model_emits_no_model_flag PASSED
test_tiered_spec_routes_the_PROPOSAL_to_the_strong_model PASSED
test_proposer_model_key_routes_the_proposal PASSED
test_aux_model_reaches_subprocesses_via_env PASSED
test_single_model_spec_sets_no_aux_env PASSED
test_aux_spend_has_its_own_bucket_and_never_inflates_optimizer_usd PASSED
test_spent_roundtrips_aux_and_stays_readable_without_it PASSED
test_max_usd_counts_aux_spend PASSED
test_max_optimizer_usd_caps_only_the_proposer_tier PASSED
test_aux_priced_at_its_own_model_and_token_profile PASSED
test_estimate_prices_proposal_with_the_proposer_tier PASSED
test_estimate_single_model_is_byte_identical_to_pre_tiering PASSED
test_calibration_keeps_the_tiers_separate PASSED
test_dashboard_reports_per_tier_cost_and_hides_it_when_untiered PASSED
============================== 20 passed in 0.09s ==============================
```

### Real end-to-end, zero API cost (`examples/toy_calc`, `mock` optimizer)

**Single-model spec** — both tiers resolve to the one model; no `model_tiers` event:

```
######## MODE=single  spec model keys:
optimizer_model: "claude-opus-4-8"
proposer_model: ""
aux_model: ""
######## PLAN (which model each call site resolves to):
model_tiers: {"proposer": "claude-opus-4-8", "aux": "claude-opus-4-8"}
PROPOSAL call-site argv: --name mock --json --workdir {workdir} --prompt {prompt} --model claude-opus-4-8
######## REAL RUN (mock optimizer, zero API cost):
{
  "best_id": "cand_0001",
  "baseline_val": 0.0,
  "test_reward": 1.0,
  "test_baseline_reward": 0.0,
  "test_delta": 1.0,
  "iterations": 3
}
######## events: model_tiers + per-tier spend:
(no model_tiers event -> single-model run, as expected)
spent: optimizer_usd=0.0 aux_usd=0.0 runner_usd=0.0
```

**Tiered spec** — proposal still resolves to the STRONG model; cheap model absent from the proposal argv:

```
######## MODE=tiered  spec model keys:
optimizer_model: "claude-opus-4-8"
proposer_model: ""
aux_model: "claude-haiku-4-5"
######## PLAN (which model each call site resolves to):
model_tiers: {"proposer": "claude-opus-4-8", "aux": "claude-haiku-4-5"}
PROPOSAL call-site argv: --name mock --json --workdir {workdir} --prompt {prompt} --model claude-opus-4-8
######## REAL RUN (mock optimizer, zero API cost):
{
  "best_id": "cand_0001",
  "baseline_val": 0.0,
  "test_reward": 1.0,
  "test_baseline_reward": 0.0,
  "test_delta": 1.0,
  "iterations": 3
}
######## events: model_tiers + per-tier spend:
{"t": 1785363095.774411, "kind": "model_tiers", "proposer": "claude-opus-4-8", "aux": "claude-haiku-4-5"}
spent: optimizer_usd=0.0 aux_usd=0.0 runner_usd=0.0
```

**The two runs differ only as intended** — identical results, one extra event:

```
$ diff single/final.json tiered/final.json
50c50
<     "seconds": 0.0007507801055908203
---
>     "seconds": 0.0005178451538085938
101c101
<     "seconds": 0.0004019737243652344
---
>     "seconds": 0.0003421306610107422

$ diff <(event kinds, single) <(event kinds, tiered)
3a4
> model_tiers
```

Tier pricing, showing the cost lever is real (aux on haiku vs a proposal on opus):

```
tier_call_cost('claude-opus-4-8', 'proposer') = 0.1
tier_call_cost('claude-haiku-4-5', 'aux')     = 0.0065      # ~15x cheaper per call
```

No live-gateway run was needed: routing is asserted on the real argv that reaches `run-optimizer`, and both e2e runs use the offline `mock` optimizer, so **$0 was spent** and no credential was read.

### Note on `test_dashboard_launch.py`

While verifying I hit `test_maybe_launch_spawns_when_available` failing. It is **pre-existing and environmental**, not from this change — reproduced on a pristine `origin/main` worktree, caused by a leftover uvicorn dashboard server holding port 7878 in my sandbox. After freeing the port it passes on both main (179 passed) and this branch (199 passed), as shown above.


OsherElhadad:
## 🔬 Evidence

Every command run, with full output. Zero API cost — the offline `mock` optimizer was used for both end-to-end runs, and no credential was read.

### 1. Baseline on pristine `origin/main`
```
$ git worktree add /tmp/wt-132-clean origin/main --detach
HEAD is now at 6fca097 evograph weakness-graph view (mounted in the cap-evolve dashboard) (#53)

$ PYTHONPATH=/tmp/wt-132-clean/core python -m pytest core/tests -q
179 passed in 60.67s (0:01:00)
```

### 2. Call-site enumeration (the basis for the tier table)

The only LLM-shelling callable in core is the edit proposer:
```
$ grep -rn "optimizer(workdir" core/ skills/
core/cap_evolve/gepa.py:575:            opt_report = optimizer(workdir, instructions)
core/cap_evolve/harness.py:1262:        opt_report = optimizer(workdir, instructions)  # mutates workdir in place
core/tests/test_empty_seed.py:251:    def noop_optimizer(workdir, instructions):          # test double
skills/algorithms/gepa/scripts/check.py:65:def _good_optimizer(...)             # test double
skills/algorithms/gepa/scripts/check.py:73:def _noop_optimizer(...)             # test double
```

And the auxiliary helpers are all pure Python (no model, no subprocess):
```
$ grep -rn "subprocess.run|subprocess.Popen" core/cap_evolve/
core/cap_evolve/store.py:33      # git
core/cap_evolve/store.py:80      # user store_commit_cmd
core/cap_evolve/harness.py:450   # optimizer_from_command -> THE proposer
core/cap_evolve/dashboard_launch.py:80  # dashboard server
core/cap_evolve/dashboard.py:211        # git log
core/cap_evolve/cli.py:220              # phase skill sequencing
```

### 3. Tier resolution smoke check
```
single  proposer= claude-opus-4-8 aux= claude-opus-4-8 tiered= False
tiered  proposer= claude-opus-4-8 aux= claude-haiku-4-5 tiered= True
empty   proposer= '' tiered= False
tier_call_cost proposer opus  = 0.1
tier_call_cost aux      haiku = 0.0065
tier_call_cost aux      opus  = 0.0325  <- same model, different tier => different price (tier label is not a no-op)
```

### 4. New tests (20), verbose
```
$ PYTHONPATH=/tmp/wt-132/core python -m pytest core/tests/test_model_tiering.py -v
test_single_model_spec_resolves_both_tiers_to_optimizer_model PASSED [  5%]
test_no_model_at_all_stays_blank_for_both_tiers PASSED [ 10%]
test_tiered_spec_splits_the_tiers PASSED [ 15%]
test_proposer_model_overrides_optimizer_model_for_the_proposal_only PASSED [ 20%]
test_unknown_tier_is_a_loud_error PASSED [ 25%]
test_single_model_optimizer_argv_is_unchanged_by_tiering PASSED [ 30%]
test_spec_with_no_model_emits_no_model_flag PASSED [ 35%]
test_tiered_spec_routes_the_PROPOSAL_to_the_strong_model PASSED [ 40%]
test_proposer_model_key_routes_the_proposal PASSED [ 45%]
test_aux_model_reaches_subprocesses_via_env PASSED [ 50%]
test_single_model_spec_sets_no_aux_env PASSED [ 55%]
test_aux_spend_has_its_own_bucket_and_never_inflates_optimizer_usd PASSED [ 60%]
test_spent_roundtrips_aux_and_stays_readable_without_it PASSED [ 65%]
test_max_usd_counts_aux_spend PASSED   [ 70%]
test_max_optimizer_usd_caps_only_the_proposer_tier PASSED [ 75%]
test_aux_priced_at_its_own_model_and_token_profile PASSED [ 80%]
test_estimate_prices_proposal_with_the_proposer_tier PASSED [ 85%]
test_estimate_single_model_is_byte_identical_to_pre_tiering PASSED [ 90%]
test_calibration_keeps_the_tiers_separate PASSED [ 95%]
test_dashboard_reports_per_tier_cost_and_hides_it_when_untiered PASSED [100%]
============================== 20 passed in 0.09s ==============================
```

### 5. Full suite on this branch
```
$ PYTHONPATH=/tmp/wt-132/core python -m pytest core/tests -q
........................................................................ [ 36%]
........................................................................ [ 72%]
.......................................................                  [100%]
199 passed in 62.97s (0:01:02)
```

### 6. compileall
```
rc=0  (clean — no output)
```

### 7. Real end-to-end: single-model spec (backward-compat)

Script: `examples/toy_calc` + `templates/project/capevolve.yaml`, `optimizer_model` set, both tier keys left blank.
```
######## MODE=single  spec model keys:
optimizer_model: "claude-opus-4-8"                      # backend-specific (e.g. a claude model id)
proposer_model: ""                       # strong tier; blank = use optimizer_model
aux_model: ""                            # cheap tier; blank = use optimizer_model (no tiering)
######## PLAN (which model each call site resolves to):
model_tiers: {"proposer": "claude-opus-4-8", "aux": "claude-opus-4-8"}
PROPOSAL call-site argv: --name mock --json --workdir {workdir} --prompt {prompt} --model claude-opus-4-8
######## REAL RUN (mock optimizer, zero API cost):
{
  "best_id": "cand_0001",
  "baseline_val": 0.0,
  "test_reward": 1.0,
  "test_baseline_reward": 0.0,
  "test_delta": 1.0,
  "test_pass_k": {
    "1": 1.0,
    "2": 0.0
  },
  "iterations": 3,
}
######## events: model_tiers + per-tier spend:
(no model_tiers event -> single-model run, as expected)
spent: optimizer_usd=0.0 aux_usd=0.0 runner_usd=0.0
```

### 8. Real end-to-end: tiered spec

Same project, plus `aux_model: claude-haiku-4-5`.
```
######## MODE=tiered  spec model keys:
optimizer_model: "claude-opus-4-8"                      # backend-specific (e.g. a claude model id)
proposer_model: ""                       # strong tier; blank = use optimizer_model
aux_model: "claude-haiku-4-5"                            # cheap tier; blank = use optimizer_model (no tiering)
######## PLAN (which model each call site resolves to):
model_tiers: {"proposer": "claude-opus-4-8", "aux": "claude-haiku-4-5"}
PROPOSAL call-site argv: --name mock --json --workdir {workdir} --prompt {prompt} --model claude-opus-4-8
######## REAL RUN (mock optimizer, zero API cost):
{
  "best_id": "cand_0001",
  "baseline_val": 0.0,
  "test_reward": 1.0,
  "test_baseline_reward": 0.0,
  "test_delta": 1.0,
  "test_pass_k": {
    "1": 1.0,
    "2": 0.0
  },
  "iterations": 3,
}
######## events: model_tiers + per-tier spend:
{"t": 1785363641.399174, "kind": "model_tiers", "proposer": "claude-opus-4-8", "aux": "claude-haiku-4-5"}
spent: optimizer_usd=0.0 aux_usd=0.0 runner_usd=0.0
```

### 9. The two runs differ ONLY as intended
```
$ diff toy-single/.capevolve/run_demo/final.json toy-tiered/.capevolve/run_demo/final.json
50c50
<     "seconds": 0.0005500316619873047
---
>     "seconds": 0.0005481243133544922
101c101
<     "seconds": 0.00039696693420410156
---
>     "seconds": 0.00041794776916503906

$ diff <(event kinds, single) <(event kinds, tiered)
3a4
> model_tiers

(only wall-clock float diffs in final.json; only the one added model_tiers event)
```

### 10. Secret-redaction safety (new keys reach the shared dashboard artifact)
```
proposer     -> redacted by nothing (OK — model ids are not secrets)
aux          -> redacted by nothing (OK — model ids are not secrets)
aux_usd      -> redacted by nothing (OK — model ids are not secrets)
aux_tokens   -> redacted by nothing (OK — model ids are not secrets)
model_tiers  -> redacted by nothing (OK — model ids are not secrets)
```

### 11. Diff scope (for merge sequencing with #109 / #134)
```
$ git diff --stat HEAD~1
 CHANGELOG.md                          |  18 ++
 core/cap_evolve/cli.py                |  54 +++++-
 core/cap_evolve/dashboard.py          |  37 +++-
 core/cap_evolve/pricing.py            |  23 ++-
 core/cap_evolve/rundir.py             |  22 ++-
 core/cap_evolve/specfile.py           |  56 ++++++
 core/tests/test_model_tiering.py      | 309 ++++++++++++++++++++++++++++++++++
 skills/phases/intake/SKILL.md         |   5 +
 skills/phases/intake/inputs/INPUTS.md |  14 ++
 templates/project/capevolve.yaml      |  18 ++
 10 files changed, 539 insertions(+), 17 deletions(-)
```

`harness.py` is **not in the list** — zero conflict surface with #109. No credential/auth code touched — zero conflict with #134.

### 12. Commit authorship
```
$ git log -1 --format="author: %an <%ae>%ncommitter: %cn <%ce>"
author: Osher Elhadad <Osher.Elhadad@ibm.com>
committer: Osher Elhadad <Osher.Elhadad@ibm.com>

$ git log -1 --format=%B | grep -ciE "co-authored-by|generated with claude"
0
0  (no Claude attribution)
```

### 13. Pre-existing `test_dashboard_launch` flake — NOT from this change

Reproduced on a pristine `origin/main` worktree; caused by a leftover uvicorn dashboard server holding port 7878 in this sandbox (started before any of this work).
```
$ cd /tmp/wt-132-clean  # pristine origin/main
$ PYTHONPATH=.../core python -m pytest core/tests/test_dashboard_launch.py -q
core/tests/test_dashboard_launch.py:56: AssertionError
1 failed, 6 passed in 0.05s

$ ps -o lstart=,command= -p 50734
Thu Jul 30 01:06:06 2026   .../Python -m uvicorn ...   # leftover dashboard server on :7878

$ kill 50734 && python -m pytest core/tests/test_dashboard_launch.py -q
.......                                                                  [100%]
7 passed in 0.02s
```

After freeing the port: main = 179 passed, this branch = 199 passed (both 0 failed).

OsherElhadad:
## 🔍 Review — PR #205

**Verdict: APPROVE WITH NITS.**

The central claim verifies. I independently enumerated every LLM/subprocess/network call site in the repo (AST scan for `subprocess`/`socket`/`urllib`/`http`/`requests`/`openai`/`anthropic`/`litellm`/`httpx`/`boto3` imports across all 200+ `.py` files, plus a call-graph check on the `OptimizerFn` type) and confirmed: **exactly two proposer call sites, every enumerated aux step is genuinely model-free.** Cost accounting holds — aux cannot inflate `optimizer_usd`, cannot escape `max_usd`, and a tier label is not a no-op. An old run dir's `Spent` still parses (verified by writing one with `origin/main` code and reading it with this branch's). The proposer cannot be downgraded under any config I could construct, including an aux-only spec.

Nothing here is blocking. Three non-blocking items, two nits.

---

### Blocking

None.

---

### Non-blocking

**N1. `dashboard/frontend/src/components/CostPanel.tsx:10-32` — the React "Cost by role" chart omits `aux_usd`, so bars sum to less than the header total.**

The PR added `aux_usd` to `reduce_run`'s `summary.cost` and to the *ANSI/HTML* dashboard, but the React `CostPanel` — which the issue's point 4 explicitly names ("Surface per-tier spend in the existing CostPanel") — was not touched. `ROLES` is a hardcoded 3-tuple:

```ts
const ROLES = [ {key:'intake'}, {key:'optimizer'}, {key:'runner'} ] as const
const usdByRole = { intake: cost?.intake_usd ?? 0, optimizer: ..., runner: ... }
const total = cost?.total_usd ?? 0     // <- includes aux
```

`total_usd` includes aux; the chart's three bars do not. Reproduced:

```
React 'Cost by role' bars sum = 0.61
React header 'total'          = 0.81
MISSING FROM CHART            = 0.2 (aux)
```

This is exactly the failure mode this PR is otherwise careful about: a cost figure the reader cannot reconcile. It is *not* a double-count (I checked the inverse — `tokens_by_role` sums to `tokens` exactly, and cost parts sum to `total_usd` exactly on the Python side), it is an **under-count in the chart**, which is the safer direction but still a visible arithmetic gap. `dashboard/frontend/src/lib/types.ts:99-106` also lacks `aux_usd` / `aux` in the `cost` and `tokens_by_role` types.

*Consequence:* a tiered run's React CostPanel shows bars that don't add to the stated total. Silently wrong-looking on the one surface the issue asked for.

*Fix:* add `{ key: 'aux', label: 'Aux', color: 'var(--muted)' }` to `ROLES`, `aux: cost?.aux_usd ?? 0` to `usdByRole`, and widen the two type entries in `types.ts`. Gate the row on `cost.aux_usd > 0` if you want the untiered panel byte-identical. Non-blocking only because aux is $0 today (see the audit below) — but this becomes user-visibly wrong the moment #128/#129 land, which is precisely the seam this PR exists to prepare.

**N2. `core/cap_evolve/cli.py:207-209` — `CAPEVOLVE_AUX_MODEL` is set unconditionally (even for untiered specs) and never unset, so it leaks across in-process runs.**

```python
aux_model = _model_for_tier(spec, "aux")
if aux_model:
    os.environ["CAPEVOLVE_AUX_MODEL"] = aux_model
```

Two issues. First, a *single-model* spec exports it too, because `model_for_tier(spec,"aux")` falls back to `optimizer_model`:

```
SINGLE-MODEL   CAPEVOLVE_AUX_MODEL='claude-opus-4-8'
NO MODEL       CAPEVOLVE_AUX_MODEL=None
TIERED         CAPEVOLVE_AUX_MODEL='claude-haiku-4-5'
```

The test `test_single_model_spec_sets_no_aux_env` (`core/tests/test_model_tiering.py:169-175`) is **named for a behavior it does not assert** — its body only checks `plan["model_tiers"]["aux"] == plan["model_tiers"]["proposer"]`, never touches the env, and its own docstring hedges ("but it may export the same one model harmlessly"). A test whose name contradicts its assertion is worse than no test.

Second, there is no `else: os.environ.pop(...)`, so within one process the value survives into a later run with a *different* spec:

```
after TIERED run:              claude-haiku-4-5
after UNTIERED run w/ gpt-5:   gpt-5           <- correct
after NO-MODEL run:            gpt-5           <- STALE, spec has no model at all
```

*Consequence:* a future aux step (the whole point of the env channel) reading a stale model id from a prior run in the same process — e.g. sequential runs from `orchestrate`, or a test suite. Today harmless because nothing reads the var.

*Fix:* one line — `else: os.environ.pop("CAPEVOLVE_AUX_MODEL", None)`. And either rename the test to match what it asserts or make it assert the env.

**N3. `core/cap_evolve/pricing.py:87-94` + `:63` — `tier_call_cost` and `TIER_ROLE` are dead code outside tests.**

```
$ grep -rn "tier_call_cost|TIER_ROLE" --exclude=test_model_tiering.py
core/cap_evolve/pricing.py:63:  TIER_ROLE = {...}
core/cap_evolve/pricing.py:94:  return call_cost(model, TIER_ROLE.get(tier, tier))
```

Zero production callers. `_estimate_core` (`cli.py:619-622`) still prices via `_pricing.call_cost(opt_model, "optimizer")` directly, never `tier_call_cost`. The PR's guard-2 argument ("`TIER_ROLE` pins each tier to a pricing role… is what stops a cheap-tier call from being silently priced at the proposer's rate") is therefore asserted by a test against a function no production path calls. The *actual* guard that holds today is the separate `Spent.aux_usd` bucket (guard 1), which is real and I verified it.

`ASSUMED_TOKENS["aux"] = (4_000, 500)` is likewise unreachable from any production code path — `call_cost(..., "aux")` is only ever invoked from tests.

*Consequence:* the mis-pricing guard is aspirational, not enforced. Anyone adding an aux estimate later can just as easily call `call_cost(m, "optimizer")` and the "explicit mapping" buys nothing.

*Fix:* either (a) route the one existing proposal-pricing call through it — `_pricing.tier_call_cost(opt_model, "proposer")` at `cli.py:620`, which is behavior-identical (`TIER_ROLE["proposer"] == "optimizer"`) and makes the seam load-bearing; or (b) drop `tier_call_cost` and keep only `ASSUMED_TOKENS["aux"]` until a caller exists. (a) is the one-line option and I'd take it.

---

### Nits

**Nit 1. `core/cap_evolve/pricing.py:56` — `ASSUMED_TOKENS["aux"] = (4_000, 500)` has no stated provenance.**

Every other value in that dict carries a one-line justification of the *shape* of the call ("reads INSTRUCTIONS/MEMORY + proposes an edit", "one agent-under-test rollout"). The aux comment justifies the *direction* ("smaller on BOTH sides than a proposal") but not the magnitude — why 4k/500 and not 2k/300 or 6k/800? There is no aux step in existence to have measured. It is a guess.

Defensible, though, for three reasons I verified: (1) the module docstring at `pricing.py:1-9` already frames the whole table as "a LAST-RESORT fallback", ranked third behind calibration and user-supplied prices; (2) `_estimate_core` prints a ±2x band (`low = expected*0.5`, `high = expected*2.0`) around it; (3) `_calibrate` supersedes it entirely once one real run exists, and the PR correctly added `usd_per_aux_call` there (`cli.py:557-558`). The number cannot silently become the headline figure.

*Fix:* one clause — "rough default; no LLM aux step exists yet to measure, calibration supersedes it after the first priced run." Say it's a guess where the guess lives, not just in the PR body.

**Nit 2. `core/cap_evolve/specfile.py:145-151` — `is_tiered` returns `True` for `{"proposer_model": "opus"}` alone (aux `""` ≠ proposer `"opus"`).**

```
proposer only   -> proposer='opus'  aux=''     tiered=True
aux only        -> proposer=''      aux='haiku' tiered=True
```

The docstring says "routes the aux tier to a DIFFERENT model than the proposer", and `""` (backend default) genuinely may resolve to a different model than `opus`, so this is arguably correct — but it means a spec that sets *only* `proposer_model` logs a `model_tiers` event whose `proposer` and `aux` read `"opus"` and `""`, and `dashboard.py:757` renders `aux ` with an empty model id. Cosmetic. Worth a one-line note in the docstring that a blank tier counts as "different".

---

### Is the aux tier really $0 today?

**TRUE — verified independently, not by reading the PR's table.**

*Method:* AST-parsed every `.py` in the repo and collected imports whose root is in `{subprocess, socket, urllib, http, requests, openai, anthropic, litellm, httpx, aiohttp, boto3}`. Complete result:

```
core/cap_evolve/cli.py                            ['subprocess']   phase-skill sequencing
core/cap_evolve/dashboard.py                      ['subprocess']   git log
core/cap_evolve/dashboard_launch.py               ['socket','subprocess']  dashboard server
core/cap_evolve/harness.py                        ['subprocess']   THE proposer (optimizer_from_command)
core/cap_evolve/store.py                          ['subprocess']   git / user store_commit_cmd
dashboard/backend/.../gitlog.py                   ['subprocess']   git
dashboard/backend/.../server.py                   ['subprocess','urllib.request']  dashboard HTTP
skills/optimizers/run-optimizer/scripts/run.py    ['subprocess']   THE proposer's backend
skills/phases/implement-and-check/scripts/run.py  ['subprocess']   runs check.py
examples/tau2_airline/adapters/rits.py            ['litellm','requests']   RUNNER (agent under test)
templates/adapters/{hf_litellm,jsonl_litellm,swe_bench,skillsbench}       RUNNER (agent under test)
examples/skillsbench/adapters/adapter.py          ['subprocess']   RUNNER
```

Every LLM-capable entry is either the **proposer** (`harness.py` → `run-optimizer/scripts/run.py`) or a **runner adapter** (the agent under test — a pre-existing, separately-priced role, untouched by this PR). No third category exists.

*Proposer call sites — exactly 2, confirmed:*
```
$ grep -rn "optimizer(" core/cap_evolve/*.py
core/cap_evolve/gepa.py:575:      opt_report = optimizer(workdir, instructions)
core/cap_evolve/harness.py:1262:  opt_report = optimizer(workdir, instructions)
core/cap_evolve/skillcheck.py:166: def make_mock_optimizer(...)   # factory, not a call
```
`skillopt.py` calls neither — it threads `optimizer` through to `harness.run_step` (`skillopt.py:297,347,470`), which is call site #2. So `skillopt` adds no third site.

*Each "aux (n/a today)" row, checked individually:* `gepa._write_reflection` (gepa.py:205), `_write_focus` (:254), `_build_merge` (:362), `harness._build_ledger` (:727), `_reconcile_journal` (:801), `_build_runmap` (:838), `skillopt._failure_patterns` (:65), `_categorize` (:169) — none appear in the AST import scan's file list for anything but `subprocess` (git/proposer), and none is reachable to an LLM. `skills/phases/diagnose/scripts/run.py` (108 lines) imports only `argparse, json, re, sys, collections, pathlib` — its "clustering" is `re.sub` normalization (`normalized_feedback_signature`, lines 40-52). **Zero model calls.** The PR's table is accurate.

So the tier is not live, and its cost accounting is **not load-bearing today**. That correctly lowers the severity of everything above — N1's chart gap and N3's dead guard are both currently unobservable.

**YAGNI verdict: justified seam, narrowly.** I went in expecting to call this speculative infrastructure — a tier that routes nothing, in a repo whose norm is "no config for a value that never varies." Three things flip it:

1. **The diff is genuinely small and additive.** `specfile.py` +2 functions, `pricing.py` +1 dict entry +1 mapping +1 alias, `rundir.py` +2 fields. No new abstraction layer, no interface with one implementation, no factory. The tier resolution is one 8-line function both consumers read. This is the shape of a seam, not a framework.
2. **The `Spent` field is the part that must ship early.** `state.json` is a durable on-disk format. Adding `aux_usd` later, after runs exist, means either a migration or a cost history with a hole in it. Getting the bucket in *before* anything spends into it is the lazy-correct ordering, not the speculative one.
3. **The author declined to invent an LLM aux step to justify the tier.** That is the honest choice and the right one — a fabricated aux step would have been real complexity added to make a config key look useful.

Against it: `tier_call_cost`/`TIER_ROLE`/`ASSUMED_TOKENS["aux"]` and the `CAPEVOLVE_AUX_MODEL` env channel *are* the speculative part — four constructs whose only callers are tests (N3). If I were trimming, I'd cut `tier_call_cost` and keep the `Spent` bucket + `model_for_tier`. But routing #128/#129 through this later is a one-line change per step, which is exactly what a seam is supposed to buy. Ship it; fix N3 by making the one existing pricing call use the seam so it isn't dead.

---

### Cost-accounting audit

**Can aux spend inflate `optimizer_usd`? No.** Separate `Spent` field, separate `update_spent` kwarg, no cross-assignment anywhere. Verified live:

```
runner 0.30 -> (False, '')
+aux 0.30  -> (False, '')
+aux 0.50  -> (True, 'max_usd reached ($1.10/$1.00; run $0.30 + opt $0.00 + intake $0.00 + aux $0.80)')
optimizer_usd still: 0.0   (max_optimizer_usd cap NOT consumed by aux)
```

`_calibrate` (`cli.py:550-558`) reads `optimizer_usd` and `aux_usd` into separate accumulators, so `usd_per_optimizer_call` cannot be skewed. Confirmed with a synthetic prior run (`optimizer_usd=0.40, aux_usd=1.00, iterations=2`): `usd_per_optimizer_call=0.2` (= 0.4/2, not 1.4/2), `usd_per_aux_call=0.5`. Correct.

**Can aux escape `max_usd`? No** — `total_usd` includes it (`rundir.py:145`), and `budget_exhausted` checks `total_usd` (`:307`). The one caveat is not new: with **no `max_usd` set at all**, aux is unbounded — but so are runner and intake, and `max_optimizer_usd` was never a total cap. Verified:

```
NO max_usd, aux=$99999 -> (False, '') total= 99999.0
```

Same pre-existing hole as `usd=$99999`. Not a regression, not introduced here.

**Can a tier label be a no-op hiding a mis-priced call?** In principle no — `tier_call_cost(STRONG,"aux") != tier_call_cost(STRONG,"proposer")` because the token profiles differ, so labeling a proposal "aux" changes the number. But this guard lives entirely in code no production path calls (N3), so today the answer is "the question can't arise, because nothing prices aux." The guard that actually holds is the separate bucket.

**Old run dir compat: PASSES.** I wrote a real run dir with `origin/main`'s code, then read it with this branch's:

```
OLD-code state.json spent keys: [intake_seconds, intake_tokens, intake_usd, iterations,
  metric_calls, optimizer_seconds, optimizer_tokens, optimizer_usd, runner_seconds,
  runner_tokens, stall, usd]        # no aux keys
OLD total_usd: 1.7

NEW reads OLD: optimizer_usd= 0.4 aux_usd= 0.0 aux_tokens= 0 intake_usd= 0.05
NEW total_usd: 1.7 (expect 1.70)          <- unchanged
budget_exhausted: (False, '')
dashboard cost: {'optimizer_usd':0.4,'runner_usd':1.25,'intake_usd':0.05,'aux_usd':0.0,'total_usd':1.7}
dashboard tokens_by_role: {'runner':99,'optimizer':1234,'intake':0,'aux':0} tokens= 1333
model_tiers: None
```

The positional-`from_dict` concern is real but correctly handled: appending last is the only safe position, and the two `.get(..., 0)` defaults make old data parse to $0. No cost history corruption. **Dashboard does not double-count** — `0.4+1.25+0.05+0.0 = 1.7 = total_usd` exactly, and `tokens_by_role` sums to `tokens` exactly (155 = 155 on a synthetic 4-role run), because `runner` is computed as the residual after subtracting the other three (`dashboard.py:573`).

**Proposer downgrade: impossible under every config I tried.** `cli.py:200-202` reads only `_model_for_tier(spec,"proposer")`, whose key chain is `("proposer_model","optimizer_model")` — `aux_model` is not in it. An aux-only spec:

```
spec {"aux_model":"claude-haiku-4-5"}
  proposer='' aux='claude-haiku-4-5'
  opt_cmd:  --name mock --json --workdir {workdir} --prompt {prompt}
  cheap model on proposal argv? False
```
No `--model` at all (backend default), *not* the cheap model. Correct behavior — a user who sets only `aux_model` does not silently get their proposals downgraded.

**Routing assertion is on real argv — confirmed.** `_plan()` (`test_model_tiering.py:76-105`) invokes `cli._cmd_run([..., "--plan-only"])` and parses stdout; `plan["optimizer_cmd"]` is the same f-string (`cli.py:192`) that becomes `optimizer_from_command`'s template. My own e2e run shows the identical string reaching `run-optimizer/scripts/run.py`. The assertion is meaningful, not tautological — `cmd.count("--model") == 1` and `CHEAP not in cmd` are both properties of the real command line.

**Security: clean.** The `model_tiers` event carries exactly `{t, kind, proposer, aux}` — verified on a real run:
```
{"t":1785364218.28,"kind":"model_tiers","proposer":"claude-opus-4-8","aux":"claude-haiku-4-5"}
```
No credential, no base URL, no endpoint. `dashboard.redact()` passes model ids through unchanged, which is correct (a model id is not a secret) — and if a user ever put a URL in a model field, `_INLINE_KV_RE` would not catch it, but that's true of `optimizer_model` today too, so no new exposure vs. #190's surface. `CAPEVOLVE_AUX_MODEL` holds a model id only.

**Docs: accurate, and the "$0 today" caveat is stated where a user reads it** — `templates/project/capevolve.yaml:33-35` (the file a user actually edits), `INPUTS.md:114-118`, `CHANGELOG.md:34-40`, `intake/SKILL.md:176-180`. That is the honest placement; a user setting `aux_model` will see the NOTE inline and not wonder why nothing changed. All 20 skill `check.py` scripts pass.

---

### Merge-order note

I ran actual three-way merges (`git merge-tree --write-tree`) against all 11 sibling branches rather than inferring from file overlap. Real conflicts: **3**.

| Branch | PR | Conflict |
|---|---|---|
| `feat/issue-134-provider-creds` | #190 | **`templates/project/capevolve.yaml`** |
| `feat/issue-142-protected-paths` | #197 | **`CHANGELOG.md`** (yaml + dashboard.py auto-merge clean) |
| `fix/issue-101-changelog` | #184 | **`CHANGELOG.md`** |
| other 8 | — | clean |

**`cli.py` is the surprise: zero conflicts.** #109 (103 lines changed in `cli.py`), #116 (192 lines), #121, #117 all auto-merge with this PR. The author's claim that keeping `harness.py` untouched avoided the #109 collision is correct but understates it — the reason `cli.py` is clean is that #205's edits cluster in three narrow, distinct regions (`opt_cmd` construction ~line 192, the `--plan-only` dump ~line 253, `_calibrate`/`_estimate_core` ~line 550+) that don't overlap the siblings' regions. Do not treat this as durable: any of those four rebasing near line 192 or 250 flips it.

**`capevolve.yaml` is the genuinely hot file** — four PRs touch it (#205, #190, #197, #198, #189) and only #190 conflicts, because #190 adds a *credentials* block near the same top-of-file model region. #198 and #189 land in different sections.

**`CHANGELOG.md` conflicts are structural, not semantic** — #184 rewrites 94/62 lines of the whole file, so it conflicts with every PR that adds an `### Added` entry. Not a #205 problem.

**Recommended order:**
1. **`fix/issue-101-changelog` (#184) FIRST** — it restructures CHANGELOG wholesale. Land it before anything else adds entries, or every subsequent PR eats the same conflict.
2. **#205 (this PR)** — second, while `cli.py` is clean. It has the smallest `cli.py` surface of the batch and no `harness.py` surface at all; landing it early means the four `cli.py`-heavy PRs rebase onto it once instead of it rebasing onto four of them.
3. `fix/issue-109-optimizer-context` (#199) and `feat/issue-116-follow-tail` (#191) — the two `cli.py`-heavy ones, in either order.
4. `feat/issue-134-provider-creds` (#190) — resolve its `capevolve.yaml` conflict against #205's model-tiering block. Trivial: both are additive blocks in the same file, just keep both. Author of whichever lands second does a 5-line resolve.
5. `#197`, `#193`, `#194`, `#204`, `#198`, `#189`, `#185` — all clean or CHANGELOG-only after step 1.

---

### Verification I re-ran

Branch confirmed: `gh pr view 205 --json headRefName` → `feat/issue-132-model-tiering`, HEAD `d067782`.

**Full suite (199 claimed, 199 observed):**
```
$ cd /tmp/rv-205 && PYTHONPATH=/tmp/rv-205/core python -m pytest core/tests -q
........................................................................ [ 36%]
........................................................................ [ 72%]
.......................................................                  [100%]
199 passed in 58.13s
```

**`test_dashboard_launch.py` (the #200 flake) — passed for me; no port-7878 contention in my sandbox:**
```
$ PYTHONPATH=core python -m pytest core/tests/test_dashboard_launch.py -q
.......                                                                  [100%]
7 passed in 0.02s
```

**compileall:**
```
compileall rc=0
```

**All 20 skill `check.py`:** every one `OK` (baseline, implement-and-check, diagnose, intake, evaluate, gate, report, finalize, orchestrate ×2, capabilities ×4, algorithms ×5, run-optimizer).

**e2e single-model vs tiered (`examples/toy_calc`, `mock` optimizer, $0 spent):**
```
######## MODE=single
model_tiers: {'proposer': 'claude-opus-4-8', 'aux': 'claude-opus-4-8'}
opt_cmd: ... run.py --name mock --json --workdir {workdir} --prompt {prompt} --model claude-opus-4-8
{"run_dir":".capevolve/run_demo","best_id":"cand_0001","baseline_val":0.0,
 "test_reward":1.0,"test_delta":1.0,"iterations":3}
events: ['splits','evaluate','baseline','evaluate','gate_warning','step',...,'finalize']
spent: optimizer_usd=0.0 aux_usd=0.0 runner_usd=0.0

######## MODE=tiered
model_tiers: {'proposer': 'claude-opus-4-8', 'aux': 'claude-haiku-4-5'}
opt_cmd: ... run.py --name mock --json --workdir {workdir} --prompt {prompt} --model claude-opus-4-8
{"run_dir":".capevolve/run_demo","best_id":"cand_0001","baseline_val":0.0,
 "test_reward":1.0,"test_delta":1.0,"iterations":3}
events: ['splits','evaluate','baseline','model_tiers','evaluate',...,'finalize']
spent: optimizer_usd=0.0 aux_usd=0.0 runner_usd=0.0
```
Note the tiered `opt_cmd` still carries `--model claude-opus-4-8` and never `claude-haiku-4-5`.

**The claimed "only wall-clock floats" diff — reproduced exactly:**
```
$ diff /tmp/e2e-single/.capevolve/run_demo/final.json /tmp/e2e-tiered/.capevolve/run_demo/final.json
50c50
<     "seconds": 0.0005068778991699219
---
>     "seconds": 0.0007648468017578125
101c101
<     "seconds": 0.00030493736267089844
---
>     "seconds": 0.0003561973571777344

$ diff <(single event kinds) <(tiered event kinds)
3a4
> model_tiers
```

**Old-`Spent` backward-compat probe** (wrote with `origin/main`, read with this branch) and **budget-escape test** — full output in the Cost-accounting audit above. Both pass.

**Merge probes:** `git merge-tree --write-tree` against all 11 siblings; 3 conflicts, enumerated above.

All of the PR's verification claims reproduced.

OsherElhadad:
## 🔧 Review fixes

All 3 non-blocking items and both nits fixed in `877b84a`. Nothing deleted — after N3, all four constructs the review flagged as speculative have a production caller.

---

### 1. N1 — React `CostPanel` omitted `aux_usd` (bars didn't sum to the total)

Confirmed and fixed. `Aux` is now a fourth cost-by-role bar + tile, rendered **only when it spent**, so an untiered panel is byte-identical. `types.ts` gains `cost.aux_usd` and `tokens_by_role.aux`.

**Before/after, read off the RENDERED DOM (not the props):**

```
######## BEFORE (PR head CostPanel.tsx)
rendered bars : Intake=0.01  Optimizer=0.5  Runner=0.1
bars sum      = 0.61
header 'total'= $0.810
MISSING       = 0.20
######## AFTER (aux added as a role)
rendered bars : Intake=0.01  Optimizer=0.5  Runner=0.1  Aux=0.2
bars sum      = 0.81
header 'total'= $0.810
MISSING       = 0.00
```

New vitest reads the tile figures out of the `Cost by role` card and asserts they sum to the *rendered* header total (`.text-accent`), plus a second test that `Aux` is absent when `aux_usd == 0`. It fails on the pre-fix component:

```
$ git stash push src/components/CostPanel.tsx && npx vitest run src/test/CostFileGit.test.tsx
TestingLibraryElementError: Unable to find an element with the text: Aux
     56|     expect(screen.getByText('Aux')).toBeInTheDocument()
 Tests  1 failed | 4 passed (5)
$ git stash pop && npx vitest run src/test/CostFileGit.test.tsx
 Tests  5 passed (5)
```

Per epic #127's policy `dist/` was **rebuilt, not hand-merged** — see the determinism check below.

---

### 2. N2 — `CAPEVOLVE_AUX_MODEL` leaked stale; the test's name lied

Both real. Now set only when `is_tiered(spec)`, and `os.environ.pop(...)` otherwise. Absent env == "no cheap tier".

```
######## CAPEVOLVE_AUX_MODEL across sequential in-process runs
  pre-seeded stale      : stale-model-from-a-previous-run
  after SINGLE-MODEL run: None
  after TIERED run      : claude-haiku-4-5
  after NO-MODEL run    : None   <- stale cleared, not inherited
  after UNTIERED gpt-5  : None
```

`test_single_model_spec_sets_no_aux_env` now asserts the behavior it is named for. The `_plan` helper **pre-seeds a stale sentinel** and returns the env as observed *inside* the run, so "unset" proves the run actively cleared a leftover rather than merely never setting one. Added `test_untiered_run_clears_a_stale_aux_env_from_an_earlier_run` for the cross-run case.

Both fail if the `cli.py` change is reverted:

```
$ # revert cli.py to the unconditional set, then:
$ pytest core/tests/test_model_tiering.py -k "aux_env or no_aux_env or stale"
>       assert plan["aux_env"] is None
E       assert 'claude-opus-4-8' is None
>       assert _plan(tmp_path / "b", {"optimizer_model": STRONG})["aux_env"] is None
E       AssertionError: assert 'claude-opus-4-8' is None
FAILED test_single_model_spec_sets_no_aux_env
FAILED test_untiered_run_clears_a_stale_aux_env_from_an_earlier_run
2 failed, 20 deselected

$ # restored:
22 passed in 0.50s
```

---

### 3. N3 — the mis-pricing guard is now LIVE in production

`_estimate_core` prices the proposal through the seam: `_pricing.tier_call_cost(opt_model, "proposer")`. Behaviour-identical (`TIER_ROLE["proposer"] == "optimizer"`), so no estimate changes — but the tier is now named at the live call site.

**Guard (b) proven end to end, not in a unit test — the seam is on the real `_estimate_core` path:**

```
######## (b) the seam is on the LIVE _estimate_core path
  tier_call_cost calls observed from _estimate_core: [('claude-opus-4-8', 'proposer')]
  estimated optimizer_usd: 0.4  == 4 x tier_call_cost(STRONG,'proposer') = 0.4

######## a MIS-TIERED call is caught
  tier_call_cost(STRONG,'proposer') = $0.100000  (10k in / 2k out)
  tier_call_cost(STRONG,'aux')      = $0.032500  (4k in / 500 out)
  differ -> True

######## simulate the mis-tier IN PRODUCTION (patch the seam to price 'aux')
  correct   optimizer_usd: 0.4
  mistiered optimizer_usd: 0.13  <- VISIBLY different, not silent
```

Production callers now exist, so **nothing was deleted**:

```
$ grep -rn --include='*.py' "tier_call_cost|TIER_ROLE" . | grep -v test_model_tiering
core/cap_evolve/cli.py:631:        po = _pricing.tier_call_cost(opt_model, "proposer")
core/cap_evolve/pricing.py:68:TIER_ROLE = {"proposer": "optimizer", "aux": "aux"}
core/cap_evolve/pricing.py:92:def tier_call_cost(...)
```

`ASSUMED_TOKENS["aux"]` is reached through the same seam whenever a caller prices the aux tier, and `_calibrate`'s `usd_per_aux_call` already supersedes it.

---

### 4. Nit 1 — `ASSUMED_TOKENS["aux"] = (4_000, 500)` provenance

Labeled explicitly as a guess, at the guess:

```python
# PROVENANCE: a ROUGH DEFAULT, not a measurement — no LLM-backed aux step exists yet
# to measure, so the magnitude (not just the direction) is a guess: ~40% of a
# proposal's input, ~25% of its output. ``_calibrate`` supersedes it with a real
# ``usd_per_aux_call`` after the first priced run, and the estimate prints a ±2x band
# around it until then. Do not read it as authoritative.
```

### 5. Nit 2 — `is_tiered` and blank tiers

Docstring now states a blank tier counts as "different" and why. Also fixed the cosmetic render you flagged (`dashboard.py:757`): a blank tier prints `(backend default)` instead of a dangling `aux ` with nothing after it — applied to both tiers.

---

### Test-name audit (all 22 tests in the file)

AST-scanned every `test_*` in `test_model_tiering.py` for assertion count and assertion subjects, checking each against its own name. **One further mismatch found beyond the one you flagged:**

- `test_aux_model_reaches_subprocesses_via_env` asserted `os.environ.get(...)` in the *parent* process and never spawned a subprocess. It now spawns a real child and asserts the child reads `CAPEVOLVE_AUX_MODEL` out of its inherited environment.

The other 20 all assert their names (each has ≥1 assertion on the subject its name claims; `test_unknown_tier_is_a_loud_error` uses `pytest.raises` rather than `assert`, which is correct for its name).

---

### Verification (literal output)

**Full Python suite — 199 → 201 (2 net new tests):**
```
........................................................................ [ 71%]
.........................................................                [100%]
201 passed in 62.98s (0:01:02)
```

**`test_dashboard_launch.py` (#200 flake) — passed, no port-7878 contention here:**
```
$ pytest core/tests/test_dashboard_launch.py -q
.......                                                                  [100%]
7 passed in 0.01s
```

**Frontend — `tsc -b --noEmit` clean, `npm test` (vitest, +2 tests; CI still never runs this — #207):**
```
$ npx tsc -b --noEmit
rc=0
$ npm test
 Test Files  13 passed (13)
      Tests  47 passed (47)
```

**`dist/` rebuilt (not hand-merged) and deterministic across three builds:**
```
$ npm ci && npm run build
dist/assets/index-C7ufbmDx.css   16.44 kB │ gzip:   4.53 kB
dist/assets/index-D74TaUDO.js   830.40 kB │ gzip: 249.95 kB
✓ built in 807ms
$ npm run build   # 2nd
$ git add -A dashboard/frontend/dist
$ npm run build   # 3rd
$ git status --porcelain dashboard/frontend/dist | grep -E "^.[^ ]"
rc=1 (no unstaged changes -> reproducible bundle)
$ grep -c "Aux" dist/assets/index-D74TaUDO.js
2
```

**compileall + all 20 skill `check.py`:**
```
compileall rc=0
$ for f in $(find skills -name check.py); do ... done
  agent-optimize / evograph / gepa / hill-climb / skillopt        rc=0 ok=True
  mcp-tool / skill-package / system-prompt / tools                rc=0 ok=True
  run-optimizer / orchestrate / using-cap-evolve                  rc=0 ok=True
  baseline / diagnose / evaluate / finalize / gate                rc=0 ok=True
  implement-and-check / intake / report                           rc=0 ok=True
any failure: 0
```

**Re-proofs of the four properties you validated (all still hold):**

Aux cannot escape `max_usd`, and does not consume the optimizer cap:
```
  runner 0.30 -> (False, '')
  +aux 0.30  -> (False, '')
  +aux 0.50  -> (True, 'max_usd reached ($1.10/$1.00; run $0.30 + opt $0.00 + intake $0.00 + aux $0.80)')
  optimizer_usd still: 0.0
```

Old `Spent` (no aux keys) parses, no double-count:
```
  OLD total_usd: 1.7  NEW reads: 1.7  aux_usd: 0.0  aux_tokens: 0
```

Proposer cannot be downgraded (aux-only spec emits no `--model` at all):
```
  spec {aux_model: claude-haiku-4-5}
    model_tiers: {'proposer': '', 'aux': 'claude-haiku-4-5'}
    optimizer_cmd tail: --name mock --json --workdir {workdir} --prompt {prompt}
    cheap model on the proposal argv? False
    any --model at all? False
```

`model_tiers` event carries no secret and no base URL — exactly `{t, kind, proposer, aux}`:
```
{"t": 1785366618.9690108, "kind": "model_tiers", "proposer": "claude-opus-4-8", "aux": "claude-haiku-4-5"}
single-model run model_tiers events: 0
```

**e2e single vs tiered — only diff is one wall-clock float, plus the extra event:**
```
$ diff /tmp/e2e-single/.capevolve/run_demo/final.json /tmp/e2e-tiered/.capevolve/run_demo/final.json
50c50
<     "seconds": 0.0004620552062988281
---
>     "seconds": 0.0005140304565429688

$ diff <(single event kinds) <(tiered event kinds)
@@ -3,0 +4 @@
+model_tiers
```

---

### Merge order (restating the reviewer's, from real `git merge-tree` runs against all 11 siblings — 3 conflicts)

1. **#184** (`fix/issue-101-changelog`) **first** — it rewrites CHANGELOG wholesale, so landing it first stops every other PR eating the same structural conflict.
2. **#205 (this PR) second, while `cli.py` is clean** — zero `cli.py` conflicts across the batch and no `harness.py` surface at all; landing it early means the four `cli.py`-heavy PRs rebase onto it once.
3. **#199** and **#191** — the two `cli.py`-heavy ones, either order.
4. **#190** — resolve its `templates/project/capevolve.yaml` conflict against the tiering block. Both are additive blocks in the same file: **keep both**.
5. #197, #193, #194, #204, #198, #189, #185 — clean or CHANGELOG-only after step 1.

Known conflicts: **#190** (`capevolve.yaml`), **#197** and **#184** (`CHANGELOG.md`). Note this PR touches `CHANGELOG.md`, so it will conflict with #184 if #184 lands second — hence #184 first.

### Files touched

`core/cap_evolve/cli.py`, `core/cap_evolve/pricing.py`, `core/cap_evolve/specfile.py`, `core/cap_evolve/dashboard.py`, `core/tests/test_model_tiering.py`, `dashboard/frontend/src/components/CostPanel.tsx`, `dashboard/frontend/src/lib/types.ts`, `dashboard/frontend/src/test/CostFileGit.test.tsx`, `dashboard/frontend/dist/*` (rebuilt), `CHANGELOG.md`.

OsherElhadad:
Closing — thanks for the unusually honest write-up; it's what makes the call easy.

Your own analysis is the reason: **every auxiliary step in core is pure Python, so the aux
tier routes zero calls and costs $0 today.** I re-verified on current main —
`grep -rn "optimizer(" core/cap_evolve/*.py` returns exactly two sites (`harness.py:1482`,
`gepa.py:595`), both the edit proposer, and the only model id that ever becomes argv is
`optimizer_model` at `cli.py:664`. The two issues that would add an LLM-backed aux step
(#128, #129) are still open.

That leaves the PR as plumbing with no consumer: `aux_model` / `CAPEVOLVE_AUX_MODEL` that
nothing reads, a `Spent.aux_usd` bucket nothing writes, a `pricing.ASSUMED_TOKENS["aux"] =
(4_000, 500)` profile invented for a call that is never made, and a CostPanel split that
always shows $0. The other new key, `proposer_model`, is an alias — `optimizer_model` is
already "the model that PROPOSES edits" (see the comment at `target_profile.py:3` and
`templates/project/capevolve.yaml:14`). And issue #132's point 3 (gate expensive aux work
behind promise) is already satisfied by gepa's minibatch-before-full-val gate, as you note.

The core additions are genuinely clean and vendor-agnostic, and the 20 tests are careful —
but the strongest of them assert that the untiered path is byte-identical, which is another
way of saying the feature does nothing yet. Fixing a config interface before its only
consumer exists costs us the chance to shape it around that consumer's real token profile.

Concrete path forward: when #128 or #129 lands the first LLM-backed aux step, bring back
`specfile.model_for_tier` (~10 lines) in *that* PR, with that step's measured token
profile, and drop `proposer_model`. Reopening #132 with that scope is welcome; the design
work here isn't wasted.

Issue #142, opened 2026-07-25
Protected-paths tamper guard: verify the optimizer never edited scoring/eval/task files


OsherElhadad:
## Problem / opportunity

cap-evolve enforces honesty on the *evaluation* side (sealed test, val-only gate, seal-on-success). But when the capability being optimized is **tool code or a skill package**, the optimizer edits files in the workspace — and there is no structural guarantee that it didn't (accidentally or via reward-hacking) edit the **scoring / eval / task-definition** files themselves. The capability layer has an action-policy / mutation-lock for *what kind* of edits are allowed, but not a tamper check that the ground-truth/scoring code is byte-for-byte untouched.

## Why it matters

Prior art in this space (Arbor, Apache-2.0 — inspiration only) hashes and OS-read-only-locks **protected paths** (the eval entrypoint / ground-truth verifier / task definition) at workspace creation and **re-verifies them after** the candidate runs; if a protected file changed, the score is **discarded** and the candidate can't advance. This makes "only edit the target, never the grader" a *structural* guarantee, not a hope — exactly the anti-self-deception posture cap-evolve already sells for the test split, extended to the scoring code. It closes a real reward-hacking hole for the `tools`/`skill-package` capabilities.

## What to change

1. Let the adapter/benchmark declare **`protected_paths`** (scorer, eval harness, task data, gold files).
2. At candidate materialize, record a SHA-256 manifest of protected paths (and ideally make them read-only in the candidate workspace).
3. After `run_target`/`score`, **re-verify** the manifest; on any mismatch, discard the candidate's score, log a `tamper_detected` event, and refuse to let it advance/seal.
4. Surface tamper events prominently in the dashboard and report (an honesty signal).

## How to test

- New test: a candidate that edits a declared protected file has its score discarded and a `tamper_detected` event logged; it cannot become best.
- A well-behaved candidate (edits only the target) passes unaffected.
- Works for both `tools` (edits tool code) and `skill-package` capabilities.

## Dependencies / parallelization

- Complements the honesty core and **#141** (the zoo verifier uses protected paths). Independent of UX/docs. **Parallel-safe.** Labeled `honesty` — brand-aligned hardening.

OsherElhadad:
[...]

Issue #209, opened 2026-07-29
Model-controlled event text can blank the generated dashboard.html (incomplete script-context escape at dashboard.py:657)


OsherElhadad:
Model-controlled event text can blank the entire generated `dashboard.html`. Not XSS — no script executes — but a denial-of-view on the primary results artifact, triggerable by the optimizer's own output.

## The sink
`core/cap_evolve/dashboard.py:657`
```python
data = json.dumps(payload, default=str).replace("</", "<\\/")
```
The reduced run payload is embedded in an inline `<script>`. The escape defends only against a literal `</` closing the script tag early — but HTML's script-data parsing also honours **comment-like sequences**. A payload containing `<!--<script>` survives this replace and shifts the parser into a state where the rest of the document is swallowed.

Measured in headless Chrome: sections rendered **5 → 0**, body text **793 → 34 chars**. The page is blank.

## Reachability
The payload carries `reason` and `optimizer_error`, which are the **optimizer subprocess's own stdout/stderr** — i.e. LLM-generated text. No attacker is needed: a model that happens to emit `<!--<script>` while explaining a rejected edit silently destroys the run's dashboard. Anything that reaches `reduce_run`'s output can carry it.

Pre-existing on `main`; not introduced by any epic #127 PR (verified — #204's only new payload field is `summary.algorithm`, sourced from the spec file, not from model output).

## Fix
Escape the characters that can shift the parser out of script-data, not just the one sequence:
```python
data = (json.dumps(payload, default=str)
        .replace("<", "\\u003c")
        .replace(">", "\\u003e")
        .replace("&", "\\u0026"))
```
`<` inside a JSON string literal parses back to `<`, so the embedded data is unchanged for the JS that reads it while becoming inert to the HTML parser. This is the standard treatment for JSON-in-`<script>` and is strictly simpler than pattern-matching `</` and `<!--` separately — the denylist approach is what failed here.

Alternative worth considering: emit the payload as `<script type="application/json">` and `JSON.parse` it, or write it to a sibling `.json` file the page fetches — though the latter breaks the single-file-artifact property that makes `dashboard.html` shareable, so probably prefer the escape.

## Acceptance
- A run whose `reason`/`optimizer_error` contains `<!--<script>`, `</script>`, and `<!--` still renders the full dashboard (assert section count and body length, not just "no exception").
[...]

PR #210, opened 2026-07-29
GEPA eval-cache hits now carry output/trace (no more hollow reflective dataset)

### LINKED ISSUE #111: GEPA eval cache drops output/trace → hollow reflective dataset
## Problem

GEPA's eval cache **drops `output`/`trace`**, which silently hollows out GEPA's own reflective dataset — its entire learning signal.

- `EvalCache` stores only `{reward, feedback}` (`cache.py:89`).
- On a cache hit, `_eval_minibatch` builds a `Score` with `raw={"cached": True}` and **no `output` / `trace`** (`gepa.py:149-155`).
- GEPA re-samples parents constantly, so the parent minibatch frequently hits the cache → `_write_reflection` then emits `Agent output: ` (empty) for those failing tasks (`gepa.py:238-245`).

So the reflective dataset that's supposed to tell the optimizer *what the agent actually did wrong* is frequently blank for cached tasks. GEPA's headline advantage (rich reflection over failures) is quietly degraded by its own cache.

## Why it matters

Reflection quality is the whole point of GEPA (arXiv:2507.19457). Feeding the optimizer empty "Generated Outputs" for cached failing tasks produces weaker edits and wastes the expensive optimizer call. This is a correctness bug in the flagship algorithm, not a nit.

## What to change (pick one)

- **Option A (recommended):** Extend `EvalCache` to persist `output` and `trace` (or a bounded, redacted slice sufficient for reflection) so cache hits reconstruct a full `Score`. Watch cache size — bound/truncate trace length.
- **Option B:** Bypass the cache for the **parent minibatch that feeds reflection** (always do a fresh eval there), while keeping the cache for the child-vs-parent local gate where only the reward is needed.

Option A is more sample-efficient; Option B is simpler and guarantees non-empty reflection.

## How to test

- New test: run GEPA so a parent minibatch is served from cache; assert the reflective dataset written for a failing cached task has a **non-empty** "Generated Outputs" section (not `Agent output: `).
- Confirm cache size stays bounded if Option A (assert a max trace length is enforced).

- Independent of the snapshot fix; **related to** "wire GEPA/SkillOpt optimizer context" (both improve GEPA reflection) but a distinct root cause. **Parallel-safe** with coordination on `gepa.py`/`cache.py` merges.

OsherElhadad:
Closes #111

## Problem

`EvalCache` stored only `{reward, feedback}`, so on a cache hit `_eval_minibatch` built a `Score` with `raw={"cached": True}` and **no `output`/`trace`**. GEPA re-samples parents constantly, so the parent minibatch that *feeds reflection* frequently hit the cache — and `_write_reflection` then emitted `- Agent output:` (empty) for exactly those failing tasks.

Reflection quality is the whole point of GEPA (arXiv:2507.19457). On cached parents it degraded to a blind hill-climb, and the expensive optimizer call was spent on a blank dataset.

## Fix

Issue #111 **Option A**, pointer flavour:

- `EvalCache.put(..., rollout_file=...)` — each entry gains `rollout_file`, the name of the rollout json under `rollouts/<split>/` that *produced* that score.
- `gepa._replay_cached(...)` — on a hit, re-read that json for the real `output`/`trace` and rebuild the full `Score`; **hardlink** it under the current eval tag so `rollouts/<split>/*__<tag>__t0.json` is complete either way and the tag-pinned `trajectories/` dir exists for a cached minibatch too.
- A hit whose pointer does **not** resolve (a pre-#111 cache, or pruned rollouts) is treated as a **MISS and re-run**. Paying one rollout beats serving an empty `Agent output:` as if it were a trace — so a score-only entry can never hollow out a reflective dataset again.

**Why a pointer, not the payload.** The traces are *already* persisted in the run dir. Copying them into `eval_cache.json` would grow the cache by the trace size *and* leave reflection truncated (whatever bound you pick). A pointer costs ~42 bytes/entry, keeps reflection **untruncated**, and the hardlink makes a cache hit free on disk. Replayed `output`/`trace` still go through the same `_short(n=1500)` truncation a fresh eval uses, so prompt size stays bounded regardless of trace size.

## Cache format + size impact

Measured on the pasted end-to-end run (3 GEPA iterations, toy_calc, 8 cache entries, 24 rollout files):

**Before**
```json
"6a1c1a0e…::a3": {
  "reward": 0.0,
  "feedback": "expected '10' but agent produced 'I think 2 * 5 is roughly some number.'; …"
}
```
**After**
```json
"6a1c1a0e…::a3": {
  "reward": 0.0,
  "feedback": "expected '10' but agent produced 'I think 2 * 5 is roughly some number.'; …",
  "rollout_file": "a3__mb_p_0000__t0.json"
}
```

| | before | after |
|---|---|---|
| `eval_cache.json` | 2110 B | 2446 B (**+336 B / 8 entries = +42 B per entry**) |
| `rollouts/` | 48 KB | **48 KB** (hardlinks — a cache hit adds a name, not bytes) |
| total run dir | 436 KB | **436 KB** |

The per-entry cost is a fixed filename, independent of trace size — a 50 KB trace still costs 42 bytes of cache. `test_trace_is_bounded_in_the_reflective_signal` pins both halves (replayed trace truncated, cache stays <1 KB with a 50 KB trace).

## Before / after: the reflective dataset on a cache-hit iteration

Same run, iteration 2, whose parent minibatch was **4/4 served from the cache** (`{"kind":"minibatch","tag":"mb_p_0001","fired":0,"cached":4}`) and **0/4 passing** — the exact scenario in the issue.

**BEFORE** — `work/gepa_0002/REFLECTION.md`
```
## 4 actionable failing task(s)
### task a6
- Agent output: 
- Feedback: expected '12' but agent produced 'I think 6 * 2 is roughly some number.'; …
### task a2
- Agent output: 
- Feedback: expected '4' but agent produced 'I think 10 - 6 is roughly some number.'; …
```
Every "Generated Outputs" blank, no `Trajectory:` line at all. Note the scorer feedback *quotes* the output the optimizer was not shown.

**AFTER** — same file, same iteration
```
## 4 actionable failing task(s)
### task a6
- Agent output: I think 6 * 2 is roughly some number.
- Trajectory: prompt_had_calc=False
- Feedback: expected '12' but agent produced 'I think 6 * 2 is roughly some number.'; …
### task a2
- Agent output: I think 10 - 6 is roughly some number.
- Trajectory: prompt_had_calc=False
- Feedback: expected '4' but agent produced 'I think 10 - 6 is roughly some number.'; …
```
Real per-task output **and** trajectory, from a 100% cache hit, at zero extra rollout cost.

## Coordination with #199 / #197

**#199 (issue #109) and this PR compose exactly as its author described:** #199 made the absence *honest*; this makes the traces *available*, so the honest-absence path becomes a genuine edge case (stale/pruned cache) instead of the normal cached path.

I did **not** edit #199's text — that would conflict for no benefit. Instead this PR makes it **self-correcting**, because both #199 branches are already conditional on the dir existing:

- `gepa._gepa_block(..., has_trajectories=...)` — the `"This minibatch was served entirely from the eval cache, so NO rollout files were persisted for it"` branch is now **not taken** on a cached minibatch (the rollouts ARE there), so the prompt makes the truthful `"./trajectories/ holds the SAME minibatch rollouts VERBATIM and untruncated"` claim instead.
- `harness._copy_step_trajectories` — the `optimizer_context_warning` event with `"no rollouts persisted for the pinned eval tag (fully-cached minibatch); trajectories/ OMITTED"` now fires **0 times** on the merged tree (was once per cached iteration).

Verified on a tree with all three PRs merged:

```
=== optimizer_context_warning events ===  → 0
=== iter2 trajectories/ (cached minibatch) ===
a2__mb_p_0001__t0.json  a3__mb_p_0001__t0.json
a5__mb_p_0001__t0.json  a6__mb_p_0001__t0.json
=== iter2 prompt claim ===
trajectories/` holds the SAME minibatch rollouts VERBATIM and untruncated — read them when REFLECTION…
```

**Two stale comments for #199's author** (theirs to change, one word each, no behavior): `harness._copy_step_trajectories` says *"the eval cache stores only reward+feedback — see #111"* and *"A fully-cached minibatch writes no rollout files"*. Both are false after this PR; the surrounding fallback logic remains correct and still-needed for a pre-#111 / pruned cache.

**#197 (issue #142):** untouched. Both `protect.verify` guards in `_eval_minibatch` (pre and post) survive verbatim — I only changed the cache-hit branch inside the loop and the `cache.put` call.

**Expected merge order:** #199 → #197 → this PR (any order works; conflicts are import-line only and *between #199 and #197*, not with this PR — see Verification).

## Verification

```
184 passed in 60.70s
```
179 baseline + 5 new (`core/tests/test_gepa_cache_traces.py`), 0 failed.

Fail-before/pass-after, by stashing only the source change:
```
$ git stash push core/cap_evolve/gepa.py core/cap_evolve/cache.py
$ pytest core/tests/test_gepa_cache_traces.py -q
5 failed in 0.53s
$ git stash pop && pytest core/tests/test_gepa_cache_traces.py -q
5 passed in 0.07s
```

Merged with **both** sibling branches (#199 then #197), conflicts resolved (import lines only, all three between #199 and #197 — none in my hunks):
```
$ pytest core/tests -q
227 passed in 74.41s
```

```
COMPILE_OK
```

Full commands + untruncated output in the `## 🔬 Evidence` comment below.

### DIFF
--- a/core/cap_evolve/cache.py
+++ b/core/cap_evolve/cache.py
@@ -1,12 +1,25 @@
 """Eval cache — skip a rollout when the same candidate was already scored on a task.
 
-Keyed by ``(hash of the candidate's editable files, task_id) -> {reward, feedback}``
-and persisted in the run dir, so re-evaluating an identical candidate (e.g. a parent
-re-sampled in GEPA, or a resumed run) costs nothing. The hash is over file CONTENTS,
-so two byte-identical candidates share cache entries even under different ids.
+Keyed by ``(hash of the candidate's editable files, task_id) ->
+{reward, feedback, rollout_file}`` and persisted in the run dir, so re-evaluating an
+identical candidate (e.g. a parent re-sampled in GEPA, or a resumed run) costs nothing.
+The hash is over file CONTENTS, so two byte-identical candidates share cache entries
+even under different ids.
+
+``rollout_file`` (#111) names the rollout json, under ``rollouts/<split>/``, that
+PRODUCED this cached score. GEPA's reflective dataset — its entire learning signal —
+needs the agent's ``output``/``trace``, not just the number; a score-only hit made
+``REFLECTION.md`` emit ``Agent output:`` (empty) for cached failing tasks and quietly
+degraded GEPA to a blind hill-climb on cached parents. Storing a POINTER rather than
+the payload keeps the cache small (~40 bytes/entry) and the reflection UNTRUNCATED:
+the rollout json is already persisted in the run dir, so a hit re-reads the real
+record instead of a redacted copy of it. A hit whose pointer no longer resolves is
+treated as a MISS by the reflection-bearing caller, so a score-only entry can never
+hollow out a reflective dataset again.
 
 Honesty notes:
-  * The cache stores only the SCORE (reward + feedback), never gold answers.
+  * The cache stores only the SCORE (reward + feedback) plus a pointer to the
+    already-persisted rollout json, never gold answers.
   * It is keyed on candidate-file content, so an edit (even whitespace) busts the
     key — a stale score can never be served for changed files.
   * It is an optimization, not a source of truth: ``events.jsonl`` still records
@@ -64,10 +77,11 @@ def hash_candidate_dir(candidate_dir: Path) -> str:
 class EvalCache:
     """A tiny JSON-file eval cache living in the run dir.
 
-    ``get(candidate_hash, task_id)`` -> ``{"reward", "feedback"}`` or ``None``;
-    ``put(candidate_hash, task_id, reward, feedback)`` persists. Persistence is a
-    single JSON object ``{ "<hash>::<task_id>": {...} }`` rewritten on each put — fine
-    for the run sizes here (a few thousand entries) and trivially portable.
+    ``get(candidate_hash, task_id)`` -> ``{"reward", "feedback", "rollout_file"}`` or
+    ``None``; ``put(candidate_hash, task_id, reward, feedback, rollout_file=...)``
+    persists. Persistence is a single JSON object ``{ "<hash>::<task_id>": {...} }``
+    rewritten on each put — fine for the run sizes here (a few thousand entries) and
+    trivially portable.
     """
 
     def __init__(self, path: Path):
@@ -86,9 +100,15 @@ class EvalCache:
     def get(self, candidate_hash: str, task_id: str) -> dict | None:
         return self._data.get(self._key(candidate_hash, task_id))
 
-    def put(self, candidate_hash: str, task_id: str, reward: float, feedback: str = "") -> None:
-        self._data[self._key(candidate_hash, task_id)] = {
-            "reward": float(reward), "feedback": str(feedback or "")}
+    def put(self, candidate_hash: str, task_id: str, reward: float, feedback: str = "",
+            rollout_file: str = "") -> None:
+        """Persist a score. ``rollout_file`` is the ``rollouts/<split>/`` filename whose
+        json holds the rollout behind this score, so a later hit can rebuild the full
+        reflective signal (output + trace) instead of only the number (#111)."""
+        entry = {"reward": float(reward), "feedback": str(feedback or "")}
+        if rollout_file:
+            entry["rollout_file"] = str(rollout_file)
+        self._data[self._key(candidate_hash, task_id)] = entry
         self._flush()
 
     def _flush(self) -> None:
--- a/core/cap_evolve/cache.py
+++ b/core/cap_evolve/cache.py
@@ -77,8 +77,10 @@ def hash_candidate_dir(candidate_dir: Path) -> str:
 class EvalCache:
     """A tiny JSON-file eval cache living in the run dir.
 
-    ``get(candidate_hash, task_id)`` -> ``{"reward", "feedback", "rollout_file"}`` or
-    ``None``; ``put(candidate_hash, task_id, reward, feedback, rollout_file=...)``
+    ``get(candidate_hash, task_id)`` -> ``{"reward", "feedback"}`` plus an OPTIONAL
+    ``"rollout_file"`` (omitted when unknown — e.g. a pre-#111 entry; readers treat its
+    absence as a miss), or ``None``;
+    ``put(candidate_hash, task_id, reward, feedback, rollout_file=...)``
     persists. Persistence is a single JSON object ``{ "<hash>::<task_id>": {...} }``
     rewritten on each put — fine for the run sizes here (a few thousand entries) and
     trivially portable.
--- a/core/cap_evolve/gepa.py
+++ b/core/cap_evolve/gepa.py
@@ -64,6 +64,7 @@ from .harness import (
 )
 from .loop import SplitResult, aggregate_scores
 from .rundir import RunDir
+from .rundir import _atomic_write  # separate line: keeps the #199 merge clean
 from .types import Rollout, Score
 
 OptimizerFn = Callable[[Path, str], None]
@@ -159,7 +160,8 @@ def _eval_minibatch(
     with _live(adapter, candidate_dir) as ctx:
         for task in tasks:
             cached = cache.get(chash, task.id) if cache is not None else None
-            replay = _replay_cached(cached, out_dir, task.id, tag) if cached else None
+            replay = (_replay_cached(cached, out_dir, task.id, tag, run_dir)
+                      if cached else None)
             if replay is not None:
                 scores.append(replay)
                 continue
@@ -179,11 +181,15 @@ def _eval_minibatch(
                      "trace": _short(getattr(rollout, "trace", None))},
             ))
             rfile = f"{task.id}__{tag}__t0.json"
-            (out_dir / rfile).write_text(
-                json.dumps({"input": task.input, "rollout": rollout.to_dict(),
-                            "score": sc.to_dict()}, default=str),
-                encoding="utf-8",
-            )
+            # ``_atomic_write`` (tmp + ``os.replace``) and NOT ``write_text``: a cache
+            # hit may have HARDLINKED a previous iteration's archived rollout to this
+            # name, and ``write_text`` truncates the shared inode — silently rewriting
+            # that earlier iteration's evidence. Replacing the directory entry breaks
+            # the link instead, so the archived record is immutable once written.
+            _atomic_write(out_dir / rfile,
+                          json.dumps({"input": task.input,
+                                      "rollout": rollout.to_dict(),
+                                      "score": sc.to_dict()}, default=str))
             if cache is not None:
                 cache.put(chash, task.id, sc.reward, sc.feedback or "",
                           rollout_file=rfile)
@@ -200,7 +206,8 @@ def _eval_minibatch(
     return result
 
 
-def _replay_cached(cached: dict, out_dir: Path, task_id: str, tag: str) -> Score | None:
+def _replay_cached(cached: dict, out_dir: Path, task_id: str, tag: str,
+                   run_dir: RunDir | None = None) -> Score | None:
     """Rebuild a full ``Score`` (output + trace) from a cache hit, or ``None`` = miss.
 
     The fix for #111. The cache entry carries ``rollout_file`` — the rollout json in
@@ -212,30 +219,52 @@ def _replay_cached(cached: dict, out_dir: Path, task_id: str, tag: str) -> Score
     Returning ``None`` (→ the caller re-runs the rollout) is deliberate for a pointerless
     or unreadable entry: a score-only hit is exactly the hollow-reflection bug, so we pay
     one rollout instead of serving empty "Agent output:" as if it were a trace.
+
+    ``eval_cache.json`` is plain JSON in the optimizer-writable run dir, so ``rollout_file``
+    is UNTRUSTED input (#142's threat model). It must be a BARE FILENAME under ``out_dir``:
+    ``Path(rfile).name != rfile`` rejects ``/etc/x``, ``../y`` and ``sub/z`` in one
+    comparison — an allowlist of the only legitimate shape, not a denylist of bad ones —
+    and the resolved-parent check confines it to ``out_dir`` even through a symlink.
+    Otherwise a tampered entry reads any json-shaped file on disk straight into the
+    optimizer's prompt (and hardlinks its inode into the run dir).
     """
     rfile = str(cached.get("rollout_file") or "")
-    if not rfile:
+    if not rfile or Path(rfile).name != rfile:
         return None
     src = out_dir / rfile
     try:
+        # Second layer: the resolved path must still live directly in out_dir.
+        if src.resolve().parent != out_dir.resolve():
+            return None
         rec = json.loads(src.read_text(encoding="utf-8"))
-    except (OSError, json.JSONDecodeError, ValueError):
-        return None
+        reward = float(cached.get("reward", 0.0))
+    except (OSError, TypeError, ValueError, json.JSONDecodeError):
+        return None  # unreadable pointer OR malformed reward → honest MISS, never a crash
     rollout = rec.get("rollout") or {}
-    reward = float(cached.get("reward", 0.0))
     # Re-materialize under this eval's tag so ``_copy_step_trajectories(tag=...)`` finds
     # this minibatch's rollouts. Same bytes, new name — no re-run, no fabrication. A
     # hardlink keeps a cache hit free on disk too (rollout jsons are write-once); the
     # copy fallback covers filesystems/paths where linking isn't available.
+    #
+    # Overwrite unconditionally (no ``not dst.exists()`` short-circuit): on a resumed run
+    # the tag can already hold a DIFFERENT parent's rollout, and keeping it would let
+    # ``trajectories/`` serve candidate A while REFLECTION.md shows candidate B — under
+    # #199's unconditional "SAME minibatch VERBATIM" claim. ``os.link`` needs a free name,
+    # and the unlink is safe because fresh writes no longer truncate a shared inode.
     dst = out_dir / f"{task_id}__{tag}__t0.json"
-    if dst != src and not dst.exists():
+    if dst != src:
         try:
+            dst.unlink(missing_ok=True)
             os.link(src, dst)
         except OSError:
             try:
-                dst.write_text(json.dumps(rec, default=str), encoding="utf-8")
-            except OSError:
-                pass  # ponytail: trajectories/ then falls back to REFLECTION.md excerpts
+                _atomic_write(dst, json.dumps(rec, default=str))
+            except OSError as e:
+                # Loud, not silent: the Score is still complete, but trajectories/ now
+                # lacks this task while the prompt claims VERBATIM completeness.
+                if run_dir is not None:
+                    run_dir.log_event("rollout_rematerialize_failed", task_id=task_id,
+                                      tag=tag, file=dst.name, error=str(e)[:300])
     return Score(
         task_id=task_id, reward=reward, feedback=str(cached.get("feedback", "")),
         n=1, stderr=0.0, trial_rewards=[reward],
--- a/core/cap_evolve/gepa.py
+++ b/core/cap_evolve/gepa.py
@@ -44,6 +44,7 @@ Pure stdlib + the cap_evolve engine.
 from __future__ import annotations
 
 import json
+import os
 import random
 import shutil
 import time
@@ -130,6 +131,17 @@ def _eval_minibatch(
       * consults the **eval cache** keyed on ``(candidate-hash, task_id)`` to skip
         a rollout that was already scored for byte-identical candidate files.
 
+    A cache hit yields the SAME reflective signal as a fresh eval (#111). The cache
+    entry points at the rollout json that produced the score; on a hit we re-read that
+    json for the real ``output``/``trace`` and re-materialize it under THIS eval's tag,
+    so ``rollouts/train/*__<tag>__t0.json`` is complete either way and the optimizer's
+    ``trajectories/`` (pinned to the tag by ``_copy_step_trajectories``) holds this
+    minibatch's rollouts verbatim. The candidate hash is over file CONTENTS, so a hit's
+    rollout genuinely IS this minibatch's record, not a lookalike from elsewhere.
+    A hit whose pointer does not resolve (a pre-#111 cache, or pruned rollouts) is
+    treated as a MISS and re-run: paying one rollout beats handing the optimizer an
+    empty "Agent output:" and calling it reflection.
+
     Minibatch tasks are drawn from TRAIN only (test stays sealed; val is for the
     honest gate). One trial per task — the minibatch is a cheap signal, not the
     significance test.
@@ -147,12 +159,9 @@ def _eval_minibatch(
     with _live(adapter, candidate_dir) as ctx:
         for task in tasks:
             cached = cache.get(chash, task.id) if cache is not None else None
-            if cached is not None:
-                reward = float(cached.get("reward", 0.0))
-                fb = str(cached.get("feedback", ""))
-                scores.append(Score(task_id=task.id, reward=reward, feedback=fb,
-                                    n=1, stderr=0.0, trial_rewards=[reward],
-                                    raw={"cached": True}))
+            replay = _replay_cached(cached, out_dir, task.id, tag) if cached else None
+            if replay is not None:
+                scores.append(replay)
                 continue
             rollout = adapter.run_target(task, ctx, seed=seed)
             if rollout is None:
@@ -169,13 +178,15 @@ def _eval_minibatch(
                      "output": _short(getattr(rollout, "output", None)),
                      "trace": _short(getattr(rollout, "trace", None))},
             ))
-            (out_dir / f"{task.id}__{tag}__t0.json").write_text(
+            rfile = f"{task.id}__{tag}__t0.json"
+            (out_dir / rfile).write_text(
                 json.dumps({"input": task.input, "rollout": rollout.to_dict(),
                             "score": sc.to_dict()}, default=str),
                 encoding="utf-8",
             )
             if cache is not None:
-                cache.put(chash, task.id, sc.reward, sc.feedback or "")
+                cache.put(chash, task.id, sc.reward, sc.feedback or "",
+                          rollout_file=rfile)
 
     elapsed = time.time() - t0
     # Count ONLY rollouts actually fired (cache hits cost nothing) toward budget.
@@ -187,6 +198,51 @@ def _eval_minibatch(
                       reward=result.reward, fired=n_called,
                       cached=len(task_ids) - n_called)
     return result
+
+
+def _replay_cached(cached: dict, out_dir: Path, task_id: str, tag: str) -> Score | None:
+    """Rebuild a full ``Score`` (output + trace) from a cache hit, or ``None`` = miss.
+
+    The fix for #111. The cache entry carries ``rollout_file`` — the rollout json in
+    ``rollouts/<split>/`` that produced the cached score. We re-read it for the real
+    ``output``/``trace``, so the reflective dataset built from a cached parent minibatch
+    is identical to one built from a fresh eval, and we copy it under THIS eval's tag so
+    the tag-pinned ``trajectories/`` dir exists too.
+
+    Returning ``None`` (→ the caller re-runs the rollout) is deliberate for a pointerless
+    or unreadable entry: a score-only hit is exactly the hollow-reflection bug, so we pay
+    one rollout instead of serving empty "Agent output:" as if it were a trace.
+    """
+    rfile = str(cached.get("rollout_file") or "")
+    if not rfile:
+        return None
+    src = out_dir / rfile
+    try:
+        rec = json.loads(src.read_text(encoding="utf-8"))
+    except (OSError, json.JSONDecodeError, ValueError):
+        return None
+    rollout = rec.get("rollout") or {}
+    reward = float(cached.get("reward", 0.0))
+    # Re-materialize under this eval's tag so ``_copy_step_trajectories(tag=...)`` finds
+    # this minibatch's rollouts. Same bytes, new name — no re-run, no fabrication. A
+    # hardlink keeps a cache hit free on disk too (rollout jsons are write-once); the
+    # copy fallback covers filesystems/paths where linking isn't available.
+    dst = out_dir / f"{task_id}__{tag}__t0.json"
+    if dst != src and not dst.exists():
+        try:
+            os.link(src, dst)
+        except OSError:
+            try:
+                dst.write_text(json.dumps(rec, default=str), encoding="utf-8")
+            except OSError:
+                pass  # ponytail: trajectories/ then falls back to REFLECTION.md excerpts
+    return Score(
+        task_id=task_id, reward=reward, feedback=str(cached.get("feedback", "")),
+        n=1, stderr=0.0, trial_rewards=[reward],
+        raw={"cached": True, "errored": bool(rollout.get("error")),
+             "output": _short(rollout.get("output")),
+             "trace": _short(rollout.get("trace"))},
+    )
 
 
 def _short(x, n: int = 1500) -> str:
--- a/core/cap_evolve/harness.py
+++ b/core/cap_evolve/harness.py
@@ -249,11 +249,14 @@ def evaluate_candidate(
             per_task_trials[tid].append(sc.reward)
             per_task_feedback[tid] = sc.feedback or per_task_feedback[tid]
             per_task_metrics[tid].append(sc.metrics)
-            (out_dir / f"{tid}__{tag}__t{k}.json").write_text(
-                json.dumps({"input": task.input, "rollout": rollout.to_dict(),
-                            "score": sc.to_dict()}, default=str),
-                encoding="utf-8",
-            )
+            # tmp + ``os.replace``, never ``write_text``: GEPA's cache replay hardlinks
+            # rollout jsons across tags, and truncating a shared inode would rewrite a
+            # PREVIOUS iteration's archived rollout in place. Replacing the directory
+            # entry breaks the link, keeping every archived record immutable.
+            _atomic_write(out_dir / f"{tid}__{tag}__t{k}.json",
+                          json.dumps({"input": task.input,
+                                      "rollout": rollout.to_dict(),
+                                      "score": sc.to_dict()}, default=str))
 
     # ``live()`` makes the candidate the one the target uses for this evaluation and
     # yields the ``ctx`` the runner consumes (default ctx == candidate_dir). Using a
--- a/core/tests/test_gepa_cache_traces.py
+++ b/core/tests/test_gepa_cache_traces.py
@@ -1,0 +1,166 @@
+"""A GEPA eval-cache HIT must yield the same reflective signal as a fresh eval (#111).
+
+The cache used to store only ``{reward, feedback}``, so ``_eval_minibatch`` rebuilt a
+``Score`` with ``raw={"cached": True}`` and no ``output``/``trace``. ``_write_reflection``
+then emitted ``- Agent output:`` (empty) for cached failing tasks — GEPA's whole learning
+signal, blank, on exactly the parents it re-samples most.
+
+These tests pin the fix at both levels: the cache entry carries a pointer to the rollout
+json that produced the score, and a hit re-reads it (and re-materializes it under the new
+eval tag so the tag-pinned ``trajectories/`` dir still exists).
+"""
+
+import json
+import sys
+from pathlib import Path
+
+import pytest
+
+REPO = Path(__file__).resolve().parents[2]
+sys.path.insert(0, str(REPO / "core"))
+
+
+def _adapter():
+    """A deterministic adapter whose rollouts carry a distinctive output AND trace."""
+    from cap_evolve import CapabilityAdapter, Rollout, Score, Task
+
+    class _A(CapabilityAdapter):
+        def tasks(self, split):  # noqa: ARG002
+            return [Task(id=f"t{i}", input=f"in-{i}", target="ok") for i in range(4)]
+
+        def run_target(self, task, ctx, *, seed=0):  # noqa: ARG002
+            return Rollout(task_id=task.id,
+                           output=f"WRONG-ANSWER-for-{task.id}",
+                           trace=f"STEP1 read {task.input}; STEP2 guessed")
+
+        def score(self, task, rollout):  # noqa: ARG002
+            return Score(task_id=task.id, reward=0.0,
+                         feedback=f"expected ok, got {rollout.output}",
+                         trial_rewards=[0.0])
+
+        def materialize(self, candidate_dir, edits=None):  # noqa: ARG002
+            return None
+
+    return _A()
+
+
+@pytest.fixture
+def setup(tmp_path):
+    from cap_evolve import Budget, RunDir
+    from cap_evolve.cache import EvalCache
+    cand = tmp_path / "cand"
+    cand.mkdir()
+    (cand / "cap.md").write_text("seed capability\n", encoding="utf-8")
+    run_dir = RunDir.create(tmp_path / ".capevolve", ts="c1", budget=Budget(max_iterations=4))
+    return _adapter(), run_dir, cand, EvalCache(run_dir.root / "eval_cache.json")
+
+
+def test_cache_hit_reflective_dataset_has_output_and_trace(setup):
+    """Fail-before / pass-after: a fully-cached minibatch must still produce a
+    REFLECTION.md with the agent's real output AND trajectory."""
+    from cap_evolve import gepa
+    adapter, run_dir, cand, cache = setup
+    ids = ["t0", "t1", "t2"]
+
+    first = gepa._eval_minibatch(adapter, cand, ids, run_dir=run_dir, cache=cache,
+                                 tag="mb_p_0000", seed=0)
+    # Second eval of the same candidate + tasks: every task is a cache hit (0 rollouts).
+    before = run_dir.spent.metric_calls
+    second = gepa._eval_minibatch(adapter, cand, ids, run_dir=run_dir, cache=cache,
+                                  tag="mb_p_0001", seed=0)
+    assert run_dir.spent.metric_calls == before, "expected a full cache hit"
+    assert second.reward == first.reward
+
+    for pt in second.per_task:
+        raw = pt.get("raw") or {}
+        assert raw.get("cached") is True
+        assert f"WRONG-ANSWER-for-{pt['task_id']}" in str(raw.get("output"))
+        assert "STEP1 read" in str(raw.get("trace"))
+
+    wd = run_dir.root / "work" / "x"
+    wd.mkdir(parents=True)
+    gepa._write_reflection(wd, second)
+    refl = (wd / "REFLECTION.md").read_text(encoding="utf-8")
+    assert "- Agent output: \n" not in refl, "hollow reflective dataset"
+    assert "WRONG-ANSWER-for-t0" in refl
+    assert "- Trajectory: STEP1 read in-0" in refl
+
+
+def test_cache_hit_rematerializes_rollouts_under_new_tag(setup):
+    """The cached hit must leave ``rollouts/train/*__<tag>__t0.json`` for the NEW tag, so
+    ``harness._copy_step_trajectories(tag=...)`` still finds this minibatch verbatim."""
+    from cap_evolve import gepa
+    adapter, run_dir, cand, cache = setup
+    gepa._eval_minibatch(adapter, cand, ["t0", "t1"], run_dir=run_dir, cache=cache,
+                         tag="mb_p_0000", seed=0)
+    gepa._eval_minibatch(adapter, cand, ["t0", "t1"], run_dir=run_dir, cache=cache,
+                         tag="mb_p_0001", seed=0)
+    got = sorted(p.name for p in (run_dir.rollouts / "train").glob("*__mb_p_0001__t0.json"))
+    assert got == ["t0__mb_p_0001__t0.json", "t1__mb_p_0001__t0.json"]
+    rec = json.loads((run_dir.rollouts / "train" / "t0__mb_p_0001__t0.json")
+                     .read_text(encoding="utf-8"))
+    assert rec["rollout"]["output"] == "WRONG-ANSWER-for-t0"
+    assert "STEP1" in rec["rollout"]["trace"]
+
+
+def test_cache_entry_stores_rollout_pointer_not_payload(setup):
+    """Cache format: reward + feedback + a POINTER. The trace is NOT copied into the
+    cache, so the cache stays tiny however large the traces get."""
+    from cap_evolve import gepa
+    adapter, run_dir, cand, cache = setup
+    gepa._eval_minibatch(adapter, cand, ["t0"], run_dir=run_dir, cache=cache,
+                         tag="mb_p_0000", seed=0)
+    raw = json.loads((run_dir.root / "eval_cache.json").read_text(encoding="utf-8"))
+    (entry,) = raw.values()
+    assert set(entry) == {"reward", "feedback", "rollout_file"}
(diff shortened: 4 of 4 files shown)


OsherElhadad:
## 🔬 Evidence

Every command + its literal output. Python `/tmp/ce-venv/bin/python`, worktree `/tmp/wt-111` off `origin/main` @ `6fca097`.

### 1. Fail-before (stash only the source change, keep the test)

```
$ git stash push core/cap_evolve/gepa.py core/cap_evolve/cache.py
Saved working directory and index state WIP on fix/issue-111-gepa-cache-traces: 6fca097 …

$ PYTHONPATH=core python -m pytest core/tests/test_gepa_cache_traces.py -q
        res = gepa._eval_minibatch(adapter, cand, ["t0"], run_dir=run_dir, cache=cache,
                                   tag="mb_p_0001", seed=0)
        trace = str((res.per_task[0]["raw"] or {}).get("trace"))
>       assert len(trace) < 2000 and trace.endswith("…[truncated]")
E       AssertionError: assert (4 < 2000 and False)
E        +  where 4 = len('None')
E        +  and   False = <built-in method endswith of str object at 0x108802250>('…[truncated]')
E        +    where <built-in method endswith of str object at 0x108802250> = 'None'.endswith

core/tests/test_gepa_cache_traces.py:164: AssertionError
FAILED core/tests/test_gepa_cache_traces.py::test_cache_hit_reflective_dataset_has_output_and_trace
FAILED core/tests/test_gepa_cache_traces.py::test_cache_hit_rematerializes_rollouts_under_new_tag
FAILED core/tests/test_gepa_cache_traces.py::test_cache_entry_stores_rollout_pointer_not_payload
FAILED core/tests/test_gepa_cache_traces.py::test_pointerless_or_missing_rollout_is_treated_as_a_miss
FAILED core/tests/test_gepa_cache_traces.py::test_trace_is_bounded_in_the_reflective_signal
5 failed in 0.53s
```

### 2. Pass-after

```
$ git stash pop
$ PYTHONPATH=core python -m pytest core/tests/test_gepa_cache_traces.py -q
.....                                                                    [100%]
5 passed in 0.07s
```

### 3. Real end-to-end, zero API cost — `examples/toy_calc`, `mock` optimizer, `gepa` algorithm

Setup (identical for the before and after runs). `noop_edit.json` is a mock edit that appends a comment with **no behavior change**, so the local minibatch gate rejects every child and GEPA keeps re-sampling the **still-failing seed as parent** — precisely the issue's scenario (a re-sampled failing parent whose minibatch is served from cache).

```bash
export CAPEVOLVE_CORE=$PWD/core PYTHONPATH=$PWD/core CAPEVOLVE_SKILLS_DIR=$PWD/skills \
       CAPEVOLVE_TOY_DATA=$PWD/examples/toy_calc CAPEVOLVE_MOCK_SCRIPT=$D/noop_edit.json
mkdir -p $D/.capevolve/project/adapters
cp examples/toy_calc/adapter.py $D/.capevolve/project/adapters/
cp -R examples/toy_calc/capability $D/seed_capability
cat > $D/noop_edit.json <<'EOF'
{"edits": [{"file": "prompt.txt", "op": "ensure_contains", "text": "\n# note: iteration touched this file (no behavior change)"}]}
EOF
sed -e 's/^algorithm_skill: hill-climb/algorithm_skill: gepa/' \
    -e 's/^max_iterations: 10/max_iterations: 3/' -e 's/^stall: 2/stall: 99/' \
    templates/project/capevolve.yaml > $D/.capevolve/project/capevolve.yaml
python -m cap_evolve.cli run --spec $D/.capevolve/project/capevolve.yaml \
       --project $D/.capevolve/project --run-ts demo --dashboard off
```

Run result (identical before/after — the fix changes the *reflective signal*, not the score):
```json
{
  "best_id": "seed",
  "baseline_val": 0.0,
  "test_reward": 0.0,
  "iterations": 3
}
```

The cache hits, from `events.jsonl` — iterations 2 and 3 fire **0** rollouts, **4** cached:
```
{"kind": "minibatch", "tag": "mb_p_0000", "ids": ["a3","a5","a2","a6"], "reward": 0.0, "fired": 4, "cached": 0}
{"kind": "minibatch", "tag": "mb_c_0000", "ids": ["a3","a5","a2","a6"], "reward": 0.0, "fired": 4, "cached": 0}
{"kind": "gepa_local_gate", "candidate": "gepa_0001", "parent": "seed", "child_sum": 0.0, "parent_sum": 0.0, "passed": false}
{"kind": "minibatch", "tag": "mb_p_0001", "ids": ["a6","a2","a3","a5"], "reward": 0.0, "fired": 0, "cached": 4}
{"kind": "minibatch", "tag": "mb_c_0001", "ids": ["a6","a2","a3","a5"], "reward": 0.0, "fired": 0, "cached": 4}
{"kind": "gepa_local_gate", "candidate": "gepa_0002", "parent": "seed", "child_sum": 0.0, "parent_sum": 0.0, "passed": false}
{"kind": "minibatch", "tag": "mb_p_0002", "ids": ["a6","a5","a3","a2"], "reward": 0.0, "fired": 0, "cached": 4}
{"kind": "minibatch", "tag": "mb_c_0002", "ids": ["a6","a5","a3","a2"], "reward": 0.0, "fired": 0, "cached": 4}
{"kind": "gepa_local_gate", "candidate": "gepa_0003", "parent": "seed", "child_sum": 0.0, "parent_sum": 0.0, "passed": false}
```

#### BEFORE — `work/gepa_0002/REFLECTION.md` (source change stashed), full file

```
# Reflective dataset (GEPA)

Parent minibatch reward: 0.000 (0/4 sampled tasks pass). Below are the FAILING tasks with the agent's actual output/trajectory and the scorer's feedback. Diagnose the COMMON root cause and edit the capability to fix the general pattern — not one task.

## 4 actionable failing task(s)
### task a6
- Agent output: 
- Feedback: expected '12' but agent produced 'I think 6 * 2 is roughly some number.'; the prompt likely lacks an explicit instruction to compute and output only the number

### task a2
- Agent output: 
- Feedback: expected '4' but agent produced 'I think 10 - 6 is roughly some number.'; the prompt likely lacks an explicit instruction to compute and output only the number

### task a3
- Agent output: 
- Feedback: expected '10' but agent produced 'I think 2 * 5 is roughly some number.'; the prompt likely lacks an explicit instruction to compute and output only the number

### task a5
- Agent output: 
- Feedback: expected '5' but agent produced 'I think 8 - 3 is roughly some number.'; the prompt likely lacks an explicit instruction to compute and output only the number
```

Four blank `Agent output:` lines, zero `Trajectory:` lines — hollow, exactly as #111 describes.

#### AFTER — same file, same iteration, same 4/4 cache hit, full file

```
# Reflective dataset (GEPA)

Parent minibatch reward: 0.000 (0/4 sampled tasks pass). Below are the FAILING tasks with the agent's actual output/trajectory and the scorer's feedback. Diagnose the COMMON root cause and edit the capability to fix the general pattern — not one task.

## 4 actionable failing task(s)
### task a6
- Agent output: I think 6 * 2 is roughly some number.
- Trajectory: prompt_had_calc=False
- Feedback: expected '12' but agent produced 'I think 6 * 2 is roughly some number.'; the prompt likely lacks an explicit instruction to compute and output only the number

### task a2
- Agent output: I think 10 - 6 is roughly some number.
- Trajectory: prompt_had_calc=False
- Feedback: expected '4' but agent produced 'I think 10 - 6 is roughly some number.'; the prompt likely lacks an explicit instruction to compute and output only the number

### task a3
- Agent output: I think 2 * 5 is roughly some number.
- Trajectory: prompt_had_calc=False
- Feedback: expected '10' but agent produced 'I think 2 * 5 is roughly some number.'; the prompt likely lacks an explicit instruction to compute and output only the number

### task a5
- Agent output: I think 8 - 3 is roughly some number.
- Trajectory: prompt_had_calc=False
- Feedback: expected '5' but agent produced 'I think 8 - 3 is roughly some number.'; the prompt likely lacks an explicit instruction to compute and output only the number
```

Genuine per-task output **and** trajectory — `prompt_had_calc=False` is the actual root cause the optimizer needs and previously never saw.

### 4. Cache format + size impact, same run

```
$ # BEFORE
$ python -c "import json,os; d=json.load(open('$B/eval_cache.json')); k=list(d)[0]; print(json.dumps({k:d[k]},indent=2)); print('bytes:',os.path.getsize('$B/eval_cache.json'),'entries:',len(d))"
{
  "6a1c1a0ed1679765499f01e0d1305e5b0af2f930d56cfd25a4e01630f3bf2667::a3": {
    "reward": 0.0,
    "feedback": "expected '10' but agent produced 'I think 2 * 5 is roughly some number.'; the prompt likely lacks an explicit instruction to compute and output only the number"
  }
}
bytes: 2110 entries: 8

$ # AFTER
{
  "6a1c1a0ed1679765499f01e0d1305e5b0af2f930d56cfd25a4e01630f3bf2667::a3": {
    "reward": 0.0,
    "feedback": "expected '10' but agent produced 'I think 2 * 5 is roughly some number.'; the prompt likely lacks an explicit instruction to compute and output only the number",
    "rollout_file": "a3__mb_p_0000__t0.json"
  }
}
bytes: 2446 entries: 8    bytes/entry: 305.8
```

`+336 B over 8 entries = +42 B per entry`, and that cost is a fixed filename — **independent of trace size**.

Disk (hardlinks make a cache hit free; `du -sk`):
```
$ du -sk $B/rollouts $A/rollouts
48	/tmp/e2e111_before/.capevolve/run_demo/rollouts
48	/tmp/e2e111c/.capevolve/run_demo/rollouts
$ du -sk $B $A
436	/tmp/e2e111_before/.capevolve/run_demo
436	/tmp/e2e111c/.capevolve/run_demo
```
24 rollout *names* after vs 8 before (`mb_p_0001`/`mb_p_0002` re-materialized), **0 KB** of new bytes.

The bounded-trace test proves the ceiling holds with a large trace:
```python
huge = "X" * 50_000   # 50 KB trace
...
assert len(trace) < 2000 and trace.endswith("…[truncated]")   # replayed trace bounded
assert (run_dir.root / "eval_cache.json").stat().st_size < 1000   # cache did NOT grow by it
```

### 5. Full suite, this branch

```
........................................                                 [100%]
184 passed in 60.70s (0:01:00)
```
179 baseline + 5 new, **0 failed**. (`test_dashboard_launch.py::test_maybe_launch_spawns_when_available` — the #200 port-7878 flake — passed here.)

### 6. Merge compatibility with #199 and #197

```
$ git merge origin/fix/issue-109-optimizer-context      # PR #199
 18 files changed, 1064 insertions(+), 181 deletions(-)   ← CLEAN, no conflicts

$ git merge origin/feat/issue-142-protected-paths        # PR #197
CONFLICT (content): Merge conflict in core/cap_evolve/__init__.py
CONFLICT (content): Merge conflict in core/cap_evolve/gepa.py
CONFLICT (content): Merge conflict in core/cap_evolve/harness.py

$ grep -n "<<<<<<<\|=======\|>>>>>>>" core/cap_evolve/{__init__,harness,gepa}.py
core/cap_evolve/__init__.py:20:<<<<<<< HEAD
core/cap_evolve/__init__.py:22:=======
core/cap_evolve/__init__.py:24:>>>>>>> origin/feat/issue-142-protected-paths
core/cap_evolve/harness.py:27:<<<<<<< HEAD
core/cap_evolve/harness.py:31:=======
core/cap_evolve/harness.py:33:>>>>>>> origin/feat/issue-142-protected-paths
core/cap_evolve/gepa.py:55:<<<<<<< HEAD
core/cap_evolve/gepa.py:57:=======
core/cap_evolve/gepa.py:59:>>>>>>> origin/feat/issue-142-protected-paths
```

All three conflicts are **import lines between #199 and #197** (`from . import optimizer_context as oc` vs `from . import protect`) — they exist independently of this PR and resolve by keeping both sides. **Zero conflicts in this PR's hunks.**

Full suite on the merged tree (main + #199 + #197 + this):
```
........................................................................ [ 31%]
........................................................................ [ 63%]
........................................................................ [ 95%]
227 passed in 74.41s (0:01:14)
```

### 7. #199's warning is now unnecessary — merged-tree e2e

Same no-op e2e run, on the tree with all three PRs:
```
$ grep -c optimizer_context_warning $D/.capevolve/run_demo/events.jsonl
0

$ ls $D/.capevolve/run_demo/work/gepa_0002/trajectories/
a2__mb_p_0001__t0.json
a3__mb_p_0001__t0.json
a5__mb_p_0001__t0.json
a6__mb_p_0001__t0.json

$ grep -o "trajectories/\` holds[^.]*\|served entirely from the eval cache" \
     $D/.capevolve/run_demo/work/gepa_0002/INSTRUCTIONS.md
trajectories/` holds the SAME minibatch rollouts VERBATIM and untruncated — read them when REFLECTION
```

On `origin/fix/issue-109-optimizer-context` alone this iteration produced an `optimizer_context_warning` with `"no rollouts persisted for the pinned eval tag (fully-cached minibatch); trajectories/ OMITTED"`, no `trajectories/` dir, and the `"served entirely from the eval cache"` prompt branch. After this PR: **0 warnings**, the dir is present with this exact minibatch's four rollouts, and the prompt makes the truthful VERBATIM claim. #199's honest-absence path is preserved and still correct for a pre-#111 or pruned cache.

### 8. compileall

```
COMPILE_OK
```

### Files touched

- `core/cap_evolve/cache.py` — `put(..., rollout_file=...)`, module + class docstrings
- `core/cap_evolve/gepa.py` — `_replay_cached()` (new), cache-hit branch of `_eval_minibatch`, `cache.put` call, `import os`, `_eval_minibatch` docstring
- `core/tests/test_gepa_cache_traces.py` — new, 5 tests

OsherElhadad:
## 🔍 Review — PR #210

The diagnosis and the *content* of the fix are right: I independently confirmed the reflective dataset is now genuinely per-task **correct** (not merely non-empty), truncation is symmetric, and the cached REFLECTION.md is byte-identical to an uncached one. Every claim in the evidence comment reproduced. But the re-materialization mechanism has two real defects: an unvalidated `rollout_file` path (a tampered cache entry reads *any* JSON-shaped file on disk into the optimizer prompt — and hardlinks it into the run dir), and the hardlink's shared inode lets a later write silently rewrite an **earlier iteration's** archived rollout. Both are two-line fixes.

### Blocking

**1. `gepa.py:216-219` — `rollout_file` is not validated or confined to the rollouts dir.**
`src = out_dir / rfile` is a bare `Path.__truediv__` on a value read from `eval_cache.json`, a plain on-disk JSON file. `../` traversal escapes; an absolute path *replaces* `out_dir` entirely (`Path("/a/b") / "/etc/x" == Path("/etc/x")`). This is precisely the threat #142 models — the guard hashes `.capevolve/project`, it does not cover `eval_cache.json`. Any JSON file with a `rollout` key is exfiltrated into `REFLECTION.md`, i.e. into the optimizer LLM's prompt:

```
$ /tmp/ce-venv/bin/python /tmp/probe_exfil.py
crafted rollout_file = ../../../../secrets.json
fired rollouts: 0
output in reflective signal: 'sk-SUPER-SECRET-KEY'
trace  in reflective signal: 'AWS_SECRET=xyz'
--- REFLECTION.md leaked? ---
True
ABSOLUTE path output: 'sk-SUPER-SECRET-KEY'      # rollout_file="/tmp/probe_exfil/secrets.json"
```

Worse, `os.link` at `gepa.py:233` then hardlinks the **out-of-tree inode** into `rollouts/train/`, and a later write on that tag writes *through* it. Pointed at a protected `gold.json`:

```
crafted pointer: ../../../project/gold.json | gold nlink before: 1
hardlink created into rollouts: True | shares gold's inode: True | gold nlink now: 2
cap_evolve.protect.TamperError: ... modified gold.json
```

The #142 guard catches the *write* (good — the run aborts loudly), but the read/exfiltration is silent and the link is created regardless. Cost accounting also reads `reward` unchecked: `_replay_cached({"reward": "not-a-number", ...})` → `ValueError: could not convert string to float`, an uncaught crash mid-loop.

**Fix** — confine the pointer, one guard at the top of `_replay_cached`:
```python
rfile = str(cached.get("rollout_file") or "")
# rollout_file is a BARE FILENAME under out_dir. The cache is plain on-disk JSON, so a
# tampered/absolute/../ entry must not read outside rollouts/ (#142's threat model).
if not rfile or Path(rfile).name != rfile:
    return None
```
`Path(rfile).name != rfile` rejects `/etc/x`, `../y`, and `sub/z` in one comparison, and every entry `put()` writes is already a bare `f"{task.id}__{tag}__t0.json"`, so nothing legitimate is lost. Also wrap the `float(cached.get("reward", 0.0))` in the existing `except` (or add `TypeError, ValueError`) so a malformed number degrades to MISS rather than aborting.

**2. `gepa.py:230-236` — the hardlink shares an inode, so a later `write_text` on a reused tag silently rewrites a PREVIOUS iteration's archived rollout.**
`_eval_minibatch`'s fresh-eval writer (`gepa.py:182`) is `Path.write_text` → `open(..., "w")`, which **truncates the shared inode** rather than replacing the directory entry. Reachable path: tags are `f"mb_p_{n:04d}"` / `f"mb_c_{n:04d}"` where `n = step_offset + len(steps)`, and `_save()` only runs at the *end* of a step (`gepa.py:728`, `gepa.py:781`) — so a run torn mid-iteration and `--resume`d **repeats the same tag**. In the e2e run I did, a no-op child edit is byte-identical to the parent, so `mb_c_0000` is a cache HIT hardlinked onto the seed's `mb_p_0000` inode (verified: `nlink=3` across all 8 inodes). On resume the optimizer proposes a real edit → new hash → MISS on that same tag → the truncate lands on iteration 0's record:

```
$ /tmp/ce-venv/bin/python /tmp/probe_corrupt3.py
iter-0 archived parent rollout   : out-of[PARENT-SEED]
mb_c_0000 shares the inode       : True nlink = 2
--- torn iteration; --resume repeats step n=0 with a REAL edit (cache MISS) ---
iter-0 archived parent rollout NOW: out-of[CHILD-REAL-EDIT]
RESULT: PAST ITERATION SILENTLY CORRUPTED
```

Consequence: the audit trail this PR exists to make trustworthy can be retroactively rewritten with *another candidate's* output — no event, no warning. `harness._copy_step_trajectories` also uses `shutil.copyfile` (`harness.py:1020`), which write-throughs a hardlinked destination the same way; and `_reconstruct_gepa` reads val results back off these files, so a corrupted rollout is silently *believed*.

**Fix** — make the fresh-eval write break the link instead of truncating it. `rundir._atomic_write` (already imported in harness, tmp+`os.replace`) does exactly this; the two-line version at `gepa.py:181`:
```python
rfile = f"{task.id}__{tag}__t0.json"
(out_dir / rfile).unlink(missing_ok=True)   # never truncate a hardlinked inode: a cache
                                            # hit may have linked a PRIOR iteration here
(out_dir / rfile).write_text(...)
```
Add a test that pins it: link a rollout under a second tag, write a fresh rollout to that second tag, assert the first tag's bytes are unchanged.

### Non-blocking

**3. `gepa.py:231` — the `not dst.exists()` guard makes a reused tag serve a *stale* trajectory, contradicting the "SAME minibatch VERBATIM" prompt claim.** On a resumed run where tag `mb_p_0000` already exists from parent A but the frontier now selects parent B, the replay skips materialization and `trajectories/` keeps A's rollout while `REFLECTION.md` shows B's — under #199's unconditional "VERBATIM" assertion:
```
mb_p_0000 on disk: OUT-A
replayed Score (what REFLECTION.md gets): OUT-B
mb_p_0000 on disk (what trajectories/ gets): OUT-A
MISMATCH: True
```
Fix: overwrite when the existing record's `rollout.task_id`/content doesn't match, or drop the `not dst.exists()` short-circuit once finding 2's unlink is in (the unlink makes the overwrite safe).

**4. `gepa.py:238` — the copy fallback swallows `OSError` to `pass`.** When both link and copy fail, the hit is still returned as a complete `Score`, so `trajectories/` silently lacks this task while the prompt (on the merged tree, `has_traj` is `True` if *any* file landed) still claims VERBATIM completeness. The `# ponytail:` comment names the intent but there's no `log_event`. Fix: `run_dir.log_event("rollout_rematerialize_failed", ...)` — cheap, and keeps the honesty posture the rest of the module has. (Requires threading `run_dir` or returning a flag.)

**5. Untested critical paths.** The 5 new tests cover the happy path, the pointer format, dangling/legacy MISS, and truncation — genuinely content-asserting, not non-emptiness theatre (`assert f"WRONG-ANSWER-for-{pt['task_id']}" in ...` is real per-task correspondence). Missing: hardlink-unavailable / cross-fs `EXDEV` fallback, tag reuse (findings 2 and 3), and path traversal. The `EXDEV` fallback does work — I verified the copy is byte-identical to the link source by monkeypatching `os.link` to raise `OSError(18)` — but nothing in CI pins it.

### Nits

**6. `cache.py:108-110` — `rollout_file` is written only when truthy, so entries are heterogeneous** (`{reward, feedback}` vs `{reward, feedback, rollout_file}`). Harmless (the reader treats absence as MISS), but the docstring at `cache.py:80` promises a 3-key shape. One word in the docstring.

**7. `gepa.py:263` — `raw={"cached": True}` is preserved (good) but has zero consumers.** `grep -rn '"cached"' core/` outside the new tests finds only this write site; dashboard and report don't read it. Nothing broken, just noting the marker is currently write-only, so no downstream behavior changed.

### Hardlink audit

| Scenario | Behavior | Acceptable? |
|---|---|---|
| Cross-filesystem (`EXDEV`) | `os.link` raises → `dst.write_text(json.dumps(rec))`. Verified byte-identical to the link source. | ✅ |
| Filesystem without hardlink support | Same `OSError` branch → copy fallback. | ✅ |
| Both link and copy fail | `pass` — Score still returned complete, no event logged. | ⚠️ (finding 4) |
| **In-place mutation of the target** | `write_text` truncates the **shared inode**; a reused tag rewrites a *previous* iteration's archived rollout. `nlink=3` observed in the real e2e run. | ❌ **(finding 2)** |
| Deleted target | Cache entry becomes a MISS → re-runs, cost accounted (`usd delta: 0.5`, `fired: 1`), score complete (`n: 1`), and the entry **self-heals** to the new `rollout_file`. No loop, no partial score. | ✅ |
| Windows | `os.link` works on NTFS; on FAT/exFAT/network shares it raises → copy fallback. Untested in CI but structurally covered. | ✅ |
| Existing `dst` from a different parent (tag reuse) | Skipped — stale trajectory served under a VERBATIM claim. | ❌ (finding 3) |
| **#197 tamper guard false-positive** | **NO false positive.** `protect.build_manifest` hashes `.capevolve/project` with `exclude=run_dir.root` (`protect.py:362,407`); `rollouts/` is inside the run dir and never hashed. Real merged-tree gepa run: `grep -c tamper_detected events.jsonl` → **0**, `protected.json` present. | ✅ |

### Is the reflection CORRECT?

**Yes — per-task correspondence holds**, and I tested it adversarially rather than taking the non-emptiness evidence at face value.

Cache populated in order `t0,t1,t2,t3`, replayed in a **different** order `t3,t0,t2,t1` (real minibatches are rng-sampled, so ordering must not be load-bearing). Every field cross-checked against its own task id:

```
fired (want 0): 0
  t3: OK  out='OUT-t3-IN-3' trace='TRACE-t3-only' fb='FB-t3'
  t0: OK  out='OUT-t0-IN-0' trace='TRACE-t0-only' fb='FB-t0'
  t2: OK  out='OUT-t2-IN-2' trace='TRACE-t2-only' fb='FB-t2'
  t1: OK  out='OUT-t1-IN-1' trace='TRACE-t1-only' fb='FB-t1'
PER-TASK CORRESPONDENCE: CORRECT
```
The re-materialized files carry the matching record, so filename ↔ `task_id` ↔ `input` all agree (no pointer mix-up):
```
  t0__mb_p_0001__t0.json -> t0 OUT-t0-IN-0 | input field: IN-0
  t1__mb_p_0001__t0.json -> t1 OUT-t1-IN-1 | input field: IN-1
  t2__mb_p_0001__t0.json -> t2 OUT-t2-IN-2 | input field: IN-2
  t3__mb_p_0001__t0.json -> t3 OUT-t3-IN-3 | input field: IN-3
```
Same in the real merged-tree gepa run's fully-cached iteration (`gepa_0002`, 0 fired / 4 cached) — each task's arithmetic matches its own input:
```
  a2__mb_p_0001__t0.json | record task_id: a2 | input: 10 - 6 | output: I think 10 - 6 is roughly...
  a3__mb_p_0001__t0.json | record task_id: a3 | input: 2 * 5  | output: I think 2 * 5 is roughly...
  a5__mb_p_0001__t0.json | record task_id: a5 | input: 8 - 3  | output: I think 8 - 3 is roughly...
  a6__mb_p_0001__t0.json | record task_id: a6 | input: 6 * 2  | output: I think 6 * 2 is roughly...
```

**Truncation is symmetric** (5 KB output + a nested-dict trace, fresh vs cached):
```
  fresh  out len 1513 trace len 1513
  cached out len 1513 trace len 1513
  IDENTICAL output: True | IDENTICAL trace: True
```
Both go through the same `_short(n=1500)`. It cuts mid-structure on a JSON-serialized dict trace (`'YYY… …[truncated]'` is not valid JSON) — but that is **pre-existing** `_short` behavior on the fresh path too, so cache state doesn't change it. Not a #210 finding.

**Determinism: byte-identical.** Cached vs uncached `REFLECTION.md`:
```
uncached sha: fc2f9e6a4884cc44
cached   sha: fc2f9e6a4884cc44
BYTE-IDENTICAL REFLECTION.md: True
raw keys fresh : ['errored', 'output', 'trace']
raw keys cached: ['cached', 'errored', 'output', 'trace']
```
The only divergence is the extra `cached` key in `raw`, which `_write_reflection` doesn't render. Published results reproduce across cache states.

### Merged-tree result

My own merge, `main` @ `6fca097` + #199 + #197 + #210. Confirmed **3 conflicts, all between #199 and #197 only** (`__init__.py:20`, `harness.py:27`, `gepa.py:54` — every one a keep-both import pair: `optimizer_context` vs `protect`). **#210 merged with zero conflicts.**

```
$ PYTHONPATH=/tmp/rv-merge/core python -m pytest core/tests -q
227 passed in 75.64s (0:01:15)
```

Real zero-API gepa run on `examples/toy_calc` + `mock` optimizer, merged tree — every #199 interaction claim reproduces:
```
$ grep -c optimizer_context_warning events.jsonl   → 0
$ grep -c tamper_detected events.jsonl             → 0        (protected.json present)
{"kind":"minibatch","tag":"mb_p_0001","ids":["a6","a2","a3","a5"],"reward":0.0,"fired":0,"cached":4}
{"kind":"minibatch","tag":"mb_p_0002","ids":["a6","a5","a3","a2"],"reward":0.0,"fired":0,"cached":4}
$ work/gepa_0002/trajectories/ → a2__mb_p_0001__t0.json a3__… a5__… a6__…
$ grep -o "trajectories/\` holds the SAME[^.]*" work/*/INSTRUCTIONS.md
  → all 3 iterations make the VERBATIM claim
```
The "trajectories/ OMITTED" branch fires **0 times**, and the VERBATIM claim is **truthful** — the dir holds exactly this minibatch's four tasks with matching inputs (see above). #199's honest-absence path remains correct for a pre-#111 / pruned cache. The author's "self-correcting, no #199 edit needed" reasoning holds.

**Recommended merge order:** `#197 → #199 → #110 → #210 → #114`. #197 and #199 are the only pair that actually conflicts, so land them first and resolve the three import lines once (keep both sides). #110 merges clean into #210. #210 last of the gepa trio — its `_eval_minibatch` hunk sits next to but not on #197's `protect.verify` call. `#114` conflicts with #210 in `core/cap_evolve/cache.py` **docstring only** (the #111 `rollout_file` paragraph vs #114's shortened header) — trivial, but rebase #114 rather than #210 since #114 is a refactor.

### Verification I re-ran

```
$ cd /tmp/rv-210 && PYTHONPATH=core python -m pytest core/tests -q
184 passed in 58.38s                                  ← 179 baseline + 5, claim reproduces

$ git checkout origin/main -- core/cap_evolve/gepa.py core/cap_evolve/cache.py
$ PYTHONPATH=core python -m pytest core/tests/test_gepa_cache_traces.py -q
FAILED ...::test_cache_hit_reflective_dataset_has_output_and_trace
FAILED ...::test_cache_hit_rematerializes_rollouts_under_new_tag
FAILED ...::test_cache_entry_stores_rollout_pointer_not_payload
FAILED ...::test_pointerless_or_missing_rollout_is_treated_as_a_miss
FAILED ...::test_trace_is_bounded_in_the_reflective_signal
5 failed in 0.54s                                     ← fail-before reproduces exactly
$ # restored
5 passed in 0.04s

COMPILE_OK

$ # (9) pre-#111 old-format eval_cache.json, real file, score-only entry
loaded OK; fired: 1  reward now: 0.0  raw: {'errored': False, 'output': 'OUT', 'trace': 'TR'}
cost accounted: 0.5 tokens: 7                         ← degrades to MISS, re-runs, complete

$ # (4) dangling pointer
fired: 1  usd delta: 0.5  reward complete: 0.0  n: 1
cache entry AFTER the miss-rerun: t0__mb_p_0001__t0.json (self-heals -> True)
                                                      ← no loop, no partial score

$ # (1) EXDEV fallback (os.link monkeypatched to raise OSError(18))
fallback file exists: True | shares inode: False
byte-identical to hardlink source: True

$ # real hardlink state, merged-tree e2e run
inode 212538709: nlink=3  ['a2__mb_p_0000__t0.json','a2__mb_p_0001__t0.json','a2__mb_p_0002__t0.json']
… 8 distinct inodes / 24 files                        ← sharing is real, hence finding 2
```

Both blocking findings are contained in `_replay_cached` + one line in `_eval_minibatch`. The reflective-dataset work itself — the part that matters for #111 — is correct and I'd take it as-is.

OsherElhadad:
## 🔧 Review fixes

All 7 findings addressed. Both blocking exploits re-run against the fixed code and **closed**. Nothing in the validated reflective-dataset work was churned — re-proved below.

### Exploits closed

| Attack | Previously | Now | Test pinning it |
|---|---|---|---|
| `rollout_file = "../../../../secrets.json"` | `fired=0`, `sk-SUPER-SECRET-KEY` + `AWS_SECRET=xyz` into `REFLECTION.md` (canary count **1**), secrets inode hardlinked into `rollouts/` (`nlink=2`) | `fired=1` (honest MISS + real re-run), canary count **0**, `secrets nlink=1` | `test_crafted_rollout_pointer_cannot_read_outside_the_rollouts_dir[../../../../secrets.json]` |
| `rollout_file = "/abs/path/secrets.json"` | same exfiltration, `nlink=3` | canary count **0**, `nlink=1` | `…[/etc/passwd]` (absolute) |
| `rollout_file = "sub/secrets.json"` | escaped `out_dir` | MISS | `…[sub/secrets.json]` |
| `reward = "not-a-number"` | uncaught `ValueError` mid-loop | degrades to MISS, no crash | `test_malformed_cached_reward_degrades_to_a_miss_instead_of_crashing` |
| Torn iteration + `--resume` tag reuse | iteration-0 archived rollout silently rewritten `out-of[PARENT-SEED]` → `out-of[CHILD-REAL-EDIT]` | archived rollout **UNCHANGED** | `test_fresh_write_never_rewrites_a_hardlinked_archived_rollout` |
| Tag reuse across parents | `trajectories/` served `OUT-A` while `REFLECTION.md` showed `OUT-B` under the VERBATIM claim | both `OUT-B`, `MISMATCH: False` | `test_tag_reuse_replaces_a_prior_parents_trajectory` |

**Exploit 1 — before → after** (same probe, same canary):
```
BEFORE
--- rel: rollout_file = ../../../../secrets.json
    fired rollouts (0 == exploit served from 'cache'): 0
    output in reflective signal : 'sk-SUPER-SECRET-KEY'
    trace  in reflective signal : 'AWS_SECRET=xyz'
    canary count in REFLECTION.md: 1
    hardlinked secrets inode into rollouts/: True  (secrets nlink=2)
--- abs: rollout_file = /var/folders/.../secrets.json
    canary count in REFLECTION.md: 1
    hardlinked secrets inode into rollouts/: True  (secrets nlink=3)
--- malformed reward ---
    malformed reward: CRASH ValueError: could not convert string to float: 'not-a-number'
EXFILTRATED (rel=1, abs=1) -> exploit OPEN

AFTER
--- rel: rollout_file = ../../../../secrets.json
    fired rollouts (0 == exploit served from 'cache'): 1
    output in reflective signal : 'WRONG-ANSWER-for-t0'
    trace  in reflective signal : 'STEP1 read in-0; STEP2 guessed'
    canary count in REFLECTION.md: 0
    hardlinked secrets inode into rollouts/: False  (secrets nlink=1)
--- abs: rollout_file = /var/folders/.../secrets.json
    canary count in REFLECTION.md: 0
    hardlinked secrets inode into rollouts/: False  (secrets nlink=1)
--- malformed reward ---
    malformed reward: handled (no crash)
EXFILTRATED (rel=0, abs=0) -> exploit CLOSED
```

**Exploit 2 + finding 3 — before → after**:
```
BEFORE                                          AFTER
iter-0 archived parent rollout   : out-of[PARENT-SEED]      out-of[PARENT-SEED]
mb_c_0000 shares the inode       : True nlink = 2           True nlink = 2
--- torn iteration; --resume repeats step n=0 with a REAL edit (cache MISS) ---
iter-0 archived parent rollout NOW: out-of[CHILD-REAL-EDIT] out-of[PARENT-SEED]
RESULT: PAST ITERATION SILENTLY CORRUPTED       archived rollout UNCHANGED (safe)

--- finding 3: stale dst.exists() under the VERBATIM claim ---
replayed Score (REFLECTION.md) : out-of[OUT-B]  out-of[OUT-B]
mb_p_0000 on disk (trajectories/): out-of[OUT-A]  out-of[OUT-B]
MISMATCH: True                                  MISMATCH: False
```

### Threat-model note

`eval_cache.json` lives in the **optimizer-writable run dir**, so every field it carries is untrusted input — **exactly the threat #142 models**. #197's guard hashes `.capevolve/project` with `exclude=run_dir.root`, so it deliberately does *not* cover this file (correctly — that's why `rollouts/` produces no false positive). The trust boundary is therefore `_replay_cached` itself, and validation has to live there. Following the lesson this epic has now learned three times (#192's 5-host CDN list, #209's single-sequence HTML escape, #197's `.pyc` skip), the guard is an **allowlist of the one legitimate shape**, not a denylist of bad ones: `Path(rfile).name != rfile` rejects absolute, `../` and `sub/` in a single comparison, and every pointer `put()` writes is already a bare `f"{task.id}__{tag}__t0.json"`, so nothing legitimate is lost. `src.resolve().parent != out_dir.resolve()` is the second layer, closing symlink indirection.

### nlink evidence

Sharing is *retained* (a cache hit is still free on disk) — what changed is that a **fresh write replaces the directory entry instead of truncating the shared inode**. Same 3-tag replay, before and after:

```
BEFORE                                                          AFTER
inode …094: nlink=3 [t0__mb_p_0000, t0__mb_p_0001, t0__mb_p_0002]   inode …586: nlink=3 [t0__mb_p_0000, t0__mb_p_0001, t0__mb_p_0002]
inode …096: nlink=3 [t1__…]                                          inode …588: nlink=3 [t1__…]
inode …098: nlink=3 [t2__…]                                          inode …590: nlink=3 [t2__…]
inode …100: nlink=3 [t3__…]                                          inode …592: nlink=3 [t3__…]
4 distinct inodes / 12 files                                         4 distinct inodes / 12 files
```

Real zero-API gepa e2e (`examples/toy_calc` + `mock`, 3 iterations), merged tree behavior unchanged:
```
--- cache hits (fired 0 / cached N) ---
  tag=mb_p_0000 fired=4 cached=0
  tag=mb_c_0000 fired=4 cached=0
  tag=mb_p_0001 fired=0 cached=4
  tag=mb_c_0001 fired=0 cached=4
  tag=mb_p_0002 fired=0 cached=4
  tag=mb_c_0002 fired=0 cached=4
--- honesty events (all must be 0) ---
  rollout_rematerialize_failed: 0
  optimizer_context_warning: 0
  tamper_detected: 0
--- nlink census in rollouts/train ---
  inode …428: nlink=3 ['a2__mb_c_0000__t0.json', 'a2__mb_c_0001__t0.json', 'a2__mb_c_0002__t0.json']
  inode …370: nlink=3 ['a2__mb_p_0000__t0.json', 'a2__mb_p_0001__t0.json', 'a2__mb_p_0002__t0.json']
  inode …424: nlink=3 ['a3__mb_c_0000__t0.json', 'a3__mb_c_0001__t0.json', 'a3__mb_c_0002__t0.json']
  8 distinct inodes / 24 files
  trajectories/ entries sharing a rollouts/ inode (must be 0): 0
--- archived iteration-0 rollouts still readable + self-consistent ---
  iteration-0 rollouts: 4 | task_id mismatches (must be 0): 0
```

### Response to all 7 findings

1. **BLOCKING — unvalidated `rollout_file`.** Fixed with your suggested shape plus a second layer. `_replay_cached` now starts `if not rfile or Path(rfile).name != rfile: return None`, and inside the `try` requires `src.resolve().parent == out_dir.resolve()`. `float(cached.get("reward", 0.0))` moved *inside* the `try` with `TypeError, ValueError` added, so a malformed reward degrades to an honest MISS instead of aborting the loop. Exfiltration count 0, `nlink=1` on the target, no crash — table above.

2. **BLOCKING — truncate-in-place through the shared inode.** Root-caused rather than patched at the one call site: the bug is *any* `write_text` into `rollouts/`, since a replay may have hardlinked a prior iteration's record to that name. Both writers now use `rundir._atomic_write` (tmp + `os.replace`, already the module's answer for exactly this hazard) — `gepa.py` `_eval_minibatch` **and** `harness.py:252`'s `_persist_trial`, which had the identical bug and is the writer for every non-GEPA eval. Replacing the directory entry breaks the link, so an archived record is immutable once written. Reused `_atomic_write` instead of adding an `unlink` + `write_text` pair: same effect, one existing helper, and it's crash-safe too.

3. **Non-blocking — stale `dst.exists()` serving a prior parent under the VERBATIM claim.** Fixed, and treated as blocking-adjacent as you asked: this is #199's dishonesty class. The short-circuit is **gone** — `dst.unlink(missing_ok=True)` then `os.link`, unconditionally. Safe precisely because finding 2's fix means a fresh write no longer writes through. `MISMATCH: False` above.

4. **Non-blocking — silent `pass` on double failure.** Now `run_dir.log_event("rollout_rematerialize_failed", task_id=…, tag=…, file=…, error=…)`. `run_dir` threaded into `_replay_cached` as an optional last arg. Pinned by `test_rematerialize_failure_is_logged_not_swallowed`, which fails both `os.link` and the copy and asserts the event is emitted while the `Score` stays complete.

5. **Non-blocking — untested critical paths.** All three added, plus more: traversal (3 shapes, parametrized), tag reuse (findings 2 *and* 3, separate tests), `EXDEV` fallback (`test_cross_filesystem_link_failure_falls_back_to_a_byte_identical_copy` — monkeypatches `os.link` to `OSError(EXDEV)` and asserts the copy is a distinct inode with equal content), malformed reward, and the loud-failure path. Also `test_copy_step_trajectories_never_writes_through_a_hardlink`: `shutil.copyfile` **does** write through a pre-existing hardlinked destination (verified: a target went `'ARCHIVED'` → `'NEW'`), so `_copy_step_trajectories` is safe *only* because `_copy_tag` `rmtree`s `dst` first — that invariant is now pinned rather than left implicit. **9 new tests.**

6. **Nit — heterogeneous cache entries vs the docstring.** Docstring corrected: `{"reward", "feedback"}` plus an **optional** `"rollout_file"` (omitted when unknown, e.g. a pre-#111 entry; readers treat its absence as a miss). Kept the shape conditional rather than writing `""` — the reader already treats absence as a MISS, and an always-present empty string would be a second thing to mean "no pointer".

7. **Nit — `raw={"cached": True}` write-only.** Confirmed and left as-is: it *is* consumed, by `test_cache_hit_reflective_dataset_has_output_and_trace`'s `assert raw.get("cached") is True`, which is what distinguishes "served from cache" from "re-run" in the regression suite. No production consumer, no behavior change — noting agreement, not adding one speculatively.

### Verification

**Both exploits re-run and closed** (literal output above). **Re-proved everything already validated**, so none of it regressed:

```
=== (A) per-task correspondence under SHUFFLED replay ===
fired (want 0): 0
  t3: OK  out='OUT-t3-IN-3' trace='TRACE-t3-only' fb='FB-t3'
  t0: OK  out='OUT-t0-IN-0' trace='TRACE-t0-only' fb='FB-t0'
  t2: OK  out='OUT-t2-IN-2' trace='TRACE-t2-only' fb='FB-t2'
  t1: OK  out='OUT-t1-IN-1' trace='TRACE-t1-only' fb='FB-t1'
PER-TASK CORRESPONDENCE HOLDS: True

=== (B) symmetric truncation (50KB trace) ===
  fresh  out len 11 trace len 1513
  cached out len 11 trace len 1513
  IDENTICAL output: True | IDENTICAL trace: True
  cache size (bytes, must stay small): 152

=== (C) byte-identical cached vs uncached REFLECTION.md ===
  uncached sha: a276fc9d05a9715b
  cached   sha: a276fc9d05a9715b
  BYTE-IDENTICAL: True
  raw keys fresh : ['errored', 'output', 'trace']
  raw keys cached: ['cached', 'errored', 'output', 'trace']

=== (D) old-format (pre-#111, score-only) cache degrades to MISS ===
  fired: 1  reward now: 0.0  raw: {'errored': 'False', 'output': 'OUT-t0-IN-0', 'trace': 'TRACE-t0-onl'}
  entry self-healed to a pointer: True

=== (E) dangling pointer self-heals, cost accounted ===
  fired: 1  reward complete: 0.0  n: 1
  cache entry AFTER the miss-rerun: t0__mb_p_0002__t0.json

=== (F) EXDEV fallback byte-identical ===
  fallback file exists: True | shares inode: False
  byte-identical to hardlink source: True
```

Suite + compileall:
```
193 passed in 60.39s (0:01:00)          ← 184 + 9 new, 0 failed

$ PYTHONPATH=core python -m pytest core/tests/test_gepa_cache_traces.py -q
14 passed in 0.11s                      ← 5 original + 9 new

COMPILE_OK
```

Merged tree (`main + #197 + #199 + #210`):
```
236 passed in 79.11s (0:01:19)          ← 227 + 9 new, 0 failed
```
Both exploits also verified closed on the merged tree (`EXFILTRATED (rel=0, abs=0) -> exploit CLOSED`, `archived rollout UNCHANGED (safe)`, `MISMATCH: False`).

**Conflicts: still 3, still all #199↔#197 import lines, keep-both** (`__init__.py:20`, `harness.py:27`, `gepa.py:54`). #210 merges with **zero** conflicts — my `_atomic_write` import is deliberately on its own line rather than appended to the existing `from .rundir import RunDir`, so it does not create a 4th conflict with #199's `optimizer_context` imports.

### Merge order

**`#197 → #199 → #110 → #210 → #114`**, per your recommendation. #197 and #199 are the only genuinely conflicting pair — land them first and resolve the three import lines once (keep both sides). #110 merges clean into #210. #210 last of the gepa trio. **#114 conflicts with #210 in a `cache.py` docstring only — rebase #114**, since #114 is the refactor.

### Files touched by these fixes

- `core/cap_evolve/gepa.py` — pointer confinement + malformed-reward MISS in `_replay_cached`, `_atomic_write` for the fresh-eval write, unconditional re-materialization, `rollout_rematerialize_failed` event, `run_dir` threaded through
- `core/cap_evolve/harness.py` — `_persist_trial` switched from `write_text` to `_atomic_write` (same truncate-in-place bug)
- `core/cap_evolve/cache.py` — docstring: `rollout_file` documented as optional
- `core/tests/test_gepa_cache_traces.py` — 9 new tests (5 original unchanged)

OsherElhadad:
Superseded and now landed. The #111 fix shipped as #387 (`2a4c3bf9`), rebuilt minimal against current main rather than rebasing this branch: the root cause was `EvalCache.put` storing only `{reward, feedback}` so the cache-hit branch rebuilt `raw={"cached": True}` while `_write_reflection` reads `output`/`trace`/`errored`. Fixed at the single chokepoint — `cache.put` has exactly one caller.

Dropped from this PR as over-built: the rollout-file pointer, hardlink re-materialization, the `harness.py` atomic-write ripple, the path-traversal validation that forced, and the 364-line test file. Final size was 27/11 across two source files plus one 65-line regression test, against this PR's +503.

Worth recording how often the bug actually fired: in a zero-API GEPA run, **4 of 6 minibatches were served entirely from cache** (`fired=0 cached=4`), and every one of those produced hollow reflection — so GEPA was learning from nothing about two-thirds of the time. A second symptom neither this PR nor #111 mentioned also turned out to be real and is fixed: a cached infra failure was being counted as an actionable capability defect instead of landing in the run/infra-errors ignore bucket.

PR #183, opened 2026-07-29
fix: finish the Acapo -> cap-evolve rebrand (#105)

OsherElhadad:
Closes #105

## What & why

The rename to **cap-evolve** was never finished. This removes the last `Acapo` /
`.agentcapo` / `AgentCapTune` residue from user-facing templates, live hook code,
and stale doc paths. Repo-wide `grep -ri acapo` is now **zero hits**.

## Every occurrence fixed

| File | Old | New |
|---|---|---|
| `templates/project/PROJECT.md:1` | `# Acapo project — <target name>` | `# cap-evolve project — <target name>` |
| `plugins/cap-evolve/hooks/_hooklib.py:56` (docstring) | `` `.capevolve/run_*` (or legacy `.agentcapo/run_*`) `` | `` `.capevolve/run_*` `` |
| `plugins/cap-evolve/hooks/_hooklib.py:70` | `for base_name in (".capevolve", ".agentcapo"):` | `base = parent / ".capevolve"` (loop removed) |
| `plugins/cap-evolve/hooks/_hooklib.py:86` (docstring) | `` `.capevolve/project` (or `.agentcapo/project`) `` | `` `.capevolve/project` `` |
| `plugins/cap-evolve/hooks/_hooklib.py:87` (comment) | `# .capevolve/ or .agentcapo/` | `# run_dir.parent is .capevolve/` |
| `plugins/cap-evolve/hooks/inject_router.py:51` | `for base in (".capevolve", ".agentcapo"):` | `if (parent / ".capevolve" / "project").is_dir():` |
| `docs/superpowers/plans/…-dashboard-explainability.md:13` | `.demo/.agentcapo/run_demo` | `.demo/.capevolve/run_demo` |
| `docs/superpowers/plans/…-dashboard-backend.md:19` | `.demo/.agentcapo/run_demo/` | `.demo/.capevolve/run_demo/` |
| `docs/superpowers/plans/…-dashboard-backend.md:7` | `agent-capo/dashboard/backend/` | `dashboard/backend/` |
| `docs/superpowers/plans/…-dashboard-frontend.md:7` | `agent-capo/dashboard/frontend/` | `dashboard/frontend/` |
| `docs/superpowers/plans/…-dashboard-frontend.md:77` | `--base .demo/.agentcapo` | `--base .demo/.capevolve` |
| `docs/superpowers/specs/…-dashboard-design.md:140,161,257` | `agent-capo/dashboard`, `agent-capo/` | `dashboard/` |
[...]

PR #223, opened 2026-07-30
Finish the Acapo→cap-evolve rebrand (PROJECT.md, inject_router.py, _hooklib.py)

### LINKED ISSUE #105: Finish the Acapo→cap-evolve rebrand (PROJECT.md, inject_router.py, _hooklib.py)
## Problem

The project was renamed to **cap-evolve** but the rename was never finished — "Acapo" / ".agentcapo" residue remains in user-facing templates and in live hook code.

- `templates/project/PROJECT.md:1` — literally "# Acapo project" (this is the file a new user's scaffold is based on).
- `plugins/cap-evolve/hooks/inject_router.py:51` and `plugins/cap-evolve/hooks/_hooklib.py:56,70,86-87` — still honor a legacy `.agentcapo` directory.

## Why it matters

A brand-new user runs intake, opens their scaffolded `PROJECT.md`, and sees a different product name. It reads as unfinished and erodes trust. The legacy `.agentcapo` path handling is either dead compatibility code or an undocumented fallback — either way it needs a decision.

## What to change

1. Replace "Acapo" in `templates/project/PROJECT.md:1` (and grep for any other occurrence) with "cap-evolve".
2. For `.agentcapo` handling in `inject_router.py` and `_hooklib.py`: either
   - remove it if no released version wrote that dir, **or**
   - keep it and add a one-line comment documenting *why* (backward-compat for pre-rename run dirs) so it isn't mistaken for a bug.
3. Repo-wide sweep: `grep -rin "acapo" . --include='*.md' --include='*.py' --include='*.yaml'` and resolve each hit.

## How to test

- `grep -rin "acapo" . ` (excluding this issue / CHANGELOG history) returns nothing user-facing, or only intentional documented compat with a comment.
- Scaffolding a fresh project produces a `PROJECT.md` titled with the correct product name.

- **Parallel-safe.** Small, self-contained. Good first issue.

YuuGR1337:
Fixes #105

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,8 +19,9 @@
 evaluation traces.**
 
 > **Note on naming.** This project was previously called *Acapo*. The current name is
-> **cap-evolve** everywhere — templates, scaffolds, and hooks. The legacy `.agentcapo`
-> directory is no longer honored; use `.cap-evolve` instead.
+> **cap-evolve** everywhere — templates, scaffolds, and hooks. New projects use the
+> `.capevolve` run directory (no hyphen); the legacy `.agentcapo` directory is still
+> honored as a fallback.
 
 You bring the agent and the eval you already have. cap-evolve runs the loop — evaluate →
 diagnose the failures → propose an edit → keep it only if it beats a held-out split by a
--- a/README.md
+++ b/README.md
@@ -18,6 +18,10 @@
 **cap-evolve improves an AI agent's prompts, tools, and skills by learning from failed
 evaluation traces.**
 
+> **Note on naming.** This project was previously called *Acapo*. The current name is
+> **cap-evolve** everywhere — templates, scaffolds, and hooks. The legacy `.agentcapo`
+> directory is no longer honored; use `.cap-evolve` instead.
+
 You bring the agent and the eval you already have. cap-evolve runs the loop — evaluate →
 diagnose the failures → propose an edit → keep it only if it beats a held-out split by a
 significant margin → commit — and reports one honest number. It optimizes what your agent


OsherElhadad:
Thanks for picking this up — and for engaging with #105.

I need to flag two problems with the note as written, plus some overlap you couldn't have known about.

### 1. The directory name in the note is wrong
The note says *"use `.cap-evolve` instead"*, but the on-disk directory is **`.capevolve`** (no hyphen). From `plugins/cap-evolve/hooks/_hooklib.py:70` on current `main`:
```python
for base_name in (".capevolve", ".agentcapo"):
```
Same spelling in `rundir.py`, `capevolve.yaml`, and every doc path (`.capevolve/project/`). A user following this note would create `.cap-evolve/`, which nothing looks for — and because the honesty hooks resolve the run dir by walking for those names, `find_run_dir()` would return `None` and the hooks would **silently no-op**, leaving the sealed test split editable. That's the exact failure mode that got PR #166 closed, so it's worth being precise here.

### 2. "The legacy `.agentcapo` directory is no longer honored" isn't true on `main`
The loop above still honors it, and `_hooklib.py:86-87` explicitly documents both. So as of this branch the note describes a state the code hasn't reached — the same docs-vs-code drift #105 is about.

### 3. Overlap with an existing PR
#105 is being addressed in **#183**, which is green and covers the three files the issue names plus two the issue didn't (`agent-capo/` paths in `docs/superpowers/`, and `AgentCapTune` in `core/README.md`, which is what PyPI renders). It **removes** the `.agentcapo` fallback rather than documenting its removal — justified by `git merge-base --is-ancestor 306ff98 v0.1.0`, which proves the rename predates the only release tag, so no released version ever wrote that directory. It also proves the honesty hooks still block a sealed-split edit (exit 2) afterwards.

So once #183 lands, claim (2) becomes true — but the note would still need `.capevolve` corrected.

### Where this could still help
A short naming note in `README.md` is genuinely useful and #183 doesn't add one. If you'd like to keep this PR:
- fix `.cap-evolve` → **`.capevolve`**;
- drop or reword the "no longer honored" sentence, or rebase onto #183 where it becomes accurate;
- consider pointing at `docs/HOST_SUPPORT.md` (added in #202) rather than restating specifics that can drift again.

Either way the `.capevolve` spelling needs fixing before merge — happy to re-review once it's updated. Flagging rather than closing, since the underlying idea is reasonable.

YuuGR1337:
Thanks for the review — I've updated the PR to address your feedback. Please take another look when you have a moment.

OsherElhadad (approved):
LGTM thank you!

PR #233, opened 2026-07-30
Benchmark zoo + `cap-evolve benchmark add/verify` + declarative manifest

OsherElhadad:
Closes #141

## What this does

Onboarding a benchmark was a from-scratch, per-user effort: hand-write a `CapabilityAdapter` subclass, prune a ~100-line `capevolve.yaml`, author a bespoke `optimizer/INSTRUCTIONS.md`. This adds the three pieces #141 asks for — a curated verifier-gated **library** (`benchmarks/`), a **`cap-evolve benchmark add|verify|list`** subcommand, and a **declarative manifest** — plus the docs.

## The boilerplate, measured first

I diffed the two *generic* bundled templates (`templates/adapters/jsonl_litellm` vs `huggingface_litellm`): 78 changed lines out of 127, and the changes are almost entirely the dataset-loading block.

| Repeats verbatim across every benchmark | Genuinely per-benchmark |
|---|---|
| module preamble + `sys.path` juggling | **how the target agent runs** |
| dataset → `Task` loop | a bespoke match predicate (sometimes) |
| `if rollout.error:` infra-noise branch of `score` | |
| the `exact`/`contains`/`regex` match helper | |
| `Score(task_id=…, reward=…, feedback=…, trial_rewards=[…])` | |
| the whole `capevolve.yaml` except ~6 values | |

### Measured reduction — same benchmark (`toy_calc`), hand-authored non-blank lines

| | Before (hand-written adapter) | After (manifest) |
|---|---|---|
| Python | 43 (`adapter.py`: 3 methods + `apply`) | **18** (`target.py`: one `run()`) |
| YAML | 35 (`capevolve.yaml` from the template) | **18** (`benchmark.yaml`) |
| `adapters/adapter.py` | — | 0 — **generated** |
| `capevolve.yaml` | — | 0 — **generated** |
| **Total** | **78** | **36** (−54%) |

For the documented generic LLM case (`jsonl_litellm`) the "before" is 101 (88 Python + 13 YAML) against the same 36.

## Declarative vs code (and why)

**`benchmark.yaml` declares:** dataset file + field mapping, scoring mode, metric direction, capability path, split seed/ratios/pinned ids, trial count, protected paths. An unknown key is a **hard error** — a silently-ignored key in an honesty-critical config is exactly how "I declared it" and "it applied" drift apart.

[...]

Issue #203, opened 2026-07-29
Docs and 5 algorithm SKILL.mds instruct agents to run cap-evolve subcommands that don't exist (finalize, report)


OsherElhadad:
Docs and skills instruct agents to run `cap-evolve` subcommands that **do not exist**. The CLI's real surface is six commands; the docs reference at least two more.

## Ground truth
`core/cap_evolve/cli.py:615-622`
```python
COMMANDS = {
    "version": _cmd_version,
    "splits":  _cmd_splits,
    "check":   _cmd_check,
    "run":     _cmd_run,
    "estimate": _cmd_estimate,
    "dashboard": _cmd_dashboard,
}
```
Anything else exits 2 with the usage string.

## Phantom commands found
| Referenced | Exists? | Where |
|---|---|---|
| `cap-evolve finalize` | ❌ | `skills/algorithms/{gepa,hill-climb,skillopt,evograph,agent-optimize}/SKILL.md`, `skills/orchestrate/orchestrate/SKILL.md`, `skills/phases/intake/inputs/INPUTS.md` |
| `cap-evolve report --terminal` | ❌ | `CHANGELOG.md` (found while auditing #101/PR #184 — never existed, at the `v0.1.0` tag or at HEAD) |

The `finalize` case is the interesting one, because **the repo already knows**. `docs/AGENT_ORCHESTRATION.md:98`:

> **Note — there is no `cap-evolve finalize` subcommand.** The orchestrate/host prose uses `cap-evolve finalize` as shorthand; sealing is done by the finalize *phase script* above…

So one doc carries a footnote disclaiming the shorthand, while five algorithm SKILL.mds still tell an agent to literally `Seal once with cap-evolve finalize, then report`. A footnote in a file the agent isn't reading does not stop the agent running the command.

## Why this matters more than a typo
These SKILL.md lines are **executable instructions consumed by an LLM agent**, not prose for humans. An agent following `gepa/SKILL.md:106` in agent-orchestration mode runs a command that exits 2 — at the *seal* step, i.e. after the entire optimization budget has been spent. The failure lands at the most expensive possible moment.

This is the same drift class as the one fixed in #104 (PR #198), which found `docs/REPRODUCE_tau2.md` and `agent-optimize/SKILL.md` instructing agents to pass `--mode paired` to a gate script whose parser rejected it — every such invocation exited 2. Two independent instances of "documented command doesn't parse" in one review batch means the pattern needs a guard, not another one-off fix.

[...]

PR #244, opened 2026-07-30
Ship a filled, runnable worked project reference (the scaffold #108 describes is gitignored, not blank)

### LINKED ISSUE #108: Ship a filled, runnable worked .capevolve/project reference (current scaffold is a blank template)
## Problem

The scaffolded example project that ships in the repo, `.capevolve/project/`, is just the **untouched template** — not a filled, runnable worked example. A user who opens it for reference sees a blank form:
- `optimizer_skill: mock`
- `PROJECT.md` full of `<placeholders>` (and `.capevolve/project/PROJECT.md:1` even says "Acapo project" — see the rebrand issue).

The real worked examples live in `examples/` (`toy_calc`, `tau2_airline`, `skillsbench`), but a newcomer inspecting the scaffold in place has no way to know that, and learns nothing from a blank template.

## Why it matters

The scaffold is the first concrete artifact a user sees after intake. A blank template teaches nothing and looks abandoned; a filled, runnable reference teaches the whole contract by example (adapter, capevolve.yaml, INSTRUCTIONS.md, PROJECT.md) and is the fastest way to understand the shape of a real integration.

## What to change (pick one)

- **Option A (recommended):** Replace `.capevolve/project/` with a **filled, runnable worked example** — the simplest end-to-end case (e.g. the `toy_calc` adapter + a real `capevolve.yaml` + a real `PROJECT.md` + a real `INSTRUCTIONS.md`), with a top-of-file comment saying "This is a worked reference; copy `templates/project/` to start your own."
- **Option B:** Delete the checked-in `.capevolve/project/` scaffold entirely and have intake generate it fresh, so there's no stale blank template in the repo. Point users to `examples/` for reference.

## How to test

- Option A: `cap-evolve check` passes in `.capevolve/project/` out of the box (it's a real, deterministic example), and there are no `<placeholder>` tokens left.
- Option B: the repo no longer ships a half-filled `.capevolve/project/`; intake still scaffolds correctly into a clean dir.

- Overlaps with the **rebrand** issue (both touch `PROJECT.md`). **Parallel-safe** but coordinate the PROJECT.md edit. Independent of code changes.

OsherElhadad:
Closes #108

## What #108 asked for, and the correction

The issue says the checked-in `.capevolve/project/` scaffold is an untouched blank
template, and offers Option A (fill it) or Option B (delete it).

**Neither is available: the path does not exist.** `.capevolve/` is gitignored
(`.gitignore`: `# cap-evolve run state (per-run, not source)`), and no path under it has
ever been tracked:

```
$ git log --all --oneline -- '.capevolve'
(no output)
$ git log --all --oneline --name-only | grep -c "capevolve/project"
0
```

So there is nothing to delete (Option B is already the status quo), and Option A as
literally written would mean committing an ignored run dir.

**The real gap survives the correction.** The repo ships `templates/project/` — a blank
form — and **zero** filled `PROJECT.md` or commented `capevolve.yaml` anywhere:

```
$ git ls-files | grep -c PROJECT.md
1                       # templates/project/PROJECT.md — the blank template itself
```

`examples/toy_calc/` did not even have a spec: `run.sh` copied
`templates/project/capevolve.yaml`. So the artifact intake produces first had no worked
counterpart to read anywhere in the repo. That is the gap this fills, in the example dir
the docs already point at.

## Judgement on #233's benchmark zoo — does it make this redundant?

**Partly, and I scoped around the overlap rather than duplicating it.**

The zoo genuinely supersedes the *hand-written adapter* teaching angle for benchmarks
that fit its manifest: it cuts 78 → 36 hand-authored lines, and `benchmark verify` is a
far stronger correctness story than anything a doc can assert. If the question is "how do
I add a benchmark", the zoo is now the answer and this PR should not compete with it.

But the zoo's `benchmarks/toy_calc/project/capevolve.yaml` is **`GENERATED from
benchmark.yaml`** and says so in its first line: *"edit the manifest and re-run
`cap-evolve benchmark add --refresh`, not this file."* A generated file is the opposite of
a worked reference — it deliberately explains nothing about why a value was chosen, and it
carries no `PROJECT.md` at all. The zoo also covers only the declarative path; a project
whose runner does not fit the manifest still writes a `CapabilityAdapter` by hand and still
has nothing filled to read.

So I scoped this to the gap the zoo does not close: **the hand-written path, and the
commentary and honesty framing a generated file structurally cannot carry.** No new
directory, no third copy of toy_calc — the two files land in the example that already
exists, and `examples/toy_calc/README.md` cross-links the zoo entry and says outright
*"Prefer the zoo when the manifest fits your benchmark."* If a reviewer disagrees and
wants this closed as zoo-covered, the only thing lost is the filled `PROJECT.md`, which
exists nowhere else.

## What the reference demonstrates

Two genuinely filled files, plus the honesty framing:

- **`examples/toy_calc/PROJECT.md`** — the filled `templates/project/PROJECT.md`: what is
  optimized, how the runner works, how the scorer works and what its feedback *says*, the
  data and splits, the optimizer and algorithm, the budget, and which RECOMMENDED inputs
  were skipped **and why**.
- **`examples/toy_calc/capevolve.yaml`** — the run spec with a note on **why** for every
  decision, rather than a copy of the template's option list.
- **`run.sh`** now uses that spec instead of the blank template, and runs
  `cap-evolve check` first, so the documented hard gate is *exercised*, not described.

### Both docs lead with what it does NOT prove

This is the part worth more than the parts that work, because it is what readers over-read:

1. **The significance gate never does real work here.** `gate_mode: paired`,
   `gate_k_se: 1.0` are set, but both val tasks move `0 → 1` *together*, so the spread of
   the per-task deltas is zero, `SE(Δ) = 0`, and the gate logs a `gate_warning` and takes
   its documented **STRICT fallback**. Verbatim from the run:
   `paired Δ̄=+1.0000 > 0 (SE=0 → STRICT fallback, warned; n=2)`.
2. **`k_se = 1.0` is stricter than it looks.** Improving exactly ONE of `n` val tasks
   gives `mean(Δ) = SE(Δ)` **exactly**, so the strict `>` **rejects** it — at every `n`,
   and however large that single gain is. **Improve ≥ 2 tasks at `k_se = 1.0`, or bank
   nothing.**
3. **`val = 2` is the harness floor** (`MIN_VAL_TASKS`), not a recommendation — and still
   under the low-confidence threshold.
4. **`check` passing proves less than it looks.** It never calls `run_target`, so a
   non-deterministic runner passes it; a raising `materialize()` can still yield
   `{"ok": true}`.
5. **The `0.0 → 1.0` improvement is engineered by construction.**

## Correct against the current (unmerged) contracts

- **#181 adapter contract** — described as **3 required** `@abstractmethod`s (`tasks`,
  `run_target`, `score`) plus an override of the **optional** `apply` hook. Not "4
  methods". Matches `docs/ADAPTER_CONTRACT.md` on `fix/issue-103-adapter-methods`
  line-for-line, and matches that branch's own edit to this README.
- **#197 `protected_paths`** — deliberately **OMITTED**, with a comment saying why: an
  empty list is a **HARD ERROR**, never a "use defaults" shorthand. Verified against
  #197's actual code (evidence below).
- **#195 split floor / `k_se` cap** — the reference's splits clear the floor
  (`honest_gate: true`) on `fix/issue-113-small-samples`, and it emits exactly the
  LOW CONFIDENCE warning `PROJECT.md` documents. `gate_k_se: 1.0` is far under the 26.5 cap.
- **#198 gate rule** — no "banks 1-task gains" claim anywhere; the `k_se = 1.0` caveat is
  stated in full, magnitude- and `n`-independent, in both docs and the CHANGELOG.
- **#183 rebrand** — `templates/project/PROJECT.md` gains **one line** at line 3 pointing
  at the filled example; the `Acapo` H1 at line 1 is left untouched so #183 owns it.
  `git merge origin/fix/issue-105-rebrand` → *"Automatic merge went well."*

## How this is prevented from rotting

Prose asserting a contract is exactly what this epic keeps finding stale (#203's
`cap-evolve finalize`, `maybe_cached_score`, "4 methods"). So the claims are tested, not
written down:

`core/tests/test_e2e_slice.py` gains two tests that drive the **committed** files through
the real CLI:

- `test_worked_reference_check_and_spec_are_honest` — runs the real `run_check`, then
  asserts the spec rules: **no** declared `protected_paths` (post-#197), `val ≥ 2`
  (post-#195), no pre-rebrand name, and no template placeholder token still present. The
  placeholder list is **derived from `templates/project/PROJECT.md` at test time**, so it
  cannot go stale against the template it mirrors — and it fails loudly if the template
  stops having placeholders at all.
- `test_worked_reference_runs_end_to_end` — a real zero-API `cap-evolve run` subprocess
  through to the **sealed test number**, and asserts the gate actually logs the
  `gate_warning` and `STRICT fallback` reason the docs claim. If the gate stops falling
  back, the doc claim is stale and **this test fails** rather than the doc lying.

## Expected merge order

Both are one-line-adjacent, not blocking, but land them first for a clean history:

1. **#183** (`fix/issue-105-rebrand`) — owns `templates/project/PROJECT.md`'s H1. Verified
   clean auto-merge with this branch.
2. **#197** (`feat/issue-142-protected-paths`) — makes `protected_paths: []` a hard error;
   this reference is already correct for it (omits the key). No conflict either way.

Neither is required for the tests here to pass on `main` today.

## Verification

Every number below is pasted literal output. Full commands + transcripts in the
**🔬 Evidence** comment.

**`cap-evolve check` on the reference:**
```
$ python3 -m cap_evolve.cli check /tmp/ref/.capevolve/project
{
  "ok": true,
  "stubs": [],
  "problems": [],
  "notes": [
    "tasks('val') -> 8 task(s)",
  ]
}
```

**A real zero-API `cap-evolve run` with the `mock` optimizer, to a sealed test number:**
```
$ python3 -m cap_evolve.cli run --spec .capevolve/project/capevolve.yaml \
    --project .capevolve/project --run-ts ref
{"dashboard": "http://127.0.0.1:7898"}
{
  "run_dir": ".capevolve/run_ref",
  "best_id": "cand_0001",
  "baseline_val": 0.0,
  "test_reward": 1.0,
  "test_baseline_reward": 0.0,
  "test_delta": 1.0,
  "test_pass_k": {
    "1": 1.0,
    "2": 0.0
  },
  "iterations": 3,
  "dashboard": ".capevolve/run_ref/dashboard.html",
  "dashboard_server": "http://127.0.0.1:7899"
}
```
Sealed test = **1.0** from `baseline_val 0.0`. `iterations: 3` — the win lands on
iteration 1, then `stall: 2` ends the run, exactly as `capevolve.yaml` documents.

**The gate claim, from that same run's events (proving the honesty caveat is real):**
```
splits       :: {"train": 4, "val": 2, "test": 2, "seed": 0}
gate_warning :: combined/paired SE is 0 (likely n_trials=1 or identical trials) — the
                significance gate cannot distinguish noise from signal and is falling
                back to STRICT (accept any Δ>0). ...
step         :: paired Δ̄=+1.0000 > 0 (SE=0 → STRICT fallback, warned; n=2)
step         :: paired Δ̄=+0.0000 <= 0 (SE=0 → STRICT fallback, warned; n=2)
step         :: paired Δ̄=+0.0000 <= 0 (SE=0 → STRICT fallback, warned; n=2)
finalize     :: {"test_reward": 1.0, "test_baseline_reward": 0.0, "test_delta": 1.0,
                 "best_id": "cand_0001"}
```

**Post-#197 `protected_paths` correctness** (run on `feat/issue-142-protected-paths`):
```
$ python3 -m cap_evolve.cli check /tmp/ref197/.capevolve/project
{"ok": true, "stubs": [], "problems": [], ...}

$ protect.resolve_protected(project)        # the key is OMITTED -> real defaults
['adapters/adapter.py', 'capevolve.yaml']

# control — what the reference deliberately avoids:
$ (append "protected_paths: []"); protect.protected_globs(project)
TamperError: cap-evolve: `protected_paths` in .../capevolve.yaml is an EMPTY list.
That would protect nothing (the scorer / eval harness / task data would all be optimi...
```

**Post-#195 split floor** (run on `fix/issue-113-small-samples`):
```
"test_reward": 1.0,
"honest_gate": true,
"warnings": [
  "val split (at baseline) has only 2 tasks (< 5) — acceptance decisions are LOW
   CONFIDENCE. The gate applies a Student-t small-sample correction (df=1), which
   widens the bar, but the honest fix is more val tasks.",
  ...
]
```
Clears the `MIN_VAL_TASKS = 2` floor, and the warning is exactly what `PROJECT.md`
documents as an honest limit.

**`run.sh` end to end (the CI gate path):**
```
{"ok": true, "stubs": [], "problems": [], ...}
"baseline_val": 0.0,
"test_reward": 1.0,
```
CI's assertion greps `"baseline_val": 0.0` + `"test_reward": 1.0` — both still present.

**Full test suite:**
```
$ PYTHONPATH=core python -m pytest core/tests -q      # THIS BRANCH
181 passed in 68.87s

$ PYTHONPATH=core python -m pytest core/tests -q      # BASELINE (origin/main, clean worktree)
179 passed in 57.19s
```
**179 baseline → 181: exactly +2** (my two new tests), **0 failures on either side**.
Earlier in the session the same command hit
`test_dashboard_launch.py::test_maybe_launch_spawns_when_available` — the known port-7878
flake tracked in **#200** — on the branch *and* on the unmodified baseline
(`1 failed, 180 passed` vs `1 failed, 178 passed`: the same +2 with the same pre-existing
failure). It is timing-dependent and unrelated to this change.

**`compileall`:**
```
$ python -m compileall -q core examples
compileall: clean
```

**Links:** all 15 relative links in the two new/edited docs resolve (checked
programmatically; full output in the Evidence comment).

## Files touched

| File | Change |
|---|---|
| `examples/toy_calc/PROJECT.md` | **new** — the filled `PROJECT.md`, incl. *Honest limits* |
| `examples/toy_calc/capevolve.yaml` | **new** — the filled spec, a note per decision |
| `examples/toy_calc/run.sh` | use the filled spec; run `cap-evolve check` first |
| `examples/toy_calc/README.md` | reframed as the worked reference; zoo cross-link; #181 wording |
| `core/tests/test_e2e_slice.py` | +2 anti-rot tests over the committed files |
| `templates/project/PROJECT.md` | +1 line pointing at the filled example (H1 untouched) |
| `CHANGELOG.md` | Unreleased / Added entry |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,30 @@ All notable changes to cap-evolve are documented here. The format follows
   longer looks like a capability regression.
 
 ### Added
+- **A filled, runnable worked reference project (`examples/toy_calc/`).** The repo shipped
+  a blank scaffold (`templates/project/`) and zero filled `PROJECT.md` / commented
+  `capevolve.yaml` anywhere — so the first concrete artifact after intake taught nothing.
+  `examples/toy_calc/` now carries both: a filled `PROJECT.md` (what is optimized, how it
+  runs, how it scores, the splits, the budget, the resolved/skipped inputs) and a filled
+  `capevolve.yaml` with a note on **why** for every decision. `run.sh` uses that spec
+  instead of the blank template and runs `cap-evolve check` first, so the documented hard
+  gate is actually exercised.
+
+  Each doc leads with what the example does **not** prove, because that is what readers
+  over-read: `gate_mode: paired` / `gate_k_se: 1.0` are set but never tested here — both
+  val tasks move together, so `SE(Δ)=0` and the gate takes its documented **STRICT
+  fallback** (`paired Δ̄=+1.0000 > 0 (SE=0 → STRICT fallback, warned; n=2)`); at
+  `k_se=1.0` improving exactly one of `n` tasks gives `mean(Δ)=SE(Δ)` *exactly* and is
+  **rejected** at any `n` and any magnitude; `val=2` is the harness floor, not a
+  recommendation; and `check` passing does not run `run_target`, so it proves less than it
+  looks. `protected_paths` is deliberately **omitted** (an empty list is a hard error,
+  never a "use defaults" shorthand).
+
+  Two tests in `core/tests/test_e2e_slice.py` drive the committed files through the real
+  CLI — `cap-evolve check` → a full zero-API `run` → the sealed test — and assert the spec
+  rules (no declared `protected_paths`, val ≥ 2, no leftover template placeholders) plus
+  the gate's `gate_warning`. A contract change now breaks a test instead of silently
+  rotting the prose.
 - **SWE-bench oracle mode + calibrated smoke selection.** The SWE-bench adapter gains
   `SWEBENCH_ORACLE=1`, which attaches the "Oracle" retrieval context (the file[s] the
   gold patch touches, from `princeton-nlp/SWE-bench_Lite_oracle`'s `text` field) to the
--- a/core/tests/test_e2e_slice.py
+++ b/core/tests/test_e2e_slice.py
@@ -8,6 +8,7 @@ architecture works without any model.
 """
 
 import os
+import re
 import shutil
 import sys
 from pathlib import Path
@@ -127,3 +128,110 @@ def test_baseline_better_than_nothing_is_gated(tmp_path):
         gate_kwargs={"mode": "significant", "k_se": 1.0},
     )
     assert step["accepted"] is False
+
+
+# --------------------------------------------------------------------------
+# The worked reference project (issue #108). These drive the COMMITTED example
+# files — capevolve.yaml, PROJECT.md, adapter.py — through the real CLI, so a
+# contract change breaks a test instead of silently rotting the docs.
+# --------------------------------------------------------------------------
+
+def _build_reference_project(dest: Path) -> Path:
+    """Assemble the project dir exactly as examples/toy_calc/run.sh does."""
+    proj = dest / ".capevolve" / "project"
+    (proj / "adapters").mkdir(parents=True)
+    shutil.copy(EXAMPLE / "adapter.py", proj / "adapters" / "adapter.py")
+    shutil.copy(EXAMPLE / "capevolve.yaml", proj / "capevolve.yaml")
+    shutil.copy(EXAMPLE / "PROJECT.md", proj / "PROJECT.md")
+    shutil.copytree(EXAMPLE / "capability", dest / "seed_capability")
+    return proj
+
+
+def test_worked_reference_check_and_spec_are_honest(tmp_path):
+    """`cap-evolve check` accepts the reference, and its spec obeys the current rules.
+
+    Guards the two rules a scaffolded example most easily gets wrong:
+      * `protected_paths: []` is a HARD ERROR (#142/#197), so the reference must OMIT
+        the key rather than declare an empty list;
+      * the val split must clear the harness floor (#113) — these ratios give val=2,
+        which is exactly MIN_VAL_TASKS.
+    """
+    from cap_evolve.check import run_check
+    from cap_evolve.specfile import read_yaml
+
+    proj = _build_reference_project(tmp_path)
+
+    rep = run_check(proj)
+    assert rep.ok, f"cap-evolve check rejected the worked reference: {rep.to_dict()}"
+    assert rep.stubs == [] and rep.problems == []
+
+    raw = (proj / "capevolve.yaml").read_text(encoding="utf-8")
+    spec = read_yaml(raw)
+    # An empty protected_paths is a hard error, so the reference must not declare one.
+    assert "protected_paths" not in spec, "the reference must OMIT protected_paths (empty = hard error)"
+    # ...and not as a live (uncommented) line either.
+    assert not [ln for ln in raw.splitlines()
+                if ln.strip().startswith("protected_paths")], "protected_paths is declared"
+
+    # Splits must clear the val floor on the real dataset.
+    import importlib.util
+    aspec = importlib.util.spec_from_file_location("ref_adapter_spec", EXAMPLE / "adapter.py")
+    toy = importlib.util.module_from_spec(aspec)
+    aspec.loader.exec_module(toy)
+    n = len(toy.Adapter().tasks("all"))
+    n_val = int(n * float(spec["split_val"]))
+    assert n_val >= 2, f"val split is {n_val} task(s); the harness floor is 2"
+
+    # No unfilled template placeholders left in either doc (issue #108's ask).
+    template = (REPO / "templates" / "project" / "PROJECT.md").read_text(encoding="utf-8")
+    tokens = set(re.findall(r"<[^<>\n]{2,40}>", template))
+    assert tokens, "template has no placeholders to compare against — did its shape change?"
+    for name in ("capevolve.yaml", "PROJECT.md"):
+        body = (proj / name).read_text(encoding="utf-8")
+        assert "Acapo" not in body, f"{name} carries the pre-rebrand name"
+        left = sorted(t for t in tokens if t in body)
+        assert not left, f"{name} still carries unfilled template placeholders: {left}"
+    project_md = (proj / "PROJECT.md").read_text(encoding="utf-8")
+    assert "Honest limits" in project_md, "the reference must keep its honest-limits section"
+
+
+def test_worked_reference_runs_end_to_end(tmp_path):
+    """A real zero-API `cap-evolve run` on the reference, through to the sealed test."""
+    import json
+    import subprocess
+
+    proj = _build_reference_project(tmp_path)
+    env = dict(os.environ)
+    env["PYTHONPATH"] = str(CORE)
+    env["CAPEVOLVE_SKILLS_DIR"] = str(REPO / "skills")
+
+    proc = subprocess.run(
+        [sys.executable, "-m", "cap_evolve.cli", "run",
+         "--spec", str(proj / "capevolve.yaml"), "--project", str(proj), "--run-ts", "ref"],
+        cwd=tmp_path, env=env, capture_output=True, text=True,
+    )
+    assert proc.returncode == 0, f"run failed:\n{proc.stdout}\n{proc.stderr}"
+    # The run prints one or more JSON objects; the summary is the last top-level one.
+    dec, objs, i = json.JSONDecoder(), [], 0
+    while (i := proc.stdout.find("{", i)) != -1:
+        try:
+            obj, end = dec.raw_decode(proc.stdout, i)
+        except json.JSONDecodeError:
+            i += 1
+            continue
+        objs.append(obj)
+        i = end
+    summary = objs[-1]
+    assert summary["baseline_val"] == 0.0
+    assert summary["test_reward"] == 1.0, summary
+    assert summary["test_delta"] == 1.0
+
+    # The gate's honest caveat: a deterministic scorer gives SE(delta)=0, so the paired
+    # gate takes its documented STRICT fallback. PROJECT.md claims exactly this; if the
+    # gate stops warning, the claim is stale and this fails.
+    events = [json.loads(l) for l in
+              (tmp_path / ".capevolve" / "run_ref" / "events.jsonl").read_text().splitlines()]
+    assert any(e.get("kind") == "gate_warning" for e in events), \
+        "PROJECT.md documents an SE=0 STRICT fallback; no gate_warning was logged"
+    accepted = [e for e in events if e.get("kind") == "step" and e.get("accept")]
+    assert accepted and "STRICT fallback" in accepted[0]["reason"], accepted
--- a/core/tests/test_e2e_slice.py
+++ b/core/tests/test_e2e_slice.py
@@ -192,7 +192,16 @@ def test_worked_reference_check_and_spec_are_honest(tmp_path):
         left = sorted(t for t in tokens if t in body)
         assert not left, f"{name} still carries unfilled template placeholders: {left}"
     project_md = (proj / "PROJECT.md").read_text(encoding="utf-8")
-    assert "Honest limits" in project_md, "the reference must keep its honest-limits section"
+    # The reference's whole value is the caveats, so assert the SECTION and its CONTENT.
+    # A bare `"Honest limits" in project_md` also matches the cross-reference earlier in
+    # the file, so it survives both a heading rename and deleting the section outright.
+    assert "\n## Honest limits" in project_md, "the reference lost its `## Honest limits` heading"
+    limits = project_md.split("\n## Honest limits", 1)[1].split("\n## ", 1)[0]
+    for claim in ("STRICT fallback", "bank nothing", "MIN_VAL_TASKS", "run_target",
+                  "engineered"):
+        assert claim in limits, f"Honest limits lost its `{claim}` caveat"
+    caveats = re.findall(r"^\d+\. \*\*", limits, flags=re.M)
+    assert len(caveats) == 5, f"Honest limits should list 5 caveats, found {len(caveats)}"
 
 
 def test_worked_reference_runs_end_to_end(tmp_path):
--- a/examples/toy_calc/PROJECT.md
+++ b/examples/toy_calc/PROJECT.md
@@ -1,0 +1,100 @@
+# cap-evolve project — toy_calc arithmetic stand-in
+
+A **filled** version of `templates/project/PROJECT.md`. The template ships as a blank
+form; this is what one looks like after `intake` has actually interviewed someone, so
+you can see the shape of a real answer instead of a `<placeholder>`. Copy the
+**template** to start your own.
+
+Records the decisions behind this run so anyone (human or agent) can reproduce it.
+
+## What we're optimizing
+- Capability: `system-prompt`
+- Artifact: `capability/prompt.txt` (copied to `<workdir>/seed_capability/` by `run.sh`)
+- Allowed edits: `[edit]` — rewrite the prompt text. No adding or deleting files; the
+  stand-in only ever reads `prompt.txt`.
+
+## How we run the target (the RUNNER)
+- Agent under test: a **deterministic Python stand-in**, not a model. It lives in
+  `adapter.py::Adapter.run_target` and computes the arithmetic correctly **only** when
+  the candidate prompt contains the marker `[CALC]`; otherwise it emits
+  `"I think <expr> is roughly some number."` and fails the exact-match scorer.
+- How `run_target` invokes it: reads `ctx/prompt.txt` (where `ctx` is the candidate dir
+  yielded by the default `live()`), branches on `"[CALC]" in prompt`, and returns a
+  `Rollout`. `seed` is accepted per the contract and unused — this runner is exact.
+- Why a stand-in: it makes the whole pipeline provable at **zero API cost** and with
+  byte-identical reruns, which is what lets this example be the CI gate. It buys
+  determinism, not realism — see *Honest limits*.
+
+## How we score
+- Metric: exact string match against `task.target`, reward `1.0` or `0.0`.
+- Feedback signal: on failure, `score()` reports what was expected vs produced **plus
+  the hypothesis** that "the prompt likely lacks an explicit instruction to compute and
+  output only the number". That sentence is the learning signal the optimizer acts on —
+  it names the *cause*, not just the miss. It does not leak the gold answer beyond the
+  expected value the optimizer would see in the trace anyway.
+- Secondary metrics: none (`metrics_display: []`). The gate reads only the primary
+  scalar reward regardless.
+
+## Data
+- Source: `adapter` — the adapter's own `tasks()` reads `tasks.jsonl` (8 arithmetic
+  tasks) from `CAPEVOLVE_TOY_DATA`. `dataset_source: adapter` means the harness does not
+  load a file itself; `tasks()` is the single source.
+- Split: train 4 / val 2 / test 2, seed `0`, frozen once into `<run>/splits.json`. Test
+  is sealed and scored exactly once at finalize — a second `finalize` raises
+  `TestSealError`.
+- `tasks(split)` deliberately returns **all** tasks and lets the harness filter by the
+  frozen split. That is the supported shape; the split is the harness's job, not the
+  adapter's.
+
+## Optimizer + algorithm
+- Optimizer (proposer): `mock` — applies `mock_script.json` verbatim
+  (`ensure_contains` the `[CALC]` line). Deterministic, zero API calls.
+- Algorithm: `hill-climb --focus all`
+- Budget: `max_iterations: 5`, `stall: 2`. The winning edit lands on iteration 1; the
+  next two are no-ops the gate rejects, so `stall` ends the run after **3** iterations.
+
+## Inputs status
+- NEEDED inputs resolved: capability + artifact path, runner (the stand-in), scorer,
+  dataset, splits, optimizer, algorithm, budget.
+- RECOMMENDED inputs skipped: `target_model` (there is no consuming LLM — the "agent"
+  is Python, so there is no reader tier to tune edits for); `runner_repo_path` and
+  `capability_sources` (the runner is 20 lines inside the adapter, nothing to surface);
+  `metrics_display` (one metric).
+
+## Honest limits — what this example does NOT prove
+
+Worth more than the parts that work, because these are the claims people over-read.
+
+1. **The significance gate never does real work here.** `gate_mode: paired`,
+   `gate_k_se: 1.0` are set, but both val tasks move `0 → 1` *together*, so the spread
+   of the per-task deltas is zero, `SE(Δ) = 0`, and the gate logs a `gate_warning` and
+   applies its documented **STRICT fallback** (accept any `Δ > 0`). The run transcript
+   says exactly that: `paired Δ̄=+1.0000 > 0 (SE=0 → STRICT fallback, warned; n=2)`.
+   A deterministic scorer has no noise to reject.
+2. **`k_se = 1.0` is stricter than it looks.** Improving exactly ONE of `n` val tasks
+   gives `mean(Δ) = SE(Δ)` *exactly*, so the strict `>` **rejects** it — at every `n`,
+   and no matter how large that single gain is. **Improve ≥ 2 tasks at `k_se = 1.0`, or
+   bank nothing.** Set `gate_k_se: 0.2` (as the other examples do) to bank a 1-of-`n`
+   gain.
+3. **val = 2 is the floor, not a recommendation.** It equals the harness's
+   `MIN_VAL_TASKS = 2`; a smaller val split is *refused*, and 2 is still under the
+   low-confidence threshold. A real benchmark wants tens of val tasks.
+4. **`cap-evolve check` passing is not proof the adapter is correct.** It proves the
+   three abstract methods are implemented and non-stubbed, `tasks()` is non-empty and
+   stable, `score()` is deterministic, and `materialize()` is callable. It does not run
+   `run_target`, so a non-deterministic runner passes `check`. Notably a `materialize()`
+   that *raises* can still yield `{"ok": true}`.
+5. **The improvement is engineered.** The stand-in is written so `[CALC]` is the one
+   thing that matters and `mock` is scripted to add it. `0.0 → 1.0` proves the loop
+   plumbing, gate wiring, and seal — not that optimization works on a real benchmark.
+   For that, see `examples/tau2_airline/` and `examples/skillsbench/`.
+
+## Where the pieces live
+| File | Role |
+|---|---|
+| [`adapter.py`](adapter.py) | the `CapabilityAdapter`: 3 required methods + one hook override |
+| [`capevolve.yaml`](capevolve.yaml) | the filled run spec, with a note per choice |
+| [`capability/prompt.txt`](capability/prompt.txt) | the seed artifact (no `[CALC]`) |
+| [`tasks.jsonl`](tasks.jsonl) | 8 arithmetic tasks |
+| [`mock_script.json`](mock_script.json) | the deterministic edit `mock` applies |
+| [`run.sh`](run.sh) | the whole thing end to end, zero API |
--- a/examples/toy_calc/PROJECT.md
+++ b/examples/toy_calc/PROJECT.md
@@ -69,7 +69,8 @@ Worth more than the parts that work, because these are the claims people over-re
    `gate_k_se: 1.0` are set, but both val tasks move `0 → 1` *together*, so the spread
    of the per-task deltas is zero, `SE(Δ) = 0`, and the gate logs a `gate_warning` and
    applies its documented **STRICT fallback** (accept any `Δ > 0`). The run transcript
-   says exactly that: `paired Δ̄=+1.0000 > 0 (SE=0 → STRICT fallback, warned; n=2)`.
+   says exactly that: `paired Δ̄=+1.0000 > 0 (SE=0 → STRICT fallback, warned; n=2`…
+   (the suffix varies by branch; the prefix is the stable part).
    A deterministic scorer has no noise to reject.
 2. **`k_se = 1.0` is stricter than it looks.** Improving exactly ONE of `n` val tasks
    gives `mean(Δ) = SE(Δ)` *exactly*, so the strict `>` **rejects** it — at every `n`,
@@ -77,12 +78,15 @@ Worth more than the parts that work, because these are the claims people over-re
    bank nothing.** Set `gate_k_se: 0.2` (as the other examples do) to bank a 1-of-`n`
    gain.
 3. **val = 2 is the floor, not a recommendation.** It equals the harness's
-   `MIN_VAL_TASKS = 2`; a smaller val split is *refused*, and 2 is still under the
-   low-confidence threshold. A real benchmark wants tens of val tasks.
+   `MIN_VAL_TASKS = 2`; a smaller val split is *refused*, and 2 is still under
+   `LOW_CONFIDENCE_VAL_TASKS = 5`, the threshold that flags acceptance decisions as LOW
+   CONFIDENCE. A real benchmark wants tens of val tasks.
 4. **`cap-evolve check` passing is not proof the adapter is correct.** It proves the
    three abstract methods are implemented and non-stubbed, `tasks()` is non-empty and
    stable, `score()` is deterministic, and `materialize()` is callable. It does not run
-   `run_target`, so a non-deterministic runner passes `check`. Notably a `materialize()`
+   `run_target` (except an opt-in degenerate-trials probe behind
+   `CAPEVOLVE_CHECK_TRIAL_PROBE=1` *and* `CAPEVOLVE_N_TRIALS>1`, which only ever appends
+   a WARNING note), so a non-deterministic runner passes `check`. Notably a `materialize()`
    that *raises* can still yield `{"ok": true}`.
 5. **The improvement is engineered.** The stand-in is written so `[CALC]` is the one
    thing that matters and `mock` is scripted to add it. `0.0 → 1.0` proves the loop
--- a/examples/toy_calc/README.md
+++ b/examples/toy_calc/README.md
@@ -68,3 +68,10 @@ section spells out all five caveats. For real benchmarks see
   adapter implements: 3 required `@abstractmethod`s, defaulted hooks, `hasattr`-probed
   optional fast paths.
 - [`templates/project/`](../../templates/project) — the blank scaffold to copy.
+- `benchmarks/toy_calc/` — the **declarative** form of this same benchmark: a
+  `benchmark.yaml` manifest plus a one-function `project/target.py`, with the spec
+  *generated* by `cap-evolve benchmark add`. Take that path when the manifest fits; this
+  directory is the reference for the hand-written `CapabilityAdapter` when it doesn't.
+  The two are deliberately paired — the zoo entry links back here for the adapter form,
+  and this is the "before" side of its boilerplate measurement. *(Lands with #233; the
+  path is intentionally unlinked until then.)*
--- a/examples/toy_calc/README.md
+++ b/examples/toy_calc/README.md
@@ -1,18 +1,40 @@
-# Example: toy_calc (zero-API, deterministic)
+# Example: toy_calc — the worked reference project (zero-API, deterministic)
 
-The smallest end-to-end proof — no model calls, fully deterministic, runs in
-seconds. A deterministic stand-in "agent" answers arithmetic tasks; it only
-succeeds when the system prompt contains the marker `[CALC]`, so the optimization
-(the `mock` optimizer adds `[CALC]`) provably raises the score. Used as the CI gate.
+The smallest end-to-end proof, and the **filled reference** for what a real
+`.capevolve/project/` looks like. No model calls, fully deterministic, runs in seconds.
+
+A deterministic stand-in "agent" answers arithmetic tasks; it only succeeds when the
+system prompt contains the marker `[CALC]`, so the optimization (the `mock` optimizer
+adds `[CALC]`) provably raises the score. Used as the CI gate.
+
+> **To start your own project, copy [`templates/project/`](../../templates/project) —
+> it carries the full option list with every knob documented.** Read *this* directory to
+> see a finished one: every choice made with a note on why, plus an explicit list of what
+> the example does **not** prove.
 
 ## Files
-- `adapter.py` — a `CapabilityAdapter`: the 3 required methods (deterministic agent +
-  exact-match scorer) plus an `apply` hook override.
+- [`PROJECT.md`](PROJECT.md) — the **filled** `templates/project/PROJECT.md`: what is
+  optimized, how it runs, how it scores, the splits, the budget, and an *Honest limits*
+  section naming what this example does not demonstrate.
+- [`capevolve.yaml`](capevolve.yaml) — the **filled** run spec, one comment per decision.
+- [`adapter.py`](adapter.py) — a `CapabilityAdapter`: the **3 required** methods
+  (`tasks`, `run_target`, `score`) plus an override of the optional `apply` hook.
 - `capability/prompt.txt` — the seed system prompt (no `[CALC]`).
 - `mock_script.json` — the deterministic edit the `mock` optimizer applies.
 - `tasks.jsonl` — 8 arithmetic tasks.
 
 ## Run it
+```bash
+bash examples/toy_calc/run.sh
+# -> {"ok": true}                             (cap-evolve check, the hard gate)
+# -> baseline_val 0.0  ->  test_reward 1.0     (gate-accepted, test sealed) + dashboard.html
+```
+`run.sh` builds the project dir, runs `cap-evolve check`, then the full pipeline.
+`core/tests/test_e2e_slice.py::test_worked_reference_runs_end_to_end` drives these same
+files through `check` → run → sealed test, so a contract change breaks a test rather than
+rotting this README.
+
+Or by hand, to see the pieces:
 ```bash
 REPO=$PWD                       # cap-evolve repo root
 export CAPEVOLVE_CORE=$REPO/core PYTHONPATH=$REPO/core CAPEVOLVE_SKILLS_DIR=$REPO/skills
@@ -20,11 +42,29 @@ export CAPEVOLVE_TOY_DATA=$REPO/examples/toy_calc
 export CAPEVOLVE_MOCK_SCRIPT=$REPO/examples/toy_calc/mock_script.json
 
 D=/tmp/toy; mkdir -p $D/.capevolve/project/adapters
-cp $REPO/examples/toy_calc/adapter.py $D/.capevolve/project/adapters/
-cp -R $REPO/examples/toy_calc/capability $D/seed_capability
-cp $REPO/templates/project/capevolve.yaml $D/.capevolve/project/capevolve.yaml   # defaults: system-prompt / mock / hill-climb (focus all)
+cp $REPO/examples/toy_calc/adapter.py     $D/.capevolve/project/adapters/
+cp -R $REPO/examples/toy_calc/capability  $D/seed_capability
+cp $REPO/examples/toy_calc/capevolve.yaml $D/.capevolve/project/capevolve.yaml
+cp $REPO/examples/toy_calc/PROJECT.md     $D/.capevolve/project/PROJECT.md
 
-python3 -m cap_evolve.cli run --spec $D/.capevolve/project/capevolve.yaml --project $D/.capevolve/project --run-ts demo
-# -> baseline_val 0.0  ->  test_reward 1.0   (gate-accepted, test sealed) + dashboard.html
+python3 -m cap_evolve.cli check $D/.capevolve/project      # {"ok": true}
+python3 -m cap_evolve.cli run --spec $D/.capevolve/project/capevolve.yaml \
+                              --project $D/.capevolve/project --run-ts demo
 ```
-This is exactly what `core/tests/test_e2e_slice.py` asserts.
+
+## What it proves, and what it doesn't
+
+Proves: the adapter contract, the frozen splits, the gate wiring, the optimizer handoff,
+the sealed test, and the run-dir artifacts — reproducibly and for free.
+
+Does **not** prove: that the significance gate rejects noise (a deterministic scorer has
+none, so the gate takes its documented `SE=0 → STRICT` fallback here), nor that
+optimization works on a real benchmark. [`PROJECT.md`](PROJECT.md)'s *Honest limits*
+section spells out all five caveats. For real benchmarks see
+[`../tau2_airline`](../tau2_airline) and [`../skillsbench`](../skillsbench).
+
+## Related
+- [`docs/ADAPTER_CONTRACT.md`](../../docs/ADAPTER_CONTRACT.md) — the contract this
+  adapter implements: 3 required `@abstractmethod`s, defaulted hooks, `hasattr`-probed
+  optional fast paths.
+- [`templates/project/`](../../templates/project) — the blank scaffold to copy.
--- a/examples/toy_calc/capevolve.yaml
+++ b/examples/toy_calc/capevolve.yaml
@@ -1,0 +1,79 @@
+# toy_calc — a FILLED cap-evolve run spec (the worked reference).
+#
+# This is the template at `templates/project/capevolve.yaml` with every decision
+# actually made, and a note on WHY. To start your own project, copy the TEMPLATE
+# (it carries the full option list); read this file to see what a finished one
+# looks like and what each choice costs you.
+#
+# `run.sh` copies this file to `<workdir>/.capevolve/project/capevolve.yaml` — the
+# project dir the harness reads. Only keys that differ from the template's default,
+# or that are load-bearing here, are kept; every omitted key takes its default.
+
+# --- what to optimize -------------------------------------------------------
+capabilities: [system-prompt]     # the artifact is one prompt file, so one capability
+capability_path: seed_capability  # resolved project-relative; run.sh puts the seed copy here
+actions: [edit]                   # edit the prompt text; no add/remove of files
+
+# --- who proposes edits -----------------------------------------------------
+optimizer_skill: mock             # `mock` = deterministic, ZERO API calls. It applies
+                                  # CAPEVOLVE_MOCK_SCRIPT verbatim, which is why this
+                                  # example is a CI gate and not a flaky model call.
+optimizer_model: ""               # unused by `mock`
+target_model: ""                  # no consuming LLM: the "agent" is a Python stand-in,
+                                  # so there is no reader tier to tune edits for
+
+# --- how to iterate ---------------------------------------------------------
+algorithm_skill: hill-climb       # parent is always the current best — the simplest loop
+algorithm_focus: all              # reflect over the whole train set each iteration
+orchestration_mode: deterministic # cap-evolve sequences the loop; honesty is code-enforced
+
+# --- data + splits ----------------------------------------------------------
+dataset_source: adapter           # tasks come from Adapter.tasks(); there is no file
+                                  # loader in this path — the adapter reads tasks.jsonl
+                                  # itself (see adapter.py), and the harness filters by
+                                  # the frozen split.
+split_seed: 0                     # frozen ONCE into <run>/splits.json; test is sealed
+split_train: 0.5                  # 8 tasks -> train 4 / val 2 / test 2
+split_val: 0.25
+split_test: 0.25
+# HONEST LIMIT: val = 2 tasks. That is exactly the floor the harness enforces
+# (`MIN_VAL_TASKS = 2`); a 3-task dataset with these ratios would be REFUSED, not
+# silently run. 2 is still below the low-confidence threshold, so a real benchmark
+# wants tens of val tasks. This example gets away with it only because the scorer is
+# exact and the stand-in is deterministic.
+
+# --- scoring ----------------------------------------------------------------
+num_trials: 1                     # the stand-in is deterministic: extra trials are
+                                  # byte-identical, so they buy no variance information
+
+# --- protected paths (tamper guard) -----------------------------------------
+# DELIBERATELY OMITTED, which means "use the defaults derived from this layout"
+# (adapters/, capevolve.yaml, the dataset, and *gold* data files). Do NOT write
+# `protected_paths: []` to mean that — an empty list is a HARD ERROR, because
+# "protects nothing" must never happen by accident.
+
+# --- acceptance gate --------------------------------------------------------
+gate_mode: paired                 # the default in every real run: mean(per-task Δ) vs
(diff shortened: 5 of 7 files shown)


OsherElhadad:
## 🔬 Evidence

Every command, verbatim, with full output. Env: macOS, Python 3.14, `/tmp/ce-venv`.

### 0. The correction — `.capevolve/project/` is gitignored and never existed
```
$ git log --all --oneline -- ".capevolve"
(no output — never tracked)

$ git ls-files | grep PROJECT.md          # BEFORE this PR (on origin/main)
templates/project/PROJECT.md             # ...the blank template, and nothing else

$ git ls-files | grep PROJECT.md          # AFTER
examples/toy_calc/PROJECT.md             # <- the filled reference this PR adds
templates/project/PROJECT.md

$ grep -n -A1 "cap-evolve run state" .gitignore
16:# cap-evolve run state (per-run, not source)
17-.capevolve/
```

### 1. `cap-evolve check` on the reference
```
$ rm -rf /tmp/ref && mkdir -p /tmp/ref/.capevolve/project/adapters
$ cp examples/toy_calc/adapter.py /tmp/ref/.capevolve/project/adapters/
$ cp examples/toy_calc/capevolve.yaml examples/toy_calc/PROJECT.md /tmp/ref/.capevolve/project/
$ cp -R examples/toy_calc/capability /tmp/ref/seed_capability
$ export CAPEVOLVE_TOY_DATA=$PWD/examples/toy_calc PYTHONPATH=$PWD/core
$ python -m cap_evolve.cli check /tmp/ref/.capevolve/project
{
  "ok": true,
  "stubs": [],
  "problems": [],
  "notes": [
    "tasks('val') -> 8 task(s)",
  ]
}
```

### 2. A real zero-API `cap-evolve run` to the sealed test number
```
$ cd /tmp/refE && python -m cap_evolve.cli run --spec .capevolve/project/capevolve.yaml --project .capevolve/project --run-ts ref
{
  "run_dir": ".capevolve/run_ref",
  "best_id": "cand_0001",
  "baseline_val": 0.0,
  "test_reward": 1.0,
  "test_baseline_reward": 0.0,
  "test_delta": 1.0,
  "test_pass_k": {
    "1": 1.0,
    "2": 0.0
  },
  "iterations": 3,
  "dashboard": ".capevolve/run_ref/dashboard.html",
  "dashboard_server": "http://127.0.0.1:7878"
}

```

### 3. Frozen splits (val = 2 = `MIN_VAL_TASKS`) and every gate decision
```
$ cat .capevolve/run_ref/splits.json
{
  "train": [
    "a5",
    "a2",
    "a6",
    "a3"
  ],
  "val": [
    "a1",
    "a4"
  ],
  "test": [
    "a8",
    "a7"
  ],
  "seed": 0,
  "test_used": true
}
$ # all splits/gate/step/finalize events from that run
splits :: {"train": 4, "val": 2, "test": 2, "seed": 0}
gate_warning :: combined/paired SE is 0 (likely n_trials=1 or identical trials) — the significance gate cannot distinguish noise from signal and is falling back to STRICT (accept any Δ>0). Increase n_trials and ensure the runner forwards the per-trial seed to get real variance.
step :: paired Δ̄=+1.0000 > 0 (SE=0 → STRICT fallback, warned; n=2)
gate_warning :: combined/paired SE is 0 (likely n_trials=1 or identical trials) — the significance gate cannot distinguish noise from signal and is falling back to STRICT (accept any Δ>0). Increase n_trials and ensure the runner forwards the per-trial seed to get real variance.
step :: paired Δ̄=+0.0000 <= 0 (SE=0 → STRICT fallback, warned; n=2)
gate_warning :: combined/paired SE is 0 (likely n_trials=1 or identical trials) — the significance gate cannot distinguish noise from signal and is falling back to STRICT (accept any Δ>0). Increase n_trials and ensure the runner forwards the per-trial seed to get real variance.
step :: paired Δ̄=+0.0000 <= 0 (SE=0 → STRICT fallback, warned; n=2)
finalize :: {"test_reward": 1.0, "test_baseline_reward": 0.0, "test_delta": 1.0, "best_id": "cand_0001"}
```

### 4. Post-#197 `protected_paths` — correct, and the control that proves the rule
Run on `origin/feat/issue-142-protected-paths` (the branch that introduces the rule).
```
$ # reference spec under #197
$ python -m cap_evolve.cli check /tmp/ref197/.capevolve/project
{
  "ok": true,
  "stubs": [],
  "problems": [],
  "notes": [
    "tasks('val') -> 8 task(s)",
  ]
}

$ # the key is OMITTED, so resolve_protected() yields the LAYOUT DEFAULTS:
['adapters/adapter.py', 'capevolve.yaml']

$ # CONTROL — what the reference deliberately avoids (append protected_paths: [])
TamperError: cap-evolve: `protected_paths` in /var/folders/zh/srgnbq_97qvb6002zsr40tgc0000gn/T/tmppank9ic5/project/capevolve.yaml is an EMPTY list. That would protect nothing (the scorer / eval harness / task data would all be optimizer-writable). Remove the key to use the defaults, or declare the paths.
```

### 5. Post-#195 split floor + `k_se` cap
Run on `origin/fix/issue-113-small-samples`.
```
$ python -m cap_evolve.cli run --spec ... --run-ts r195   (tail)
{
  "run_dir": ".capevolve/run_r195b",
  "best_id": "cand_0001",
  "baseline_val": 0.0,
  "test_reward": 1.0,
  "test_baseline_reward": 0.0,
  "test_delta": 1.0,
  "test_pass_k": {
    "1": 1.0,
    "2": 0.0
  },
  "iterations": 3,
  "honest_gate": true,
  "warnings": [
    "val split (at baseline) has only 2 tasks (< 5) \u2014 acceptance decisions are LOW CONFIDENCE. The gate applies a Student-t small-sample correction (df=1), which widens the bar, but the honest fix is more val tasks.",
    "val split (at split freeze) has only 2 tasks (< 5) \u2014 acceptance decisions are LOW CONFIDENCE. The gate applies a Student-t small-sample correction (df=1), which widens the bar, but the honest fix is more val tasks."
  ],
  "dashboard": ".capevolve/run_r195b/dashboard.html",
  "dashboard_server": "http://127.0.0.1:7878"
}

```
Clears the floor (`honest_gate: true`), and the LOW CONFIDENCE warning is exactly
what `PROJECT.md` documents as honest limit #3. `gate_k_se: 1.0` is far under the 26.5 cap.

### 6. #181 adapter contract, side by side

`docs/ADAPTER_CONTRACT.md` on `origin/fix/issue-103-adapter-methods`:
```
## The three required methods

These three are `@abstractmethod` — `cap-evolve check` refuses to run until all
three are real (no `IMPLEMENT ME` stub):

```python
class Adapter(CapabilityAdapter):
    def tasks(self, split: str) -> list[Task]: ...
    def run_target(self, task: Task, ctx, *, seed: int = 0) -> Rollout: ...
    def score(self, task: Task, rollout: Rollout) -> Score: ...
```

- **`tasks(split)`** — where evaluation data comes from, for `split` in
...

You only override these when the default behavior doesn't fit:

```python
def materialize(self, candidate_dir, edits=None) -> None   # PURE write of {component: text} edits into candidate_dir
def live(self, candidate_dir)                              # @contextmanager: make the candidate live for ONE eval, yield ctx
def apply(self, candidate_dir, edits=None) -> None         # back-compat inject hook (env var / config patch / copy)
def trajectories(self, split, ctx=None) -> Path | None     # the runner's NATIVE trace dir for the last eval (default: None)
def runner_model(self) -> str | None                       # the CONSUMING model id, for check's mismatch note (default: None)
```

Three more optional fast paths are **not** on the base class — the harness probes for them
with `hasattr` (`core/cap_evolve/harness.py`) and uses them when present:

```python
def run_batch(self, tasks, ctx, *, seed=0) -> ...                                  # drive a benchmark's OWN batch runner INSTEAD of run_target (as tau2 does)
def run_trials(self, tasks, ctx, *, n_trials, base_seed) -> {id: [Rollout, ...]}   # batched fast path: ALL trials in ONE run
def score_batch(self, tasks, rollouts) -> {id: Score}                             # batched fast path: score a WHOLE trial in ONE call (e.g. one Docker harness invocation, as swebench does)
```
```

My reference (`examples/toy_calc/README.md`):
```
20:- [`adapter.py`](adapter.py) — a `CapabilityAdapter`: the **3 required** methods
21:  (`tasks`, `run_target`, `score`) plus an override of the optional `apply` hook.
68:  adapter implements: 3 required `@abstractmethod`s, defaulted hooks, `hasattr`-probed
```

My reference (`examples/toy_calc/PROJECT.md` — describing the same adapter):
```
18:  `adapter.py::Adapter.run_target` and computes the arithmetic correctly **only** when
21:- How `run_target` invokes it: reads `ctx/prompt.txt` (where `ctx` is the candidate dir
30:- Feedback signal: on failure, `score()` reports what was expected vs produced **plus
45:- `tasks(split)` deliberately returns **all** tasks and lets the harness filter by the
84:   stable, `score()` is deterministic, and `materialize()` is callable. It does not run
85:   `run_target`, so a non-deterministic runner passes `check`. Notably a `materialize()`
```

And the adapter itself implements exactly that set:
```
$ grep -n "    def " examples/toy_calc/adapter.py
34:    def tasks(self, split: str) -> list[Task]:
44:    def run_target(self, task: Task, ctx, *, seed: int = 0) -> Rollout:
60:    def score(self, task: Task, rollout: Rollout) -> Score:
70:    def apply(self, candidate_dir: Path, edits: dict | None = None) -> None:
```
Three required abstract methods + one optional-hook override. No "4 methods" anywhere:
```
$ grep -rn "4 method\|four method" examples/toy_calc/
(no matches)
```

### 7. #198 gate rule — no unqualified "banks 1-task gains"
```
$ grep -rn "banks\|k_se = 1.0\|k_se = 1\.0\|>= 2 tasks\|≥ 2 tasks" examples/toy_calc/*.md examples/toy_calc/capevolve.yaml
examples/toy_calc/PROJECT.md:75:   gives `mean(Δ) = SE(Δ)` *exactly*, so the strict `>` **rejects** it — at every `n`,
examples/toy_calc/PROJECT.md:76:   and no matter how large that single gain is. **Improve ≥ 2 tasks at `k_se = 1.0`, or
examples/toy_calc/PROJECT.md:77:   bank nothing.** Set `gate_k_se: 0.2` (as the other examples do) to bank a 1-of-`n`
examples/toy_calc/capevolve.yaml:60:#  1. At k_se = 1.0, improving exactly ONE of n val tasks gives mean(Δ) = SE(Δ)
examples/toy_calc/capevolve.yaml:62:#     gain is. You need >= 2 improved tasks to bank anything at 1.0. (The examples
```

### 8. #183 rebrand merges cleanly (H1 untouched)
```
$ git merge --no-commit --no-ff origin/fix/issue-105-rebrand
Auto-merging templates/project/PROJECT.md
Automatic merge went well; stopped before committing as requested

$ git diff origin/main -- templates/project/PROJECT.md
diff --git a/templates/project/PROJECT.md b/templates/project/PROJECT.md
index 32d1c447..80a6b351 100644
--- a/templates/project/PROJECT.md
+++ b/templates/project/PROJECT.md
@@ -3,6 +3,9 @@
 Filled by the `intake` skill. Records the decisions behind this run so anyone
 (human or agent) can understand and reproduce it.
 
+> For a **filled** worked example of this file (and of `capevolve.yaml`), see
+> `examples/toy_calc/` — a runnable, zero-API reference project.
+
 ## What we're optimizing
 - Capability: <skill-package | tools | mcp-tool | system-prompt | ...>
 - Artifact: `<path>`
```
#183 edits line 1 (the H1); mine adds at line 3. Non-adjacent, hence the clean auto-merge.

### 9. `run.sh` end to end (the CI gate path)
```
Working directory: /var/folders/zh/srgnbq_97qvb6002zsr40tgc0000gn/T/toy_calc.XXXXXX.QxcJlLN1lH
{
  "ok": true,
  "stubs": [],
  "problems": [],
  "notes": [
    "tasks('val') -> 8 task(s)",
  ]
}
{
  "best_id": "cand_0001",
  "baseline_val": 0.0,
  "test_reward": 1.0,
  "test_baseline_reward": 0.0,
  "test_delta": 1.0,
  "test_pass_k": {
    "1": 1.0,
    "2": 0.0
  },
  "iterations": 3,
  "dashboard_server": "skipped"
}

```
CI (`.github/workflows/ci.yml`) greps `"baseline_val": 0.0` and `"test_reward": 1.0` — both present.

### 10. Full test suite — this branch
```
.....................................                                    [100%]
181 passed in 68.87s (0:01:08)
```

### 10b. Full test suite — the unmodified baseline (`origin/main`), same command
```
179 passed in 57.19s
```

**179 baseline -> 181 on this branch: exactly +2, my two new tests, 0 failures either
side.** Both runs are green here; earlier in this session the same command hit
`test_dashboard_launch.py::test_maybe_launch_spawns_when_available` (the known port-7878
flake, **#200**) on the branch *and* on the unmodified baseline — `1 failed, 180 passed`
vs `1 failed, 178 passed`, the same +2 with the same pre-existing failure. It is timing-
dependent and unrelated to anything here.

### 11. Just the new tests (no `CAPEVOLVE_*` env preset — they build the project themselves)
```
$ PYTHONPATH=core python -m pytest core/tests/test_e2e_slice.py -v
core/tests/test_e2e_slice.py::test_full_slice PASSED                     [ 20%]
core/tests/test_e2e_slice.py::test_cyclic_variant_also_improves PASSED   [ 40%]
core/tests/test_e2e_slice.py::test_baseline_better_than_nothing_is_gated PASSED [ 60%]
core/tests/test_e2e_slice.py::test_worked_reference_check_and_spec_are_honest PASSED [ 80%]
core/tests/test_e2e_slice.py::test_worked_reference_runs_end_to_end PASSED [100%]
============================== 5 passed in 6.61s ===============================
```

### 12. `compileall`
```
$ python -m compileall -q core examples && echo clean
clean
```

### 13. All relative links resolve
```
OK   README.md -> ../../templates/project
OK   README.md -> PROJECT.md
OK   README.md -> capevolve.yaml
OK   README.md -> adapter.py
OK   README.md -> PROJECT.md
OK   README.md -> ../tau2_airline
OK   README.md -> ../skillsbench
OK   README.md -> ../../docs/ADAPTER_CONTRACT.md
OK   README.md -> ../../templates/project
OK   PROJECT.md -> adapter.py
OK   PROJECT.md -> capevolve.yaml
OK   PROJECT.md -> capability/prompt.txt
OK   PROJECT.md -> tasks.jsonl
OK   PROJECT.md -> mock_script.json
OK   PROJECT.md -> run.sh

broken: none
```

### 14. Anti-rot: the placeholder check derives its tokens from the template at test time
So it cannot go stale against the file it mirrors, and it fails loudly if the template
stops carrying placeholders at all:
```python
    template = (REPO / "templates" / "project" / "PROJECT.md").read_text(encoding="utf-8")
    tokens = set(re.findall(r"<[^<>\n]{2,40}>", template))
    assert tokens, "template has no placeholders to compare against — did its shape change?"
    for name in ("capevolve.yaml", "PROJECT.md"):
        body = (proj / name).read_text(encoding="utf-8")
        assert "Acapo" not in body, f"{name} carries the pre-rebrand name"
        left = sorted(t for t in tokens if t in body)
        assert not left, f"{name} still carries unfilled template placeholders: {left}"
    project_md = (proj / "PROJECT.md").read_text(encoding="utf-8")
```

### 15. Files touched
```
$ git show --stat --oneline HEAD
7d862c29 Ship a filled, runnable worked project reference (examples/toy_calc)
 CHANGELOG.md                     |  24 +++++++++
 core/tests/test_e2e_slice.py     | 108 +++++++++++++++++++++++++++++++++++++++
 examples/toy_calc/PROJECT.md     | 100 ++++++++++++++++++++++++++++++++++++
 examples/toy_calc/README.md      |  66 +++++++++++++++++++-----
 examples/toy_calc/capevolve.yaml |  79 ++++++++++++++++++++++++++++
 examples/toy_calc/run.sh         |   7 ++-
 templates/project/PROJECT.md     |   3 ++
 7 files changed, 373 insertions(+), 14 deletions(-)
```

### 16. Commit authorship — no Claude attribution
```
$ git log -1 --format="%an <%ae> | committer: %cn <%ce>"
Osher Elhadad <Osher.Elhadad@ibm.com> | committer: Osher Elhadad <Osher.Elhadad@ibm.com>
$ git log -1 --format=%B | grep -ci "co-authored\|generated with claude"
0
(0 = none)
```

OsherElhadad:
## 🔍 Review — PR #244

**APPROVE WITH NITS.** The premise rejection is correct and evidenced. The contract claims all match their sibling branches' ground truth — I checked each one against the unmerged branch, not against `main`. The two honesty disclosures reproduce verbatim. One anti-rot assertion is weaker than it reads (non-blocking, below), and the `check`-never-runs-`run_target` claim is true by default but has a documented opt-in escape the reference doesn't mention.

Nothing blocking.

### Blocking

None.

### Non-blocking

1. **`core/tests/test_e2e_slice.py:195` — the *Honest limits* guard passes on a gutted section.** The assertion is `"Honest limits" in project_md`, but `examples/toy_calc/PROJECT.md:26` contains the incidental prose `— see *Honest limits*`. I deleted the **entire** section body (lines 64–90, all five caveats) and the test still passed:

   ```
   --- PROBE 5: delete the ENTIRE Honest limits section ---
   1                       # remaining "Honest limits" occurrences (the line-26 cross-ref)
   ..                                                                       [100%]
   2 passed, 3 deselected in 3.29s
   ```

   Consequence: the one property this PR's whole value rests on — that the caveats stay in the file — is the one property the anti-rot test does not actually protect. Every *other* asserted property failed when broken (see below); this is the single hole.

   Fix: anchor on the heading and require the caveats survive, e.g.

   ```python
   assert "## Honest limits" in project_md, "the reference must keep its honest-limits section"
   body = project_md.split("## Honest limits", 1)[1].split("\n## ", 1)[0]
   for claim in ("STRICT fallback", "MIN_VAL_TASKS", "run_target"):
       assert claim in body, f"Honest limits lost its {claim} caveat"
   ```

2. **`examples/toy_calc/PROJECT.md:85` — "It does not run `run_target`" is true by default but not unconditionally.** `core/cap_evolve/check.py:178-182` *does* call `run_target` twice, gated behind `CAPEVOLVE_N_TRIALS>1` **and** `CAPEVOLVE_CHECK_TRIAL_PROBE=1` (the degenerate-trials probe). Consequence: a reader who sets both env vars sees `check` fire real rollouts, contradicting the reference's flat claim — mild, since the probe only ever appends a WARNING note and never fails the check. Fix: "It does not run `run_target` (except an opt-in degenerate-trials probe behind `CAPEVOLVE_CHECK_TRIAL_PROBE=1`, which only warns), so a non-deterministic runner passes `check`."

3. **No forward link to the benchmark zoo (#233).** #233's `benchmarks/toy_calc/README.md` links *back* here ("`examples/toy_calc/` keeps the hand-written adapter form... the reference for what a custom `CapabilityAdapter` looks like when the manifest does not fit"), but nothing in this PR points at `benchmarks/toy_calc/`. I grepped `examples/toy_calc/` and `CHANGELOG.md` for `benchmark|zoo` — only incidental prose hits. Consequence: the cross-link is one-directional, so a reader landing here doesn't learn the declarative path exists. Fix: one line in `examples/toy_calc/README.md`'s *Related* section once #233 lands. Ordering makes this unavoidable in this PR; worth a follow-up note rather than a rebase.

### Nits

4. `examples/toy_calc/PROJECT.md:72` and `capevolve.yaml:67` quote the transcript as `paired Δ̄=+1.0000 > 0 (SE=0 → STRICT fallback, warned; n=2)`. That is byte-exact on **main** and on #197/#198, but **#195 appends `; no t-correction on this path`**:

   ```
   #195 branch:  "paired Δ̄=+1.0000 > 0 (SE=0 → STRICT fallback, warned; n=2; no t-correction on this path)"
   this branch:  "paired Δ̄=+1.0000 > 0 (SE=0 → STRICT fallback, warned; n=2)"
   ```

   Not a defect today and the test uses a substring (`"STRICT fallback" in reason`), so it survives the merge — but the two quoted strings go subtly stale when #195 lands. Prefer quoting the stable prefix `(SE=0 → STRICT fallback, warned; n=2` or dropping the closing paren.

5. `examples/toy_calc/PROJECT.md:81` says val=2 "is still under the low-confidence threshold" without naming it. `LOW_CONFIDENCE_VAL_TASKS = 5` (`core/cap_evolve/splits.py:27` on #195). Naming the constant makes the caveat checkable.

### Is the premise rejection right?

**Yes, on all three facts, and I verified each independently:**

```
$ grep -n "capevolve" .gitignore
17:.capevolve/
$ git log --all --oneline -- .capevolve        # (no output — never tracked on any ref)
$ git ls-files | grep -i "PROJECT.md"
templates/project/PROJECT.md                   # exactly one, the template
```

So the issue's factual claim — "the scaffolded example project that **ships in the repo**, `.capevolve/project/`" — is wrong. Nothing ships there. Option A ("replace `.capevolve/project/` with a filled worked example") would mean committing a path that `.gitignore:17` excludes, and Option B ("delete the checked-in scaffold") is a no-op because there is nothing to delete. The issue was written against a developer's local run-state dir, not the repo.

**Is filling `examples/toy_calc/` the right substitute?** Yes, and this is the correct call rather than a dodge. Read past the wrong path, #108's actual complaint is *"a newcomer has no filled reference; the template teaches nothing"* — and its own "How to test" is `no <placeholder> tokens left` plus `cap-evolve check` passes, both of which this PR satisfies at the new location. I also confirmed the gap was real: on `origin/main`, `git ls-tree origin/main examples/toy_calc/` has **no `capevolve.yaml` and no `PROJECT.md`**, and `run.sh:19` copied `templates/project/capevolve.yaml` — the blank template. So the repo genuinely had zero filled spec and zero filled `PROJECT.md` anywhere. Claim verified.

**Does the epic instead want a `.gitignore` change?** I don't think so, and the PR is right to flag rather than assume. Committing a scaffold at `.capevolve/project/` would mean un-ignoring a live run-state directory — every subsequent `cap-evolve run` writes `.capevolve/run_*/` siblings, so you'd need a negative-pattern dance (`!.capevolve/project/`) and users' real runs would then fight the tracked reference on every `git status`. That is a genuinely different decision with a maintenance cost, and it should not be made silently inside a docs PR. The `templates/project/PROJECT.md:6-8` pointer this PR adds is the cheap version of the same discovery win: a reader who opens the blank template is told where the filled one lives. If the epic owner *does* want the checked-in path, that's a separate issue about `.gitignore` policy — not this one.

### Is this additive over #233's zoo, or should it be closed?

**Additive — keep it. But the margin is thinner than the PR's framing implies, and the honest reason to keep it is not the one the PR leads with.**

I read `benchmarks/toy_calc/` on #233's branch. The PR's three factual claims about it hold:

- `benchmarks/toy_calc/project/capevolve.yaml:1-2` is headed `# GENERATED from benchmark.yaml by 'cap-evolve benchmark add' — edit the manifest and re-run ... not this file.` So it is structurally disqualified as a "worked reference you read to learn the spec" — it tells you not to read it as authored.
- No `PROJECT.md` anywhere under `benchmarks/toy_calc/` (`find` returns 9 files; none is a `PROJECT.md`).
- The zoo's code is `project/target.py` — "one function, `run(task, ctx, *, seed=0)`" per its own README — i.e. the declarative path only. There is no hand-written `CapabilityAdapter` subclass there.

The decisive point, though, is that **#233's own README asks for this PR to exist**: *"`examples/toy_calc/` keeps the hand-written adapter form of the same benchmark — it is the 'before' side of the boilerplate measurement, and the reference for what a custom `CapabilityAdapter` looks like when the manifest does not fit."* #233 depends on `examples/toy_calc/` remaining the hand-written reference. Closing this PR as zoo-covered would leave #233's cross-link pointing at a directory with no filled spec and no `PROJECT.md` — a worse outcome than either branch alone.

Now the honest deduction. The PR says "the only thing lost would be the filled `PROJECT.md`." That undersells its own case slightly and oversells it in another way:

- **Genuinely unique and worth the PR:** the filled `PROJECT.md`, the per-decision commentary in `capevolve.yaml` (the zoo's is machine-generated and says don't read it), the five *Honest limits* caveats, and the two anti-rot tests. None of these exist anywhere in the repo today, on either branch.
- **The commentary is the real payload, not the `PROJECT.md`.** `capevolve.yaml:39-43` (why val=2 is the floor), `:49-53` (why `protected_paths` is omitted rather than `[]`), and `:59-69` (the two distinct gate caveats) encode three contract rules that a reader otherwise learns by breaking them. I reproduced all three against their sibling branches. That is teaching content, and it is not derivable from a generated file.

Verdict: **additive, keep open.** ~370 lines for the repo's only filled reference plus two tests that pin four contract rules is a fair trade, and the alternative leaves a dangling promise in #233. This is not manufactured work.

### Contract accuracy

Every claim checked against the **sibling branch**, not `main`.

| Claim in this PR | Sibling branch ground truth | Match? |
|---|---|---|
| "the **3 required** methods (`tasks`, `run_target`, `score`)" — `README.md:20-21` | **#181** `core/cap_evolve/adapter.py`: exactly 3 `@abstractmethod` at lines 77/86/99 → `tasks`, `run_target`, `score`. `docs/ADAPTER_CONTRACT.md:8` "## The three required methods" | ✅ |
| "plus an override of the optional `apply` hook" — `README.md:21`, `PROJECT.md:95` | **#181** `docs/ADAPTER_CONTRACT.md:41` "## Optional hooks (working defaults provided)", line 48 lists `apply` | ✅ `apply` correctly labelled optional |
| Any "4 methods" claim | grep for `four methods\|4 methods\|4 required` across all new files → **no hits** | ✅ none present |
| "3 required `@abstractmethod`s, defaulted hooks, `hasattr`-probed optional fast paths" — `README.md:68-69` | **#181** doc lines 41-58: defaulted hooks + 3 `hasattr`-probed paths (`run_batch`, `run_trials`, `score_batch`) | ✅ |
| "at `k_se = 1.0`, improving exactly ONE of `n` gives `mean(Δ) = SE(Δ)` *exactly*, so strict `>` **rejects** it — at every `n`, and however large that one gain is" — `capevolve.yaml:60-63`, `PROJECT.md:74-78` | **#198** `docs/ADAPTER_CONTRACT.md:113`: "(`mean(Δ) = SE(Δ)` exactly, at any `n` and any magnitude) — ≥2 improved tasks, or `gate_k_se < 1.0`" | ✅ exact, incl. both quantifiers |
| "`gate_mode: paired` … the default in every real run" — `capevolve.yaml:56` | **#198** doc:99 "**`paired` — the default.**" | ✅ |
| Unqualified "banks 1-task gains" anywhere | grep → **no such claim**; both docs state the opposite | ✅ absent |
| "`protected_paths` DELIBERATELY OMITTED … an empty list is a HARD ERROR" — `capevolve.yaml:49-53` | **#197** reproduced both halves on its branch — see *Verification* | ✅ |
| "`val = 2` … equals the harness's `MIN_VAL_TASKS = 2`; a smaller val split is *refused*" — `PROJECT.md:79-81` | **#195** `core/cap_evolve/splits.py:22` `MIN_VAL_TASKS = 2`; `gate.py:130-144` raises `TinyValSplitError` below it | ✅ |
| "2 is still under the low-confidence threshold" — `PROJECT.md:81` | **#195** `splits.py:27` `LOW_CONFIDENCE_VAL_TASKS = 5`; run emits `honest_gate: true` + the LOW CONFIDENCE warning | ✅ (nit 5: name the 5) |
| "`materialize()` that *raises* can still yield `{"ok": true}`" — `PROJECT.md:86` | Reproduced directly — see *Verification* | ✅ |
| "`check` … does not run `run_target`" — `PROJECT.md:85` | `core/cap_evolve/check.py:178-182` runs it behind `CAPEVOLVE_N_TRIALS>1` **and** `CAPEVOLVE_CHECK_TRIAL_PROBE=1` | ⚠️ true by default only — non-blocking #2 |
| Transcript `paired Δ̄=+1.0000 > 0 (SE=0 → STRICT fallback, warned; n=2)` | byte-exact here and on #197/#198; **#195** appends `; no t-correction on this path` | ⚠️ nit 4 |

### Anti-rot tests

**Real, with one hole.** I broke each asserted property and re-ran. 7 of 8 failed as they must:

| # | Property broken | Result |
|---|---|---|
| 1 | appended `protected_paths: []` to the spec | ✅ **FAILED** at `:171` |
| 2 | `split_val: 0.25` → `0.10` (val floor) | ✅ **FAILED** at `:183` |
| 3 | reinserted a `<path>` placeholder into `PROJECT.md` | ✅ **FAILED** at `:193` |
| 4 | `## Honest limits` → `## Caveats` (heading only) | ⚠️ **passed** — matched the line-26 cross-ref |
| 5 | deleted the **whole** *Honest limits* section (all 5 caveats) | ⚠️ **passed** — non-blocking #1 |
| 6 | patched `gate.py` to stop logging `gate_warning` | ✅ **FAILED** at `:234` |
| 7 | patched the `STRICT fallback` reason string out of `gate.py` | ✅ **FAILED** at `:237` |
| 8 | stripped every `<...>` placeholder from `templates/project/PROJECT.md` | ✅ **FAILED** at `:188` |

**The placeholder list is genuinely derived at test time — verified, not taken on trust.** Probe 8 is the proof: `test_e2e_slice.py:186-188` reads the template and `re.findall(r"<[^<>\n]{2,40}>", ...)`, then asserts `tokens` is non-empty. Removing the template's placeholders makes the token set empty and the test fails with "template has no placeholders to compare against — did its shape change?" — so the list cannot silently become vacuous. Probe 7 (adding a *new* placeholder to the template that isn't in the reference) correctly passed, which is the right behavior.

The subprocess test is honest work: `:208-212` shells out to the real CLI with the committed files, `:215-224` parses the last top-level JSON object, and `:232-237` reads `events.jsonl` rather than trusting stdout.

### Merge-order note

Trial-merged this branch against all six siblings:

```
MERGE origin/fix/issue-105-rebrand           -> exit=0 conflicts=[none]
MERGE origin/docs/issue-104-gate-modes       -> exit=0 conflicts=[none]
MERGE origin/fix/issue-113-small-samples     -> exit=0 conflicts=[none]
MERGE origin/fix/issue-103-adapter-methods   -> exit=1 conflicts=[examples/toy_calc/README.md]
MERGE origin/feat/issue-142-protected-paths  -> exit=1 conflicts=[CHANGELOG.md]
MERGE origin/feat/issue-141-benchmark-zoo    -> exit=1 conflicts=[CHANGELOG.md]
```

**#183 auto-merges clean, as the PR claims.** Verified the geometry: #183's only hunk is `templates/project/PROJECT.md:1` (`-# Acapo project` / `+# cap-evolve project`); this PR's hunk starts at line 6 (`@@ -3,6 +3,9 @@`). Non-adjacent, so git takes both.

**One conflict the PR does not mention.** #181 also edits `examples/toy_calc/README.md` — the same "## Files" bullet list this PR rewrites:

```
<<<<<<< HEAD
- [`adapter.py`](adapter.py) — a `CapabilityAdapter`: the **3 required** methods
  (`tasks`, `run_target`, `score`) plus an override of the optional `apply` hook.
=======
- `adapter.py` — a `CapabilityAdapter`: the 3 required methods (deterministic agent +
  exact-match scorer) plus an `apply` hook override.
>>>>>>> origin/fix/issue-103-adapter-methods
```

Trivial to resolve — **take HEAD** (this PR's version is the more specific and equally correct one; it names the three methods and labels `apply` optional). Worth calling out so it isn't a surprise. The two `CHANGELOG.md` conflicts are the ordinary "everyone appends under `### Added`" kind.

**Recommended order:** `#183 → #181 → #198 → #195 → #197 → #244 → #233`. Rationale: the doc/contract branches land first so this PR's cross-references (`docs/ADAPTER_CONTRACT.md`, the gate rules) resolve against merged truth rather than a branch tip; #244 lands after them (resolving the one-line #181 README conflict and, if #195 is in, relaxing the quoted transcript per nit 4); #233 lands last so its `examples/toy_calc/` back-link resolves and its author can add the missing forward link (non-blocking #3) in the same pass.

### Verification I re-ran

**Tests — exactly +2, as claimed:**
```
origin/main:  179 passed in 65.90s
this branch:  181 passed in 65.78s
```
No flakes; `test_dashboard_launch.py` (#200) passed on both runs.

**`compileall` clean; all 15 relative links resolve:**
```
relative links checked=15 broken=0
compileall OK
```

**No third copy of toy_calc** — `find -name tasks.jsonl` → `examples/toy_calc/` and `templates/adapters/jsonl_litellm/` (an unrelated template). Only `benchmarks/toy_calc/` (#233) + `examples/toy_calc/` exist as toy_calc dirs.

**`bash examples/toy_calc/run.sh` — `check` green, and the claimed sealed number reached:**
```
{
  "ok": true,
  "stubs": [],
  "problems": [],
  "notes": [
    "tasks('val') -> 8 task(s)",
  ]
}
{
  "run_dir": ".capevolve/run_demo", "best_id": "cand_0001",
  "baseline_val": 0.0,
  "test_reward": 1.0,
  "test_baseline_reward": 0.0, "test_delta": 1.0,
  "iterations": 3
}
```
`baseline_val 0.0 → test_reward 1.0`, `iterations: 3` — matches `PROJECT.md:52-54` and `capevolve.yaml:72-74` (win on iteration 1, then `stall: 2`). **CI's greps still match**: `.github/workflows/ci.yml:86-87` greps `'"baseline_val": 0.0'` and `'"test_reward": 1.0'` — both present verbatim in the output above.

**The gate really does degrade to STRICT — reproduced from `events.jsonl`, not stdout:**
```
Counter({'evaluate': 6, 'gate_warning': 3, 'step': 3, 'splits': 1, 'baseline': 1, 'finalize': 1})

GATE_WARNING: {"kind": "gate_warning", "mode": "paired", "context": "n=2",
  "reason": "combined/paired SE is 0 (likely n_trials=1 or identical trials) — the
   significance gate cannot distinguish noise from signal and is falling back to
   STRICT (accept any Δ>0). ..."}

STEP accept=True  | paired Δ̄=+1.0000 >  0 (SE=0 → STRICT fallback, warned; n=2)
STEP accept=False | paired Δ̄=+0.0000 <= 0 (SE=0 → STRICT fallback, warned; n=2)
STEP accept=False | paired Δ̄=+0.0000 <= 0 (SE=0 → STRICT fallback, warned; n=2)

SPLIT: {"kind": "splits", "train": 4, "val": 2, "test": 2, "seed": 0}
```
Three `gate_warning`s, all three steps on the STRICT path, and the accepted step's reason is **byte-identical** to the string quoted at `PROJECT.md:72` and `capevolve.yaml:67`. Splits are `4/2/2` as documented.

**Is the degradation a teaching hazard or a strength here? A strength.** The disclosure is adequate and then some — it is the *lead* item of *Honest limits* (`PROJECT.md:68-72`), repeated in the spec at the point of use (`capevolve.yaml:59-69`, directly under the `gate_k_se: 1.0` line a copier would copy), in `README.md:60-62`, and in the CHANGELOG. Critically, the reference does **not** overstate: `README.md:57-58` claims only "the gate **wiring**", and `PROJECT.md:89` says `0.0 → 1.0` "proves the loop plumbing, gate wiring, and seal — not that optimization works." A reader cannot come away believing the significance test was exercised. And `test_e2e_slice.py:232-237` asserts the warning still fires, so the disclosure can't rot into a lie.

**#197 control — both halves reproduce on `origin/feat/issue-142-protected-paths`:**
```
### A) reference AS-IS (protected_paths omitted) ###
  "test_reward": 1.0, "test_delta": 1.0, "iterations": 3        # runs clean, defaults apply

### B) CONTROL: same spec + "protected_paths: []" ###
cap_evolve.protect.TamperError: cap-evolve: `protected_paths` in
.capevolve/project/capevolve.yaml is an EMPTY list. That would protect nothing
(the scorer / eval harness / task data would all be optimizer-writable).
Remove the key to use the defaults, or declare the paths.
  raised at core/cap_evolve/protect.py:152 via harness.baseline
```
So `capevolve.yaml:49-53`'s comment is exactly right, and the omission is load-bearing rather than an oversight.

**#195 control on `origin/fix/issue-113-small-samples` — `honest_gate: true` and the LOW CONFIDENCE warning both appear:**
```
"baseline_val": 0.0,
"honest_gate": true,
"warnings": [
  "val split (at baseline) has only 2 tasks (< 5) — acceptance decisions are LOW
   CONFIDENCE. The gate applies a Student-t small-sample correction (df=1), which
   widens the bar, but the honest fix is more val tasks.",
  "val split (at split freeze) has only 2 tasks (< 5) — ..."
]
split_warning | {"val": 2, "min_recommended": 5, ...}
STEP accept=true | paired Δ̄=+1.0000 > 0 (SE=0 → STRICT fallback, warned; n=2;
                                          no t-correction on this path)
```
`val=2` == `MIN_VAL_TASKS`, `honest_gate: true`, warning present. Note the reason-string suffix (nit 4).

**Raising `materialize()` still yields `ok: true` — reproduced:**
```
{
  "ok": true, "stubs": [], "problems": [],
  "notes": [ ..., "materialize() probe raised (may be expected): BOOM" ]
}
```
`PROJECT.md:86` states this, and `PROJECT.md:82-86` is careful that `check` "is not proof the adapter is correct." The reference does not imply `check` proves more than it does.

**Net: every verification claim in the PR description reproduced.** The only corrections are the `run_target` opt-in probe (non-blocking #2), the weak *Honest limits* assertion (non-blocking #1), and the unmentioned #181 README conflict (merge note).

OsherElhadad:
## 🔧 Review fixes

Commit `617bc37a` (new commit, not an amend). Every claim below is literal output.

### Finding 1 — the *Honest limits* guard now asserts the content [FIXED]

You were right and the hole was exactly where you said. `"Honest limits" in project_md`
matched the cross-reference at `PROJECT.md:26`, so the section could be gutted or
renamed under it. The guard now anchors the heading, slices the section body, asserts a
distinctive phrase per caveat, and pins the count:

```python
assert "\n## Honest limits" in project_md, "the reference lost its `## Honest limits` heading"
limits = project_md.split("\n## Honest limits", 1)[1].split("\n## ", 1)[0]
for claim in ("STRICT fallback", "bank nothing", "MIN_VAL_TASKS", "run_target",
              "engineered"):
    assert claim in limits, f"Honest limits lost its `{claim}` caveat"
caveats = re.findall(r"^\d+\. \*\*", limits, flags=re.M)
assert len(caveats) == 5, f"Honest limits should list 5 caveats, found {len(caveats)}"
```

**The three break-probes you asked for — all three now FAIL.**

```
=== BASELINE (unbroken) ===
.                                                                        [100%]
1 passed, 4 deselected in 0.07s

=== PROBE A: delete the ENTIRE 'Honest limits' section (all 5 caveats) ===
remaining "Honest limits" occurrences: 1
E       assert '\n## Honest limits' in '# cap-evolve project — toy_calc arithmetic stand-in\n\nA **filled** version of `templates/project/PROJECT.md`. The te...cript.json) | the deterministic edit `mock` applies |\n| [`run.sh`](run.sh) | the whole thing end to end, zero API |\n'

core/tests/test_e2e_slice.py:198: AssertionError
FAILED core/tests/test_e2e_slice.py::test_worked_reference_check_and_spec_are_honest
1 failed, 4 deselected in 0.53s

=== PROBE B: rename the heading only (## Honest limits -> ## Caveats) ===
remaining "Honest limits" occurrences: 1
E       assert '\n## Honest limits' in '# cap-evolve project — toy_calc arithmetic stand-in\n\nA **filled** version of `templates/project/PROJECT.md`. The te...cript.json) | the deterministic edit `mock` applies |\n| [`run.sh`](run.sh) | the whole thing end to end, zero API |\n'

core/tests/test_e2e_slice.py:198: AssertionError
FAILED core/tests/test_e2e_slice.py::test_worked_reference_check_and_spec_are_honest
1 failed, 4 deselected in 1.04s

=== PROBE C: delete ONE caveat (#3, the val=2 floor) ===
caveats remaining: 4
E           assert 'MIN_VAL_TASKS' in ' — what this example does NOT prove\n\nWorth more than the parts that work, because these are the claims people over-...that optimization works on a real benchmark.\n   For that, see `examples/tau2_airline/` and `examples/skillsbench/`.\n'

core/tests/test_e2e_slice.py:202: AssertionError
FAILED core/tests/test_e2e_slice.py::test_worked_reference_check_and_spec_are_honest
1 failed, 4 deselected in 0.06s

=== RESTORED ===
.                                                                        [100%]
1 passed, 4 deselected in 0.04s
```

Note probe B: **one** `Honest limits` occurrence survives the rename (the line-26
cross-ref) — the exact string that used to keep the old assertion green — and the test
now fails anyway. And the count guard is not decorative; adding a **6th** caveat while
leaving all five phrases intact also fails, so the guard cannot silently drift out of
sync with the section:

```
=== PROBE C2: count guard is live (append a 6th caveat, all 5 phrases intact) ===
core/tests/test_e2e_slice.py:204: AssertionError
FAILED core/tests/test_e2e_slice.py::test_worked_reference_check_and_spec_are_honest
1 failed, 4 deselected in 0.05s
```

Point taken on this being the seventh instance in the epic of a check matching the wrong
thing. The general shape of that mistake is asserting on a *substring of a whole file*
when the property is scoped to a *section*; slicing the section first is what makes the
assertion mean what it says.

### The other 7 anti-rot properties — re-run, none weakened

```
=== P1: append protected_paths: [] to the spec ===
FAILED core/tests/test_e2e_slice.py::test_worked_reference_check_and_spec_are_honest
1 failed, 1 passed, 3 deselected in 3.42s

=== P2: split_val 0.25 -> 0.10 (val floor) ===
37:split_val: 0.10
FAILED core/tests/test_e2e_slice.py::test_worked_reference_check_and_spec_are_honest
1 failed, 1 passed, 3 deselected in 3.41s

=== P3: reinsert a <path> placeholder into PROJECT.md ===
FAILED core/tests/test_e2e_slice.py::test_worked_reference_check_and_spec_are_honest
1 failed, 1 passed, 3 deselected in 3.37s

=== P8: strip every <...> placeholder from templates/project/PROJECT.md ===
FAILED core/tests/test_e2e_slice.py::test_worked_reference_check_and_spec_are_honest
1 failed, 1 passed, 3 deselected in 3.38s

=== P6: patch gate.py to stop logging gate_warning ===
FAILED core/tests/test_e2e_slice.py::test_worked_reference_runs_end_to_end
1 failed, 1 passed, 3 deselected in 3.46s

=== P7: patch the 'STRICT fallback' reason string out of gate.py ===
FAILED core/tests/test_e2e_slice.py::test_worked_reference_runs_end_to_end
1 failed, 1 passed, 3 deselected in 3.56s

=== P7b (control): add a NEW placeholder to the template, absent from the reference — must PASS ===
..                                                                       [100%]
2 passed, 3 deselected in 3.33s

=== ALL RESTORED ===
..                                                                       [100%]
2 passed, 3 deselected in 3.53s
```

So it is now **10 of 10** broken properties detected (your 7, plus heading rename,
section deletion, and single-caveat deletion), with the false-positive control still
correctly passing.

### Finding 2 — the `run_target` claim, qualified against the code [FIXED]

`examples/toy_calc/PROJECT.md:84-89` now reads:

```
84:4. **`cap-evolve check` passing is not proof the adapter is correct.** It proves the
85-   three abstract methods are implemented and non-stubbed, `tasks()` is non-empty and
86-   stable, `score()` is deterministic, and `materialize()` is callable. It does not run
87-   `run_target` (except an opt-in degenerate-trials probe behind
88-   `CAPEVOLVE_CHECK_TRIAL_PROBE=1` *and* `CAPEVOLVE_N_TRIALS>1`, which only ever appends
```

Matched against the code it describes:

```
177:    import os as _os
178:    if (int(_os.environ.get("CAPEVOLVE_N_TRIALS", "1")) > 1
179:            and _os.environ.get("CAPEVOLVE_CHECK_TRIAL_PROBE") == "1"):
180:        try:
181:            r1 = adapter.run_target(probe_task, Path(project_dir), seed=0)
182:            r2 = adapter.run_target(probe_task, Path(project_dir), seed=1)
183:            d1 = json.dumps(r1.to_dict(), default=str, sort_keys=True)
```

Both env vars, the `and` conjunction, and the warn-only outcome (`rep.notes.append`, no
`rep.problems`) are all now stated. Your framing is the right one: an honesty doc making
an unqualified claim the code contradicts is precisely the failure this reference exists
to prevent.

### Finding 3 — cross-link is now bidirectional [FIXED]

Added to `examples/toy_calc/README.md`'s *Related* section, and made explicit that the
two directories are a deliberate pair rather than duplicates:

```
- `benchmarks/toy_calc/` — the **declarative** form of this same benchmark: a
  `benchmark.yaml` manifest plus a one-function `project/target.py`, with the spec
  *generated* by `cap-evolve benchmark add`. Take that path when the manifest fits; this
  directory is the reference for the hand-written `CapabilityAdapter` when it doesn't.
  The two are deliberately paired — the zoo entry links back here for the adapter form,
  and this is the "before" side of its boilerplate measurement. *(Lands with #233; the
  path is intentionally unlinked until then.)*
```

Deliberately **not** a markdown link, since `benchmarks/toy_calc/` doesn't exist until
#233 merges and the link-resolution check (15/15 below) would otherwise fail on this
branch. It becomes linkable in #233's pass.

### Nit 4 — quoted transcript now uses the stable prefix [FIXED]

`PROJECT.md:72` and `capevolve.yaml:67` dropped the closing paren so #195's
`; no t-correction on this path` suffix can't make them stale:

- `PROJECT.md`: ``…warned; n=2`… (the suffix varies by branch; the prefix is the stable part).``
- `capevolve.yaml`: ``…warned; n=2` (the stable prefix; the suffix after `n=2` varies by branch).``

### Nit 5 — constant named [FIXED]

`PROJECT.md:81` now: *"2 is still under `LOW_CONFIDENCE_VAL_TASKS = 5`, the threshold
that flags acceptance decisions as LOW CONFIDENCE."* Matches `core/cap_evolve/splits.py:27`
on #195 (`LOW_CONFIDENCE_VAL_TASKS = 5`).

### Verification

**`cap-evolve check` green and the sealed number reproduces — `bash examples/toy_calc/run.sh`:**
```
{
  "ok": true,
  "stubs": [],
  "problems": [],
  "notes": [
    "tasks('val') -> 8 task(s)",
  ]
}
{
  "best_id": "cand_0001",
  "baseline_val": 0.0,
  "test_reward": 1.0,
  "test_baseline_reward": 0.0,
  "test_delta": 1.0,
  "iterations": 3
}
```

**Full suite — still exactly 181, 0 failed (no new tests; the existing guard got teeth):**
```
.....................................                                    [100%]
181 passed in 69.63s (0:01:09)
```
`test_dashboard_launch.py` (#200) passed this run.

**`compileall` clean; all 15 relative links still resolve** (the new zoo pointer is
deliberately unlinked, so the count is unchanged):
```
compileall clean
relative links checked=15 broken=0
```

**Authorship:**
```
Osher Elhadad <Osher.Elhadad@ibm.com> | Osher Elhadad <Osher.Elhadad@ibm.com>
```

### Merge order — corrected, with the conflict I missed

Adopting your order:

**`#183 → #181 → #198 → #195 → #197 → #244 → #233`**

Three conflicts to expect, all trivial:

1. **`#181` also edits `examples/toy_calc/README.md`** — the same `## Files` bullet list
   this PR rewrites. **I did not mention this and should have.** Resolution: **take HEAD**
   (this PR's version, which names the three required methods and labels `apply` optional).
2. **`#197` → `CHANGELOG.md`** — ordinary "everyone appends under `### Added`".
3. **`#233` → `CHANGELOG.md`** — same.

`#183` auto-merges clean (its only hunk is `templates/project/PROJECT.md:1`, the H1;
this PR's hunk starts at line 6 — non-adjacent).

Two order-dependent follow-ups: once **#195** is in, the quoted transcript prefix above
is already suffix-tolerant, so nothing further is needed; once **#233** is in, its author
(or a one-line follow-up) turns the `benchmarks/toy_calc/` pointer into a real link.

Thanks for the review — particularly for breaking the assertion rather than reading it,
and for the point that I undersold the PR's own case. The per-decision commentary being
the real payload (not the `PROJECT.md`) is the better framing, and it is what #233's
back-link depends on.

OsherElhadad:
Closing — the issue is moot and the surviving gap is much smaller than this PR.
**The issue's premise is gone.** `.capevolve/` was never tracked (`git log --all -- '.capevolve'`
is empty) and is ignored (`.gitignore:19` → `.capevolve*/`), so Option B is already the
status quo. This PR says so itself and then re-scopes to a different deliverable.
**The re-scoped gap is mostly already covered.** The claim of "zero commented
`capevolve.yaml` anywhere" is not true on main: `templates/project/capevolve.yaml` documents
every knob inline, and `examples/tau2_airline/capevolve.yaml` and
`examples/skillsbench/capevolve.yaml` are already filled worked specs with a why-note per
decision. The only artifact genuinely missing is a filled `PROJECT.md`.
**Three of the five headline caveats are stale against main** — the exact failure mode this
PR was written to prevent:
- `MIN_VAL_TASKS = 2` / `LOW_CONFIDENCE_VAL_TASKS = 5`: no such symbols exist
  (`grep -rn "MIN_VAL\|LOW_CONFIDENCE" core/cap_evolve/` → nothing; `splits.py` has no floor).
  They are from unmerged #195.
- `protected_paths: []` "is a HARD ERROR": on main `cli.py:904` is
  `pp = spec.get("protected_paths") or []` — empty is silently off. From unmerged #197.
- The README links `benchmarks/toy_calc/`, which does not exist (#233 unmerged).
**The anti-rot tests are prose tripwires**: `len(caveats) == 5`, a literal `## Honest limits`
heading, and required substrings including `MIN_VAL_TASKS` — pinning the docs to a symbol the
code does not have. Plus a second full `cap-evolve run` subprocess duplicating what
`test_e2e_slice.py` already covers.
~390 lines to ship one missing markdown file. If we still want it, the follow-up is
`examples/toy_calc/PROJECT.md` written against the contracts that actually exist today,
plus the one-line pointer in `templates/project/PROJECT.md` — no new spec, no prose-lint
tests. Reopening #108 for that narrow scope is fine; this branch is not the shape to
land it in.

PR #222, opened 2026-07-30
feat(algorithm): re-inject rejected approaches as optimizer constraints (#129)

OsherElhadad:
Closes #129

## The honesty gap first: was "never re-proposed" true?

**No — it was false on both halves.** `RUN.md` claimed:

> Rejected approaches are remembered and never re-proposed.

Before this PR:
1. **Nothing re-injected them.** `rejected.jsonl` was written but its only reader was the dashboard. PR #212 (issue #114) deleted the *write-only* memory whose `render()` never reached a prompt and left an explicit note: `_augment_instructions` is the only function whose output reaches the optimizer. Nothing routed rejections there.
2. **Nothing could prevent a re-proposal** even in principle — the optimizer is a black-box agent CLI.

The e2e transcript below shows the failure concretely: with a mock optimizer proposing the same bad edit, **hill-climb re-proposed the identical rejected edit 5 times** and the prompt never mentioned it once. That's 4 wasted full-val evals per run.

`RUN.md` and `docs/COMPARISON.md` are corrected to state what's actually true.

## What changed

**`harness.approach_signature(parent_dir, cand_dir)`** — a stable, compact signature of *what* an edit changed, from the capability diff (`_diff_capabilities`, the same source the dashboard/RUNMAP use, so it never picks up injected read-context or algorithm scratch): per touched file, whitespace-collapsed added/removed lines. Cosmetic variants of one idea collapse to one signature. A no-op edit (optimizer errored → workdir is a verbatim parent copy) yields `""` and is not injected.

**`harness.dead_end_constraints(run_dir)`** — the `## ALREADY TRIED & REJECTED` block: deduped signature + gate reason + repeat count, plus the actual constraint ("do not re-propose; if you revisit one, state in `PROCESS.md` what is materially different and which lesson it counters").

**Wired into `_augment_instructions`** — the one function whose output reaches the prompt (#114) and which **all three algorithms route through**, so hill-climb / GEPA / SkillOpt get it with zero per-algorithm plumbing. Every rejection site now records `approach`: `run_step`, GEPA's local minibatch gate, GEPA's val gate, and GEPA's two merge gates.

Per the task brief I used **`RunDir.iteration_events()` semantics throughout and hand-filtered no event kinds** — `dead_end_constraints` reads `rejected.jsonl`, which is kind-agnostic by construction, so #216's class of bug cannot recur here.

### How constraints are bounded

[...]

PR #219, opened 2026-07-30
Durable synthesized priors (INSIGHTS.md) fed to every proposal, all three algorithms

OsherElhadad:
Closes #128.

A compact, bounded, framework-synthesized "what we've learned so far" block — `INSIGHTS.md` — that survives across iterations independent of the transcript, and reaches the optimizer prompt for **all three** deterministic algorithms.

## What the insight actually is

Three sections, all derived from the run's own numbers:

| Section | Content |
| --- | --- |
| **What HELPED** | gate-ACCEPTED iterations, largest val \|Δ\| first, with the exact tasks each fixed (and any it broke anyway) |
| **What HURT** | gate-REJECTED iterations, largest \|Δ\| first, with the reject reason and the exact tasks broken |
| **Still OPEN** | the val tasks the current best does NOT pass |

Written to **both** the run dir (`run_<ts>/INSIGHTS.md` — the durable copy, and what the dashboard Insights tab can read) and each iteration's workdir (the copy that reaches the prompt).

## Is an LLM call involved? No.

Zero LLM calls. The synthesis is pure Python over `events.jsonl` + the persisted rollouts, matching the profile #205/#132 established for every auxiliary step in core (reflection distillation, ledger/runmap, failure clustering, the whole `diagnose` phase). It reads `RunDir.iteration_events()` and `_candidate_task_impact()` — the same reads `LEDGER.md` uses, so the two artifacts can never disagree.

#205's `aux_model` tier is deliberately **unused**: the signal is already fully determined by the run's own numbers, so a per-iteration model call on every run would buy prose, not information. If a future version wants narrative priors, that is where it belongs.

## Built on the #199 seam

Its reviewer called #128 a "✅ two-line" extension, and it is: the entire wiring is two lines in `harness._augment_instructions` — the ONE function whose output reaches the prompt, and which all three algorithms route through (the note #212 left in `memory.py`). No per-algorithm block, no signature change to `render_instructions`.

Critically, `_insight_rows` reads **`RunDir.iteration_events()`**, not `kind == "step"`. That hand-filtering is the bug #199 fixed in three consumers (GEPA emits `gepa_val_gate`); a fourth instance is still open as #216. Filtering by hand would have made the priors block permanently empty for the flagship algorithm.

## Growth / eviction policy and cap

The block is **re-derived every iteration**, so it does not grow monotonically — but its input does, and an unbounded render would silently eat #199's `MAX_INSTRUCTIONS_CHARS` budget over a long run.

[...]

PR #212, opened 2026-07-29
refactor(core): drop dead optimizer-memory API + unused params; fix misleading cache docstring

OsherElhadad:
Closes #114

## The issue's premise is only PARTLY right — scoped accordingly

#114 says `memory.py` is write-only and the jsonl machinery can go. I checked empirically before deleting, and the issue itself asked to "confirm the `rejected.jsonl`/`history.jsonl` files aren't relied on by the dashboard before removing writes … (the dashboard reads `events.jsonl`, not these — verify)".

**Verified: that parenthetical is wrong.** The jsonl files have a live reader. So I removed only the genuinely-unread surface and **kept the writes**.

### Write-only proof

The prompt-facing API — `render()`, `entries()`, `_render_impact` — has zero callers outside `memory.py` itself:

```
$ grep -rn "\.render()\|\.entries()\|_render_impact" --include="*.py" core/cap_evolve skills examples
core/cap_evolve/memory.py:40:def _render_impact(e: dict) -> Optional[str]:
core/cap_evolve/memory.py:86:        items = self.entries()[-limit:]
core/cap_evolve/memory.py:102:            imp = _render_impact(e)
core/cap_evolve/memory.py:133:        items = self.entries()[-limit:]
core/cap_evolve/memory.py:142:            imp = _render_impact(e)
```

Every hit is `memory.py` calling itself. `LEDGER.md`/`JOURNAL.md`/`RUNMAP.md` (via `_augment_instructions`) fully replaced this as the optimizer's cross-iteration channel. **Dead → removed.**

Same for the per-record `note` / `broke` / `fixed` fields — written on every iteration, read by nobody:

```
$ grep -rn "broke\|fixed" dashboard/frontend/src dashboard/backend --include="*.tsx" --include="*.ts" --include="*.py"
dashboard/frontend/src/test/insights.test.ts:36:  { candidate_id: 'c', summary: '', reason: 'broke correctness gate', val: 0 },
dashboard/frontend/src/components/Trajectories.tsx:115:  <div className="fixed inset-0 z-40 ...
dashboard/frontend/src/lib/api.ts:71: * first snapshot) from a genuinely missing/broken run. */
dashboard/backend/capevolve_dashboard/runs.py:19:  prefixed ``run_`` or lacking ``events.jsonl``. ...
```

Three CSS/prose coincidences, zero real readers. Field-level audit against the dashboard's consumers:

| field | readers in dashboard | verdict |
|---|---|---|
| `candidate_id` | 5 | **live** |
[...]

PR #246, opened 2026-07-30
feat(algorithm): on-demand reasoning skills + an advisory proposal-quality declaration (#140)

### LINKED ISSUE #140: On-demand reasoning skills + a "mechanism-not-knob" proposal quality gate
## Problem / opportunity

cap-evolve's skills are structured by **role** (phases / capabilities / algorithms / optimizers). There's no notion of a small, **on-demand "reasoning skill"** loaded at a specific step to counter a specific optimizer failure mode — e.g. the classic "the model skips analysis and jumps to a plausible one-line tweak" or "it proposes a config knob instead of a mechanism." The `diagnose`/proposal step relies on prose instructions rather than a reusable, testable quality gate.

## Why it matters

Prior art in this space (Arbor, Apache-2.0 — inspiration only) uses two ideas worth adopting *as techniques*:
1. **On-demand reasoning skills** — tiny skills loaded only at the step where they matter, each targeting a named failure mode, toggleable.
2. A **"mechanism, not knob" ideation gate** — before a proposal is accepted for evaluation, the optimizer must declare a mechanism/hypothesis/observable and pass a self-check that rejects "could this just be one config tweak?" ideas.

For cap-evolve, whose sample-efficiency depends on each expensive proposal being *bold and well-reasoned* (its own design tenet), a reusable proposal-quality gate raises edit quality and reduces wasted rollouts — directly sharpening `diagnose` and the optimizer prompt.

## What to change

1. Introduce a lightweight **reasoning-skill** mechanism: small skills (e.g. `mechanism-probe`, `first-principles`) that the proposal step can load on demand, each documenting the failure mode it counters. Keep them within the existing skill registry/validation.
2. Add a **proposal quality gate**: require the optimizer to state hypothesis + expected observable for its edit, and reject/flag proposals that are trivial knob-tweaks when a mechanism change is warranted. Log the declaration alongside the candidate.
3. Wire it into `diagnose`/the proposal prompt for all algorithms (via #109's shared path).

## How to test

- New test: a proposal missing a hypothesis/observable declaration is flagged (mock optimizer emits a bare tweak → gate marks it low-quality).
- The reasoning skills register and pass `skills/_registry/build_manifest.py` validation.
- The declaration is persisted per candidate and visible in the dashboard Insights.

- Builds on **#109** (shared injection) and complements **#128/#129** (priors/constraints). Touches the `diagnose` phase + skill registry. **Parallel-safe** with UX/site/install issues.

OsherElhadad:
Closes #140

Builds on **#199** (the algorithm hub) and stacks on **#222** (`feat/issue-129-failure-memory`, whose base already carries #199 + #212). The reviewer's "✅ one `copytree`" assessment held: the file half of this is literally one `copytree` on #199's `inject` seam.

## What the skills are and when they load

A new **`reasoning` skill component**: tiny skills loaded at ONE step to counter ONE named optimizer failure mode. Unlike a phase they are never sequenced by the orchestrate DAG — they reach the optimizer as injected read-context, at `./guidance/reasoning/<skill>/`, plus native placement in the agent's own skills dir (`.claude/skills/…`) so a headless CLI auto-loads them.

**One skill, `mechanism-probe`, loaded at the proposal step.** The failure mode it counters is the one this framework keeps paying for: the optimizer reads a few traces, recognizes a familiar shape, and ships the first plausible edit — another prose rule for a rule the agent already skips, a retuned threshold, a reworded docstring. It asks one question while the edit is still cheap to throw away: *could this whole proposal be replaced by changing one existing value or restating one existing rule? If yes, it is a knob.*

I did **not** add a second `first-principles` skill. Nothing loads it and nothing would; a registered skill with no caller is a knob of its own. Add it when a second named failure mode actually needs it.

## Whether the gate is advisory or enforcing

**Advisory.** Nothing in this PR can reject a candidate. The exact honest wording, verbatim from the prompt every optimizer receives:

> **How this is judged:** the declaration is RECORDED per candidate, not enforced — a missing or knob-shaped declaration does NOT reject your edit, and a well-declared one does not get it accepted. The val significance gate remains the only thing that accepts or rejects. Declare it anyway: an edit you cannot state a mechanism and an observable for is the edit that historically wasted the iteration.

Why not enforcing: "is this a mechanism or a knob?" is a judgement no regex can make. A heuristic strict enough to reject knobs would also reject a real one-line in-body guard, and **a false rejection discards a genuine improvement invisibly** — nobody sees the gain that never happened. So this adopts #129's resolution exactly: **advisory at the prompt, hard at the val gate.** `RUN.md` and `docs/ARCHITECTURE.md` say "recorded, never enforced", not "rejects low-quality proposals" — the #222 lesson about not claiming enforcement nothing enforces.

What *is* precisely checkable is the **declaration**: presence of three named fields (`Mechanism:` / `Hypothesis:` / `Expected observable:`) in the `PROCESS.md` the optimizer already writes. `cap_evolve.proposal_quality` parses those and logs a `proposal_quality` event per candidate, surfaced in the dashboard's existing annotations stream.

### False-rejection probe

The test that matters, pinned from both sides:

- `test_a_genuine_mechanism_proposal_is_not_rejected` — a declared, genuinely-improving edit is **accepted**, and the gate reason contains none of `mechanism` / `knob` / `declar` / `proposal quality`.
- `test_an_undeclared_proposal_is_also_not_rejected` — the bare mock edit, with **no declaration at all**, is also accepted on its val delta. A missing declaration is a signal, not a verdict.

`mechanism-probe`'s own `check.py` carries the same probe as a behavioral contract.

## LLM calls

**Zero.** Pure stdlib regex over a markdown file — #205's rule that every auxiliary step in core is pure Python is preserved, and `test_the_gate_makes_no_model_call` pins it (no `anthropic`/`openai`/`requests`/`urllib`/`aux_model`/`subprocess` in the module). No `aux_model` tier needed.

## Composed-prompt measurement

Routed through **#222's single shared `cap_instructions`** — no second cap. `test_the_bar_is_capped_by_the_shared_cap_not_a_second_one` asserts `_augment_instructions` applies `cap_instructions` exactly once.

```
rendered template + #222 dead-ends (200 rejections) + #140 bar  :  24416
  of which #140's PROMPT_BLOCK                                  :   1320
  of which #222's dead-end block @ 200 rejections               :   2251
+ #219's INSIGHTS pointer bullet                                :    463
+ #221's diversify block (39 lineages -> bounded to 6)          :    887
-------------------------------------------------------------------------------
COMPOSED TOTAL (#219 + #221 + #222 + #140)                      :  25766
CAP (MAX_INSTRUCTIONS_CHARS)                                    :  60000
headroom                                                        :  34234  (42.9% of cap)
```

**Truncation never silently drops a whole block.** The block is 1,320 chars and sits LAST, inside the kept 30% tail (18 KB at the default ceiling), next to #222's constraints. `test_an_overflowing_prompt_keeps_the_bar_whole` **crosses the bound** (the #219 lesson — a pinning test whose fixture never overflows proves nothing): it feeds 110,000 chars through the real `_augment_instructions` and asserts the composed output is ≤ cap, the elision notice is present, and *both* #140's and #222's blocks survive **entire**.

## Per-algorithm evidence

Real `cap-evolve run` on `examples/toy_calc` with the `mock` optimizer, **zero API cost**, all three deterministic algorithms. `diff -r` against source is rc=0 for every one:

| algorithm | workdir | `diff -r` vs source | composed prompt | bar in prompt | `proposal_quality` logged |
|---|---|---|---|---|---|
| hill-climb | `cand_0001` | **rc=0 (identical)** | 22,265 | ✅ | ✅ |
| gepa | `gepa_0001` | **rc=0 (identical)** | 23,634 | ✅ | ✅ |
| skillopt | `so_e01s01` | **rc=0 (identical)** | 22,793 | ✅ | ✅ |

## No-leak proof

`grep` for each sealed test id across every file in each optimizer workdir: **0 files** for all three algorithms. `test_no_test_split_id_reaches_any_injected_workdir_file` pins it, and asserts the test-id list is non-empty first so the probe cannot be vacuous.

The injected subtree is also invisible to the snapshot / eval-cache hash / GEPA component list — `guidance/` is already in #199's `INJECTED_DIRS`, and `test_the_reasoning_skill_is_not_mistaken_for_a_capability_edit` proves the hash and component list are unchanged by its presence.

## Expected merge order

`#199` → `#212` → `#211` → `#219` / `#221` / **#222** → **this**. This branch is cut from `origin/feat/issue-129-failure-memory` because it must route through #222's `cap_instructions`. #219 and #221 touch `_augment_instructions` / the same prompt tail; whichever lands second resolves a small conflict in that one function. #213's skill lint should land before or with this so the new skill is linted in CI from day one (it already passes).

## Verification

```
240 passed in 103.91s
```
214 on the #222 base + 19 new + 7 `test_dashboard_launch` (which passed here). **0 failed.**

```
$ python -m compileall -q core/cap_evolve core/tests skills
rc=0

wrote skills/_registry/manifest.json (21 skill(s))
  reasoning: mechanism-probe

$ python skills/_registry/lint_skills.py skills      # #213's lint
skill authoring lint — 21 skill package(s)
  # zero errors AND zero advisories for reasoning/mechanism-probe
  # the 3 remaining errors (capabilities/tools body size, using-cap-evolve XML tags)
  # are pre-existing on this base and are exactly what #213 fixes on its own branch.

$ for f in skills/*/*/scripts/check.py; do ...; done
rc=0 ok=True   × 21 / 21   (including skills/reasoning/mechanism-probe)
```

**Fail-before / pass-after** — test file present, implementation stashed:
```
15 failed, 3 passed in 9.80s
```
then restored: `19 passed`.

### DIFF
--- a/RUN.md
+++ b/RUN.md
@@ -74,7 +74,12 @@ artifact.
 - Acceptance is gated on **val**, never on the data the optimizer edited against.
 - Only the **primary** metric (the scalar reward) gates; shown-only secondaries (e.g. `cost_usd`, `db_match`) are displayed but never affect accept/reject.
 - Multi-trial scoring reports mean + stderr; pass^k is reported when trials > 1.
-- Rejected approaches are remembered and never re-proposed.
+- Rejected approaches are remembered (`rejected.jsonl`) and re-injected into every later
+  proposal prompt as an explicit "already tried & rejected — do not re-propose" constraint
+  block, carrying the exact edit signature and why the gate killed it. Enforcement is
+  **advisory** at the prompt: the optimizer is a black-box agent CLI, so cap-evolve cannot
+  forbid it from re-emitting an edit. What is **hard** is the gate — a re-proposed dead end
+  is still rejected on val, and the repeat is counted in the constraint block.
 
 ## One-command alternative
 If you'd rather not drive it turn-by-turn:
--- a/core/cap_evolve/__init__.py
+++ b/core/cap_evolve/__init__.py
@@ -17,6 +17,7 @@ from .cache import EvalCache, hash_candidate_dir
 from .gate import GateDecision, TrainGateError, decide
 from .lr_schedule import build_schedule
 from .memory import History, RejectedMemory
+from .optimizer_context import OptimizerContext
 from .rundir import Budget, RunDir, Spent
 from .selection import PICKERS, STRATEGIES, pick, validate_strategy
 from .splits import Splits, TestSealError, make_splits
@@ -41,6 +42,7 @@ __all__ = [
     "decide",
     "History",
     "RejectedMemory",
+    "OptimizerContext",
     "Budget",
     "RunDir",
     "Spent",
--- a/core/cap_evolve/__init__.py
+++ b/core/cap_evolve/__init__.py
@@ -4,7 +4,7 @@ This is the *only* shipped code in cap-evolve. Everything user-facing is an
 Agent Skill; those skills' ``run.py`` scripts call into here for the things that
 must be consistent and honest across every run: task/score types, seeded splits
 with a sealed test set, variance-aware statistics, the acceptance gate,
-optimizer memory, the run directory, and the adapter contract.
+the run directory, and the adapter contract.
 
 Import it directly when Python is available, or invoke ``python -m
 cap_evolve <command>`` and parse the JSON it prints.
--- a/core/cap_evolve/cache.py
+++ b/core/cap_evolve/cache.py
@@ -22,13 +22,16 @@ import hashlib
 import json
 from pathlib import Path
 
-from .rundir import SCRATCH_NAMES
+from .rundir import NON_CAPABILITY_NAMES
 
 # Files that are NOT part of the capability (optimizer scratch, memory, vcs); they
 # must not perturb the content hash or every iteration would miss the cache.
-# ``rundir.SCRATCH_NAMES`` is the shared definition (see the note there); INSTRUCTIONS
-# and PROCESS are added because they ARE snapshotted but still are not capability bytes.
-_IGNORE_NAMES = {"INSTRUCTIONS.md", "PROCESS.md"} | set(SCRATCH_NAMES)
+# ``rundir.NON_CAPABILITY_NAMES`` is the shared definition (see the note there): the
+# union of live + legacy scratch plus INSTRUCTIONS/PROCESS, which ARE snapshotted but
+# still are not capability bytes. This is a read-side filter (skip bytes when hashing),
+# so it takes the whole union — including the legacy names, so caches written before
+# they were retired keep resolving.
+_IGNORE_NAMES = set(NON_CAPABILITY_NAMES)
 _IGNORE_DIRS = {".git", "__pycache__", "prior_iterations"}
 
 
@@ -48,9 +51,15 @@ def hash_candidate_dir(candidate_dir: Path) -> str:
     for p in cdir.rglob("*"):
         if not p.is_file():
             continue
-        if p.name in _IGNORE_NAMES:
+        rel = p.relative_to(cdir)
+        # Root-anchored: every ignored name is a root-level framework injection, so a
+        # NESTED file that merely shares one (``src/prompts/STATE.md``) is capability
+        # content and MUST fold into the digest — otherwise deleting it leaves the hash
+        # unchanged and the next iteration serves the parent's cached rewards for a
+        # materially different candidate (a stale hit on a mutilated candidate).
+        if len(rel.parts) == 1 and p.name in _IGNORE_NAMES:
             continue
-        if any(part in _IGNORE_DIRS for part in p.relative_to(cdir).parts):
+        if any(part in _IGNORE_DIRS for part in rel.parts):
             continue
         files.append(p)
     for p in sorted(files, key=lambda x: str(x.relative_to(cdir))):
--- a/core/cap_evolve/cache.py
+++ b/core/cap_evolve/cache.py
@@ -30,7 +30,7 @@ import hashlib
 import json
 from pathlib import Path
 
-from .optimizer_context import INJECTED_DIRS, INJECTED_NAMES
+from .optimizer_context import INJECTED_DIRS, INJECTED_NAMES, SCRATCH_NAMES
 
 # Files that are NOT part of the capability (optimizer scratch, memory, vcs); they
 # must not perturb the content hash or every iteration would miss the cache. The
@@ -39,10 +39,7 @@ from .optimizer_context import INJECTED_DIRS, INJECTED_NAMES
 # definition in ``optimizer_context``, folded in here as a plain constant expression
 # (no import-time set mutation, so there is no import-order dependence to reason about;
 # ``optimizer_context`` imports only ``.loop`` and ``.rundir`` at module level).
-_IGNORE_NAMES = {"MEMORY.md", "STATE.md", "INSTRUCTIONS.md", "REJECTED.md", "FOCUS.md",
-                 "REFLECTION.md",
-                 # cross-iteration state files (clean-ownership redesign) — scratch, not capability
-                 "LEDGER.md", "JOURNAL.md", "PROCESS.md", "RUNMAP.md"} | set(INJECTED_NAMES)
+_IGNORE_NAMES = set(SCRATCH_NAMES) | set(INJECTED_NAMES)
 _IGNORE_DIRS = {".git", "__pycache__"} | set(INJECTED_DIRS)
 
 
--- a/core/cap_evolve/cache.py
+++ b/core/cap_evolve/cache.py
@@ -1,16 +1,18 @@
 """Eval cache — skip a rollout when the same candidate was already scored on a task.
 
 Keyed by ``(hash of the candidate's editable files, task_id) -> {reward, feedback}``
-and persisted in the run dir. The hash is over file CONTENTS, so two byte-identical
-candidates share cache entries even under different ids.
+and persisted in the run dir, so re-evaluating an identical candidate (e.g. a parent
+re-sampled in GEPA, or a resumed run) costs nothing. The hash is over file CONTENTS,
+so two byte-identical candidates share cache entries even under different ids.
 
-**Scope: GEPA only.** The single consumer is ``gepa._eval_minibatch``, where the same
-parent is re-sampled from the Pareto frontier across iterations and re-scored on
-overlapping minibatches — that repetition is what the cache pays for.
-``harness.evaluate_candidate`` (every full-val and sealed-test eval, in every algorithm)
-does NOT consult it and always pays full price, so re-scoring an identical candidate on
-full val — common on ``--resume`` or a seed re-eval — is not deduplicated. Wiring it in
-there is a possible perf win, not existing behavior; there is no flag for it today.
+**Scope: GEPA only** — which bounds the "costs nothing" above. The single consumer is
+``gepa._eval_minibatch``, where the same parent is re-sampled from the Pareto frontier
+across iterations and re-scored on overlapping minibatches; that repetition is what the
+cache pays for. ``harness.evaluate_candidate`` (every full-val and sealed-test eval, in
+every algorithm) does NOT consult it and always pays full price, so re-scoring an
+identical candidate on full val — including on ``--resume`` or a seed re-eval — is NOT
+deduplicated. Wiring it in there is a possible perf win, not existing behavior, and
+there is no flag for it today (``test_w1_engine.py`` guards that claim).
 
 Honesty notes:
   * The cache stores only the SCORE (reward + feedback), never gold answers.
--- a/core/cap_evolve/cache.py
+++ b/core/cap_evolve/cache.py
@@ -1,17 +1,23 @@
 """Eval cache — skip a rollout when the same candidate was already scored on a task.
 
 Keyed by ``(hash of the candidate's editable files, task_id) -> {reward, feedback}``
-and persisted in the run dir, so re-evaluating an identical candidate (e.g. a parent
-re-sampled in GEPA, or a resumed run) costs nothing. The hash is over file CONTENTS,
-so two byte-identical candidates share cache entries even under different ids.
+and persisted in the run dir. The hash is over file CONTENTS, so two byte-identical
+candidates share cache entries even under different ids.
+
+**Scope: GEPA only.** The single consumer is ``gepa._eval_minibatch``, where the same
+parent is re-sampled from the Pareto frontier across iterations and re-scored on
+overlapping minibatches — that repetition is what the cache pays for.
+``harness.evaluate_candidate`` (every full-val and sealed-test eval, in every algorithm)
+does NOT consult it and always pays full price, so re-scoring an identical candidate on
+full val — common on ``--resume`` or a seed re-eval — is not deduplicated. Wiring it in
+there is a possible perf win, not existing behavior; there is no flag for it today.
 
 Honesty notes:
   * The cache stores only the SCORE (reward + feedback), never gold answers.
   * It is keyed on candidate-file content, so an edit (even whitespace) busts the
     key — a stale score can never be served for changed files.
   * It is an optimization, not a source of truth: ``events.jsonl`` still records
-    every evaluation. Wiring into ``evaluate_candidate`` is OFF by default and gated
-    behind a flag (see ``maybe_cached_score``) so it cannot silently change behavior.
+    every evaluation.
 
 Pure stdlib (hashlib + json).
 """
--- a/core/cap_evolve/cache.py
+++ b/core/cap_evolve/cache.py
@@ -22,12 +22,13 @@ import hashlib
 import json
 from pathlib import Path
 
+from .rundir import SCRATCH_NAMES
+
 # Files that are NOT part of the capability (optimizer scratch, memory, vcs); they
 # must not perturb the content hash or every iteration would miss the cache.
-_IGNORE_NAMES = {"MEMORY.md", "STATE.md", "INSTRUCTIONS.md", "REJECTED.md", "FOCUS.md",
-                 "REFLECTION.md",
-                 # cross-iteration state files (clean-ownership redesign) — scratch, not capability
-                 "LEDGER.md", "JOURNAL.md", "PROCESS.md", "RUNMAP.md"}
+# ``rundir.SCRATCH_NAMES`` is the shared definition (see the note there); INSTRUCTIONS
+# and PROCESS are added because they ARE snapshotted but still are not capability bytes.
+_IGNORE_NAMES = {"INSTRUCTIONS.md", "PROCESS.md"} | set(SCRATCH_NAMES)
 _IGNORE_DIRS = {".git", "__pycache__", "prior_iterations"}
 
 
--- a/core/cap_evolve/cache.py
+++ b/core/cap_evolve/cache.py
@@ -31,18 +31,20 @@ import json
 from pathlib import Path
 
 from .optimizer_context import INJECTED_DIRS, INJECTED_NAMES
+from .rundir import NON_CAPABILITY_NAMES
 
 # Files that are NOT part of the capability (optimizer scratch, memory, vcs); they
-# must not perturb the content hash or every iteration would miss the cache. The
-# INJECTED_* halves are the optimizer-context read-context (``trajectories/``,
-# ``guidance/``, the native per-agent skill dirs and instructions files) — one
-# definition in ``optimizer_context``, folded in here as a plain constant expression
-# (no import-time set mutation, so there is no import-order dependence to reason about;
-# ``optimizer_context`` imports only ``.loop`` and ``.rundir`` at module level).
-_IGNORE_NAMES = {"MEMORY.md", "STATE.md", "INSTRUCTIONS.md", "REJECTED.md", "FOCUS.md",
-                 "REFLECTION.md",
-                 # cross-iteration state files (clean-ownership redesign) — scratch, not capability
-                 "LEDGER.md", "JOURNAL.md", "PROCESS.md", "RUNMAP.md"} | set(INJECTED_NAMES)
+# must not perturb the content hash or every iteration would miss the cache.
+# ``rundir.NON_CAPABILITY_NAMES`` is the shared definition (see the tier note there):
+# the union of live + legacy scratch plus INSTRUCTIONS/PROCESS, which ARE snapshotted
+# but still are not capability bytes. This is a read-side filter (skip bytes when
+# hashing), so it takes the whole union — including the legacy names, so caches written
+# before they were retired keep resolving. The INJECTED_* halves are the
+# optimizer-context read-context (``trajectories/``, ``guidance/``, the native per-agent
+# skill dirs and instructions files) — one definition in ``optimizer_context``, folded in
+# here as a plain constant expression (no import-time set mutation, so there is no
+# import-order dependence to reason about).
+_IGNORE_NAMES = set(NON_CAPABILITY_NAMES) | set(INJECTED_NAMES)
 _IGNORE_DIRS = {".git", "__pycache__"} | set(INJECTED_DIRS)
 
 
@@ -62,9 +64,15 @@ def hash_candidate_dir(candidate_dir: Path) -> str:
     for p in cdir.rglob("*"):
         if not p.is_file():
             continue
-        if p.name in _IGNORE_NAMES:
+        rel = p.relative_to(cdir)
+        # Root-anchored: every ignored name is a root-level framework injection, so a
+        # NESTED file that merely shares one (``src/prompts/STATE.md``) is capability
+        # content and MUST fold into the digest — otherwise deleting it leaves the hash
+        # unchanged and the next iteration serves the parent's cached rewards for a
+        # materially different candidate (a stale hit on a mutilated candidate).
+        if len(rel.parts) == 1 and p.name in _IGNORE_NAMES:
             continue
-        if any(part in _IGNORE_DIRS for part in p.relative_to(cdir).parts):
+        if any(part in _IGNORE_DIRS for part in rel.parts):
             continue
         files.append(p)
     for p in sorted(files, key=lambda x: str(x.relative_to(cdir))):
--- a/core/cap_evolve/cache.py
+++ b/core/cap_evolve/cache.py
@@ -5,13 +5,21 @@ and persisted in the run dir, so re-evaluating an identical candidate (e.g. a pa
 re-sampled in GEPA, or a resumed run) costs nothing. The hash is over file CONTENTS,
 so two byte-identical candidates share cache entries even under different ids.
 
+**Scope: GEPA only** — which bounds the "costs nothing" above. The single consumer is
+``gepa._eval_minibatch``, where the same parent is re-sampled from the Pareto frontier
+across iterations and re-scored on overlapping minibatches; that repetition is what the
+cache pays for. ``harness.evaluate_candidate`` (every full-val and sealed-test eval, in
+every algorithm) does NOT consult it and always pays full price, so re-scoring an
+identical candidate on full val — including on ``--resume`` or a seed re-eval — is NOT
+deduplicated. Wiring it in there is a possible perf win, not existing behavior, and
+there is no flag for it today (``test_w1_engine.py`` guards that claim).
+
 Honesty notes:
   * The cache stores only the SCORE (reward + feedback), never gold answers.
   * It is keyed on candidate-file content, so an edit (even whitespace) busts the
     key — a stale score can never be served for changed files.
   * It is an optimization, not a source of truth: ``events.jsonl`` still records
-    every evaluation. Wiring into ``evaluate_candidate`` is OFF by default and gated
-    behind a flag (see ``maybe_cached_score``) so it cannot silently change behavior.
+    every evaluation.
 
 Pure stdlib (hashlib + json).
 """
--- a/core/cap_evolve/cache.py
+++ b/core/cap_evolve/cache.py
@@ -22,29 +22,20 @@ import hashlib
 import json
 from pathlib import Path
 
+from .optimizer_context import INJECTED_DIRS, INJECTED_NAMES
+
 # Files that are NOT part of the capability (optimizer scratch, memory, vcs); they
-# must not perturb the content hash or every iteration would miss the cache.
+# must not perturb the content hash or every iteration would miss the cache. The
+# INJECTED_* halves are the optimizer-context read-context (``trajectories/``,
+# ``guidance/``, the native per-agent skill dirs and instructions files) — one
+# definition in ``optimizer_context``, folded in here as a plain constant expression
+# (no import-time set mutation, so there is no import-order dependence to reason about;
+# ``optimizer_context`` imports only ``.loop`` and ``.rundir`` at module level).
 _IGNORE_NAMES = {"MEMORY.md", "STATE.md", "INSTRUCTIONS.md", "REJECTED.md", "FOCUS.md",
                  "REFLECTION.md",
                  # cross-iteration state files (clean-ownership redesign) — scratch, not capability
-                 "LEDGER.md", "JOURNAL.md", "PROCESS.md", "RUNMAP.md"}
-_IGNORE_DIRS = {".git", "__pycache__"}
-
-
-def _load_injected_ignores() -> None:
-    """Fold in everything the optimizer-context seam injects into a workdir.
-
-    Injected read-context (``trajectories/``, ``guidance/``, the native per-agent skills
-    dirs and instructions files) is NOT capability content, so it must not perturb the
-    hash — otherwise a GEPA minibatch would miss the cache on every iteration. Imported
-    lazily so this module stays at the bottom of the import graph.
-    """
-    from .optimizer_context import INJECTED_DIRS, INJECTED_NAMES
-    _IGNORE_DIRS.update(INJECTED_DIRS)
-    _IGNORE_NAMES.update(INJECTED_NAMES)
-
-
-_load_injected_ignores()
+                 "LEDGER.md", "JOURNAL.md", "PROCESS.md", "RUNMAP.md"} | set(INJECTED_NAMES)
+_IGNORE_DIRS = {".git", "__pycache__"} | set(INJECTED_DIRS)
 
 
 def hash_candidate_dir(candidate_dir: Path) -> str:
--- a/core/cap_evolve/cache.py
+++ b/core/cap_evolve/cache.py
@@ -30,16 +30,21 @@ import hashlib
 import json
 from pathlib import Path
 
-from .optimizer_context import INJECTED_DIRS, INJECTED_NAMES, SCRATCH_NAMES
+from .optimizer_context import INJECTED_DIRS, INJECTED_NAMES
+from .rundir import NON_CAPABILITY_NAMES
 
 # Files that are NOT part of the capability (optimizer scratch, memory, vcs); they
-# must not perturb the content hash or every iteration would miss the cache. The
-# INJECTED_* halves are the optimizer-context read-context (``trajectories/``,
-# ``guidance/``, the native per-agent skill dirs and instructions files) — one
-# definition in ``optimizer_context``, folded in here as a plain constant expression
-# (no import-time set mutation, so there is no import-order dependence to reason about;
-# ``optimizer_context`` imports only ``.loop`` and ``.rundir`` at module level).
-_IGNORE_NAMES = set(SCRATCH_NAMES) | set(INJECTED_NAMES)
+# must not perturb the content hash or every iteration would miss the cache.
+# ``rundir.NON_CAPABILITY_NAMES`` is the shared definition (see the tier note there):
+# the union of live + legacy scratch plus INSTRUCTIONS/PROCESS, which ARE snapshotted
+# but still are not capability bytes. This is a read-side filter (skip bytes when
+# hashing), so it takes the whole union — including the legacy names, so caches written
+# before they were retired keep resolving. The INJECTED_* halves are the
+# optimizer-context read-context (``trajectories/``, ``guidance/``, the native per-agent
+# skill dirs and instructions files) — one definition in ``optimizer_context``, folded in
+# here as a plain constant expression (no import-time set mutation, so there is no
+# import-order dependence to reason about).
+_IGNORE_NAMES = set(NON_CAPABILITY_NAMES) | set(INJECTED_NAMES)
 _IGNORE_DIRS = {".git", "__pycache__"} | set(INJECTED_DIRS)
 
 
@@ -59,9 +64,15 @@ def hash_candidate_dir(candidate_dir: Path) -> str:
     for p in cdir.rglob("*"):
         if not p.is_file():
             continue
-        if p.name in _IGNORE_NAMES:
+        rel = p.relative_to(cdir)
+        # Root-anchored: every ignored name is a root-level framework injection, so a
+        # NESTED file that merely shares one (``src/prompts/STATE.md``) is capability
+        # content and MUST fold into the digest — otherwise deleting it leaves the hash
+        # unchanged and the next iteration serves the parent's cached rewards for a
+        # materially different candidate (a stale hit on a mutilated candidate).
+        if len(rel.parts) == 1 and p.name in _IGNORE_NAMES:
             continue
-        if any(part in _IGNORE_DIRS for part in p.relative_to(cdir).parts):
+        if any(part in _IGNORE_DIRS for part in rel.parts):
             continue
         files.append(p)
     for p in sorted(files, key=lambda x: str(x.relative_to(cdir))):
--- a/core/cap_evolve/cache.py
+++ b/core/cap_evolve/cache.py
@@ -28,7 +28,23 @@ _IGNORE_NAMES = {"MEMORY.md", "STATE.md", "INSTRUCTIONS.md", "REJECTED.md", "FOC
                  "REFLECTION.md",
                  # cross-iteration state files (clean-ownership redesign) — scratch, not capability
                  "LEDGER.md", "JOURNAL.md", "PROCESS.md", "RUNMAP.md"}
-_IGNORE_DIRS = {".git", "__pycache__", "prior_iterations"}
+_IGNORE_DIRS = {".git", "__pycache__"}
+
+
+def _load_injected_ignores() -> None:
+    """Fold in everything the optimizer-context seam injects into a workdir.
+
+    Injected read-context (``trajectories/``, ``guidance/``, the native per-agent skills
+    dirs and instructions files) is NOT capability content, so it must not perturb the
+    hash — otherwise a GEPA minibatch would miss the cache on every iteration. Imported
+    lazily so this module stays at the bottom of the import graph.
+    """
+    from .optimizer_context import INJECTED_DIRS, INJECTED_NAMES
+    _IGNORE_DIRS.update(INJECTED_DIRS)
+    _IGNORE_NAMES.update(INJECTED_NAMES)
+
+
+_load_injected_ignores()
 
 
 def hash_candidate_dir(candidate_dir: Path) -> str:
--- a/core/cap_evolve/cli.py
+++ b/core/cap_evolve/cli.py
@@ -86,6 +86,53 @@ def _resolve_algorithm(name: str) -> tuple[str, str | None]:
     return name, None
 
 
+# Algorithms whose run.py declares the shared optimizer-context flag set (via
+# ``OptimizerContext.add_arguments``). The agent-driven ones (evograph, agent-optimize)
+# have no deterministic loop and only tolerate the base flags, so they are excluded —
+# an explicit list, not a silent per-flag drop (issue #109).
+_OPTIMIZER_CONTEXT_ALGORITHMS = ("hill-climb", "gepa", "skillopt")
+
+
+def _optimizer_context_flags(spec: dict, project: str, optimizer_name: str) -> list[str]:
+    """Build the optimizer-context CLI flags from the spec, for ANY algorithm.
+
+    Every value here is "what context does the optimizer get" — the capability brief +
+    guidance, the intake-authored instructions template, the benchmark repo, supporting
+    source files, the optimizer's own features reference, and the consuming-LLM profile.
+    Paths are resolved project-relative when not absolute.
+    """
+    out: list[str] = []
+
+    def _project_rel(value) -> Path:
+        p = Path(str(value))
+        return p if p.is_absolute() or p.exists() else Path(project) / str(value)
+
+    def _csv(value) -> list[str]:
+        if isinstance(value, str):
+            return [v.strip() for v in value.split(",") if v.strip()]
+        return [str(v).strip() for v in (value or []) if str(v).strip()]
+
+    caps = _csv(spec.get("capabilities"))
+    if caps:
+        out += ["--capabilities", ",".join(caps)]
+    if optimizer_name:
+        out += ["--optimizer-name", str(optimizer_name)]
+    instr_p = _project_rel(spec.get("optimizer_instructions_file")
+                           or "optimizer/INSTRUCTIONS.md")
+    if instr_p.exists():
+        out += ["--instructions-file", str(instr_p)]
+    if spec.get("runner_repo_path"):
+        out += ["--bench-repo", str(_project_rel(spec["runner_repo_path"]))]
+    csrc = _csv(spec.get("capability_sources"))
+    if csrc:
+        out += ["--capability-sources", ",".join(csrc)]
+    if spec.get("target_model"):
+        out += ["--target-model", str(spec["target_model"])]
+    if spec.get("target_profile_file"):
+        out += ["--target-profile-file", str(_project_rel(spec["target_profile_file"]))]
+    return out
+
+
 def _resolve_skills(skills_dir: Path) -> dict:
     manifest = skills_dir / "_registry" / "manifest.json"
     if not manifest.exists():
@@ -345,54 +392,14 @@ def _cmd_run(argv):
         alg_cmd += ["--resume"]
     if algorithm_focus is not None:
         alg_cmd += ["--focus", algorithm_focus]
-    # Surface the selected capability skills to the optimizer prompt so it knows the
-    # allowed edit space (e.g. tools → may add composite tools). hill-climb consumes
-    # --capabilities; algorithms without the flag ignore the extra arg via argparse error,
-    # so only pass it to those that accept it.
-    caps = spec.get("capabilities") or []
-    if isinstance(caps, str):
-        caps = [c.strip() for c in caps.split(",") if c.strip()]
-    if caps and algorithm_name == "hill-climb":
-        alg_cmd += ["--capabilities", ",".join(str(c) for c in caps)]
-    # Thread the resolved optimizer NAME so the harness can copy that optimizer's
-    # features reference (parallel-subagent capabilities etc.) into each iteration's
-    # workdir. Only hill-climb accepts the flag; other algorithms ignore it.
-    if algorithm_name == "hill-climb":
-        alg_cmd += ["--optimizer-name", str(optimizer_name)]
-    # Optimizer-instructions template (intake-authored, per benchmark) + benchmark repo
-    # as read-only optimizer context. Both are resolved project-relative if not absolute.
-    # The instructions file defaults to the scaffolded project/optimizer/INSTRUCTIONS.md.
-    if algorithm_name == "hill-climb":
-        instr = spec.get("optimizer_instructions_file") or "optimizer/INSTRUCTIONS.md"
-        instr_p = Path(instr)
-        if not instr_p.is_absolute() and not instr_p.exists():
-            instr_p = Path(project) / instr
-        if instr_p.exists():
-            alg_cmd += ["--instructions-file", str(instr_p)]
-        repo = spec.get("runner_repo_path")
-        if repo:
-            repo_p = Path(str(repo))
-            if not repo_p.is_absolute() and not repo_p.exists():
-                repo_p = Path(project) / str(repo)
-            alg_cmd += ["--bench-repo", str(repo_p)]
-        # Supporting source files (data models / types the tools import) copied verbatim
-        # into the optimizer's ./guidance/sources/ so it can write correct code. Resolved
-        # project-relative by the harness; we pass them through as given.
-        csrc = spec.get("capability_sources") or []
-        if isinstance(csrc, str):
-            csrc = [c.strip() for c in csrc.split(",") if c.strip()]
-        if csrc:
-            alg_cmd += ["--capability-sources", ",".join(str(c) for c in csrc)]
-        # Consuming/runtime LLM the capabilities are optimized FOR (distinct from the
-        # optimizer model). A model id or a tier keyword; steers the optimizer prompt.
-        if spec.get("target_model"):
-            alg_cmd += ["--target-model", str(spec["target_model"])]
-        tpf = spec.get("target_profile_file")
-        if tpf:
-            tpf_p = Path(str(tpf))
-            if not tpf_p.is_absolute() and not tpf_p.exists():
-                tpf_p = Path(project) / str(tpf)
-            alg_cmd += ["--target-profile-file", str(tpf_p)]
+    # Optimizer-context flags — what the optimizer is GIVEN. These used to be gated
+    # behind `algorithm_name == "hill-climb"`, so a user who configured a capability
+    # brief, an instructions template, a bench repo, capability sources or a target model
+    # had them SILENTLY DROPPED on gepa/skillopt (issue #109). Every deterministic
+    # algorithm now declares the same flag set via
+    # OptimizerContext.add_arguments, so they are passed unconditionally.
+    if algorithm_name in _OPTIMIZER_CONTEXT_ALGORITHMS:
+        alg_cmd += _optimizer_context_flags(spec, project, optimizer_name)
     # gepa treats metric-calls as its PRIMARY budget; forward it explicitly (hill-climb
     # has no such flag and enforces the same cap via run_dir.budget_exhausted()).
     if algorithm_name == "gepa" and spec.get("max_metric_calls"):
--- a/core/cap_evolve/dashboard.py
+++ b/core/cap_evolve/dashboard.py
@@ -281,6 +281,18 @@ def reduce_run(run_dir) -> dict:
             gate_warnings.append({"reason": ev.get("reason"), "context": ev.get("context"),
                                   "mode": ev.get("mode")})
             continue
+        if kind == "proposal_quality":
+            # #140: the per-candidate proposal declaration, surfaced in the existing
+            # annotations stream (no new panel). ADVISORY — it never explains an
+            # accept/reject, so it must not read like a gate outcome.
+            missing = ", ".join(ev.get("missing") or [])
+            text = (f"declared mechanism: {ev.get('mechanism')} | expected observable: "
+                    f"{ev.get('observable')}" if not missing else
+                    f"no mechanism declaration (missing: {missing}) — advisory, "
+                    "the val gate still decided this candidate")
+            diagnoses.append({"kind": kind, "candidate": _step_candidate(ev),
+                              "text": text})
+            continue
         if kind in ("diagnose", "optimizer_error"):
             diagnoses.append({
                 "kind": kind,
--- a/core/cap_evolve/dashboard.py
+++ b/core/cap_evolve/dashboard.py
@@ -52,8 +52,9 @@ import shutil
 import subprocess
 from pathlib import Path
 
-from .optimizer_context import SCRATCH_NAMES as _oc_scratch
-from .rundir import ITERATION_EVENT_KINDS, iteration_candidate as _step_candidate
+from .optimizer_context import INJECTED_DIRS, INJECTED_NAMES
+from .rundir import (ITERATION_EVENT_KINDS, NON_CAPABILITY_NAMES,
+                     iteration_candidate as _step_candidate)
 
 # ---------------------------------------------------------------------------
 # Secret redaction
@@ -588,9 +589,22 @@ def reduce_run(run_dir) -> dict:
 # snapshot but are NOT capability edits — skipped when diffing iterations so the diff
 # shows only the real change. (The big read-context dirs trajectories/ and guidance/
 # are already excluded from the snapshot itself; see harness._SNAPSHOT_IGNORE.)
-# ONE definition (optimizer_context.SCRATCH_NAMES) — this copy had drifted from GEPA's
-# and was missing FOCUS.md/REFLECTION.md, so a GEPA candidate's diff showed scratch.
-_DIFF_SKIP = set(_oc_scratch)
+# Derived from ``rundir.NON_CAPABILITY_NAMES`` — a read-side FILTER like the cache and
+# component lists, so it takes the whole union (live + legacy scratch + the two
+# snapshotted explainability files). It must NOT be shared with
+# ``harness._SNAPSHOT_IGNORE``, which is DESTRUCTIVE and takes ``SCRATCH_NAMES`` only:
+# feeding this list to the snapshot would DELETE PROCESS.md, the explainability record
+# we deliberately keep. Same predicate, different operation. See the tier note in
+# rundir.py; the split is pinned by
+# test_gepa.py::test_scratch_ignores_are_one_shared_definition.
+#
+# The INJECTED_* halves are here for the same reason as in ``harness._CAP_DIFF_SKIP``:
+# the parent side of the diff is a snapshot (injected read-context stripped) and the
+# child side may be a live workdir (it is not), so without them an injected
+# ``CLAUDE.md``/``.claude/skills/`` file reads as a real capability addition. Derived,
+# never enumerated.
+_DIFF_SKIP = set(NON_CAPABILITY_NAMES) | set(INJECTED_NAMES)
+_DIFF_SKIP_DIRS = set(INJECTED_DIRS)
 
 
 def _read_dir_files(d: Path) -> dict[str, str]:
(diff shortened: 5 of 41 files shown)


OsherElhadad:
## 🔬 Evidence

All commands run in `/tmp/wt-140` — a worktree on `feat/issue-140-reasoning-skills`, cut from `origin/feat/issue-129-failure-memory` (= #222, which already carries #199 + #212 + #211). Python is `/tmp/ce-venv/bin/python`, always with `PYTHONPATH=/tmp/wt-140/core`. Every optimizer invocation is the deterministic `mock` optimizer on `examples/toy_calc` — **zero API cost, zero model calls**.

---

### 1. Baseline on the #222 base, before my change

```
$ cd /Users/osherelhadad/Documents/capevo/cap-evolve && git fetch origin
$ git worktree add /tmp/wt-140 -b feat/issue-140-reasoning-skills origin/feat/issue-129-failure-memory
Preparing worktree (new branch 'feat/issue-140-reasoning-skills')
HEAD is now at 6d6cc52d fix(algorithm): make the rejected-approach signature reflect the real edit (#129 review)

$ PYTHONPATH=/tmp/wt-140/core python -m pytest core/tests -q --ignore=core/tests/test_dashboard_launch.py
214 passed in 90.17s (0:01:30)
```

---

### 2. Real end-to-end through the actual CLI — all three deterministic algorithms

Each is a genuine `cap-evolve run` with `algorithm_skill` swapped in the spec. Shows the reasoning skill reaching the optimizer's workdir (`find` + `diff -r` rc=0 against source, as #213 and #185 did), the composed prompt size, the bar's presence, the logged declaration, and the sealed-split grep.

```
==============================================================
### cap-evolve run --algorithm hill-climb   (mock optimizer, $0)
==============================================================
  "iterations": 1,
}

--- reasoning skill in each optimizer workdir:
work/cand_0001/guidance/reasoning/mechanism-probe/SKILL.md
    diff rc=0 (identical) : cand_0001
    prompt    22265 chars; bar present: 1
--- proposal_quality events:
"kind": "proposal_quality", "candidate": "cand_0001", "declared": false, "missing": ["mechanism", "hypothesis", "observable"], "mechanism": "", "hypothesis": "", "observable": "", "enforcement": "advisory"
--- sealed test ids leaked into work/ :
    test ids = [a8 a7]
    files in work/ containing a test id: 0
==============================================================
### cap-evolve run --algorithm gepa   (mock optimizer, $0)
==============================================================
  "iterations": 1,
}

--- reasoning skill in each optimizer workdir:
work/gepa_0001/guidance/reasoning/mechanism-probe/SKILL.md
    diff rc=0 (identical) : gepa_0001
    prompt    23634 chars; bar present: 1
--- proposal_quality events:
"kind": "proposal_quality", "candidate": "gepa_0001", "declared": false, "missing": ["mechanism", "hypothesis", "observable"], "mechanism": "", "hypothesis": "", "observable": "", "enforcement": "advisory"
--- sealed test ids leaked into work/ :
    test ids = [a8 a7]
    files in work/ containing a test id: 0
==============================================================
### cap-evolve run --algorithm skillopt   (mock optimizer, $0)
==============================================================
  "iterations": 1,
}

--- reasoning skill in each optimizer workdir:
work/so_e01s01/guidance/reasoning/mechanism-probe/SKILL.md
    diff rc=0 (identical) : so_e01s01
    prompt    22793 chars; bar present: 1
--- proposal_quality events:
"kind": "proposal_quality", "candidate": "so_e01s01", "declared": false, "missing": ["mechanism", "hypothesis", "observable"], "mechanism": "", "hypothesis": "", "observable": "", "enforcement": "advisory"
--- sealed test ids leaked into work/ :
    test ids = [a8 a7]
    files in work/ containing a test id: 0
```

`diff -r` **rc=0 (identical)** for all three; `files in work/ containing a test id: 0` for all three.

---

### 3. The same three algorithms driven directly through the library loops

Adds the verbatim prompt text of the gate and the full `proposal_quality` event.

```

==============================================================================
### hill-climb  workdir=cand_0001
==============================================================================
--- find guidance/reasoning:
    guidance/reasoning/mechanism-probe/SKILL.md
    guidance/reasoning/mechanism-probe/meta.yaml
    guidance/reasoning/mechanism-probe/scripts/_bootstrap.py
    guidance/reasoning/mechanism-probe/scripts/check.py
    guidance/reasoning/mechanism-probe/scripts/run.py
--- diff -r vs source: rc=0 (identical)
--- composed prompt: 22893 chars / cap 60000 (38.2%)
    [x] #140 bar
    [x] #140 honest wording
    [x] #222 dead-ends
    [x] skill pointer
--- the bar, verbatim from the prompt:
    ## Proposal quality — declare the MECHANISM, not a knob (advisory bar)
    Read `./guidance/reasoning/mechanism-probe/SKILL.md` BEFORE you decide what to edit. It counters the failure mode this framework keeps paying for: skipping the analysis and shipping a plausible one-line tweak.
    Fill the **Proposal declaration** block in `./PROCESS.md` — three fields:
    - `Mechanism:` what in the system now behaves differently, and WHY that changes the outcome. A *knob* restates an existing rule or retunes a value the agent already ignores; a *mechanism* changes what is structurally possible (an in-code guard, a computation, a new/composite tool, a narrowed decision rule).
    - `Hypothesis:` which failure cluster this fixes and why it generalizes beyond the exact failing inputs.
    - `Expected observable:` the concrete change you expect in the NEXT iteration's trajectories if the hypothesis holds — something a reader could check.
    **How this is judged:** the declaration is RECORDED per candidate, not enforced — a missing or knob-shaped declaration does NOT reject your edit, and a well-declared one does not get it accepted. The val significance gate remains the only thing that accepts or rejects. Declare it anyway: an edit you cannot state a mechanism and an observable for is the edit that historically wasted the iteration.
    
--- proposal_quality events: [{"t": 1785447002.5906901, "kind": "proposal_quality", "candidate": "cand_0001", "declared": false, "missing": ["mechanism", "hypothesis", "observable"], "mechanism": "", "hypothesis": "", "observable": "", "enforcement": "advisory"}]
--- sealed test ids ['a8', 'a7'] in workdir files: 1 hits [('RUNMAP.md', 'a7')]

==============================================================================
### gepa  workdir=gepa_0001
==============================================================================
--- find guidance/reasoning:
    guidance/reasoning/mechanism-probe/SKILL.md
    guidance/reasoning/mechanism-probe/meta.yaml
    guidance/reasoning/mechanism-probe/scripts/_bootstrap.py
    guidance/reasoning/mechanism-probe/scripts/check.py
    guidance/reasoning/mechanism-probe/scripts/run.py
--- diff -r vs source: rc=0 (identical)
--- composed prompt: 23900 chars / cap 60000 (39.8%)
    [x] #140 bar
    [x] #140 honest wording
    [x] #222 dead-ends
    [x] skill pointer
--- the bar, verbatim from the prompt:
    ## Proposal quality — declare the MECHANISM, not a knob (advisory bar)
    Read `./guidance/reasoning/mechanism-probe/SKILL.md` BEFORE you decide what to edit. It counters the failure mode this framework keeps paying for: skipping the analysis and shipping a plausible one-line tweak.
    Fill the **Proposal declaration** block in `./PROCESS.md` — three fields:
    - `Mechanism:` what in the system now behaves differently, and WHY that changes the outcome. A *knob* restates an existing rule or retunes a value the agent already ignores; a *mechanism* changes what is structurally possible (an in-code guard, a computation, a new/composite tool, a narrowed decision rule).
    - `Hypothesis:` which failure cluster this fixes and why it generalizes beyond the exact failing inputs.
    - `Expected observable:` the concrete change you expect in the NEXT iteration's trajectories if the hypothesis holds — something a reader could check.
    **How this is judged:** the declaration is RECORDED per candidate, not enforced — a missing or knob-shaped declaration does NOT reject your edit, and a well-declared one does not get it accepted. The val significance gate remains the only thing that accepts or rejects. Declare it anyway: an edit you cannot state a mechanism and an observable for is the edit that historically wasted the iteration.
    
--- proposal_quality events: [{"t": 1785447003.814673, "kind": "proposal_quality", "candidate": "gepa_0001", "declared": false, "missing": ["mechanism", "hypothesis", "observable"], "mechanism": "", "hypothesis": "", "observable": "", "enforcement": "advisory"}]
--- sealed test ids ['a8', 'a7'] in workdir files: 1 hits [('RUNMAP.md', 'a7')]

==============================================================================
### skillopt  workdir=so_e01s01
==============================================================================
--- find guidance/reasoning:
    guidance/reasoning/mechanism-probe/SKILL.md
    guidance/reasoning/mechanism-probe/meta.yaml
    guidance/reasoning/mechanism-probe/scripts/_bootstrap.py
    guidance/reasoning/mechanism-probe/scripts/check.py
    guidance/reasoning/mechanism-probe/scripts/run.py
--- diff -r vs source: rc=0 (identical)
--- composed prompt: 23421 chars / cap 60000 (39.0%)
    [x] #140 bar
    [x] #140 honest wording
    [x] #222 dead-ends
    [x] skill pointer
--- the bar, verbatim from the prompt:
    ## Proposal quality — declare the MECHANISM, not a knob (advisory bar)
    Read `./guidance/reasoning/mechanism-probe/SKILL.md` BEFORE you decide what to edit. It counters the failure mode this framework keeps paying for: skipping the analysis and shipping a plausible one-line tweak.
    Fill the **Proposal declaration** block in `./PROCESS.md` — three fields:
    - `Mechanism:` what in the system now behaves differently, and WHY that changes the outcome. A *knob* restates an existing rule or retunes a value the agent already ignores; a *mechanism* changes what is structurally possible (an in-code guard, a computation, a new/composite tool, a narrowed decision rule).
    - `Hypothesis:` which failure cluster this fixes and why it generalizes beyond the exact failing inputs.
    - `Expected observable:` the concrete change you expect in the NEXT iteration's trajectories if the hypothesis holds — something a reader could check.
    **How this is judged:** the declaration is RECORDED per candidate, not enforced — a missing or knob-shaped declaration does NOT reject your edit, and a well-declared one does not get it accepted. The val significance gate remains the only thing that accepts or rejects. Declare it anyway: an edit you cannot state a mechanism and an observable for is the edit that historically wasted the iteration.
    
--- proposal_quality events: [{"t": 1785447005.03246, "kind": "proposal_quality", "candidate": "so_e01s01", "declared": false, "missing": ["mechanism", "hypothesis", "observable"], "mechanism": "", "hypothesis": "", "observable": "", "enforcement": "advisory"}]
--- sealed test ids ['a8', 'a7'] in workdir files: 1 hits [('RUNMAP.md', 'a7')]
```

---

### 4. Composed-prompt measurement — #219 + #221 + #222 + #140, and the overflow probe

#219's INSIGHTS pointer bullet is taken verbatim from `origin/feat/issue-128-persist-insight`; #221's diversify block is produced by its own `plateau.prompt_block()` from `origin/feat/issue-130-plateau-detection` at its bounded widest (39 exhausted lineages → bounded to 6); #222's dead-end block is rendered live at 200 rejections.

```
rendered template + #222 dead-ends (200 rejections) + #140 bar  :  24416
  of which #140's PROMPT_BLOCK                                  :   1320
  of which #222's dead-end block @ 200 rejections               :   2251
+ #219's INSIGHTS pointer bullet                                :    463
+ #221's diversify block (39 lineages -> bounded to 6)          :    887
-------------------------------------------------------------------------------
COMPOSED TOTAL (#219 + #221 + #222 + #140)                      :  25766
CAP (MAX_INSTRUCTIONS_CHARS)                                    :  60000
headroom                                                        :  34234  (42.9% of cap)

overflow probe: _augment_instructions(110000 chars) -> 59906 chars; <= cap: True
  #140's block survives WHOLE in the kept tail : True
  #222's dead-end block survives WHOLE too     : True
  elision notice present (nothing silently cut): True
```

**25,766 / 60,000 = 42.9% of the cap**, 34,234 chars of headroom. #140's own contribution is **1,320 chars** — well inside the kept 30% tail (18 KB), which is why the overflow probe shows both it and #222's block surviving **whole**. The overflow probe crosses the bound with a real 110,000-char input, not a fixture that never overflows (the #219 lesson).

---

### 5. The new test file, verbose — including the false-rejection probe

```
cachedir: .pytest_cache
rootdir: /private/tmp/wt-140/core
configfile: pyproject.toml
plugins: anyio-4.14.2
collecting ... collected 19 items

core/tests/test_proposal_quality.py::test_a_filled_declaration_parses_all_three_fields PASSED [  5%]
core/tests/test_proposal_quality.py::test_the_seeded_placeholders_count_as_missing PASSED [ 10%]
core/tests/test_proposal_quality.py::test_partial_and_placeholder_values_are_reported_field_by_field PASSED [ 15%]
core/tests/test_proposal_quality.py::test_markup_and_marker_variants_still_parse PASSED [ 21%]
core/tests/test_proposal_quality.py::test_a_missing_process_md_never_raises PASSED [ 26%]
core/tests/test_proposal_quality.py::test_a_genuine_mechanism_proposal_is_not_rejected PASSED [ 31%]
core/tests/test_proposal_quality.py::test_an_undeclared_proposal_is_also_not_rejected PASSED [ 36%]
core/tests/test_proposal_quality.py::test_the_bar_states_it_is_recorded_not_enforced PASSED [ 42%]
core/tests/test_proposal_quality.py::test_the_declaration_is_recorded_per_candidate PASSED [ 47%]
core/tests/test_proposal_quality.py::test_the_declaration_is_surfaced_in_the_dashboard PASSED [ 52%]
core/tests/test_proposal_quality.py::test_hill_climb_gets_the_reasoning_skill PASSED [ 57%]
core/tests/test_proposal_quality.py::test_gepa_gets_the_reasoning_skill PASSED [ 63%]
core/tests/test_proposal_quality.py::test_skillopt_gets_the_reasoning_skill PASSED [ 68%]
core/tests/test_proposal_quality.py::test_the_reasoning_skill_is_not_mistaken_for_a_capability_edit PASSED [ 73%]
core/tests/test_proposal_quality.py::test_the_composed_prompt_is_measured_and_under_the_cap PASSED [ 78%]
core/tests/test_proposal_quality.py::test_an_overflowing_prompt_keeps_the_bar_whole PASSED [ 84%]
core/tests/test_proposal_quality.py::test_the_bar_is_capped_by_the_shared_cap_not_a_second_one PASSED [ 89%]
core/tests/test_proposal_quality.py::test_no_test_split_id_reaches_any_injected_workdir_file PASSED [ 94%]
core/tests/test_proposal_quality.py::test_the_gate_makes_no_model_call PASSED [100%]

============================= 19 passed in 12.52s ==============================
```

The three that matter for the "risky half":
- `test_a_genuine_mechanism_proposal_is_not_rejected` — **false-rejection probe**: a declared, genuinely-improving mechanism edit is ACCEPTED, and the gate reason contains none of `mechanism`/`knob`/`declar`/`proposal quality`.
- `test_an_undeclared_proposal_is_also_not_rejected` — the symmetric half: no declaration at all is also accepted on its val delta.
- `test_the_bar_states_it_is_recorded_not_enforced` — pins the honest wording, so nobody can quietly upgrade the claim.

---

### 6. Fail-before / pass-after

Test file kept, implementation stashed (`git stash -u`, then the test file copied back):

```
$ PYTHONPATH=/tmp/wt-140/core python -m pytest core/tests/test_proposal_quality.py -q
...
E       ImportError: cannot import name 'proposal_quality' from 'cap_evolve'
FAILED core/tests/test_proposal_quality.py::test_a_filled_declaration_parses_all_three_fields
FAILED core/tests/test_proposal_quality.py::test_the_seeded_placeholders_count_as_missing
FAILED core/tests/test_proposal_quality.py::test_partial_and_placeholder_values_are_reported_field_by_field
FAILED core/tests/test_proposal_quality.py::test_markup_and_marker_variants_still_parse
FAILED core/tests/test_proposal_quality.py::test_a_missing_process_md_never_raises
FAILED core/tests/test_proposal_quality.py::test_an_undeclared_proposal_is_also_not_rejected
FAILED core/tests/test_proposal_quality.py::test_the_bar_states_it_is_recorded_not_enforced
FAILED core/tests/test_proposal_quality.py::test_the_declaration_is_recorded_per_candidate
FAILED core/tests/test_proposal_quality.py::test_hill_climb_gets_the_reasoning_skill
FAILED core/tests/test_proposal_quality.py::test_gepa_gets_the_reasoning_skill
FAILED core/tests/test_proposal_quality.py::test_skillopt_gets_the_reasoning_skill
FAILED core/tests/test_proposal_quality.py::test_the_reasoning_skill_is_not_mistaken_for_a_capability_edit
FAILED core/tests/test_proposal_quality.py::test_the_composed_prompt_is_measured_and_under_the_cap
FAILED core/tests/test_proposal_quality.py::test_an_overflowing_prompt_keeps_the_bar_whole
FAILED core/tests/test_proposal_quality.py::test_the_gate_makes_no_model_call
15 failed, 3 passed in 9.80s
```

(That run predates the dashboard test, hence 18 collected; with it, 19.) After `git stash pop`: **19 passed**.

---

### 7. Full suite, including the #200-flaky dashboard test

```
$ PYTHONPATH=/tmp/wt-140/core python -m pytest core/tests -q
........................................................................ [ 30%]
........................................................................ [ 60%]
........................................................................ [ 90%]
........................                                                 [100%]
240 passed in 99.76s (0:01:39)
```

**240 passed, 0 failed.** 214 (base) + 19 (new) + 7 (`test_dashboard_launch`, which passed on this machine — port 7878 was free).

---

### 8. `compileall`

```
$ find . -name __pycache__ -type d -exec rm -rf {} +
$ python -m compileall -q core/cap_evolve core/tests skills
compileall rc=0
```

---

### 9. `build_manifest.py`

```
$ PYTHONPATH=/tmp/wt-140/core python skills/_registry/build_manifest.py skills
wrote skills/_registry/manifest.json (21 skill(s))
  optimizer: run-optimizer
  reasoning: mechanism-probe
```

21 skills, 0 validation errors. The new `reasoning` component is added to `build_manifest.COMPONENTS`, so a typo in it still fails the build loudly.

---

### 10. #213's skill authoring lint

`lint_skills.py` taken verbatim from `origin/feat/issue-107-skill-lint-ci` and pointed at this tree:

```
$ python /tmp/lint_skills.py /tmp/wt-140/skills
skill authoring lint — 21 skill package(s) under /private/tmp/wt-140/skills
  ERROR   capabilities/tools: SKILL.md body is 673 lines (>500); split detail into references/ (progressive disclosure)
  ERROR   capabilities/tools: SKILL.md body is ~10934 tokens (>5000); it is a recurring per-session cost — move detail into references/
  ERROR   orchestrate/using-cap-evolve: description must not contain XML tags
  advise  algorithms/evograph: description should say WHEN to use the skill ('Use when …') — it is the primary triggering signal
  advise  algorithms/hill-climb: description should say WHEN to use the skill ('Use when …') — it is the primary triggering signal
  advise  optimizers/run-optimizer: description should say WHEN to use the skill ('Use when …') — it is the primary triggering signal
  advise  orchestrate/using-cap-evolve: description should say WHEN to use the skill ('Use when …') — it is the primary triggering signal
  advise  phases/baseline: description should say WHEN to use the skill ('Use when …') — it is the primary triggering signal
  advise  phases/diagnose: description should say WHEN to use the skill ('Use when …') — it is the primary triggering signal
  advise  phases/finalize: description should say WHEN to use the skill ('Use when …') — it is the primary triggering signal
  advise  phases/gate: description should say WHEN to use the skill ('Use when …') — it is the primary triggering signal
  advise  phases/implement-and-check: description should say WHEN to use the skill ('Use when …') — it is the primary triggering signal
  advise  phases/intake: description should say WHEN to use the skill ('Use when …') — it is the primary triggering signal
  advise  phases/report: description should say WHEN to use the skill ('Use when …') — it is the primary triggering signal
FAIL — 3 authoring violation(s) in 2 skill(s)
```

`reasoning/mechanism-probe` has **zero errors and zero advisories** — including the "say WHEN" description heuristic, and #213's `_manifest_drift` path-set check agrees (21 on disk == 21 in `manifest.json`).

The 3 remaining errors are **pre-existing on this base**, proved by stashing my change:

```
$ git stash -u && python /tmp/lint_skills.py /tmp/wt-140/skills | grep ERROR
  ERROR   capabilities/tools: SKILL.md body is 673 lines (>500); split detail into references/ (progressive disclosure)
  ERROR   capabilities/tools: SKILL.md body is ~10934 tokens (>5000); it is a recurring per-session cost — move detail into references/
  ERROR   orchestrate/using-cap-evolve: description must not contain XML tags
  ERROR   discovery: 21 skill package(s) on disk do not match the 20 in manifest.json ...
```

Those first three are exactly what #213 fixes on its own branch (it splits `tools/SKILL.md` and rewrites the `using-cap-evolve` description). Nothing here regresses the lint.

---

### 11. Every skill's `check.py`

```
$ for f in skills/*/*/scripts/check.py; do (cd $(dirname $f) && python check.py); done
rc=0 "ok": True   skills/algorithms/agent-optimize/scripts/check.py
rc=0 "ok": True   skills/algorithms/evograph/scripts/check.py
rc=0 "ok": True   skills/algorithms/gepa/scripts/check.py
rc=0 "ok": True   skills/algorithms/hill-climb/scripts/check.py
rc=0 "ok": True   skills/algorithms/skillopt/scripts/check.py
rc=0 "ok": True   skills/capabilities/mcp-tool/scripts/check.py
rc=0 "ok": True   skills/capabilities/skill-package/scripts/check.py
rc=0 "ok": True   skills/capabilities/system-prompt/scripts/check.py
rc=0 "ok": True   skills/capabilities/tools/scripts/check.py
rc=0 "ok": True   skills/optimizers/run-optimizer/scripts/check.py
rc=0 "ok": True   skills/orchestrate/orchestrate/scripts/check.py
rc=0 "ok": True   skills/orchestrate/using-cap-evolve/scripts/check.py
rc=0 "ok": True   skills/phases/baseline/scripts/check.py
rc=0 "ok": True   skills/phases/diagnose/scripts/check.py
rc=0 "ok": True   skills/phases/evaluate/scripts/check.py
rc=0 "ok": True   skills/phases/finalize/scripts/check.py
rc=0 "ok": True   skills/phases/gate/scripts/check.py
rc=0 "ok": True   skills/phases/implement-and-check/scripts/check.py
rc=0 "ok": True   skills/phases/intake/scripts/check.py
rc=0 "ok": True   skills/phases/report/scripts/check.py
rc=0 "ok": True   skills/reasoning/mechanism-probe/scripts/check.py
```

**21 / 21 print `"ok": true` and exit 0**, including the new one. `mechanism-probe`'s own check, in full:

```
{
  "skill": "mechanism-probe",
  "ok": true,
  "problems": [],
  "notes": [
    "run entry exposes main()",
    "all three declared fields parse from PROCESS.md",
    "false-rejection probe: a real mechanism proposal is never rejected",
    "declaration is recorded per candidate as advisory",
    "unfilled placeholders count as missing (no vacuous pass)"
  ]
}
rc=0
```

Note `false-rejection probe: a real mechanism proposal is never rejected` — the probe is a shipped behavioral contract, not only a pytest.

---

### 12. No sealed-test leak

Per §2 and §3 above: for every algorithm, `grep` for each sealed test id across every file in the optimizer workdir → **0 hits**. `test_no_test_split_id_reaches_any_injected_workdir_file` pins it in CI and asserts the test-id list is non-empty first, so the probe cannot pass vacuously.

The injected subtree is also invisible to the three read-side filters (`guidance/` is already in #199's `INJECTED_DIRS`):

```
$ python -m pytest core/tests/test_proposal_quality.py::test_the_reasoning_skill_is_not_mistaken_for_a_capability_edit -q
1 passed
```
— i.e. `hash_candidate_dir(a) == hash_candidate_dir(b)` and `_components(a) == _components(b)` with the skill present in one and not the other. Without that, every candidate would look edited and the eval cache would never hit.

---

### 13. Zero LLM calls

```
$ python -m pytest core/tests/test_proposal_quality.py::test_the_gate_makes_no_model_call -q
1 passed
```
Asserts none of `anthropic`, `openai`, `requests`, `urllib.request`, `http.client`, `aux_model`, `subprocess` appear in `core/cap_evolve/proposal_quality.py`. Pure stdlib `re` over a markdown file — #205's invariant that every auxiliary step in core is pure Python is intact, and the cost profile of a run is unchanged.

---

### 14. Commit authorship

```
$ git log --format="%h %an <%ae>%n%s" -1
2cb4d1dc Osher Elhadad <Osher.Elhadad@ibm.com>
feat(algorithm): on-demand reasoning skills + an advisory proposal-quality declaration (#140)

$ git log -1 --format=%B | grep -ci "co-authored-by\|generated with"
0
0
```

Authored and committed as `Osher Elhadad <Osher.Elhadad@ibm.com>`; no `Co-Authored-By` and no "Generated with" line.

OsherElhadad:
## 🔍 Review — PR #246

**Verdict: APPROVE WITH NITS.**

The central claim holds. I traced every consumer of the `proposal_quality` signal and re-ran every claim in the evidence section. "Advisory" is true in the code, not just in prose: nothing branches on the parsed declaration, at any acceptance site, in any algorithm. The composed prompt stays capped with all four blocks at their bounded widest and nothing is silently dropped. The skill reaches all three algorithms byte-identically with the cache hash and component list unchanged. Two real defects below, both non-blocking because neither can affect a run's honesty; the first is a broken documented command.

---

### Blocking

None.

---

### Non-blocking

**1. `skills/reasoning/mechanism-probe/SKILL.md:115` — the documented self-check command cannot run from the optimizer's workdir, and neither can the injected copy.**

The SKILL.md tells the optimizer:

```bash
python skills/reasoning/mechanism-probe/scripts/run.py --process ./PROCESS.md
```

The optimizer's cwd is the workdir, which has no `skills/` tree — only `guidance/reasoning/mechanism-probe/`. Reproduced from a real hill-climb workdir:

```
$ cd .../run_hill/work/cand_0001
$ python skills/reasoning/mechanism-probe/scripts/run.py --process ./PROCESS.md
can't open file '.../work/cand_0001/skills/reasoning/mechanism-probe/scripts/run.py': [Errno 2] No such file or directory
```

And the path that *does* exist fails too, because `run.py` imports `cap_evolve.proposal_quality` and the bootstrap's upward walk from `work/<cid>/guidance/reasoning/mechanism-probe/scripts/` never reaches a `core/`:

```
$ python guidance/reasoning/mechanism-probe/scripts/run.py --process ./PROCESS.md
ImportError: cannot import name 'proposal_quality' from 'cap_evolve'
```

(That import resolved to my editable install of `main`'s core — i.e. the walk found *some* `cap_evolve`, just not this branch's. `harness.optimizer_from_command` (`core/cap_evolve/harness.py:455`) passes `dict(os.environ)` through, so `CAPEVOLVE_CORE` reaches the subprocess only when the caller already had it exported; a plain `cap-evolve run` does not set it.)

**Consequence:** the only executable artifact in the new skill is dead on arrival for the agent that is told to run it. The PR's own comment ("the probe's own `run.py` is what the SKILL.md tells the optimizer to run on its PROCESS.md", `core/cap_evolve/harness.py:1271`) is the stated reason `scripts/` is kept in the injected copy — so the copy exists specifically to serve a command that cannot execute. Every `check.py` passes because it runs from the repo, where both problems are absent.

**Fix (pick one):** either change line 115 to the workdir-relative path (`python ./guidance/reasoning/mechanism-probe/scripts/run.py --process ./PROCESS.md`) and export `CAPEVOLVE_CORE` in `optimizer_from_command`'s env; or drop `scripts/` from the injected copy (add `"scripts"` to the existing `ignore_patterns` at `core/cap_evolve/harness.py:1281`, matching the native copy which already excludes it) and delete the "Checking your own declaration" section. The second is smaller and loses nothing the prompt needs — presence-checking is exactly what the harness already logs.

**2. `core/cap_evolve/proposal_quality.py:51` — `_field` takes the FIRST match in the file, so an appended declaration is shadowed by the seed's placeholder above it.**

`re.search` with `re.M` returns the earliest match. The seed (`core/cap_evolve/harness.py:569-577`) writes `- Mechanism: <what now behaves DIFFERENTLY...>` near the top. An agent that *appends* a filled block instead of editing in place records as fully undeclared:

```
fills in place:      missing=['hypothesis', 'observable']   (only the field it edited parses)
appends below seed:  missing=['mechanism','hypothesis','observable']  declared=False
                     -> mechanism=''
```

**Consequence:** advisory-only, so no run outcome changes — but the recorded signal is wrong in a plausible-and-common agent behaviour, and the dashboard row then reads "no mechanism declaration" for a candidate that declared one properly. That inverts the one thing the feature exists to observe. It also means the "declared" rate this feature is meant to surface will read artificially low.

**Fix:** prefer the LAST match (`m = None; for m in re.finditer(...): pass`), or scope the search to the text after the last `## Proposal declaration` heading. Either is a two-line change. Add a test with `_PROCESS_SEED + <filled block>` as the fixture — the existing `test_the_seeded_placeholders_count_as_missing` uses the seed alone, so it cannot catch this.

**3. `core/tests/test_proposal_quality.py:141-172` — the false-rejection probe covers hill-climb only; GEPA's local gate is unguarded by #140's own tests.**

I injected an enforcement into GEPA's local gate:

```python
        local_pass = _sum_reward(child_mb) > _sum_reward(parent_mb)
        if not proposal_quality.parse(workdir)["declared"]:
            local_pass = False  # PROBE-INJECTED ENFORCEMENT
```

All 19 of #140's tests still pass:

```
$ PYTHONPATH=/tmp/rv-246/core python -m pytest core/tests/test_proposal_quality.py -q
19 passed in 11.65s
```

It *is* caught, but by pre-existing tests in a different file (`core/tests/test_gepa.py`: 5 failures — `test_gepa_loop_accepts_and_seals_test`, `test_per_instance_frontier_used`, both merge tests, `test_candidate_snapshots_are_clean_for_every_algorithm[gepa]`). So the invariant is protected today, incidentally, by tests that would also break for a dozen unrelated reasons. `test_gepa_gets_the_reasoning_skill` (line 294) asserts injection and the presence of an event; it never asserts the candidate was accepted.

**Consequence:** if someone later "upgrades" #140 to gate GEPA locally, its own test file green-lights it and the failure surfaces as five confusing GEPA test failures rather than "the advisory gate became enforcing".

**Fix:** add one assertion to `test_gepa_gets_the_reasoning_skill` — the undeclared candidate passed the local gate (`_events(run_dir, "gepa_local_gate")[-1]["passed"] is True`), mirroring the hill-climb probe.

---

### Nits

- `core/cap_evolve/proposal_quality.py:41` — `_PLACEHOLDER_RE`'s `[-–—.]*` alternative matches the empty string, so the branch is redundant with the `^\s*...\s*$` anchoring; harmless, but it makes the pattern read as if it rejects more than it does.
- `core/cap_evolve/proposal_quality.py:38` — `expected\s+observable` requires "Expected"; a bare `Observable:` records as missing. Verified. Arguably correct (the seed says "Expected observable"), but it's a brittleness the docstring at line 47 claims tolerance for.
- One-character values count as declared (`Mechanism: a` → `declared=True`). Consistent with "presence only, by design", just worth knowing the signal's floor.

---

### Is "advisory" true in the code?

Yes. Every path traced:

**Producers.** The only two call sites of `record()` are `core/cap_evolve/harness.py:1508` (in `run_step`, between `update_spent` and `evaluate_candidate`) and `core/cap_evolve/gepa.py:638` (after the optimizer returns, before the local-gate eval). Both discard the return value — the call is a bare expression statement, so there is no variable for a later branch to read even by accident. `grep -rn "proposal_quality"` over the whole repo returns 22 hits and no third producer.

**Acceptance sites, each checked:**
- `gate.decide` (`core/cap_evolve/harness.py:1520`) — arguments are `current_val.reward`, `cand_val.reward`, stderrs, `paired_deltas`, `run_dir`, `**gate_kwargs`. No declaration data in any of them.
- `run_step`'s accept branch (`harness.py:1531`) — `accepted = decision.accept`, then only the `no_regression` per-task check can flip it. Nothing else.
- GEPA's local gate (`gepa.py:643`) — `local_pass = _sum_reward(child_mb) > _sum_reward(parent_mb)`. Rewards only.
- GEPA's val gate (`gepa.py:670`, `_full_val_gate`) — same shape as `run_step`'s.
- GEPA's merge path (`gepa.py:787 _try_merge`) — routes to `_full_val_gate`; no `record` call, no read.
- SkillOpt — both the per-step path (`skillopt.py:306`) and the epoch-boundary slow update (`skillopt.py:500`) delegate to `harness.run_step`, so they inherit the same non-branching record and nothing else.
- `plateau` (#221) — reads `iteration_events()` only.
- `insights` (#219) / dead-end memory (#222) — `rejected.add(...)` is fed `decision.reason`, `cand_val.reward` and `approach_signature(parent_dir, workdir)`. `approach_signature` (`harness.py:918`) is built from `_diff_capabilities`, which filters `_CAP_DIFF_SKIP_DIRS ⊇ INJECTED_DIRS ∋ guidance` — so neither the declaration nor the injected skill can enter a dead-end signature.

**Event-consumer surface.** `RunDir.iteration_events()` (`rundir.py:469`) filters on `ITERATION_EVENT_KINDS = ("step", "skillopt_step", "gepa_val_gate")` (`rundir.py:81`). `proposal_quality` is not in that tuple, so the LEDGER (`harness.py:746`), RUNMAP/`prior_iterations` (`harness.py:846`) and `_parent_map` (`harness.py:669`) cannot see it. The only reader anywhere is `dashboard.py:284`, which appends to the read-only `diagnoses` list.

**Wording-pinning test — does it fail on a reword?** Yes, and it fails on the *substantive* claim, not a stray phrase. I rewrote `PROMPT_BLOCK`'s judgement paragraph into an enforcing claim while preserving structure:

```
- "**How this is judged:** the declaration is RECORDED per candidate, not enforced — ..."
+ "**How this is judged:** the declaration is ENFORCED per candidate — a missing or
+  knob-shaped declaration WILL reject your edit before it reaches evaluation."

4 failed, 15 passed
FAILED test_the_bar_states_it_is_recorded_not_enforced
FAILED test_hill_climb_gets_the_reasoning_skill
FAILED test_gepa_gets_the_reasoning_skill
FAILED test_skillopt_gets_the_reasoning_skill
```

Four tests, not one — and three of them fail because `_assert_probe_in_workdir` (line 277) checks the wording in the *rendered prompt*, not in the constant. That is the right ground truth: it would also catch someone who kept the honest constant but stopped appending it. This is not #244's "Honest limits" failure shape.

Better still, the behavioural half is guarded too. I injected a real enforcement into `run_step` while leaving the honest wording intact:

```python
    accepted = decision.accept
    if not _q["declared"]:
        accepted = False

FAILED test_an_undeclared_proposal_is_also_not_rejected
    AssertionError: an UNDECLARED proposal was rejected — the gate is not advisory
```

So the claim is pinned by wording *and* by behaviour. The gap is GEPA's local gate (non-blocking #3).

**Is the framing honest and useful?** Honest: yes, verified above, and stated in the prompt, the SKILL.md, the module docstring and `docs/ARCHITECTURE.md:53` in the same terms. No surface overclaims — I grepped `docs/`, `README.md`, `CHANGELOG.md`, `RUN.md`, `CONTRIBUTING.md`; the only mention is the ARCHITECTURE.md block, which says "recorded, never enforced". The dashboard is careful in both branches: declared → `"declared mechanism: … | expected observable: …"`, undeclared → `"no mechanism declaration (…) — advisory, the val gate still decided this candidate"`. It rides the existing `diagnoses` stream, is rendered as plain text in both the static HTML (`dashboard.py:1280`) and the React `Insights.tsx:110`, and never sits near an accept/reject badge. `reduce_run`'s whole return passes through `redact` (`dashboard.py:593`), so a declaration that quotes a secret is scrubbed.

Useful: the claim "an edit you cannot state a mechanism for is the edit that historically wasted the iteration" is an **assertion, not a finding**. The SKILL.md gestures at evidence ("Prior runs in this repo lost iterations exactly this way (see any run's `LEDGER.md`)") but cites no run, no number, no committed artifact. Nothing in-repo measures knob-vs-mechanism edit outcomes, and nothing could — the `proposal_quality` event this PR adds is the first data that *would* let you test it, and it has zero rows so far. So: the mechanism is honest and cheap (1,320 chars, ~2% of the prompt budget, zero API calls, zero runtime risk), the rationale is unvalidated, and the feature is the instrument you'd need to validate it. That is a defensible order of operations, but the PR should say "unvalidated hypothesis; this event is how we'd test it" rather than "historically wasted", which reads as though a measurement exists. A one-line honesty edit to the SKILL.md's opening claim would close it.

---

### Composed prompt at the boundary

**PR's overflow probe, reproduced exactly:**

```
probe 110000 -> 59906  <=cap True  block whole True  notice True
```

It genuinely crosses the bound (input 110,000 > cap 60,000 → the elision branch fires, notice present). Not #219's vacuous-fixture shape.

**Boundary sweep** through the real `_augment_instructions`:

```
  in= 57000 out= 59284 <=cap= True whole= True notice=False
  in= 58500 out= 59904 <=cap= True whole= True notice= True   <- first crossing
  in= 58600 out= 59905 <=cap= True whole= True notice= True
  in= 59999 out= 59905 <=cap= True whole= True notice= True
  in= 60000 out= 59905 <=cap= True whole= True notice= True
  in= 60001 out= 59905 <=cap= True whole= True notice= True
  in= 61000 out= 59905 <=cap= True whole= True notice= True
```

No overshoot at any input size, unlike #219's original 98-char miss. The arithmetic is sound because `cap_instructions` truncates *and* inserts the notice in one expression: `keep = max_chars - 200 = 59,800`; head 41,860 + tail 17,940 + notice 107 = **59,907 ≤ 60,000**, with 93 chars of slack. The 200-char reserve genuinely covers the notice.

**All four blocks at their bounded widest, simultaneously.** I saturated `rejected.jsonl` with 400 records at `_MAX_APPROACH_CHARS`/200-char reasons to force #222's block to its ceiling, built #221's diversify block at 6 truncated lineages + "(+33 more)", and #219's INSIGHTS bullet verbatim from `origin/feat/issue-128-persist-insight`:

```
#222 dead-ends @widest :   7123
#221 diversify @widest :   1194
#219 INSIGHTS bullet   :    463
#140 bar               :   1320
SUM of the four        :  10100  (tail budget 17940)

base= 20000 -> out= 31065 <=cap=True |#140 whole=True |#222 whole=True |#221 whole=True |#219 whole=True |notice=False
base= 45000 -> out= 56065 <=cap=True |#140 whole=True |#222 whole=True |#221 whole=True |#219 whole=True |notice=False
base= 59000 -> out= 59906 <=cap=True |#140 whole=True |#222 whole=True |#221 whole=True |#219 whole=True |notice= True
base= 60000 -> out= 59906 <=cap=True |#140 whole=True |#222 whole=True |#221 whole=True |#219 whole=True |notice= True
```

All four survive **whole** even when the head is over-cap and the elision fires — 10,100 of 17,940 tail chars, 7,840 to spare. Nothing is cut mid-structure and nothing is silently dropped (the notice is always present when elision happens). Note the tail is now 56% consumed at the widest; a fifth block over ~7.8 KB would start eating #219's bullet, which sorts first in the tail. Worth a comment for whoever adds the next one, not a change here.

**Double-cap check.** Feeding an already-capped 59,907-char render into `_augment_instructions` produces one notice, not two: `after augment: 59905, notices: 1, #140 whole: True`. And the single-cap invariant is pinned by `test_the_bar_is_capped_by_the_shared_cap_not_a_second_one` (line 378), which asserts `src.count("cap_instructions") == 1` over `inspect.getsource`. Verified there is no second cap: `harness.py:1092` is the only `_oc.cap_instructions` in `_augment_instructions`, and the sibling branches add none (#219 returns unwrapped; #221 routes through `render_instructions(extra=...)`).

**Composed-tree caveat:** `git merge origin/feat/issue-128-persist-insight` into #246 conflicts in `harness.py` (3 hunks), `dashboard.py` and `skillopt.py` — see merge-order note. My all-four measurement therefore composes the sibling blocks at their real widths through #246's real `_augment_instructions` rather than through a merged tree. The arithmetic is unaffected (all four land in the same tail, and `cap_instructions` is a pure function of length), but the merged tree should be re-measured once the conflicts are resolved.

---

### Injection safety

**Eval-cache hash: unchanged.** `hash_candidate_dir` (`cache.py:73-76`) skips any path with a `guidance` component via `_IGNORE_DIRS ⊇ INJECTED_DIRS`. Crucially this is a **directory-part** check (`any(part in _IGNORE_DIRS for part in rel.parts)`), not the root-anchored basename check that PR #211 fixed — so the depth of the injected subtree is irrelevant. Verified empirically by `test_the_reasoning_skill_is_not_mistaken_for_a_capability_edit` (line 316), which copies the whole `mechanism-probe` tree into `b/guidance/reasoning/` and asserts hash equality. That test is one of only 4 that pass on the base — i.e. it pins a pre-existing invariant rather than new behaviour, which is exactly right for a regression guard.

**GEPA component list: unchanged.** `_NON_COMPONENT_DIRS` (`gepa.py:83`) is likewise `{".git","__pycache__"} | set(oc.INJECTED_DIRS)`, checked per path part. Same test asserts `_components(a) == _components(b)`. So an injected read-context file cannot become an editable component and the optimizer cannot rewrite its own instructions. Both sets derive from the one `INJECTED_DIRS` tuple (`optimizer_context.py:62`), so no new enumeration was added.

**Snapshot: excluded.** `"guidance" in _SNAPSHOT_IGNORE` asserted at line 332; both `run_step` (`harness.py:1546`) and GEPA's accept branch pass it, so candidates stay capability-only.

**Sealed-test leak: 0.** Real zero-API runs, all three algorithms, greping every readable file in each workdir:

```
### hill-climb  sealed test ids ['a8','a7'] in workdir files: 0 []
### gepa        sealed test ids ['a8','a7'] in workdir files: 0 []
### skillopt    sealed test ids ['a8','a7'] in workdir files: 0 []
```

Cleaner than the PR's own library-level run, which reported `1 hits [('RUNMAP.md','a7')]` — that is a substring collision (a candidate id containing `a7`), not a leak, and the shipped test (line 393) uses the same substring method and finds none. Either way 0.

**Tamper guard (#197/#142): cannot false-fire.** `protect.project_dir_for` scopes to `.capevolve/project`, `resolve_protected` takes `exclude=<run dir>` and returns early for anything under it, and `_DEFAULT_GLOBS` is `("adapters", "capevolve.yaml", *gold*)`. The injected subtree lives at `run_dir/work/<cid>/guidance/reasoning/` — disjoint from the protected set by construction, and inside the excluded run dir besides.

**`diff -r` rc=0 for all three, reproduced:**

```
### hill-climb  workdir=cand_0001   diff -r vs source: rc=0 (identical)
    prompt 22164 / cap 60000 (36.9%)   bar present: True; honest wording: True
### gepa        workdir=gepa_0001   diff -r vs source: rc=0 (identical)
    prompt 23171 / cap 60000 (38.6%)   bar present: True; honest wording: True
### skillopt    workdir=so_e01s01   diff -r vs source: rc=0 (identical)
    prompt 22692 / cap 60000 (37.8%)   bar present: True; honest wording: True
```

**Native placement verified separately** — the `mock` optimizer has no `skills_dir` row, so my run-based probe only saw the `guidance/` copy. Driving `_inject_native_skills` with `claude-code` confirms the second placement, and that the native copy correctly excludes `scripts/` (the shared `ignore` at `harness.py:1339`):

```
 DIR .claude/skills/diagnose
 DIR .claude/skills/mechanism-probe
 DIR .claude/skills/tools
files under mechanism-probe: SKILL.md, meta.yaml
```

The `glob("*/SKILL.md")` loop (`harness.py:1365`) means a second reasoning skill needs no wiring, matching the claim.

---

### Is a new component type warranted?

**Marginally — I'd keep it, but the justification in the code is stronger than the justification in the PR.**

For it: `reasoning` is genuinely a different *kind* of thing. Every other component is either sequenced by the orchestrate DAG (`phase`), selected as the optimization target (`capability`), chosen as the loop (`algorithm`), resolved as the proposer (`optimizer`), or the driver itself. `mechanism-probe` is none of those: it is injected read-context at one step, has `provides: []`, and is never invoked as a subprocess by the framework. Filing it under `phase` would make it eligible for DAG sequencing it must never enter; filing it under `capability` would make it look like an optimization target. The comment at `build_manifest.py:25-29` says exactly this, and it's correct.

The cost is genuinely one line — `COMPONENTS` gained one string. `build_manifest.py` handles it with no other change (it reads `component` from `meta.yaml` and groups); the count is right; #213's `_manifest_drift` uses a `*/*/SKILL.md` glob and a path-set comparison, so it picks the new tree up with no bump; nothing in `core/` iterates the manifest by component (`cli.py:136` resolves by name only).

Against it: one member, and the PR's own admission that it skipped `first-principles` because "nothing would load it" concedes the plural is aspirational. A component type whose population is 1 and whose second member has no loader is a taxonomy built for a future that may not arrive. But the alternative — mislabeling it as a phase — is worse than a slightly-early category, and unwinding it later is a one-line revert plus a manifest rebuild. Keep it.

The issue asked for `first-principles` too. Skipping it is the right call for the reason given, and the PR is honest about it. Do not add it until a loader exists.

---

### Merge-order note

Recommended: **#199 → #213 → #222 → #246 → #219 → #221.**

- **#199** first, unconditionally — it owns `INJECTED_DIRS`, `cap_instructions`, `MAX_INSTRUCTIONS_CHARS`. Every other PR here derives from it.
- **#213** early and before #246. #213 is the only branch on which the skill lint passes clean (`OK — 20 skill packages pass`); #246 lands on a base with 3 pre-existing lint errors. Merging #213 second means #246 arrives into green CI. **#213 must be rebased after #246 lands** to regenerate the manifest — its CI step now runs `build_manifest.py` followed by `git diff --exit-code`, so #246's 21st skill would fail that gate on a stale #213.
- **#222** before #246 — #246 is cut from it (`6d6cc52d`), so this is forced. #246 merges into #222 as a fast-forward.
- **#246** next. It is the smallest of the prompt-appending siblings (1,320 chars) and touches `_augment_instructions`'s tail with one `+=`, so it is the cheapest to rebase over.
- **#219 then #221 last.** #219 conflicts with #246 in three files today (`harness.py` ×3 hunks, `dashboard.py`, `skillopt.py`) — mostly the shared `_augment_instructions` pointer block, which both PRs rewrite, and `_build_ledger`'s signature, which #219 changes to take `rejected, history`. Whichever of {#219, #221, #246} merges last pays the rebase; #246 should not be it, because its rebase is the one that has to re-verify a cap measurement. After the composed merge, re-run the all-four-at-max probe on the merged tree.

Note the PR body says its base is 214 tests; the actual tip of `origin/feat/issue-129-failure-memory` (`6d6cc52d`) is **221** — the 214 figure predates #222's own review fixes. 221 + 19 = 240, so the arithmetic is consistent; the stated base number is just stale.

---

### Verification I re-ran

**Full suite on the branch (240, as claimed):**
```
$ cd /tmp/rv-246 && PYTHONPATH=/tmp/rv-246/core python -m pytest core/tests -q
240 passed in 102.42s (0:01:42)
```

**Full suite on the true base #222 (`6d6cc52d`) — 221, not the claimed 214:**
```
$ cd /tmp/rv-222 && git log --oneline -1
6d6cc52d fix(algorithm): make the rejected-approach signature reflect the real edit (#129 review)
$ PYTHONPATH=/tmp/rv-222/core python -m pytest core/tests -q
221 passed in 90.37s (0:01:30)
```
(`da9ad44d`, the commit before that review fix, is 213. The PR's 214 matches neither; harmless.)

**Fail-before, implementation stashed (`git checkout 6d6cc52d` + test file restored) — 15 failed / 4 passed, matching the claim exactly:**
```
E  ImportError: cannot import name 'proposal_quality' from 'cap_evolve'
15 failed, 4 passed in 12.03s
```
The 4 that pass without the fix are regression guards on pre-existing invariants, correctly so:
```
test_a_genuine_mechanism_proposal_is_not_rejected PASSED
test_the_reasoning_skill_is_not_mistaken_for_a_capability_edit PASSED
test_the_bar_is_capped_by_the_shared_cap_not_a_second_one PASSED
test_no_test_split_id_reaches_any_injected_workdir_file PASSED
```
(Caveat: `skills/reasoning/` survives a `git checkout` to the base only as untracked `__pycache__`, so the injection tests fail on ImportError rather than on a missing tree — the fail-before is real either way.)

**compileall:**
```
$ python -m compileall -q core/cap_evolve skills   →  rc=0
```

**build_manifest.py — 21 skills, and the committed manifest is byte-identical:**
```
wrote /private/tmp/rv-246/skills/_registry/manifest.json (21 skill(s))
  optimizer: run-optimizer
  reasoning: mechanism-probe
$ git status --short   →  (empty)
```

**#213's lint on the branch — the new skill has zero errors AND zero advisories:**
```
skill authoring lint — 21 skill package(s) under /private/tmp/rv-246/skills
  ERROR   capabilities/tools: SKILL.md body is 673 lines (>500); ...
  ERROR   capabilities/tools: SKILL.md body is ~10934 tokens (>5000); ...
  ERROR   orchestrate/using-cap-evolve: description must not contain XML tags
  advise  algorithms/evograph … phases/report          (11 advisories, none on reasoning/)
FAIL — 3 authoring violation(s) in 2 skill(s)
```

**The 3 errors are genuinely pre-existing and genuinely #213's:**
```
$ cd /tmp/rv-222   (base, 20 skills)
  ERROR   capabilities/tools: SKILL.md body is 673 lines (>500); ...
  ERROR   capabilities/tools: SKILL.md body is ~10934 tokens (>5000); ...
  ERROR   orchestrate/using-cap-evolve: description must not contain XML tags
FAIL — 3 authoring violation(s) in 2 skill(s)

$ cd /tmp/rv-213   (#213's own branch)
OK — 20 skill packages pass the authoring bar (0 advisory note(s))
```
Identical 3 on the base, and #213 clears all 3 plus every advisory. Nothing introduced here.

**All 21 `check.py` print `ok: true`** (run from each script's own dir; the new one in full):
```
{
  "skill": "mechanism-probe",
  "ok": true,
  "problems": [],
  "notes": [
    "run entry exposes main()",
    "all three declared fields parse from PROCESS.md",
    "false-rejection probe: a real mechanism proposal is never rejected",
    "declaration is recorded per candidate as advisory",
    "unfilled placeholders count as missing (no vacuous pass)"
  ]
}
```

**Zero LLM calls — add-an-import probe (would a regression be caught?):**
```
$ # added `import requests` to proposal_quality.py
$ python -m pytest core/tests/test_proposal_quality.py -k no_model -q
FAILED test_the_gate_makes_no_model_call
1 failed, 18 deselected
```
The guard is real. The module imports only `re` and `pathlib`; the banned list covers `anthropic`, `openai`, `requests`, `urllib.request`, `http.client`, `aux_model`, `subprocess`.

**Parser edge cases (probed, informational):**
```
multi-line value (continuation dropped)              declared= True   (first line only)
value is a bare code span `x`                        declared= True
one-char values                                      declared= True
'Observable:' without 'Expected'                     declared=False missing=['observable']
value >MAX_VALUE_CHARS                               truncated to 400
seed placeholder ABOVE an appended filled block      declared=False   <- non-blocking #2
```

OsherElhadad:
## 🔧 Review fixes

Commit `86514a79` on `feat/issue-140-reasoning-skills`. Verdict was APPROVE WITH NITS (0 blocking); all 3 non-blocking findings fixed, all 3 nits addressed, both claim corrections made.

---

### The placeholder inversion — before/after

The shape: seed block untouched, filled declaration appended below it.

```
## Proposal declaration
- Mechanism: <what now behaves differently, and why>
- Hypothesis: <which cluster this fixes>
- Expected observable: <what will look different next iteration>

### My declaration
- Mechanism: quote_price recomputes the total in-body and refuses a mismatch...
- Hypothesis: the four tax-line failures share one root cause...
- Expected observable: get_total precedes every quote_price call...
```

**BEFORE** (`2cb4d1dc`, `re.search` — first match wins):
```
declared = False | missing = ['hypothesis', 'mechanism', 'observable']
mechanism = ''
```

**AFTER** (`86514a79`, `re.finditer` — placeholders skipped):
```
declared = True | missing = []
mechanism = 'quote_price now recomputes the total in-body and refuses a m'
```

Pinned by `test_a_declaration_appended_below_the_seed_placeholders_still_counts`. On the reviewer's point about `finditer` as the default: yes — this is the second first-match-only defect in the batch after #189's guard, and the whole parser is now occurrence-scanning rather than first-hit, so a future field added to `_LABELS` inherits the correct behaviour instead of re-introducing the bug.

---

### The GEPA-local-gate enforcement probe

Injected a real enforcement at `gepa.py:643`, honest wording untouched:

```python
        local_pass = _sum_reward(child_mb) > _sum_reward(parent_mb)
        if not _q140["declared"]:
            local_pass = False  # INJECTED ENFORCEMENT (probe)
```

The new probe catches it directly, and names the algorithm:

```
FAILED core/tests/test_proposal_quality.py::test_gepas_local_gate_is_not_enforcing_the_declaration
E  AssertionError: an UNDECLARED candidate was stopped at GEPA's LOCAL gate — the advisory
   guarantee is broken for gepa: {'kind': 'gepa_local_gate', 'candidate': 'gepa_0001',
   'parent': 'seed', 'child_sum': 3.0, 'parent_sum': 0.0, 'passed': False}
1 failed, 22 passed
```

Enforcement removed; `git diff --stat core/cap_evolve/gepa.py` is empty and the file is back to 23/23 green. Before this, that injection passed all 19 tests and surfaced only as five confusing `test_gepa.py` failures.

---

### Numbered response to all six findings

**1. FINDING 1 — dead self-check, and `scripts/` copied only to serve it. Fixed.** Both halves gone. `scripts` added to the existing `ignore_patterns` at `harness.py:1281` (one word, now identical to the capability/diagnose copies) and the `## Checking your own declaration` bash block replaced with a re-read instruction that needs no runnable command. The reviewer is right that it could not work either way: not from the optimizer's workdir (no `skills/` tree) and not from the injected copy (the bootstrap's upward walk never finds `core/`). The stale comment claiming `scripts/` was deliberately kept is replaced with the real reason it is excluded. Injected copy is now 2 files instead of 5 — a small read-context saving that matters more now four blocks share the tail.

**2. FINDING 2 — the signal inverts on append-below. Fixed.** Evidence above. `re.finditer` + placeholder skip, so the last/filled declaration wins over an unfilled one wherever it sits.

**3. FINDING 3 — the false-rejection probe was hill-climb only. Fixed.** Two per-algorithm probes added:
- `test_gepas_local_gate_is_not_enforcing_the_declaration` — asserts on GEPA's own `gepa_local_gate` event that an undeclared candidate passed, plus that the verdict equals the reward comparison exactly (`(child_sum > parent_sum) is passed`).
- `test_skillopts_step_does_not_enforce_the_declaration` — asserts on SkillOpt's own step record, not on the fact that it delegates to `run_step` today. That delegation is an implementation detail the guarantee should not depend on, and this epic has repeatedly found per-algorithm gaps (#199's GEPA-blind consumers, #216's journal, the SkillOpt double-count).

**4. Nit — `_PLACEHOLDER_RE`'s `[-–—.]*` matches "". Fixed.** The empty case is now its own alternative (`^\s*$|^\s*(?:<.*>|[-–—.]+|...)\s*$`) and the dash branch requires at least one dash, so the pattern reads as what it rejects.

**5. Nit — `expected\s+observable` required the adjective. Fixed.** Now `(?:expected\s+)?observable`, so a bare `Observable:` is the same field. The reviewer flagged this as brittleness the docstring claimed tolerance for; the docstring is now true. New test: `test_a_bare_observable_label_is_the_same_field`.

**6. Nit — one-character values count as declared. Declining, deliberately.** The declaration is presence-only by design, exactly because the judgement half is not mechanically decidable. Any minimum length would be an arbitrary number doing a knob's job — a 20-char bar rejects a terse-but-real "in-body guard in `quote_price`" while passing 20 chars of filler. The floor is documented in `MAX_VALUE_CHARS`'s neighbourhood and the reviewer's own framing ("worth knowing the signal's floor") is the right disposition. If the recorded rows ever show one-char values being used to game a bar, that bar does not exist yet to game.

---

### Claim correction 1 — the base number

The PR body's **214 is wrong**. The true tip of #222 (`6d6cc52d`) is **221**; 214 predates #222's own review fix (`da9ad44d` is 213, matching neither). So the arithmetic is: **221 base + 19 original = 240**, and **221 + 23 = 244** with the 4 tests added here. Restated.

### Claim correction 2 — the rationale is a hypothesis, and now reads as one

The reviewer is right that "an edit you cannot state a mechanism for is the edit that historically wasted the iteration" cited "any run's `LEDGER.md`" with no run, no number and no artifact, and that nothing in-repo measures knob-versus-mechanism outcomes. Given this epic's history with unsubstantiated claims (#99's headline, #241's 8 removals, #244's "check does not run run_target"), the wording mattered. `SKILL.md` now says, verbatim:

> **This is an unvalidated hypothesis, stated as one:** we believe an edit whose mechanism cannot be stated is the edit that wastes the iteration, but nothing in this repo has measured knob-shaped versus mechanism-shaped edit outcomes — the `proposal_quality` event this skill produces is the instrument that would test it, and it currently has zero rows.

`PROMPT_BLOCK` carries the same correction in one clause: "our working hypothesis (unmeasured so far; this record is how we would test it)". The honest-wording tests still pin the enforcement claim; a reword to enforcing language still fails 4 tests.

---

### Verification

```
$ PYTHONPATH=/tmp/fx-246/core python -m pytest core/tests -q
244 passed in 120.00s (0:01:59)
```
244 = 221 (true #222 base) + 19 (original) + 4 (new: append-below, bare-Observable, GEPA local gate, SkillOpt step). 0 failed.

**Injected copy — `scripts/` out, `diff -r` still rc=0 for all three, sealed-test leak 0:**
```
hill-climb  scripts/ present: False  files: ['SKILL.md', 'meta.yaml']
            diff -r (-x scripts) rc=0 (identical)
            prompt 22240/60000  bar whole: True  no run.py cmd in SKILL: True
            sealed test ids ['a8', 'a7'] in workdir files: 0 []
gepa        scripts/ present: False  files: ['SKILL.md', 'meta.yaml']
            diff -r (-x scripts) rc=0 (identical)
            prompt 23247/60000  bar whole: True  no run.py cmd in SKILL: True
            sealed test ids ['a8', 'a7'] in workdir files: 0 []
skillopt    scripts/ present: False  files: ['SKILL.md', 'meta.yaml']
            diff -r (-x scripts) rc=0 (identical)
            prompt 22768/60000  bar whole: True  no run.py cmd in SKILL: True
            sealed test ids ['a8', 'a7'] in workdir files: 0 []
```
Nothing else depended on `scripts/`: the two remaining files are byte-identical to source, and the only reference to `run.py` anywhere was the SKILL.md line now deleted. `skills/reasoning/mechanism-probe/scripts/` stays in the repo — `check.py` runs there, from a checkout where the bootstrap does find `core/`, and it still prints `ok: true`.

**Cache hash + GEPA component list unchanged:**
```
hash equal: True | components equal: True | guidance in _SNAPSHOT_IGNORE: True
```

**Composed prompt capped, bar whole, notice present when elision fires:**
```
  in=  20000 out= 22360 <=cap=True bar whole=True notice=False
  in=  45000 out= 47360 <=cap=True bar whole=True notice=False
  in=  59000 out= 59905 <=cap=True bar whole=True notice=True
  in=  60000 out= 59905 <=cap=True bar whole=True notice=True
  in= 110000 out= 59906 <=cap=True bar whole=True notice=True
bar size: 1396 chars
```
The bar grew 1,320 → 1,396 chars from the honesty rewording (+76, still ~2.3% of the cap and well inside the tail's 17,940). The four-blocks-at-widest measurement is unaffected: 10,100 → 10,176 of 17,940, 7,764 to spare.

**compileall, manifest, checks, lint:**
```
$ python -m compileall -q core/cap_evolve skills   →  rc=0

$ python skills/_registry/build_manifest.py
wrote /private/tmp/fx-246/skills/_registry/manifest.json (21 skill(s))
  reasoning: mechanism-probe
$ git status --short   →  manifest.json unchanged (byte-identical)

all 21 check.py:  ok:true = 21   not-ok = 0

#213's lint on this branch:
  ERROR   capabilities/tools: SKILL.md body is 673 lines (>500)
  ERROR   capabilities/tools: SKILL.md body is ~10934 tokens (>5000)
  ERROR   orchestrate/using-cap-evolve: description must not contain XML tags
FAIL — 3 authoring violation(s) in 2 skill(s)
```
Same 3 pre-existing errors the reviewer confirmed on the base and on #213's own branch — **zero errors and zero advisories on `reasoning/`**.

---

### Merge order

**#199 → #213 → #222 → #246 → #219 → #221.**

- **#199** first, unconditionally — it owns `INJECTED_DIRS`, `cap_instructions`, `MAX_INSTRUCTIONS_CHARS`.
- **#213** early, so #246 lands into green CI (#213 is the only branch where the skill lint passes clean). **#213 must be rebased after #246 lands** to regenerate the 21-skill manifest — its `build_manifest.py` + `git diff --exit-code` gate fails on a stale one.
- **#222** before #246 — forced; #246 is cut from `6d6cc52d` and fast-forwards.
- **#246** next, not last: it is the smallest prompt-appending sibling (1,396 chars, one `+=` in `_augment_instructions`'s tail) and its rebase is the one that has to re-verify a cap measurement, so it should not be the one paying for the composed merge.
- **#219 then #221 last.** #219 conflicts with #246 today in `harness.py` (×3 hunks), `dashboard.py` and `skillopt.py`. Re-run the all-four-at-widest probe on the merged tree afterwards.

---

### Files touched

- `core/cap_evolve/proposal_quality.py` — `finditer` + placeholder skip, bare `Observable:`, `_PLACEHOLDER_RE` empty alternative, `PROMPT_BLOCK` honesty clause
- `core/cap_evolve/harness.py` — `scripts` added to the reasoning-copy `ignore_patterns`; stale comment corrected
- `core/tests/test_proposal_quality.py` — 4 new tests (19 → 23)
- `skills/reasoning/mechanism-probe/SKILL.md` — dead self-check command removed, rationale reworded as an explicit unvalidated hypothesis

OsherElhadad:
Closing — the substance of this landed on main by other means, and what is left is ceremony.

**The bar already exists where the optimizer reads it.** `skills/phases/diagnose/SKILL.md:129-137`
already says "ideate a DRASTIC, generalizing edit … not a tiny tweak … PREFER writing a new
code-bearing tool over a docstring tweak — a deterministic tool can't be forgotten the way a
prompt rule can", and `harness.py:1243` copies that skill into `guidance/diagnose/` with
`harness.py:2064` requiring it be read IN FULL before diagnosing. `_algorithm_brief`
(`harness.py:1767`) adds "a tiny or lucky change is rejected". That is knob-vs-mechanism,
already injected, already in every algorithm's prompt.

**The three-field declaration already exists in the same file, under other names.**
`harness._PROCESS_SEED` (`harness.py:719-739`) already requires, per edit: "shared root cause"
+ "what & why it generalizes" (= Mechanism/Hypothesis) and "Verify-the-fix — the trace it
targets → what the guard/computation/new-tool now does on those exact inputs" (= Expected
observable). It is REQUIRED, snapshotted with the candidate, and re-fed to the next iteration
via `./prior_iterations/` + `RUNMAP.md` (`harness.py:1000-1049`). So the net proposal here is
renaming three headings and adding a 132-line regex that counts whether they were filled.

**And the instrument measures a hypothesis nobody has tested.** By this PR's own words, shipped
in the prompt and the skill: "our working hypothesis (unmeasured so far)" and "nothing in this
repo has measured knob-shaped versus mechanism-shaped edit outcomes — the `proposal_quality`
event … currently has zero rows." Honest, and exactly the reason not to merge ~800 lines of
core + a new one-member `reasoning` skill tier + 502 lines of tests for it, while it competes
for the capped prompt tail with #129's dead-end constraints and duplicates `diagnose`'s rule so
the optimizer reads it twice.

Mechanically it is also `CONFLICTING`/`DIRTY`, stacked on unmerged #199/#212/#222 (roughly a
third of the diff is theirs), and main has moved a long way since 2026-07-30.

If the wording delta is still wanted, it is a ~6-line edit to the existing `_PROCESS_SEED` rows
in `harness.py` — reopen as that, not as a module, a skill tier, and an event. Closing #140 with
this.

PR #249, opened 2026-07-30
Add the cheap first-real-run on-ramp between toy_calc and the ~$148 tau2 run (closes #124)

### LINKED ISSUE #124: Add a cheap first real run on-ramp between toy_calc and the multi-hour ~$148 tau2 run
## Problem

There's a **cliff** between the free 2-minute `toy_calc` proof and the first *real* run. The next step advertised is the τ²-bench airline example, which `site/run-end-to-end.html:177` says "takes hours" and `:244` reports ~**$148** optimizer spend. There is no small/cheap real run in between.

## Why it matters

A newcomer who finishes `toy_calc` (deterministic, no model) and wants to see cap-evolve work on a *real* LLM task is immediately asked to commit hours and ~$148. That's a conversion killer. Prior art in this space lowers the "use it for real" barrier with free/local provider presets (e.g. Gemini/Groq free tiers, local Ollama) and a tiny real task — turning the second step from *pay hours* into *free minutes*.

## What to change

1. Add a **cheap first-real-run example** under `examples/` — a small task set (e.g. 5–10 tasks, few iterations, capped budget) against a **free or local** provider (a free-tier API key, or local Ollama via the existing provider-agnostic `model_config.py` one-line `MODEL=` switch). Target: a few minutes and ~$0.
2. Document it as the explicit "step 2" between `toy_calc` and tau2 in `docs/GETTING_STARTED.md` and the site getting-started page.
3. Ship its committed run artifact so the dashboard shows a real (small) result.

## How to test

- `bash examples/<cheap_real>/run.sh` completes in minutes on a free/local provider and produces a gate decision + a dashboard.
- Getting-started docs present a clear toy → cheap-real → full-benchmark ladder.
- The example's committed artifact loads in the dashboard.

- **Parallel-safe.** Independent example + docs. Pairs with the "demo-first replay" and `doctor` issues as the onboarding-improvement cluster. No code changes to core required (uses existing adapter templates + provider config).

OsherElhadad:
Closes #124

There was a cliff between rung 1 and rung 2 of onboarding: `toy_calc` is free but calls
no model, and the next advertised step was the τ² airline run — hours and ~$148. This
adds the missing rung: `examples/cheap_real`, a **real** run (real LLM, real paired
gate, real sealed test split) that finishes in minutes for **$0** on a local model, or
cents with a hosted proposer.

The task is date normalization — `"the 22nd of November, 1963"` → `1963-11-22`, scored
by exact match. The seed prompt is a generic "you are a helpful assistant", so a small
model answers in prose and scores 0; the fix is an output contract, which is exactly the
edit an optimizer is good at proposing. That makes the demonstration honest: the gain
comes from a general rule, not a memorized answer.

## The three-rung ladder

All three are now visible in `docs/GETTING_STARTED.md` and `site/getting-started.html`,
each with its cost, runtime, and derivation.

| Rung | What is real | Runtime | Cost |
|---|---|---|---|
| **1. free** — `toy_calc` | pipeline, gate, sealed test — no model | seconds | **$0** |
| **2. cheap-real** — `cheap_real` | **a real LLM**, real gate, real sealed test | **36 s – 5.4 min** MEASURED | **$0** local / **$0.5427** MEASURED hosted proposer |
| **3. full** — τ²-bench airline | a published benchmark | hours | ~$148 |

### Derivation (rung 2)

20 tasks → train 10 / val 5 / test 5, `num_trials: 1`, `max_iterations: 3`:

```
runner calls   = val 5 x trials 1 x (1 baseline + 3 candidates)  = 20
               + test 5 x (best + baseline seed)                 = 10   → 30 calls
proposer calls = max_iterations                                  =  3
```

- **Free variant** (`ollama/llama3.2:3b` runner + `mock` proposer) — **MEASURED: 36 s,
  $0.00.** $0 by construction: a local model is unmetered and `mock` makes no network
  call at all.
- **Cheap variant** (same local runner, `claude-code`/Haiku proposer) — **MEASURED**
  from the run's own accounting (`state.json` → `spent`): **$0.5427 total, 321 s
  (5.4 min) wall clock, and exactly the 30 `metric_calls` the formula above predicts.**
  Per-proposal spend was $0.159 / $0.199 / $0.185, so budget **~$0.15–0.35 per
  iteration**. Proposal latency dominates: 293 s of the 321 s was the optimizer, only
  27 s the runner. All 30 runner calls cost $0 because the runner is local.
- **Hosted-runner variant** (no local model) — 30 runner calls at the bundled table's
  Haiku rate and cap-evolve's assumed 3 000 in / 800 out tokens per rollout:
  `30 x (3000 x $1.00 + 800 x $5.00) / 1e6` ≈ **$0.21 — an ESTIMATE** from
  `core/cap_evolve/pricing.py`, explicitly labelled as such in the docs. I did not run
  this variant.

`max_usd: 3.0` / `max_optimizer_usd: 2.5` are hard stops, so rung 2 cannot quietly
become rung 3.

## Standalone example, not a zoo entry — and why

#233's zoo was the natural home, and I checked it properly rather than assuming. It
cannot host this one: `verify` step 5 runs every val task **twice** and fails the
benchmark if any rollout fingerprint differs
(`NON-DETERMINISTIC: … cannot produce a reproducible number`). That guard is correct and
worth keeping, but **no real LLM can satisfy it**, even at `temperature=0` with a fixed
seed. This example exists specifically to call a real model, so making it a zoo entry
would mean weakening the zoo's determinism guarantee to admit the one benchmark that
structurally cannot meet it. Standalone example; the zoo keeps its invariant. Reasoning
is recorded in the example's README so the next person does not re-litigate it.

## Entry point for #133

`bash examples/cheap_real/run.sh`. Every knob is an env var with a default, and the last
object on stdout is the run's summary JSON:

| Var | Default |
|---|---|
| `CHEAP_REAL_MODEL` | `ollama/llama3.2:3b` (free, local) |
| `CHEAP_REAL_API_BASE` | `http://localhost:11434` |
| `CHEAP_REAL_OPTIMIZER` | `mock` (zero-API) |
| `CHEAP_REAL_OPT_MODEL` | `""` |
| `CHEAP_REAL_WORKDIR` | fresh `mktemp` dir |
| `CHEAP_REAL_MAX_USD` | spec value (3.0) |
| `CHEAP_REAL_PYTHON` | `python3` |

So #133 can call it with `CHEAP_REAL_WORKDIR` pinned and parse the last JSON object; no
import, no new API surface.

## How it composes with the sibling PRs

- **#190 / #134** — no new credential resolution. The runner goes through the bundled
  `model_config.py`, and nothing this PR prints echoes an endpoint: the preflight
  reports a failed local endpoint *without* the URL, deliberately, per #190's rule that
  a non-default `base_url` is confidential.
- **#193 / #121** — `doctor` covers credential presence; this preflight covers the
  orthogonal thing (is the model actually reachable and pulled). No overlap.
- **#205 / #132** — the preset sets `proposer_model` **and** `optimizer_model` to
  `claude-haiku-4-5` on purpose. `optimizer_model` is what main reads today;
  `proposer_model` is #132's tier name and falls back to `optimizer_model` when blank,
  so setting both is correct before *and* after that merge. `aux_model` is deliberately
  unset, because every auxiliary step is still pure Python and an aux tier would route
  nothing.
- **#195 / #113** — the 20-task size is a **floor, not a preference**. val 5 is the
  smallest split that clears both `MIN_VAL_TASKS = 2` (below which `gate.decide`
  refuses) and `LOW_CONFIDENCE_VAL_TASKS = 5` (below which every decision is branded
  LOW CONFIDENCE). Verified against #195's branch, not just described.
- **#197 / #142** — `protected_paths` is **omitted**, never `[]`. `run.sh` places the
  adapter, `model_config.py` and `tasks.jsonl` *inside* the project dir and the spec
  names `dataset_source: tasks.jsonl`, so #197's layout defaults cover all four grader
  files without the preset declaring anything (a declared list would *replace* the
  defaults).

**Expected merge order:** none required — this is additive (one new example dir, one new
test file, three edited docs; zero core changes). It composes with all five siblings in
any order.

## Three bugs found while proving this works

Worth recording, because each one produced a *clean-looking* run whose numbers were
meaningless — the failure mode this epic keeps catching.

1. **A missing `litellm` made a broken run look like an honest 0.0.** The adapter
   correctly turns a failed model call into `error:` + reward 0.0 (infra noise must not
   be optimized against), but that means a stopped Ollama yields `baseline_val 0.0 →
   test_reward 0.0` with no visible failure. `run.sh` now preflights the interpreter,
   the endpoint, *and* whether the model is actually pulled, before spending anything.
2. **`optimizer_max_turns: 12` silently discarded every candidate.** claude-code exits
   non-zero with `Reached max turns (12)`, which cap-evolve correctly reports as a
   failed iteration — so the run read as "the optimizer proposed nothing" rather than
   "the cap was too tight". Now 40, with the measured reason in a comment.
3. **A relative `optimizer_instructions_file` silently falls back to the generic
   template.** `cli.py` resolves it against its own cwd and then a cwd-relative
   `.capevolve/project`, neither of which is the run's project dir when the workdir is
   elsewhere. The optimizer consequently received tau2-flavored "edit the tool code"
   advice for a project with no tools, and proposed a prompt for the wrong task
   (it made the agent a *date-trivia* assistant). `run.sh` writes an absolute path.
   This is a latent main-branch sharp edge for any out-of-tree project dir; I worked
   around it here rather than widen this PR's scope into `cli.py`.

## Verification

Every command and its full output is in the `## 🔬 Evidence` comment below. Summary:

**`cap-evolve check` on the preset — `{"ok": true}`:**

```json
{
  "ok": true,
  "stubs": [],
  "problems": [],
  "notes": [
    "tasks('val') -> 20 task(s)",
  ]
}
```

**Zero-cost proof first — real local LLM runner, `mock` optimizer, through to a sealed
test number (36 s, $0):**

```json
{
  "run_dir": ".capevolve/run_cheap",
  "best_id": "cand_0001",
  "baseline_val": 0.0,
  "test_reward": 1.0,
  "test_baseline_reward": 0.0,
  "test_delta": 1.0,
  "iterations": 3
}
```

**Then with a REAL agent proposing the edit (claude-code / Haiku) — completed end to
end to a sealed test number, MEASURED:**

```json
{
  "run_dir": ".capevolve/run_cheap",
  "best_id": "cand_0001",
  "baseline_val": 0.0,
  "test_reward": 1.0,
  "test_baseline_reward": 0.0,
  "test_delta": 1.0,
  "iterations": 3
}
```

Its own spend accounting — note `metric_calls: 30`, exactly what the derivation predicts:

```json
{"iterations": 3, "metric_calls": 30, "usd": 0.0,
 "optimizer_usd": 0.5426701500000001,
 "runner_seconds": 26.98, "optimizer_seconds": 293.47}
```

The gate accepted on iteration 1 and correctly rejected the two that followed (val was
already 1.0, so there was nothing left to gain):

```
{"kind": "step", "candidate": "cand_0001", "accept": true,
 "reason": "paired \u0394\u0304=+1.0000 > 0 (SE=0 \u2192 STRICT fallback, warned; n=5)",
 "val": 1.0, "parent": "seed", "parent_val": 0.0, "opt_cost_usd": 0.159017}
```

The agent's actual proposed edit — a general rule, no memorized dates:

```text
Convert the date to ISO 8601 format: YYYY-MM-DD.
Four-digit year, two-digit month, two-digit day.
Output only the date. No explanation, no greeting, no text.
```

**#195's val floor:**

```
MIN_VAL_TASKS=2  LOW_CONFIDENCE_VAL_TASKS=5
cheap_real realized split: train=10 val=5 test=5
check_val_size warning: NONE (clears both floors)
gate.decide(n=5 pairs): True | paired Δ̄=+0.8000 > 1.1416·SE=0.2283 (SE=0.2000, n=5, df=4, t-corrected from 1.0·SE)
```

**#197's `protected_paths` rule** — key omitted, all four grader files covered by the
layout defaults:

```
protected_paths key present in preset? False
resolved protected paths:
   adapters/adapter.py
   adapters/model_config.py
   capevolve.yaml
   tasks.jsonl
```

**Full suite — 183 passed (179 baseline + 4 new), 0 failed:**

```
183 passed in 63.41s (0:01:03)
```

`compileall` clean, `bash -n run.sh` clean, every relative link in the touched docs
resolves.

## Honest limits, stated in the README

- The gate reports `SE=0 → STRICT fallback` at `num_trials: 1`, so the accept decision
  is "Δ > 0", not a significance test. That is documented behavior, not a defect of the
  preset — but it is stated plainly rather than glossed.
- A 3B local model is a weak reader. The point is watching the machinery work on a real
  model, not producing a publishable number.
- I did not run the hosted-runner variant; its figure is labelled an estimate with its
  arithmetic shown.

### DIFF
--- a/core/tests/test_cheap_real_preset.py
+++ b/core/tests/test_cheap_real_preset.py
@@ -1,7 +1,7 @@
 """The cheap_real preset's honesty invariants — the things a "make it cheaper" edit breaks.
 
 The preset (``examples/cheap_real/capevolve.yaml``) exists to be the CHEAP rung, so the
-obvious future edit is to shrink it further. Two of its values are load-bearing and
+obvious future edit is to shrink it further. Several of its values are load-bearing and
 would fail silently or confusingly if changed:
 
 * the dataset size and split ratios must land ``val`` at or above the floors in
@@ -9,10 +9,21 @@ would fail silently or confusingly if changed:
   (#113) *after* the budget is spent, and below ``LOW_CONFIDENCE_VAL_TASKS`` every
   decision is branded low-confidence. So val is pinned by a test, not by a comment.
 * ``protected_paths`` must stay ABSENT. Since #142/#197 an empty list is a hard error
-  and a present list REPLACES the layout defaults, so the only way this preset gets the
-  adapter + dataset + spec guarded is to omit the key entirely.
+  and a present list REPLACES the layout defaults. Precisely: an omitted key resolves
+  to ``[*_DEFAULT_GLOBS, *spec['dataset_source']]`` (``protect.py:168-170``), and
+  ``tasks.jsonl`` matches **none** of the default globs (``adapters``,
+  ``capevolve.yaml``, ``*gold*`` suffixes) — the answer key is guarded *by the
+  dataset_source fold-in*, which only happens when the key is absent. Omission is
+  load-bearing for the dataset specifically, not merely tidier.
+* ``optimizer_max_turns`` must stay at or above the MEASURED floor: at 12 the agent
+  exits ``Reached max turns``, which cap-evolve correctly reports as a failed
+  iteration, so every candidate is discarded and the run looks like "the optimizer
+  proposed nothing".
+* ``run.sh``'s YAML rewrite is a line-PREFIX string rewriter over this preset, so it
+  silently no-ops if a key it targets is ever indented or renamed. Pinned here because
+  a no-op means the paid rung runs with ``optimizer_skill: mock``.
 
-Both are cheap file reads: no model, no run.
+All are cheap file reads: no model, no run.
 """
 
 from pathlib import Path
@@ -82,3 +93,44 @@ def test_budget_is_bounded():
     assert 0 < float(spec["max_usd"]) <= 5.0
     assert 0 < float(spec["max_optimizer_usd"]) <= float(spec["max_usd"])
     assert 0 < int(spec["max_iterations"]) <= 5
+
+
+def test_optimizer_max_turns_clears_the_measured_floor():
+    # MEASURED, not guessed: at 12 the agent exits `Reached max turns (12)`, cap-evolve
+    # reports a failed iteration, and every candidate is discarded — the run looks like
+    # "the optimizer proposed nothing". A "tighten the caps" edit must not go back there.
+    assert int(_spec()["optimizer_max_turns"]) >= 40, (
+        "optimizer_max_turns below 40 was MEASURED to make every iteration fail with "
+        "`Reached max turns`, so no candidate survives. Raise it back.")
+
+
+def test_run_sh_yaml_rewrite_still_matches_the_preset():
+    # run.sh flips the optimizer keys with a line-PREFIX rewriter, so it silently
+    # no-ops if a targeted key is ever indented or renamed — and a no-op means the
+    # paid rung quietly runs `optimizer_skill: mock`. Run the real heredoc, not a copy.
+    import re
+    import tempfile
+    run_sh = (REPO / "examples" / "cheap_real" / "run.sh").read_text(encoding="utf-8")
+    m = re.search(r"<<'SED'\n(.*?)\nSED\n", run_sh, re.S)
+    assert m, "run.sh no longer has the SED heredoc this test pins"
+
+    with tempfile.TemporaryDirectory() as td:
+        target = Path(td) / "capevolve.yaml"
+        target.write_text(PRESET.read_text(encoding="utf-8"), encoding="utf-8")
+        old_argv = sys.argv
+        sys.argv = ["-", str(target), "claude-code", "claude-haiku-4-5"]
+        try:
+            exec(compile(m.group(1), "run.sh:SED", "exec"), {"__name__": "__main__"})
+        finally:
+            sys.argv = old_argv
+        out = target.read_text(encoding="utf-8")
+
+    assert "optimizer_skill: claude-code\n" in out, "the optimizer_skill flip no-opped"
+    assert "optimizer_model: claude-haiku-4-5\n" in out
+    assert "proposer_model: claude-haiku-4-5\n" in out
+    instr = [ln for ln in out.splitlines()
+             if ln.startswith("optimizer_instructions_file:")]
+    assert len(instr) == 1, f"expected exactly one instructions line, got {instr}"
+    assert Path(instr[0].split(":", 1)[1].strip()).is_absolute(), (
+        "the instructions path must be ABSOLUTE: cli.py resolves a relative one against "
+        "its own cwd and silently falls back to the generic template (see #252)")
--- a/core/tests/test_cheap_real_preset.py
+++ b/core/tests/test_cheap_real_preset.py
@@ -1,0 +1,84 @@
+"""The cheap_real preset's honesty invariants — the things a "make it cheaper" edit breaks.
+
+The preset (``examples/cheap_real/capevolve.yaml``) exists to be the CHEAP rung, so the
+obvious future edit is to shrink it further. Two of its values are load-bearing and
+would fail silently or confusingly if changed:
+
+* the dataset size and split ratios must land ``val`` at or above the floors in
+  ``splits`` — below ``MIN_VAL_TASKS`` a run dies mid-flight inside ``gate.decide``
+  (#113) *after* the budget is spent, and below ``LOW_CONFIDENCE_VAL_TASKS`` every
+  decision is branded low-confidence. So val is pinned by a test, not by a comment.
+* ``protected_paths`` must stay ABSENT. Since #142/#197 an empty list is a hard error
+  and a present list REPLACES the layout defaults, so the only way this preset gets the
+  adapter + dataset + spec guarded is to omit the key entirely.
+
+Both are cheap file reads: no model, no run.
+"""
+
+from pathlib import Path
+import sys
+
+REPO = Path(__file__).resolve().parents[2]
+PRESET = REPO / "examples" / "cheap_real" / "capevolve.yaml"
+TASKS = REPO / "examples" / "cheap_real" / "tasks.jsonl"
+
+sys.path.insert(0, str(REPO / "core"))
+
+
+def _spec() -> dict:
+    from cap_evolve.specfile import read_yaml
+    return read_yaml(PRESET.read_text(encoding="utf-8"))
+
+
+def _task_ids() -> list[str]:
+    import json
+    return [json.loads(ln)["id"]
+            for ln in TASKS.read_text(encoding="utf-8").splitlines() if ln.strip()]
+
+
+def test_split_clears_the_honest_gate_floors():
+    from cap_evolve.splits import make_splits
+    import cap_evolve.splits as splits
+
+    spec = _spec()
+    ids = _task_ids()
+    assert len(ids) == len(set(ids)), "duplicate task ids would corrupt the split"
+    sp = make_splits(ids, seed=int(spec["split_seed"]),
+                     ratios=(float(spec["split_train"]), float(spec["split_val"]),
+                             float(spec["split_test"])))
+    # MIN_VAL_TASKS is the hard floor (gate.decide refuses below it); it only exists
+    # once #113 lands, so tolerate its absence rather than skip the whole assertion.
+    hard = getattr(splits, "MIN_VAL_TASKS", 2)
+    soft = getattr(splits, "LOW_CONFIDENCE_VAL_TASKS", 5)
+    assert len(sp.val) >= hard, (
+        f"cheap_real val split is {len(sp.val)}, below the hard floor {hard}: the run "
+        "would die inside gate.decide after spending its budget")
+    assert len(sp.val) >= soft, (
+        f"cheap_real val split is {len(sp.val)}, below {soft}: every gate decision "
+        "would be branded LOW CONFIDENCE. Add tasks rather than lowering this.")
+    assert sp.test, "the sealed test split must be non-empty"
+    assert not (set(sp.train) & set(sp.val)), "train/val overlap makes val a fit metric"
+    assert not (set(sp.test) & (set(sp.train) | set(sp.val))), "the test split leaked"
+
+
+def test_protected_paths_is_omitted_not_empty():
+    # An empty list is a hard error since #142/#197, and a present list replaces the
+    # layout defaults — omission is what protects adapters/, the dataset and the spec.
+    for line in PRESET.read_text(encoding="utf-8").splitlines():
+        assert not line.strip().startswith("protected_paths:"), (
+            "cheap_real must OMIT protected_paths: an empty list is a hard error and a "
+            "declared list would silently replace the defaults that cover the grader.")
+
+
+def test_dataset_source_names_the_real_file():
+    # #142/#197's guard hashes the path `dataset_source` names, so `adapter` here would
+    # leave the answer key unprotected.
+    assert _spec()["dataset_source"] == "tasks.jsonl"
+
+
+def test_budget_is_bounded():
+    # The whole point of the rung: it cannot silently become the multi-hour run.
+    spec = _spec()
+    assert 0 < float(spec["max_usd"]) <= 5.0
+    assert 0 < float(spec["max_optimizer_usd"]) <= float(spec["max_usd"])
+    assert 0 < int(spec["max_iterations"]) <= 5
--- a/docs/GETTING_STARTED.md
+++ b/docs/GETTING_STARTED.md
@@ -12,7 +12,7 @@ its arithmetic — no rung is advertised on a guess.
 | Rung | What is real | Runtime | Cost | How |
 |---|---|---|---|---|
 | **1. free** | pipeline, gate, sealed test — **no model** | seconds | **$0** | [`examples/toy_calc`](../examples/toy_calc/) · §3 |
-| **2. cheap-real** | **a real LLM**, real gate, real sealed test | **36 s**–**5.4 min** measured | **$0** local, or **$0.54** measured with a hosted proposer | [`examples/cheap_real`](../examples/cheap_real/) · §4 |
+| **2. cheap-real** | **a real LLM**, a real accept/reject decision, a real sealed test | **~30 s**–**5.4 min** measured | **$0** local, **$0.0115** with a hosted runner, or **$0.54** with a hosted proposer — all measured | [`examples/cheap_real`](../examples/cheap_real/) · §4 |
 | **3. full** | a published benchmark (τ²-bench airline) | hours | ~$148 | [`REPRODUCE_tau2.md`](REPRODUCE_tau2.md) |
 
 ### Where rung 2's numbers come from
@@ -27,8 +27,10 @@ proposer calls = max_iterations                                  =  3
 ```
 
 - **Free variant** — local `ollama/llama3.2:3b` runner + `mock` proposer.
-  **MEASURED: 36 s wall clock, $0.00.** A local model is not metered and the `mock`
-  proposer makes no network call at all, so this is $0 by construction, not by luck.
+  **MEASURED: ~30 s wall clock, $0.00** (30.4 s, 33.9 s and 34.2 s across three runs on
+  two machines — single-run timing is noise at this scale, so treat it as "about half a
+  minute", not as a figure). A local model is not metered and the `mock` proposer makes
+  no network call at all, so the $0 is by construction, not by luck.
 - **Cheap variant** — same local runner, `claude-code`/Haiku proposer. The 30 runner
   calls stay $0; only the 3 proposals cost anything. **MEASURED** from the run's own
   accounting (`state.json` → `spent`): **$0.5427 total, 321 s (5.4 min) wall clock,
@@ -36,14 +38,25 @@ proposer calls = max_iterations                                  =  3
   $0.159 / $0.199 / $0.185, so budget **~$0.15–0.35 per iteration**. Proposal latency
   dominates: 293 s of the 321 s was the optimizer, and only 27 s the runner.
 - **Hosted-runner variant** — no local model available, so the runner is
-  `claude-haiku-4-5` too. 30 runner calls at the bundled price table's Haiku rate and
-  cap-evolve's assumed 3 000 in / 800 out tokens per rollout:
-  `30 x (3000 x $1.00 + 800 x $5.00) / 1e6` ≈ **$0.21 — an ESTIMATE**, derived from
-  `core/cap_evolve/pricing.py`, not measured.
+  `claude-haiku-4-5` and `mock` proposes for free. **MEASURED: $0.011478 total, 98 s
+  wall clock, the same 30 `metric_calls`, `baseline_val 0.0` → sealed `test_reward
+  1.0`.** The six eval costs sum to the total exactly
+  (`0.003466 + 3 x 0.000681 + 0.000687 + 0.005282 = 0.011478`); the two seed evals
+  dominate because the unoptimized prompt answers in prose and burns output tokens.
+  This is **20x cheaper than the prior estimate** of $0.21, which assumed
+  `pricing.py`'s generic 3 000-in/800-out rollout — a one-line date is far smaller than
+  that. The estimate is retired: this row is now measured.
 
 The preset's `max_usd: 3.0` / `max_optimizer_usd: 2.5` are hard stops, so rung 2
 cannot quietly become rung 3. Rung 3's ~$148 is the committed τ² run's own reported
 optimizer spend ([`RESULTS.md`](RESULTS.md)).
+
+**What rung 2 does *not* show.** Val saturates at 1.0 on iteration 1 in every variant,
+so the two candidates that follow are rejected against a ceiling, not on merit — and at
+`num_trials: 1` the decision is `Δ > 0`, not a significance test (the run says so:
+`SE=0 → STRICT fallback`). The gate runs and decides for real; it is never asked a hard
+question. For the significance machinery under load, see `examples/skillsbench`
+(`num_trials: 3`).
 
 ## Prerequisites
 - Python **3.10+** and **git**.
@@ -112,17 +125,25 @@ CHEAP_REAL_OPTIMIZER=claude-code CHEAP_REAL_OPT_MODEL=claude-haiku-4-5 \
   bash examples/cheap_real/run.sh
 ```
 
-No local model? Point the runner at a cheap hosted one — one env var, no code change:
+No local model? Point the runner at a cheap hosted one — one env var, no code change.
+Needs that provider's credential in the environment (e.g. `ANTHROPIC_API_KEY`).
+**MEASURED: 98 s, $0.0115.**
 
 ```bash
 CHEAP_REAL_MODEL=claude-haiku-4-5 bash examples/cheap_real/run.sh
 ```
 
 `run.sh` is also the **programmatic entry point**: every knob is an env var with a
-default and the last object on stdout is the run's summary JSON. It preflights the
-model endpoint before spending anything, because the adapter (correctly) turns a
-failed model call into reward `0.0`, which otherwise makes a broken run look like a
-clean one. Details, all three variants and the honest limits:
+default and **all of stdout is the run's summary JSON** — progress goes to stderr
+(#116) and the script keeps only the last object, so `json.loads(stdout)` works even
+under `CAPEVOLVE_DASHBOARD=auto`, where the CLI prints a second object (#217).
+
+It preflights **every** variant before spending anything, because the adapter
+(correctly) turns a failed model call into reward `0.0`, which otherwise makes a broken
+run look like a clean one. Local: the endpoint answers *and* the model is actually
+pulled. Hosted: one real 1-token completion through the same wiring the adapter uses,
+so a rejected parameter, a missing credential or a typo'd model name fails loudly
+instead of scoring 0.0 thirty times. Details, all variants and the honest limits:
 [`examples/cheap_real/README.md`](../examples/cheap_real/README.md).
 
 ## 5. Where to next
--- a/docs/GETTING_STARTED.md
+++ b/docs/GETTING_STARTED.md
@@ -1,6 +1,49 @@
 # Getting started
 
-Your first successful cap-evolve run, in two minutes, with **no API key**.
+Your first successful cap-evolve run, in two minutes, with **no API key** — then a
+real one, cheaply, before committing to anything expensive.
+
+## The ladder
+
+Three rungs, so "see it work" and "run it for real" are not the same decision. Every
+figure is either **measured** on the run it describes or labelled an **estimate** with
+its arithmetic — no rung is advertised on a guess.
+
+| Rung | What is real | Runtime | Cost | How |
+|---|---|---|---|---|
+| **1. free** | pipeline, gate, sealed test — **no model** | seconds | **$0** | [`examples/toy_calc`](../examples/toy_calc/) · §3 |
+| **2. cheap-real** | **a real LLM**, real gate, real sealed test | **36 s**–**5.4 min** measured | **$0** local, or **$0.54** measured with a hosted proposer | [`examples/cheap_real`](../examples/cheap_real/) · §4 |
+| **3. full** | a published benchmark (τ²-bench airline) | hours | ~$148 | [`REPRODUCE_tau2.md`](REPRODUCE_tau2.md) |
+
+### Where rung 2's numbers come from
+
+The preset is 20 tasks → train 10 / val 5 / test 5, `num_trials: 1`,
+`max_iterations: 3`. So the call counts are fixed and checkable:
+
+```
+runner calls   = val 5 x trials 1 x (1 baseline + 3 candidates)  = 20
+               + test 5 x (best + baseline seed)                 = 10   → 30 calls
+proposer calls = max_iterations                                  =  3
+```
+
+- **Free variant** — local `ollama/llama3.2:3b` runner + `mock` proposer.
+  **MEASURED: 36 s wall clock, $0.00.** A local model is not metered and the `mock`
+  proposer makes no network call at all, so this is $0 by construction, not by luck.
+- **Cheap variant** — same local runner, `claude-code`/Haiku proposer. The 30 runner
+  calls stay $0; only the 3 proposals cost anything. **MEASURED** from the run's own
+  accounting (`state.json` → `spent`): **$0.5427 total, 321 s (5.4 min) wall clock,
+  exactly the 30 `metric_calls` the formula above predicts.** Per-proposal spend was
+  $0.159 / $0.199 / $0.185, so budget **~$0.15–0.35 per iteration**. Proposal latency
+  dominates: 293 s of the 321 s was the optimizer, and only 27 s the runner.
+- **Hosted-runner variant** — no local model available, so the runner is
+  `claude-haiku-4-5` too. 30 runner calls at the bundled price table's Haiku rate and
+  cap-evolve's assumed 3 000 in / 800 out tokens per rollout:
+  `30 x (3000 x $1.00 + 800 x $5.00) / 1e6` ≈ **$0.21 — an ESTIMATE**, derived from
+  `core/cap_evolve/pricing.py`, not measured.
+
+The preset's `max_usd: 3.0` / `max_optimizer_usd: 2.5` are hard stops, so rung 2
+cannot quietly become rung 3. Rung 3's ~$148 is the committed τ² run's own reported
+optimizer spend ([`RESULTS.md`](RESULTS.md)).
 
 ## Prerequisites
 - Python **3.10+** and **git**.
@@ -43,11 +86,51 @@ This is exactly what `core/tests/test_e2e_slice.py` asserts. The script prints a
 directory; open the `dashboard.html` it writes in any browser to see the run (KPIs,
 per-iteration diffs, the tasks × iterations heatmap).
 
-## 4. Where to next
+## 4. Run the cheap FIRST REAL example (rung 2)
+
+`toy_calc` proves the machinery but calls no model. This one calls a real one — and
+still finishes in minutes for $0, so there is a rung between it and the multi-hour
+τ² run.
+
+The task: normalize a human-written date (`"the 22nd of November, 1963"`) to ISO
+(`1963-11-22`), scored by exact match. The seed prompt is a generic "you are a helpful
+assistant", so a small model replies in prose and scores 0; the fix is an output
+contract — exactly what an optimizer is good at proposing.
+
+```bash
+pip install litellm                      # the runner's provider shim
+ollama pull llama3.2:3b                  # ~2 GB, free, local
+
+bash examples/cheap_real/run.sh          # rung 2, free variant
+```
+
+For the paid variant, let a real agent propose the edit instead of `mock` (the runner
+stays local and free):
+
+```bash
+CHEAP_REAL_OPTIMIZER=claude-code CHEAP_REAL_OPT_MODEL=claude-haiku-4-5 \
+  bash examples/cheap_real/run.sh
+```
+
+No local model? Point the runner at a cheap hosted one — one env var, no code change:
+
+```bash
+CHEAP_REAL_MODEL=claude-haiku-4-5 bash examples/cheap_real/run.sh
+```
+
+`run.sh` is also the **programmatic entry point**: every knob is an env var with a
+default and the last object on stdout is the run's summary JSON. It preflights the
+model endpoint before spending anything, because the adapter (correctly) turns a
+failed model call into reward `0.0`, which otherwise makes a broken run look like a
+clean one. Details, all three variants and the honest limits:
+[`examples/cheap_real/README.md`](../examples/cheap_real/README.md).
+
+## 5. Where to next
 
 | You want to… | Go to |
 |---|---|
 | Understand what cap-evolve optimizes and how | [`../README.md`](../README.md) · [`ARCHITECTURE.md`](ARCHITECTURE.md) |
+| See a real (cheap) LLM run end to end | [`../examples/cheap_real/`](../examples/cheap_real/) |
 | Set up a real optimizer/runner (credentials, dashboard) | [`INSTALL.md`](INSTALL.md) |
 | Optimize your own agent + benchmark | [`OPTIMIZE_YOUR_OWN.md`](OPTIMIZE_YOUR_OWN.md) |
 | See real benchmark results | [`RESULTS.md`](RESULTS.md) |
--- a/docs/OPTIMIZATION_EXAMPLES.md
+++ b/docs/OPTIMIZATION_EXAMPLES.md
@@ -8,6 +8,11 @@ under the champion `cand_0007`. Rollouts live at
 `<run_dir>/rollouts/val/<task_id>__<tag>__t<k>.json` (full message transcript +
 `reward_info`); every example below was read directly from those files.
 
+> These come from the **full** ~$148 run. To watch an optimizer propose an edit on a
+> real model for cents and minutes first, run rung 2 of the ladder —
+> [`examples/cheap_real/`](../examples/cheap_real/) — and read the `candidates/*/`
+> diffs it writes. Same mechanism, three orders of magnitude cheaper.
+
 The throughline: **argument-level feedback from failing rollouts is converted into
 executable guards inside the existing tools** — code where the agent already "knew"
 the rule but skipped it, prose only for genuine knowledge gaps. Across the run
--- a/examples/cheap_real/README.md
+++ b/examples/cheap_real/README.md
@@ -1,0 +1,85 @@
+# Example: cheap_real — a REAL LLM run in minutes, for $0 or cents
+
+The missing middle rung. `toy_calc` is free but calls no model; the τ²-bench airline
+example calls a real model but [takes hours and ~$148](../../docs/RESULTS.md). This
+one is a **real** run — real LLM, real paired gate, real sealed test split — that
+finishes in **minutes** and costs **$0** on a local model.
+
+The task: normalize a human-written date (`"the 22nd of November, 1963"`) to ISO
+(`1963-11-22`), scored by exact match. The seed prompt is a generic "you are a helpful
+assistant", so a small model answers in prose and scores **0**; the fix is an output
+contract, which is exactly the kind of edit an optimizer proposes.
+
+## The three rungs
+
+| Rung | Runner | Optimizer | Runtime | Cost |
+|---|---|---|---|---|
+| free | `ollama/llama3.2:3b` | `mock` | **36 s** measured | **$0** measured |
+| cheap-real | `ollama/llama3.2:3b` | `claude-code` (Haiku) | **5.4 min** measured | **$0.54** measured |
+| full | tau2-bench airline | `claude-code` | hours | ~$148 |
+
+Both measured rungs reached the same honest outcome — `baseline_val 0.0` →
+`test_reward 1.0` on the sealed test split. The cheap rung's $0.54 is entirely the 3
+proposals ($0.159 / $0.199 / $0.185 from the run's own `opt_cost_usd`); all 30 runner
+calls were free because the runner is local.
+
+Derivation and provenance of every figure: [`../../docs/GETTING_STARTED.md`](../../docs/GETTING_STARTED.md).
+
+## Run it
+
+```bash
+# Rung 1 — free. Needs Ollama (`ollama pull llama3.2:3b`, ~2 GB) and `pip install litellm`.
+bash examples/cheap_real/run.sh
+
+# Rung 2 — real agent proposes the edit; the runner stays free and local.
+CHEAP_REAL_OPTIMIZER=claude-code CHEAP_REAL_OPT_MODEL=claude-haiku-4-5 \
+  bash examples/cheap_real/run.sh
+
+# No local model? Use a cheap hosted one as the runner instead.
+CHEAP_REAL_MODEL=claude-haiku-4-5 bash examples/cheap_real/run.sh
+```
+
+`run.sh` is the **programmatic entry point** — every knob is an env var with a default
+and the last object on stdout is the run's summary JSON. See its header for the full
+list (`CHEAP_REAL_MODEL`, `CHEAP_REAL_API_BASE`, `CHEAP_REAL_OPTIMIZER`,
+`CHEAP_REAL_OPT_MODEL`, `CHEAP_REAL_WORKDIR`, `CHEAP_REAL_MAX_USD`,
+`CHEAP_REAL_PYTHON`).
+
+## Files
+
+- `capevolve.yaml` — the preset. Only the keys that differ from
+  [`templates/project/capevolve.yaml`](../../templates/project/capevolve.yaml) are set.
+- `tasks.jsonl` — 20 dates. **20 is a floor, not a preference:** at the default
+  0.5/0.25/0.25 ratios that is train 10 / val 5 / test 5, and val 5 is the smallest
+  split that clears *both* of #113's bars — `MIN_VAL_TASKS = 2` (below which
+  `gate.decide` refuses outright) and `LOW_CONFIDENCE_VAL_TASKS = 5` (below which
+  every decision is stamped LOW CONFIDENCE). Shrinking the task count to save money
+  would buy a cheaper run by giving up the honest gate.
+- `capability/prompt.txt` — the seed system prompt that gets optimized.
+- `mock_script.json` — the deterministic edit the `mock` optimizer applies, so rung 1
+  is reproducible and zero-API.
+- `optimizer_INSTRUCTIONS.md` — the per-iteration prompt for the real-agent rung.
+- **No adapter.** `run.sh` copies
+  [`templates/adapters/jsonl_litellm/adapter.py`](../../templates/adapters/jsonl_litellm/adapter.py)
+  and `model_config.py` verbatim; a JSONL dataset scored by exact match is precisely
+  what that bundled template is for.
+
+## Why this is not a benchmark-zoo entry
+
+The benchmark zoo (`docs/BENCHMARK_ZOO.md`, landing in #233) would be the natural home
+— declarative manifest, one code file, a `verify` command. It cannot host this one:
+`verify`'s step 5 runs every val task **twice** and fails the benchmark if any rollout
+fingerprint differs (`NON-DETERMINISTIC: … cannot produce a reproducible number`).
+That guard is correct and worth keeping, but a real LLM cannot satisfy it even at
+`temperature=0` and a fixed seed. This example exists to call a real model, so it is
+a standalone example and the zoo keeps its determinism guarantee intact.
+
+## Honest limits
+
+- The gate reports `SE=0 → STRICT fallback` at `num_trials: 1`, because a single
+  deterministic-ish trial per task gives the paired gate no variance to work with.
+  That is the documented behavior, not a defect of the preset — but it means the
+  accept decision here is "Δ > 0", not a significance test. Raise `num_trials` to get
+  a real bar, at a proportional increase in runner calls.
+- A 3B local model is a weak reader. The point is to see the machinery work end to
+  end on a real model, not to produce a publishable number.
--- a/examples/cheap_real/README.md
+++ b/examples/cheap_real/README.md
@@ -2,8 +2,8 @@
 
 The missing middle rung. `toy_calc` is free but calls no model; the τ²-bench airline
 example calls a real model but [takes hours and ~$148](../../docs/RESULTS.md). This
-one is a **real** run — real LLM, real paired gate, real sealed test split — that
-finishes in **minutes** and costs **$0** on a local model.
+one is a **real** run — real LLM, a real accept/reject decision, a real sealed test
+split — that finishes in **minutes** and costs **$0** on a local model.
 
 The task: normalize a human-written date (`"the 22nd of November, 1963"`) to ISO
 (`1963-11-22`), scored by exact match. The seed prompt is a generic "you are a helpful
@@ -14,14 +14,17 @@ contract, which is exactly the kind of edit an optimizer proposes.
 
 | Rung | Runner | Optimizer | Runtime | Cost |
 |---|---|---|---|---|
-| free | `ollama/llama3.2:3b` | `mock` | **36 s** measured | **$0** measured |
+| free | `ollama/llama3.2:3b` | `mock` | **~30 s** measured (30.4 / 33.9 / 34.2 s on three runs) | **$0** measured |
+| hosted-runner | `claude-haiku-4-5` | `mock` | **98 s** measured | **$0.0115** measured |
 | cheap-real | `ollama/llama3.2:3b` | `claude-code` (Haiku) | **5.4 min** measured | **$0.54** measured |
 | full | tau2-bench airline | `claude-code` | hours | ~$148 |
 
-Both measured rungs reached the same honest outcome — `baseline_val 0.0` →
+All three measured rungs reached the same honest outcome — `baseline_val 0.0` →
 `test_reward 1.0` on the sealed test split. The cheap rung's $0.54 is entirely the 3
-proposals ($0.159 / $0.199 / $0.185 from the run's own `opt_cost_usd`); all 30 runner
-calls were free because the runner is local.
+proposals ($0.159 / $0.199 / $0.185 from the run's own `opt_cost_usd`); its 30 runner
+calls were **metered at $0** because the runner is local — the runner's cost there is
+*time* (29.5 s, 3 863 tokens), not zero across the board. The hosted-runner rung is the
+mirror image: `mock` proposes for $0 and the 30 runner calls cost $0.0115 in total.
 
 Derivation and provenance of every figure: [`../../docs/GETTING_STARTED.md`](../../docs/GETTING_STARTED.md).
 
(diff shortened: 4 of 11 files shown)


OsherElhadad:
## 🔬 Evidence

Every command below was run in a clean worktree of this branch (`/tmp/wt-124b`) with
`PYTHONPATH` set, as the repo requires. Output is pasted literally.

### 1. Free rung — real local LLM runner, `mock` proposer, end to end to a sealed test

```console
$ CHEAP_REAL_WORKDIR=/tmp/cr9 bash examples/cheap_real/run.sh
cheap_real preflight: OK
Working directory: /tmp/cr9
Runner model: ollama/llama3.2:3b   Optimizer: mock
{
  "run_dir": ".capevolve/run_cheap",
  "best_id": "cand_0001",
  "baseline_val": 0.0,
  "test_reward": 1.0,
  "test_baseline_reward": 0.0,
  "test_delta": 1.0,
  "test_pass_k": {
    "1": 1.0,
    "2": 0.0
  },
  "iterations": 3,
  "dashboard": ".capevolve/run_cheap/dashboard.html"
}

real  0m30.423s
```

MEASURED: 30 s, $0.00. Zero cost, and it reaches a real sealed test number (1.0) from
a real baseline (0.0) — the runner is a genuine LLM, only the proposer is deterministic.

### 2. `cap-evolve check` on the preset

```console
$ cap-evolve check /tmp/cr9/.capevolve/project
{
  "ok": true,
  "stubs": [],
  "problems": [],
  "notes": [
    "tasks('val') -> 20 task(s)",
  ]
}
```

`{"ok": true}`.

### 3. Real-agent rung — claude-code / Haiku proposes the edit

Run via the gateway. Neither `ANTHROPIC_BASE_URL` nor `ANTHROPIC_AUTH_TOKEN` is echoed
anywhere in this PR, its commits, or the run artifacts.

```console
$ CHEAP_REAL_OPTIMIZER=claude-code CHEAP_REAL_OPT_MODEL=claude-haiku-4-5 \
    CHEAP_REAL_WORKDIR=/tmp/cr8 bash examples/cheap_real/run.sh
cheap_real preflight: OK
Working directory: /tmp/cr8
Runner model: ollama/llama3.2:3b   Optimizer: claude-code
{
  "run_dir": ".capevolve/run_cheap",
  "best_id": "cand_0001",
  "baseline_val": 0.0,
  "test_reward": 1.0,
  "test_baseline_reward": 0.0,
  "test_delta": 1.0,
  "test_pass_k": {
    "1": 1.0,
    "2": 0.0
  },
  "iterations": 3,
  "dashboard": ".capevolve/run_cheap/dashboard.html"
}

```

Every `step` event, from the run itself:

```json
{"kind": "splits", "train": 10, "val": 5, "test": 5, "seed": 0}
{"kind": "baseline", "val": 0.0, "stderr": 0.0}
{"kind": "step", "candidate": "cand_0001", "accept": true, "reason": "paired \u0394\u0304=+1.0000 > 0 (SE=0 \u2192 STRICT fallback, warned; n=5)", "val": 1.0, "parent": "seed", "parent_val": 0.0, "optimizer_seconds": 81.58, "runner_seconds": 3.57, "cost_usd": 0.0, "tokens": 476, "opt_cost_usd": 0.159017, "opt_tokens": 5761}
{"kind": "step", "candidate": "cand_0002", "accept": false, "reason": "paired \u0394\u0304=+0.0000 <= 0 (SE=0 \u2192 STRICT fallback, warned; n=5)", "val": 1.0, "parent": "cand_0001", "parent_val": 1.0, "optimizer_seconds": 106.69, "runner_seconds": 1.59, "cost_usd": 0.0, "tokens": 491, "opt_cost_usd": 0.198835, "opt_tokens": 6442}
{"kind": "step", "candidate": "cand_0003", "accept": false, "reason": "paired \u0394\u0304=+0.0000 <= 0 (SE=0 \u2192 STRICT fallback, warned; n=5)", "val": 1.0, "parent": "cand_0001", "parent_val": 1.0, "optimizer_seconds": 105.2, "runner_seconds": 1.64, "cost_usd": 0.0, "tokens": 496, "opt_cost_usd": 0.184818, "opt_tokens": 7060}
{"kind": "finalize", "test_reward": 1.0, "test_baseline_reward": 0.0, "test_delta": 1.0, "best_id": "cand_0001"}
```

Its own spend accounting (`state.json` → `spent`) — note `metric_calls: 30`, exactly
what the derivation in the PR body predicts:

```json
{
  "iterations": 3,
  "metric_calls": 30,
  "usd": 0.0,
  "stall": 2,
  "runner_tokens": 3658,
  "runner_seconds": 26.9793119430542,
  "optimizer_seconds": 293.4690363407135,
  "optimizer_usd": 0.5426701500000001,
  "optimizer_tokens": 19263,
  "intake_usd": 0.0,
  "intake_tokens": 0,
  "intake_seconds": 0.0
}
```

MEASURED: **321 s (5.4 min), $0.5427.** Accepted on iteration 1 (val 0.0 → 1.0), then
correctly rejected two candidates that could not improve on a val of 1.0, and sealed
the test split at 1.0.

The agent's actual proposed edit — a general rule, no memorized dates:

```text
Convert the date to ISO 8601 format: YYYY-MM-DD.
Four-digit year, two-digit month, two-digit day.
Output only the date. No explanation, no greeting, no text.
Only output the date in YYYY-MM-DD format.
```

### 4. #113/#195 — the val floor

Run against `origin/fix/issue-113-small-samples`, where the floors actually exist:

```console
MIN_VAL_TASKS=2  LOW_CONFIDENCE_VAL_TASKS=5
cheap_real realized split: train=10 val=5 test=5
check_val_size warning: NONE (clears both floors)
gate.decide(n=5 pairs): True | paired Δ̄=+0.8000 > 1.1416·SE=0.2283 (SE=0.2000, n=5, df=4, t-corrected from 1.0·SE)
```

val=5 clears both bars with **zero warnings**, and the paired gate produces a real
t-corrected decision on 5 pairs.

The new test file asserts this, and passes on BOTH branches (on `main` the constants
do not exist yet, so it falls back to the documented values):

```console
$ # against origin/fix/issue-113-small-samples
$ pytest core/tests/test_cheap_real_preset.py -q
....                                                                     [100%]
4 passed in 0.03s
```

### 5. #142/#197 — `protected_paths`

Run against `origin/feat/issue-142-protected-paths`, on the real project dir the free
rung produced:

```console
protected_paths declared in the preset? False -> KEY OMITTED
resolved protected paths (from the layout defaults alone):
    adapters/adapter.py
    adapters/model_config.py
    capevolve.yaml
    tasks.jsonl

And proof that an empty list — what this preset deliberately avoids — is a hard error:
  TamperError: cap-evolve: `protected_paths` in /private/var/folders/zh/srgnbq_97qvb6002zsr40tgc0000gn/T/tmp1zz1f1o9/project/capevolve.yaml is an EMPTY list. That wo...
```

All four grader files — the adapter, its model wiring, the spec and the answer key —
are covered by #197's layout defaults, with nothing declared. That only works because
`run.sh` places them inside the project dir and the spec names
`dataset_source: tasks.jsonl`.

### 6. The preflight actually fires

This is the guard against the bug I hit: a broken setup that looks like an honest 0.0.

```console
$ CHEAP_REAL_API_BASE=http://localhost:59999 bash examples/cheap_real/run.sh
cheap_real preflight: the local Ollama endpoint did not answer (URLError). Start it (`ollama serve`) or set CHEAP_REAL_MODEL to a hosted model. Endpoint withheld (see docs/TROUBLESHOOTING.md).

$ CHEAP_REAL_MODEL=ollama/nonexistent-model bash examples/cheap_real/run.sh
cheap_real preflight: Ollama is up but does not have 'nonexistent-model'. Run `ollama pull nonexistent-model` (~2 GB), or set CHEAP_REAL_MODEL.
```

Note the first message withholds the endpoint deliberately, per #134's rule that a
non-default `base_url` is confidential.

### 7. Full suite, compileall, shell syntax, links

```console
$ PYTHONPATH=$PWD/core python -m pytest core/tests -q
183 passed in 63.48s (0:01:03)

$ python -m compileall -q core examples && echo OK
OK

$ bash -n examples/cheap_real/run.sh && echo OK
OK
```

183 passed (179 baseline + 4 new), **0 failed**. `test_dashboard_launch.py` (flaky per
#200) passed on this run.

Relative links in every touched doc resolve:

```console
broken relative links: NONE
```

### What I did NOT verify

- The **hosted-runner** variant (`CHEAP_REAL_MODEL=claude-haiku-4-5`). Its ~$0.21 is an
  **estimate**, derived in the PR body from `core/cap_evolve/pricing.py`, and labelled
  as an estimate in the docs. I did not spend money to measure it.
- Rung 3 (~$148). That figure is the committed tau2 run's own reported spend, quoted,
  not re-measured.
- The `SE=0 → STRICT fallback` shown in every `step` above is real and documented in
  the example's README: at `num_trials: 1` the accept decision is "Δ > 0", not a
  significance test. I did not paper over it.

OsherElhadad:
## 🔍 Review — PR #249

**CHANGES REQUESTED** — one blocking defect, and it is the *exact* failure mode this PR
exists to eliminate. The three bugs the PR found are real and well-documented, the cost
figures hold up to the arithmetic, the zoo declination is correct, and the free rung
reproduced on my machine to the digit. But the **third advertised variant — the hosted
runner, the one for users with no local model — produces a clean-looking run whose every
number is 0.0**, on the very first command the docs give them, and this PR's preflight
does not fire. That is bug (a) reintroduced through the door the preflight does not
cover.

---

### Blocking

**B1 — `examples/cheap_real/run.sh:43-63` — the hosted-runner variant is broken, and the
preflight is blind to it: a clean run with `test_reward: 0.0` from an unconditional
`seed` kwarg litellm rejects.**

`templates/adapters/jsonl_litellm/adapter.py:78` forwards `seed=seed` to every
`litellm.completion` call. Anthropic does not accept `seed`. The adapter (correctly)
turns the failure into `error:` + reward 0.0, and `run.sh`'s preflight only runs its
endpoint/pull checks under `if MODEL.startswith("ollama/")` (`run.sh:50`), so the hosted
path gets `preflight: OK` and nothing else. I ran the command the docs advertise
verbatim:

```console
$ CHEAP_REAL_MODEL=claude-haiku-4-5 bash examples/cheap_real/run.sh
cheap_real preflight: OK
Working directory: /tmp/cr-hosted
Runner model: claude-haiku-4-5   Optimizer: mock
{
  "run_dir": ".capevolve/run_cheap",
  "best_id": "seed",
  "baseline_val": 0.0,
  "test_reward": 0.0,
  "test_baseline_reward": 0.0,
  "test_delta": 0.0,
  "test_pass_k": {"1": 0.0, "2": 0.0},
  "iterations": 2
}
```

Exit 0. Empty stderr. Zero warnings. The reason is only visible if you go read a rollout
file by hand:

```console
$ cat .capevolve/run_cheap/rollouts/val/d05__cand_0001__t0.json
"error": "LLM call failed: litellm.UnsupportedParamsError: anthropic does not support
 parameters: ['seed'], for model=claude-haiku-4-5. To drop these, set
 `litellm.drop_params=True` ..."
```

Consequence: this is precisely bug (a) from the PR body — *"a broken run made
indistinguishable from an honest 0.0"* — shipped, in the variant the docs offer as the
fallback for the user with no Ollama. And it is worse than (a) was, because the preflight
now prints `OK` and so actively vouches for the run. This example's whole reason to exist
is being the rung a newcomer can trust; the newcomer without a 2 GB local model gets a
$0-but-silently-dead run and no way to tell.

Three separate places advertise this exact command as working: `README.md:39`,
`docs/GETTING_STARTED.md:118`, `site/getting-started.html:130` — all as *"one env var, no
code change"*.

Note the PR body says of the hosted variant *"I did not run this variant"* — that is
honestly stated, and the ~$0.21 figure is correctly labelled an ESTIMATE. But
"unmeasured" is being used to cover "unexercised", and running it costs a fraction of a
cent, not money worth withholding. It is also the *one* claim in this PR whose failure
mode is invisible.

Fix (cheapest that holds, in `run.sh`, no core change):

```bash
export LITELLM_DROP_PARAMS=1   # litellm honours this; unsupported kwargs are dropped
```

or, if you prefer to fail loudly rather than fix silently, extend the preflight past the
`ollama/` branch to do one real 1-token completion through `model_config.llm_kwargs()`
and exit non-zero on error. Either way, **run the hosted variant once** before merge — it
costs under a cent and it is the only variant nobody has exercised. Please also check
whether `run.sh:35`'s unconditional `export API_BASE=http://localhost:11434` reaches a
hosted provider: `model_config.py:95` reads the *generic* `API_BASE` first, before any
provider special-casing, so a hosted `MODEL` inherits a localhost `api_base` unless the
user also clears it — a second trap on the same path. That one deserves a line in the
docs at minimum.

---

### Non-blocking

**N1 — `examples/cheap_real/capevolve.yaml:63` — the spec's own comment contradicts the
PR's (correct) derivation.** It says *"3 iterations x 5 val tasks x 1 trial = 15 runner
calls"*. The real number is 30, as the PR body derives and as `state.json` reports; the
comment omits the baseline val eval (5) and both test evals (10). A future reader sizing
the budget from the file rather than the PR gets half the truth. Use the PR's own
derivation here.

**N2 — `examples/cheap_real/run.sh:14` — header says `CHEAP_REAL_MAX_USD` defaults to
"capevolve.yaml: 2.0"; `capevolve.yaml:70` says `max_usd: 3.0`.** Stale after the cap was
raised. One-character class of error, but this is a *budget* number in a PR whose subject
is budget figures.

**N3 — `examples/cheap_real/README.md:17` and `docs/GETTING_STARTED.md:15,30` cite "36 s
measured" for the free rung; the PR's own evidence comment shows `real 0m30.423s` and I
measured 33.9 s.** Neither is 36. Not a material claim — all three are "half a minute" —
but #238 got dinged in this very epic for presenting single-run timing noise as a
measurement, and quoting a third number that appears in neither run is the same class of
sloppiness. Either say "~30 s" or state the spread.

**N4 — `core/tests/test_cheap_real_preset.py` docstring:11-13 and `capevolve.yaml:4-7`
both assert "a present `protected_paths` list REPLACES the layout defaults".** True, and
I verified it (`protect.py:145-171` on #197's branch: a non-empty declared list returns
`out` alone and never reaches `[*_DEFAULT_GLOBS, *extra]`). But the *reason given* for
omitting the key is incomplete in a way that will mislead: what omission actually buys is
`[*_DEFAULT_GLOBS, *spec['dataset_source']]` — the dataset is covered **because it is
omitted**, i.e. `dataset_source` is folded into the default set at `protect.py:169`, not
because `tasks.jsonl` matches any default glob (it matches none; `_DEFAULT_GLOBS` is
`adapters`, `capevolve.yaml`, and `*gold*` data suffixes). The current wording reads as
though the defaults cover the answer key on their own. Worth one clarifying clause,
because it is the load-bearing fact.

**N5 — `examples/cheap_real/capevolve.yaml:24,51` — `proposer_model` and `dataset_source`
are read by nothing in `core/` today.** The spec's own comments say so for
`dataset_source` (line 48-50) and explain `proposer_model` as forward-compat with #132,
so this is documented, not hidden. Flagging only so it is a conscious call: two of the
preset's 20-odd keys are inert on `main`, and one of them (`dataset_source`) only becomes
load-bearing once #197 merges.

**N6 — untested paths.** The 4 new tests are all static file reads of the preset (see
"Test quality" below). Nothing tests `run.sh` itself: not the preflight branches, not the
YAML-rewrite `SED` heredoc at `run.sh:84-104`, not the absolute-path append. The
YAML-rewrite is the one I would want a test on — it is a line-prefix string rewriter over
a config file (`line.startswith("optimizer_skill:")` etc.), it is the fix for bug (c),
and it silently no-ops if a key is ever indented or renamed. A ~10-line test that runs
the heredoc over the committed preset and asserts the three resulting keys would pin it.

---

### Nits

- `examples/cheap_real/README.md:24` — "all 30 runner calls were free because the runner
  is local" is right, but the free rung's `state.json` shows `runner_tokens: 3863` and
  `runner_seconds: 29.5`, i.e. the runner cost is *time*, not $0 across the board. Fine
  as written; just noting the $0 is a metering artifact, not a physics claim.
- `test_pass_k: {"2": 0.0}` at `num_trials: 1` is pre-existing #112 (pass^k should read
  N/A above `num_trials`), not this PR's. But this PR is putting that JSON in front of
  first-time users on three doc pages, so it is now a first-impression bug. Consider a
  one-line note in the README's "Honest limits" until #112 lands.

---

### Do the cost figures hold up?

**Yes — every number reproduces, and the accounting model is the right one. This is the
most carefully-costed PR in the batch.** Six independent checks:

**1. The 30-call derivation is arithmetically correct AND matches the spec's actual
values.** I re-derived it from the committed preset rather than from the PR's prose:

```
val 5 x trials 1 x (1 baseline + 3 candidates)  = 20
test 5 x (best FINAL + baseline FINAL_seed)     = 10
                                          total = 30
```

And I confirmed each of the six evals independently from my own run's `events.jsonl`,
because a formula that happens to total 30 is not the same as 30 calls actually happening
in that shape:

```console
$ grep -o '"kind": "evaluate"[^}]*' state events
"split": "val",  "tag": "seed",       "reward": 0.0
"split": "val",  "tag": "cand_0001",  "reward": 1.0
"split": "val",  "tag": "cand_0002",  "reward": 1.0
"split": "val",  "tag": "cand_0003",  "reward": 1.0
"split": "test", "tag": "FINAL",      "reward": 1.0
"split": "test", "tag": "FINAL_seed", "reward": 0.0
```

Six evals × 5 tasks × 1 trial = 30. The "best + baseline seed" term the PR asserts for
the test split is real (`FINAL` + `FINAL_seed`), not a fudge to make the total land on
30. `harness.py:307` credits `len(tasks) * n_trials` per eval, which is the same
arithmetic.

**2. My own free-rung run reports `metric_calls: 30` — independently, on a different
machine, three weeks later.**

```console
$ cat /tmp/cr-canary/.capevolve/run_cheap/state.json | jq .spent
{"iterations": 3, "metric_calls": 30, "usd": 0.0, "stall": 2,
 "runner_tokens": 3863, "runner_seconds": 29.5, "optimizer_usd": 0.0}
```

**3. The three per-proposal costs sum to the total, to 1.5e-07.**

```console
$ python -c "print(sum([0.159017,0.198835,0.184818]))"
0.54267
$ # state.json optimizer_usd: 0.5426701500000001   → diff = 1.5e-07
```

Exact to float rounding. The three `opt_cost_usd` values in the `step` events *are* the
$0.5427, with nothing unaccounted.

**4. Is $0.5427 the total or only the optimizer share? Both — and that is verifiable,
not a coincidence.** `rundir.py:135-137`: `total_usd = usd + optimizer_usd + intake_usd`.
The run reports `usd: 0.0` (local runner, unmetered) and `intake_usd: 0.0` (no intake
phase), so `total_usd == optimizer_usd == 0.5427`. The PR calls it "total" and the README
calls it "entirely the 3 proposals" — both are correct *for this configuration*, and the
PR says why (line: "all 30 runner calls cost $0 because the runner is local"). This is
the accounting gap #234's review found (a burn meter understating GEPA spend 3.9× by
reading one role's field as the total) and this PR does not have it: it names the role,
shows the other roles are zero, and the identity closes.

**5. The $0.21 hosted-runner ESTIMATE reproduces from the stated inputs.**

```console
$ python -c "print(30*(3000*1.0+800*5.0)/1e6)"
0.21
```

Correctly labelled ESTIMATE in the PR body, in `GETTING_STARTED.md:40-44`, and excluded
from the "measured" columns. No estimate is presented as a measurement anywhere in this
PR — which, given this epic's history, is worth stating explicitly.

**6. The one figure I could not check, and the one that is slightly off.** Rung 3's ~$148
is quoted from `RESULTS.md`, not re-measured — appropriately, and the PR says so. The
"36 s" free-rung timing is the one number that reproduces neither in the PR's own
evidence (30.4 s) nor mine (33.9 s) — see N3. Non-material; the *cost* figures, which are
what this section is about, are clean.

**Verdict on the figures: they hold.** The derivation is independently correct, the
call count reproduces on a fresh machine, the per-proposal costs sum to the total, the
accounting model is the all-roles one, and estimates are labelled. I went looking for the
#234 accounting gap and the #238 single-run-noise problem and found neither in the money
numbers.

---

### Is this a real optimization or a formatting fix?

**It is a real optimization, and I'll defend that against the obvious objection — but the
PR is overselling it as an *optimization demo* when its actual value is as a *pipeline
demo*, and one number in it is a red flag the PR does not flag.**

The objection is fair on its face: the seed scores 0 only because it answers in prose,
the fix is an output contract, and `mock_script.json` proves a hard-coded one-line append
solves it. That is a formatting fix by any plain reading.

Why it is nonetheless real:

- **The gain is a general rule scored on held-out data.** The accepted edit
  ("four-digit year, two-digit month, two-digit day. Output only the date.") is scored on
  a sealed test split of 5 dates the optimizer never saw, and the optimizer's
  `INSTRUCTIONS.md:22-24` explicitly forbids memorizing a date with the tamper guard
  behind it. Baseline 0.0 → test 1.0 on unseen inputs is a real generalizing improvement,
  not a lookup.
- **"Formatting" is not a lesser category of prompt optimization — it is the modal
  finding.** Output-contract violations are the single most common reason a capable model
  scores 0 on a benchmark it can actually do. Demonstrating the machinery on the most
  common real failure class is a feature.
- **The pipeline exercised is the full one, not a subset.** Real LLM rollouts, real paired
  gate, real accept/reject, real sealed finalize, real tamper guard, real spend
  accounting. Nothing is stubbed but the *proposer*, in the free rung, and the PR labels
  that.

Where the PR oversells, and what it should have said:

**The two rejected iterations prove nothing about the gate, and the PR presents them as
if they do.** It says the gate "correctly rejected the two that followed" — true but
vacuous: val was already 1.0, so there was nothing to reject *between*. The run
demonstrates one accept from a floor and two no-ops from a ceiling. It never once
demonstrates the thing the gate exists for: **rejecting a plausible-looking candidate
that does not actually beat the parent.** A task that saturates at 1.0 on iteration 1
cannot show that, and this one does, twice over (both the free and the paid rung end at
1.0 after one step).

Compounding it: at `num_trials: 1` the gate reports `SE=0 → STRICT fallback`, so the
decision is `Δ > 0`, not a significance test. The PR *does* disclose this plainly
(README:79-83, PR body "Honest limits") — real credit for that, most PRs would have
buried it. But stack the two facts and the honest description of rung 2 is: *a real LLM,
a real sealed test, and a gate that was never asked a hard question.* "Real paired gate"
appears five times across the PR body, README and docs; it is true that the gate ran, and
misleading as a claim that the gate was demonstrated.

**Verdict: a real optimization on a task too easy to exercise the interesting half of the
machinery.** As "the cheap rung that proves the pipeline runs end to end on a real model
for $0", which is what #124 asked for, it delivers. Two changes would close the gap
without adding cost: (1) drop "real paired gate" to "real accept/reject decision" and add
one sentence to Honest limits saying the run never exercises a *rejection on merit*
because val saturates at iteration 1; (2) consider whether a slightly harder task — a few
ambiguous or malformed dates that a 3B model gets wrong even *with* the contract — would
land baseline ~0.0 → best ~0.6 instead of 0.0 → 1.0, at identical cost and runtime, and
give the gate something to actually decide. (2) is a suggestion, not a request; (1) I
would want before merge.

---

### Was declining the zoo right?

**Yes, and the claim is accurate against #233's code — I verified it rather than taking
it.** `core/cap_evolve/zoo.py:1111-1153` on `origin/feat/issue-141-benchmark-zoo`:

```python
for attempt in (1, 2):
    ...
    got[t.id] = (_rollout_fingerprint(r), round(float(s.reward), 9))
    passes.append(got)
drift = sorted(k for k in passes[0] if passes[0][k] != passes[1][k])
if drift:
    rep.problems.append(f"NON-DETERMINISTIC: {len(drift)} task(s) produced a different
        rollout or reward on an identical re-run with seed=0 ...")
```

Every val task, twice, and `_rollout_fingerprint` (`zoo.py:972-982`) is a SHA-256 over
the whole `rollout.to_dict()` with only `cost_usd`/`tokens` popped. So the comparison is
**byte-exact on the model's output text**. A real LLM cannot satisfy that: temperature 0
is not a determinism guarantee across batching, kv-cache state or server version, and
`_rollout_fingerprint` has no tolerance knob — the exclusion list is two keys, hardcoded.
There is no `allow_nondeterministic` escape either; I grepped: the only opt-out flag in
the manifest schema is `allow_saturated_baseline` (`zoo.py:120`), which addresses a
different problem. So the PR's claim — *unsatisfiable for a live model, and admitting
this benchmark would mean weakening the invariant* — is literally correct.

Could a tolerance have accommodated it? Not cheaply. The natural knob is
"reward-only, ignore the rollout hash", but the reward *also* drifts for a real model
(different phrasing → different exact-match score), so the tolerance would have to be a
numeric band on the mean reward — which is a different guarantee (statistical stability)
wearing determinism's name. Building that to admit one example is the wrong trade.

Where I'd push slightly: the PR frames this as binary ("the zoo keeps its invariant"),
but the durable answer is the one the task brief names — the zoo eventually needs an
explicit **non-deterministic grade**, because `cheap_real` is the first of a class, not a
one-off. Every future live-model benchmark hits this same wall. The right shape is a
declared `determinism: exact | statistical | none` in the manifest, where `statistical`
substitutes a variance check for the fingerprint check and the zoo *reports* the grade
rather than pretending everything in it is exact. That is #233's or a follow-up's work,
not this PR's — but the README's "the zoo keeps its determinism guarantee intact"
(README:75) reads as though the question is settled, and it is deferred. One sentence
pointing at the open shape would age better.

On consistency with #244: both declinations are on the same principle — don't weaken a
host's invariant to admit one guest — and both are argued from the host's actual code
rather than asserted. Consistent, and correctly so.

---

### The relative-instructions-path bug

**Mechanism confirmed on `main`. Workaround acceptable in this PR. Yes, it needs its own
issue — and it is worse than the PR describes.**

`core/cap_evolve/cli.py:393-398` (unchanged on `main`; this PR touches zero core files —
`git diff origin/main...HEAD -- core/` returns only the new test file):

```python
instr = spec.get("optimizer_instructions_file") or "optimizer/INSTRUCTIONS.md"
instr_p = Path(instr)
if not instr_p.is_absolute() and not instr_p.exists():
    instr_p = Path(project) / instr        # ← project is RELATIVE: ".capevolve/project"
if instr_p.exists():
    alg_cmd += ["--instructions-file", str(instr_p)]
```

`project` is deliberately relativized at `cli.py:194-196` (`proj_abs.relative_to(workdir)`
→ the literal string `".capevolve/project"`), and both probes — `instr_p.exists()` and
`Path(project) / instr` — resolve against **the cwd of `cap-evolve run`**, which is not
`workdir`. Reproduced:

```console
cwd            = /private/tmp
workdir        = /private/tmp/bugc/proj
project (rel)  = .capevolve/project
instr_p abs?    False   exists from cwd? False
fallback       = .capevolve/project/MY_INSTRUCTIONS.md   exists? False
=> --instructions-file passed? False
```

Then the failure is **silent by construction**, at two layers:
1. `cli.py:397` — `if instr_p.exists()` — a *missing* path simply omits the flag. No
   warning, no log line.
2. `harness.py:1790` — `tmpl_path = Path(instructions_file) if instructions_file else
   _DEFAULT_INSTRUCTIONS_TEMPLATE`, and `harness.py:1792-1795` wraps the read in a
   bare `except Exception: tmpl = None` — so even a *present but unreadable* file
   degrades quietly to the generic template.

And `_DEFAULT_INSTRUCTIONS_TEMPLATE` (`templates/project/optimizer/INSTRUCTIONS.md`) is
not a neutral fallback — it is tau2-shaped, and aggressively so: *"EDIT BOTH ARTIFACTS,
AND PREFER CODE"*, *"leaving `tools.py` logic untouched … is an under-used iteration"*.
So a project with no tools gets told its iteration is wasted unless it edits tool code
that does not exist. The PR's account of the consequence — the optimizer proposed a
*date-trivia assistant*, i.e. optimized the wrong task — is entirely plausible given that
text.

Severity beyond what the PR claims: `skills/phases/implement-and-check/scripts/pipeline_selftest.py:73-77`
resolves the same key **differently** — `instr_path = project / instr_rel`, project-dir-relative,
unconditionally — and *problems* on a missing file. So `cap-evolve check` **passes** a
relative `optimizer_instructions_file` that `cap-evolve run` then silently ignores. The
gate designed to catch this disagrees with the runtime about what the path means. That is
the part that makes it a latent sharp edge for every user, not just out-of-tree workdirs:
you can pass the hard gate and still optimize against the wrong instructions.

Workaround: **acceptable here.** `run.sh:96-104` appends an absolute
`optimizer_instructions_file` and documents why in six lines of comment. I verified it
resolves (`absolute: True exists: True => --instructions-file passed: True`). Widening
this PR into `cli.py` would put a core behavior change inside a docs-and-example PR that
currently has the rare virtue of touching zero core files — right call.

**File the issue.** Suggested fix, one line: resolve against the *absolute* project dir
and warn when the declared file is missing rather than silently omitting it —
`instr_p = (proj_abs / instr)` when not absolute, plus `else: print(..., file=sys.stderr)`
on the miss. The `except Exception → generic template` at `harness.py:1792` deserves the
same treatment: a declared-but-unreadable instructions file should be loud. Include the
`pipeline_selftest.py` disagreement in the issue — the two resolvers should share one
helper, or the gate is checking a different thing than the runtime does. I'd call it P1:
silent, produces confidently wrong optimization, and passes the gate meant to catch it.

---

### Canary sweep

Mine, not the PR's. Six shapes across two secret-looking names, two innocent-looking
names, and URL userinfo:

```
ANTHROPIC_AUTH_TOKEN  = sk-ant-CANARY-AAAA1111
ANTHROPIC_BASE_URL    = https://user:pw-CANARY-BBBB2222@gw.canary-host.invalid/v1
MY_INNOCENT_SETTING   = CANARY-CCCC3333
DEPLOY_REGION         = CANARY-DDDD4444
OPENAI_API_KEY        = sk-CANARY-EEEE5555
CAPEVOLVE_HOME_NOTE   = CANARY-FFFF6666
```

Full free-rung run with all six exported, then swept every written file plus both
streams:

```console
$ grep -rIl -E 'CANARY-(AAAA1111|BBBB2222|CCCC3333|DDDD4444|EEEE5555|FFFF6666)|canary-host|pw-CANARY' .
(no output)
$ grep -c -E 'CANARY|canary-host' /tmp/cr-stdout.txt /tmp/cr-stderr.txt
/tmp/cr-stdout.txt:0
/tmp/cr-stderr.txt:0
$ find . -type f | wc -l
     116
```

**Zero leaks across 116 written files, stdout and stderr.** The URL-userinfo shape
(`user:pw@host`) is the one that usually slips through a naive redactor and it did not
appear either. I also confirmed the preflight's deliberate withholding actually holds
under a failure, which is the case where an endpoint most wants to escape:

```console
$ CHEAP_REAL_API_BASE=http://localhost:59999 bash examples/cheap_real/run.sh
cheap_real preflight: the local Ollama endpoint did not answer (URLError). Start it
(`ollama serve`) or set CHEAP_REAL_MODEL to a hosted model. Endpoint withheld
(see docs/TROUBLESHOOTING.md).
```

Type name only, no URL. Correct per #134.

I confirm the separate pre-existing committed-hostname finding the PR reports elsewhere in
the repo; per the review brief I am not naming the file or the host here, and it is not
this PR's to fix.

---

### #195 / #197 compliance

**Both reproduce on the sibling branches.**

**#195** — against `origin/fix/issue-113-small-samples`
(`core/cap_evolve/splits.py:22,27`), using the committed `tasks.jsonl` and the committed
ratios:

```console
MIN_VAL_TASKS=2  LOW_CONFIDENCE_VAL_TASKS=5
cheap_real realized split: train=10 val=5 test=5
check_val_size warning: None
```

val=5, both floors cleared, **zero warnings** — exactly as claimed. And the control, so
"None" means something: a 12-task dataset at the same ratios lands val=3 and *does* warn
(`'val split (control) has only 3 tasks (< 5) — acceptance decisions are LOW
CONFIDENCE...'`).

**The test genuinely pins the floor — it is not a comment.** I dropped `tasks.jsonl` to
12 lines and re-ran:

```console
E  AssertionError: cheap_real val split is 3, below 5: every gate decision would be
   branded LOW CONFIDENCE. Add tasks rather than lowering this.
E  assert 3 >= 5
1 failed, 3 passed in 0.06s
```

Fails, on the right assertion, with a message that tells you what to do. This is the
correct way to pin a load-bearing constant and I'd like to see it copied.

One caveat: the test's `getattr(splits, "MIN_VAL_TASKS", 2)` fallback (line 51-52) means
that on `main` — where the constants do not exist — it asserts against hardcoded 2/5
rather than the real values. Deliberate and documented in the comment, and it makes the
test pass on both branches, which is the right trade pre-merge. Worth deleting the
fallback once #195 lands so the test tracks the real constants; otherwise it silently
stops testing the thing if #195 ever changes them.

**#197** — against `origin/feat/issue-142-protected-paths`. `protected_paths` is absent
from the preset (confirmed: `grep protected_paths examples/cheap_real/capevolve.yaml` →
no match), and `protect.py:145-171` confirms all three claims:
- empty list → `TamperError` (line 152-157) ✓
- non-empty declared list → `return out`, defaults never reached (line 150-151) ✓
- omitted → `[*_DEFAULT_GLOBS, *extra]` where `extra` pulls in `dataset_source` (line
  168-170) ✓

So the four grader files are covered: `adapters/adapter.py` and
`adapters/model_config.py` via the `adapters` default, `capevolve.yaml` via its own
default, and `tasks.jsonl` via the `dataset_source` fold-in. `run.sh:73-78` is what makes
that true — all four are copied *inside* the project dir. Correct, and the reasoning in
`run.sh:67-72` is accurate. See N4 for the one clarification I'd want in the wording.

---

### The #248 entry point

**The contract holds, including under #217.** From my own run:

```console
$ python -c "<scan stdout for JSON objects>"
objects on stdout: 1
last object keys: ['baseline_val','best_id','dashboard','iterations','run_dir',
                   'test_baseline_reward','test_delta','test_pass_k','test_reward']
```

**Exactly one** JSON object on stdout — #217's two-object bug does not fire here, because
`run.sh:114` passes `--dashboard "${CAPEVOLVE_DASHBOARD:-off}"`, and `off` skips the
`maybe_launch` print at `cli.py:205-207` that #217 is about. Worth noting that this is
*incidental* immunity: a user who sets `CAPEVOLVE_DASHBOARD=auto` gets two objects and
their parse breaks. Since `run.sh` is being sold as the programmatic entry point, one line
of comment at `run.sh:114` saying the `off` default is load-bearing for the
single-object contract (until #217 lands) would stop someone "improving" it later.

A naive whole-stdout `json.load` still fails, because three human-readable lines precede
the object (`preflight: OK`, `Working directory:`, `Runner model:`). That is consistent
with #116's convention only in spirit — those three lines are progress output and belong
on **stderr**; they are on stdout. The documented contract is "the *last* object on
stdout", which is honestly stated in `run.sh:7` and `README.md:42-43`, and #248 can
implement it in three lines. But moving the three `echo`s to `>&2` would make the whole of
stdout parseable and cost nothing. Recommended, non-blocking.

Could #248 genuinely call it? Yes — every knob is an env var with a default
(`run.sh:25,34-36,65,80`), `CHEAP_REAL_WORKDIR` lets the caller pin the location, and exit
codes are meaningful (I got exit 1 with a single-line reason from three separate preflight
failures). I checked `origin/feat/issue-133-quickstart` and it does not reference
`cheap_real` yet — but `core/cap_evolve/quickstart.py:67` already carries the comment
*"#124's cheap-real-run preset is one more row"*, so the seam is planned from the other
side. **Caveat: if #248 adds the hosted variant as a preset row, it inherits B1** — a
preset that silently returns all-zeros is much worse inside `quickstart` than in an
example dir.

---

### Test quality

4 new tests in `core/tests/test_cheap_real_preset.py`, all asserting **outcomes**, not
that code ran:

- `test_split_clears_the_honest_gate_floors:39` — runs the *real* `make_splits` with the
  committed ratios and asserts `val >= hard/soft`, plus train/val disjointness and test
  isolation. Verified it fails on a shrunk dataset (above). The strongest of the four.
- `test_protected_paths_is_omitted_not_empty:64` — asserts absence. Coarse (a line-prefix
  scan, so an indented `  protected_paths:` inside another block would false-positive),
  but it pins the right invariant.
- `test_dataset_source_names_the_real_file:73` — asserts the value is `tasks.jsonl`, which
  is what makes #197's fold-in cover the answer key. Right assertion.
- `test_budget_is_bounded:79` — bounds `max_usd`, `max_optimizer_usd <= max_usd`, and
  `max_iterations <= 5`. Would catch someone raising the cap toward rung 3.

They are cheap (0.03 s, no model) and each one fails for a real reason. Untested: all of
`run.sh` — see N6. The gap that matters is the YAML-rewrite heredoc, which is the fix for
bug (c) and is unpinned.

On bug (b): `capevolve.yaml:30` is now `optimizer_max_turns: 40` with the measured reason
in a comment (lines 25-29), and `optimizer_usd_per_iter: 1.0` is likewise set above the
observed spend for the same reason (lines 31-36). A user copying this example gets 40, not
12, so the trap is not reachable from the committed preset. Nothing pins it — a test
asserting `optimizer_max_turns >= 40` would be one line in the existing file and would
stop a future "tighten the caps" edit from silently reintroducing a run where every
candidate is discarded. Non-blocking, but it is the same shape as the floor test that this
PR got right, so it is a cheap consistency win.

---

### Merge-order note

**Recommend: #195 → #197 → #233 → #249 → #248.**

- **#195 and #197 first.** #249 depends on both *semantically* — its val floor is only
  real once `MIN_VAL_TASKS`/`LOW_CONFIDENCE_VAL_TASKS` exist, and `dataset_source:
  tasks.jsonl` only protects anything once `protect.py`'s fold-in exists. Neither is a
  hard dependency (the test's `getattr` fallback and the inert `dataset_source` key both
  degrade gracefully), so #249 *can* merge first — but it would be advertising two
  guarantees that are not yet enforced, and the floor test would be asserting hardcoded
  constants. Land the guarantees, then the example that claims them.
- **#233 before or after — genuinely independent.** #249 declines to be a zoo entry and
  touches no zoo file. If #233 later grows a non-deterministic grade (see above),
  `cheap_real` becomes a candidate to migrate; that is a follow-up either way.
- **#249 before #248.** #248 consumes `run.sh` as an entry point and
  `quickstart.py:67` already reserves a row for it. Landing the callee first means #248
  can add the row in the same PR rather than referencing a path that does not exist.
  **But B1 must be fixed before #248 adds a hosted row** — a `quickstart` preset that
  returns all-zeros silently is a much worse first impression than an example that does.
- The PR's own claim of "no merge order required — additive, zero core changes" is
  factually right (I confirmed `git diff origin/main...HEAD -- core/` shows only the new
  test file). The ordering above is about *claim validity*, not mechanical conflicts.

---

### Verification I re-ran

```console
$ gh pr view 249 --json headRefName
{"headRefName":"feat/issue-124-cheap-onramp"}   # 8f727e8e

$ PYTHONPATH=/tmp/rv-249/core python -m pytest core/tests -q
183 passed in 63.07s (0:01:03)
```

183 passed, 0 failed — matches the claim exactly (179 + 4). `test_dashboard_launch.py`
(flaky per #200) passed on this run.

```console
$ python -m compileall -q core examples && echo COMPILEALL_OK
COMPILEALL_OK
$ bash -n examples/cheap_real/run.sh && echo BASH_N_OK
BASH_N_OK
```

**Zero-cost path, end to end, on a real local model:**

```console
$ CHEAP_REAL_WORKDIR=/tmp/cr-canary CHEAP_REAL_PYTHON=/tmp/ce-venv/bin/python \
    time bash examples/cheap_real/run.sh
cheap_real preflight: OK
Working directory: /tmp/cr-canary
Runner model: ollama/llama3.2:3b   Optimizer: mock
{
  "run_dir": ".capevolve/run_cheap",
  "best_id": "cand_0001",
  "baseline_val": 0.0,
  "test_reward": 1.0,
  "test_baseline_reward": 0.0,
  "test_delta": 1.0,
  "iterations": 3,
  "dashboard": ".capevolve/run_cheap/dashboard.html"
}
33.947 total
```

Reproduces the PR's free-rung result exactly (same `best_id`, same
`baseline_val`/`test_reward`, same `iterations`) in 33.9 s at $0. I spent no money.

**All three preflight branches fire:**

```console
$ bash examples/cheap_real/run.sh          # system python3, no litellm
cheap_real preflight: litellm is not importable by this interpreter. `pip install
litellm`, or point CHEAP_REAL_PYTHON at one that has it.

$ CHEAP_REAL_MODEL=ollama/nonexistent-model bash examples/cheap_real/run.sh
cheap_real preflight: Ollama is up but does not have 'nonexistent-model'. Run
`ollama pull nonexistent-model` (~2 GB), or set CHEAP_REAL_MODEL.

$ CHEAP_REAL_API_BASE=http://localhost:59999 bash examples/cheap_real/run.sh
cheap_real preflight: the local Ollama endpoint did not answer (URLError). ...
```

All three exit non-zero before spending anything. To answer the brief's question
directly: **yes, the preflight checks the model is actually *pulled*, not merely that
`litellm` imports** — `run.sh:53-61` hits `/api/tags` and set-compares the model name
(with a `:latest` fallback), and the "does not have" branch above is that check firing.
That is a genuinely better guard than "does the import work". Its scope limit is B1: the
whole block is inside `if MODEL.startswith("ollama/")`, so the hosted path gets none of
it.

**Bug (c) mechanism, on `main`'s `cli.py:393-398`** — reproduced above under "The
relative-instructions-path bug"; `--instructions-file` is silently omitted and
`harness.py:1790` falls back to the tau2-flavored generic template.

**#195 floors, #197 defaults, the shrunk-dataset control, the 30-call breakdown, the
per-proposal sum, and the canary sweep** — all pasted in their sections above.

**Docs links:**

```console
broken relative links: NONE
```

OsherElhadad:
## 🔧 Review fixes

Commit `7a91227`. Every claim below is pasted literal output. **The deliverable first: the
hosted variant has now been run for real, and it reaches a non-zero sealed number.**

---

### B1 (BLOCKING) — fixed. The hosted variant, run once for real.

You were right on both counts, and on the part that stung: the preflight printing `OK`
made it worse than the bug it was built to prevent. Two causes, both in `run.sh`, no core
change:

1. `export LITELLM_DROP_PARAMS=1` — litellm honours it at `litellm/__init__.py:227`, so
   the `seed=` the bundled adapter forwards is dropped for providers that reject it.
2. The preflight's `else` branch now covers the hosted path with **one real 1-token
   completion through `model_config.llm_kwargs()` and the same `seed=` kwarg the adapter
   uses** — the wiring, not a proxy for it.

**The hosted run.** `CHEAP_REAL_MODEL=claude-haiku-4-5`, credentials read from the
environment and never printed:

```console
$ CHEAP_REAL_MODEL=claude-haiku-4-5 bash examples/cheap_real/run.sh
cheap_real preflight: OK                          # ← on stderr now
Working directory: /tmp/fx-hosted-wd              # ← on stderr now
Runner model: claude-haiku-4-5   Optimizer: mock   # ← on stderr now
{
  "run_dir": ".capevolve/run_cheap",
  "best_id": "cand_0001",
  "baseline_val": 0.0,
  "test_reward": 1.0,
  "test_baseline_reward": 0.0,
  "test_delta": 1.0,
  "iterations": 3,
  "dashboard": ".capevolve/run_cheap/dashboard.html"
}

1:38.13 total
```

`baseline_val 0.0` → sealed `test_reward 1.0`, not the all-zeros you got. Its own
accounting:

```console
$ jq .spent .capevolve/run_cheap/state.json
{
  "iterations": 3, "metric_calls": 30, "usd": 0.011478000000000002, "stall": 2,
  "runner_tokens": 3910, "runner_seconds": 91.76, "optimizer_seconds": 0.55,
  "optimizer_usd": 0.0, "optimizer_tokens": 0, "intake_usd": 0.0, "intake_seconds": 0.0
}
```

**MEASURED: 98 s, $0.011478, 30 `metric_calls`.** The six eval costs sum to the total
exactly, so nothing is unaccounted:

```console
$ grep -o '"kind": "evaluate"[^}]*' events.jsonl
"split": "val",  "tag": "seed",       "reward": 0.0, "cost_usd": 0.003466
"split": "val",  "tag": "cand_0001",  "reward": 1.0, "cost_usd": 0.000681
"split": "val",  "tag": "cand_0002",  "reward": 1.0, "cost_usd": 0.000681
"split": "val",  "tag": "cand_0003",  "reward": 1.0, "cost_usd": 0.000681
"split": "test", "tag": "FINAL",      "reward": 1.0, "cost_usd": 0.000687
"split": "test", "tag": "FINAL_seed", "reward": 0.0, "cost_usd": 0.005282

$ python -c "print(0.003466+0.000681*3+0.000687+0.005282)"
0.011478            # state.json usd: 0.011478000000000002
```

And a rollout, to show the call actually landed rather than erroring into a 0:

```console
$ cat rollouts/test/d02__FINAL__t0.json
{"input": "the 5th of July 1999",
 "rollout": {"output": "1999-07-05", "cost_usd": 0.000138, "tokens": 98, "error": null,
             "metadata": {"model": "claude-haiku-4-5", "seed": 0}},
 "score": {"reward": 1.0, "feedback": "correct"}}
```

**This retires the $0.21 ESTIMATE — and it was 20x high.** `pricing.py` assumes a
3 000-in/800-out rollout; a one-line date is far smaller. The docs now carry $0.0115 as
**measured**, and the estimate is gone rather than kept alongside. Note the seed evals
cost 8x the optimized ones ($0.003466 and $0.005282 vs $0.000681) — the unoptimized
prompt answers in prose and burns output tokens, which is the same fact the task's
premise rests on, now visible in the money.

**The preflight fails loudly on the hosted path.** A bad model name:

```console
$ CHEAP_REAL_MODEL=claude-haiku-4-5-NONEXISTENT bash examples/cheap_real/run.sh
cheap_real preflight: the hosted model 'claude-haiku-4-5-NONEXISTENT' did not answer a
1-token probe (BadRequestError). Every rollout would score 0.0. Check the model name and
its credential; endpoint/keys withheld (see docs/TROUBLESHOOTING.md). Detail:
litellm.BadRequestError: LLM Provider NOT provided. ...
```

A missing credential (run under `env -i`, so no key is present at all):

```console
$ env -i ... CHEAP_REAL_MODEL=anthropic/claude-haiku-4-5 bash examples/cheap_real/run.sh
cheap_real preflight: the hosted model 'anthropic/claude-haiku-4-5' did not answer a
1-token probe (AuthenticationError). Every rollout would score 0.0. ...
```

All five preflight branches, exit code and stdout cleanliness:

```console
exit=1 stdout_bytes=0   <- system python3 (no litellm)
exit=1 stdout_bytes=0   <- CHEAP_REAL_API_BASE=http://localhost:59999
exit=1 stdout_bytes=0   <- CHEAP_REAL_MODEL=ollama/nonexistent-model
exit=1 stdout_bytes=0   <- CHEAP_REAL_MODEL=claude-haiku-4-5-NOPE   (new)
exit=1 stdout_bytes=0   <- hosted, no credential                    (new)
```

Note **stdout is 0 bytes on every failure**. It was not before: litellm writes some of
its own banners (`Provider List: ...`) straight to stdout, which would have landed in a
caller's JSON stream. The whole preflight block is now `{ ... } >&2`.

**The `API_BASE` trap is fixed.** Your read of `model_config.py:95` was exact — the
generic `API_BASE` is read before any provider special-casing, so a hosted MODEL
inherited `localhost:11434`. Now conditional:

```console
MODEL=claude-haiku-4-5      API_BASE=<UNSET>
MODEL=ollama/llama3.2:3b    API_BASE=http://localhost:11434
```

`CHEAP_REAL_API_BASE` still wins if the user sets it explicitly, for a self-hosted
OpenAI-compatible endpoint.

---

### The nine other findings

**N1 — fixed.** `capevolve.yaml` now carries your derivation, not the 15:

```yaml
# 30 runner calls, plus 3 proposer calls. The full derivation — the baseline val eval
# and both test evals are easy to forget:
#   val  5 x trials 1 x (1 baseline + 3 candidates)  = 20
#   test 5 x (best FINAL + baseline FINAL_seed)      = 10
#                                             total  = 30   (= state.json metric_calls)
```

**N2 — fixed.** `run.sh:16` now says 3.0, matching `capevolve.yaml`.

**N3 — fixed.** "36 s" is gone from all three places. Replaced with "~30 s" plus the
spread, and stated as spread rather than as a figure: *"30.4 s, 33.9 s and 34.2 s across
three runs on two machines — single-run timing is noise at this scale, so treat it as
'about half a minute', not as a figure."* You are right that this is the #238 shape.

**N4 — fixed, and this was the most useful correction in the review.** My wording implied
the defaults cover the answer key. They do not. Verified against `#197`'s branch:

```console
=== OMITTED key (what the preset ships)
globs   : ['adapters', 'capevolve.yaml', '*gold*.json', ..., 'tasks.jsonl']
resolved: ['adapters/adapter.py', 'adapters/model_config.py', 'capevolve.yaml', 'tasks.jsonl']

tasks.jsonl matches a _DEFAULT_GLOB on its own? False
    <- it is the dataset_source fold-in that covers the answer key

=== EMPTY list -> TamperError: `protected_paths` ... is an EMPTY list ...
=== DECLARED [adapters] resolved: ['adapters/adapter.py', 'adapters/model_config.py']
    <- tasks.jsonl DROPPED
```

That third line is the one that makes the point: declaring the key at all, even
correctly, silently drops the dataset. The README, the spec comment and the test
docstring now all say the fold-in at `protect.py:168-170` is the mechanism.

**N5 — acknowledged, no change.** A conscious call, as you read it: `dataset_source` is
inert on `main` and load-bearing after #197, `proposer_model` is forward-compat with
#132. Both documented in the spec.

**N6 — fixed, on the path you named.** A test that runs the **real heredoc** (extracted
from `run.sh` by regex, so it cannot drift from a copy) over the committed preset:

```console
$ # with optimizer_max_turns lowered to 12 and optimizer_skill indented one space
E  AssertionError: optimizer_max_turns below 40 was MEASURED to make every iteration
   fail with `Reached max turns`, so no candidate survives. Raise it back.
E  assert 12 >= 40
E  AssertionError: the optimizer_skill flip no-opped
2 failed, 4 passed in 0.04s
```

Both new tests fail on the real breakage, in the shape you praised on the floor test.
The heredoc test also asserts the instructions path is absolute, referencing #252. And
`optimizer_max_turns >= 40` is the one-line consistency win you asked for.

**Nit 1 — fixed.** The README now says the local rung's 30 calls were *"metered at $0*
... *the runner's cost there is time (29.5 s, 3 863 tokens), not zero across the board"*.

**Nit 2 — fixed.** New Honest-limits entry naming `{"2": 0.0}` as pre-existing #112 and
telling the reader to ignore the `"2"` entry until it lands.

---

### The gate claim — you were right, and here is the corrected wording

I over-claimed and the two "correct rejections" I touted are vacuous, exactly as you
say: val was already 1.0, so there was nothing to reject *between*. "Real paired gate"
is gone from all five places, replaced with **"a real accept/reject decision"**. And
Honest limits now says it outright rather than leaving it to be inferred:

> **This run never exercises a rejection on merit.** Val saturates at 1.0 on iteration 1,
> so the two candidates that follow are rejected against a ceiling — there was nothing
> left to improve, not a plausible candidate found wanting. The gate *ran* and made real
> decisions; it was never asked a hard question. A task the model still gets partly wrong
> *with* the output contract would land baseline ~0.0 → best ~0.6 at the same cost and
> give the gate something to decide. Do not read this example as a demonstration that
> the significance machinery works — read `examples/skillsbench` (`num_trials: 3`) for
> that.

Same line in `GETTING_STARTED.md`. On your suggestion (2) — a harder task — I agree it
would be strictly better and I am **not** doing it in this PR: it changes the committed
dataset, and therefore the measured numbers in three docs and the split test, on the last
turnaround of a PR whose blocking bug was in the variant nobody had run. It belongs in a
follow-up with its own measurement, not bundled here. Say the word and I will file it.

**Zoo line softened.** README:75 read as settled; it now names the shape you identified:

> That is the answer for *this* PR, not the durable one: `cheap_real` is the first of a
> class [...] The open shape is an explicit `determinism: exact | statistical | none`
> grade in the manifest, where `statistical` substitutes a variance check for the
> byte-exact fingerprint and the zoo *reports* the grade rather than implying everything
> in it is exact. Deferred to #233 or a follow-up — not settled.

---

### Filed, not fixed here

- **#252** — the relative `optimizer_instructions_file` bug. Thank you for finding the
  worse half: `pipeline_selftest.py:73` resolves project-relative and *reports a
  problem*, while `cli.py` resolves against the caller's cwd and silently omits the flag,
  so **`check` passes what `run` ignores**. The absolute-path workaround stays and the
  comment now names that disagreement and references #252. The new heredoc test asserts
  the path is absolute, so the workaround cannot silently rot.
- **#251** — optimizer stderr discarded on success by three layers. Referenced in
  `run.sh` as the reason a warning alone would not have surfaced B1: the reward was the
  only signal, which is precisely why this bug survived to review.

---

### Re-verification

```console
$ PYTHONPATH=$PWD/core python -m pytest core/tests -q
........................................................................ [ 38%]
.........................................                                [100%]
185 passed in 64.06s (0:01:04)

$ python -m compileall -q core examples && echo OK
OK
$ bash -n examples/cheap_real/run.sh && echo OK
OK
```

**185 passed (183 + 2 new), 0 failed.**

**Stdout is exactly one object with `CAPEVOLVE_DASHBOARD=auto`** — your incidental-immunity
point is fixed properly, not by leaving `off` load-bearing. `run.sh` filters the CLI's
stdout to the last object, so it holds regardless of the dashboard mode:

```console
$ CAPEVOLVE_DASHBOARD=auto bash examples/cheap_real/run.sh
objects on stdout: 1
whole-stdout json.loads: OK
keys: ['baseline_val','best_id','dashboard','dashboard_server','iterations','run_dir',
       'test_baseline_reward','test_delta','test_pass_k','test_reward']
```

Note `json.loads(stdout)` now succeeds outright, because the three progress lines moved
to stderr — the naive parse you noted was failing is the one that now works. The filter
carries a comment saying to delete it when #217 lands.

**Free rung, unchanged** (33.6 s this run, $0):

```console
{"run_dir": ".capevolve/run_cheap", "best_id": "cand_0001", "baseline_val": 0.0,
 "test_reward": 1.0, "test_baseline_reward": 0.0, "test_delta": 1.0, "iterations": 3}
metric_calls: 30
```

**#195 compliance** — against `origin/fix/issue-113-small-samples`:

```console
MIN_VAL_TASKS=2  LOW_CONFIDENCE_VAL_TASKS=5
cheap_real realized split: train=10 val=5 test=5
check_val_size warning: None
CONTROL 12 tasks: val=3  warning: val split has only 3 tasks (< 5) — acceptance
  decisions are LOW CONFIDENCE. ...
```

**#197 compliance** — pasted under N4 above, including the new
declared-list-drops-the-dataset control.

**`cap-evolve check` on the hosted run's project dir:**

```console
{"ok": true, "stubs": [], "problems": [],
 "notes": ["tasks('val') -> 20 task(s)", "scorer deterministic (probe reward=0.0000)",
           "materialize() callable (dry-run into temp copy; host untouched)"]}
```

**No canary leaks — including on the newly-added hosted path,** which is where it now
matters most, since the preflight touches real credentials and embeds provider errors in
its message. Your six shapes on the free rung:

```console
$ find . -type f | wc -l
     116
$ grep -rIl -E 'CANARY-(AAAA1111|...|GGGG7777)|canary-host|pw-CANARY' .
(no output)
$ grep -c -E 'CANARY|canary-host' stdout stderr
stdout:0   stderr:0
```

And two new shapes aimed at the new code, both forcing the preflight to *fail* so the
error message is the escape route:

```console
$ ANTHROPIC_API_KEY=sk-ant-CANARY-HHHH8888 \
  ANTHROPIC_API_BASE='https://user:pw-CANARY-IIII9999@gw.canary-host.invalid/v1' ...
cheap_real preflight: the hosted model 'anthropic/claude-haiku-4-5' did not answer a
1-token probe (InternalServerError). ... Detail: litellm.InternalServerError:
AnthropicException - [Errno 8] nodename nor servname provided, or not known.
canary count: stdout:0  stderr:0

$ # reachable endpoint, so the 404 path — the one that usually echoes the URL
$ ANTHROPIC_API_BASE='http://user:pw-CANARY-LLLL5678@127.0.0.1:11434/CANARY-PATH-MMMM9012'
cheap_real preflight: ... (NotFoundError). ... Detail: litellm.NotFoundError:
AnthropicException - 404 page not found
canary count: stdout:0  stderr:0
```

Zero leaks, including the URL-userinfo shape, on the failure path. The preflight replaces
each of six credential/endpoint vars in the provider's own message before printing it.

**Broken relative links: NONE.**

---

Every finding is fixed or, in one case (N5), consciously kept with the reason. The one
suggestion I declined — the harder task — I declined for scope, not merit, and it should
be filed. Thanks for running the hosted variant; that finding was the whole value of this
review, and it is exactly the failure mode this example claims to prevent.

OsherElhadad:
Closing — the value here does not clear the cost, and the cheap rung this claims to add
is toy_calc with a network call.
The onboarding cliff is real (GETTING_STARTED still ends at a link table, and the next
real run is tau2 at hours/~$148), but this is not the fix:
- **The demo is toy_calc again.** A deliberately sabotaged seed prompt on 20 exact-match
  date tasks going 0.0 → 1.0 on iteration 1 shows the same gate/split-seal/sealed-test
  shape `examples/toy_calc` already shows in a 23-line `run.sh`, for $0 with no
  prerequisites. The only new thing proved is that the model call path works, which is
  `cap-evolve doctor`'s job.
- **The "$0" path is not free of setup.** It requires installing Ollama and pulling
  `llama3.2:3b` — a heavier prerequisite than the repo has today. The measured real-agent
  path is $0.5427 / 5.4 min (293s of it optimizer latency), not "free minutes."
- **734 added lines for an onboarding step**, including a 179-line `run.sh` with 7 env
  knobs, and +106 lines onto a 93-line GETTING_STARTED. Both existing example runners are
  23 lines. This is the opposite of the on-ramp being easy to read.
- Issue requirement 3 (ship a committed run artifact for the dashboard) is not done.
- One of the three "bugs found" — relative `optimizer_instructions_file` — is already
  handled on main at `core/cap_evolve/cli.py:871-876`, so `run.sh` carries a workaround
  for nothing.
Keeping #124 open. If we want a real rung 2, the shape is a handful of lines in
GETTING_STARTED pointing `examples/toy_calc`'s spec at a real `MODEL=` via the existing
provider-agnostic `model_config.py` switch — no new example dir, no new prerequisite,
no new run.sh.

PR #266, opened 2026-07-31
feat(spreadsheetbench): held-out split, seeds, 30 turns + parallel recalc for SkillOpt comparison

eranra:
Closes the gaps that stop the full tier from producing a SkillOpt-comparable number (internal issue #5, arXiv 2605.23904).

**Zero changes under `core/`** — `split_ids_file` and `split_seed` were already supported there, so this is wiring plus benchmark-local work.

| Change | Files | Core? |
|---|---|:--:|
| Held-out split + generator | `ci/benchmarks/spreadsheetbench/{full/split_ids.json,utils/make_split.py}` | No |
| Split hook + `split_seed` | `ci/benchmarks/lib/run_suite.sh` (shared, no-op until opted into) | No |
| 30 turns | `run_suite.sh`, inside the `spreadsheetbench)` arm only | No |
| Parallel recalc | `templates/adapters/spreadsheetbench/adapter.py` | No |
| Scope labeling | `ci/benchmarks/spreadsheetbench/README.md` | No |
| `BENCH_SPLIT_SEED` env | `.github/workflows/benchmarks.yml` (1 line) | No |

## Why the full tier couldn't be compared

Every tier used the **no-holdout FIT split** (`train == val == test == all tier tasks`). Honest for smoke — the report labels it a FIT metric — but SkillOpt reports **held-out test** scores, so putting our fit number beside their 80.7 would be apples-to-oranges and optimistically biased.

## What the paper says — and what it doesn't

Reconstructed from the paper:
- **`split_seed = 42`**, stated globally.
- **2:1:7** train/selection/test — Appendix C's "default … when no benchmark-specific split is stated", repeated by the train-size ablation protocol.
- SpreadsheetBench run as "multi-round codegen with up to **30 turns**".

**Not published:** any SpreadsheetBench-specific split, the task count, or a per-task trial count. Table 2's caption says 4:1:5 — but for an *ablation panel*.

> So `full/split_ids.json` is a **documented reconstruction of SkillOpt's stated default, not a reproduction of their split.** The tier README says so explicitly, and `make_split.py --ratios 4,1,5 --write` switches to the alternative reading in one command.

## The split

182 / 91 / 639 over all 912 tasks. The loader **fails loudly** on overlap, missing or unknown tier tasks, or an empty val/test split — a stale split quietly evaluating a different task set would invalidate an entire comparison run.

## Parallel recalc — the full-tier bottleneck

The vendored `just_open_libreoffice` lets soffice use its default user profile, so concurrent instances conflict and it had to sit behind a process-global lock: **912 × 3 = 2,736 serialized soffice startups per evaluation**, completely unaffected by `SPREADSHEETBENCH_CONCURRENCY`. `_recalc_workbook` gives each call its own `-env:UserInstallation` profile. It also captures output (the vendored helper prints on every failure path), leaves the workbook intact when recalc fails, and reports failure rather than raising.

## Two things worth reviewing closely

- **`actions: [edit]` is not machine-enforced.** Nothing in `core/` or `skills/` reads it. Scope comes from `capabilities: [system-prompt]` over a single `prompt.md` — a *closer* match to SkillOpt's "skill text only" than `skill-package`, which also ships scripts. Documented rather than faked.
- **Do not equate our seed baseline with their 41.8.** Their "No skill" is the frozen model with *the benchmark's default system prompt*, and their own run started from a 224-token seed skill whose score is unreported. Ours is ~190 tokens, so seed-to-seed the starting points are comparable — but the 41.8 cell is not our baseline. Compare our sealed test % against **80.7 / 73.6 (GEPA) / 72.9 (human)**.

## Tests

`234 passed` (CI-equivalent, no pandas) · `239` with pandas+openpyxl · adapter self-checks OK · actionlint and `bash -n` clean.

- The split hook is tested by **executing the real bash block** from `run_suite.sh`: every guard rail, the actual 912-task file, and a regression test that a tier *without* a committed split still gets the byte-identical no-holdout split (so tau2/swebench/skillsbench are provably unaffected).
- Recalc is tested against a **fake soffice** — profile uniqueness, 24-way concurrency, in-place replacement, timeout/nonzero-exit/missing-binary handling, and stdout silence — so it verifies anywhere, no LibreOffice needed.
- The committed split is asserted to be byte-reproducible from the generator, so the provenance claim is checkable.

## Not verified here — please gate on this

**Recalc equivalence on real workbooks needs the runner** (no LibreOffice or dataset locally). I'd run the **smoke tier** before merging and confirm per-task rewards match the last serialized run ([30641201209](https://github.com/skillberry-ai/cap-evolve/actions/runs/30641201209): baseline 0.440, best `cand_0001`, finalize 0.553).

Also: **re-derive cost/runtime from a ~50-task pilot** before launching 912 × 3 seeds. Per-iteration evaluation only hits the 91-task selection split (with `focus: all` train is never evaluated; finalize evaluates test twice), so ~2,279 rollouts ≈ $110–150 and ~10 h per seed — **but that anchor predates `MAX_TURNS=30`**, which raises both. The shared gateway also has a spend cap that 429s a four-figure run.

## How to run it

Set repo variable `BENCH_SPLIT_SEED` to 42, then 43, then 44 between the three dispatches:

```
benchmark=spreadsheetbench  tier=full  trials=1  iterations=10
agent_model=azure/gpt-5.5   optimizer_model=claude-opus-4-8
gate_k_se=0.2  algorithm_focus=all  optimizer_usd_per_iter=0  optimizer_max_turns=80
```


eranra:
## Cross-benchmark isolation audit (requested before merge)

Verified **empirically**, not by inspection: the old (`main`) and new split blocks were extracted and executed side by side for every benchmark × tier, and their output compared.

| bench | tier | split output |
|---|---|---|
| tau2 | smoke | **identical** |
| tau2 | full | **identical** |
| swebench | smoke | **identical** |
| swebench | full | **identical** |
| skillsbench | smoke | **identical** |
| skillsbench | full | **identical** |
| spreadsheetbench | smoke | **identical** |
| spreadsheetbench | full | **changed — intended** |

The single change is the one this PR exists for:

```
before (no-holdout FIT): train=912 val=912 test=912   train==val==test: True
after  (held-out)      : train=182 val=91  test=639   disjoint: True
```

### Every shared-file hunk, accounted for

**1. Split hook — generic feature, opt-in per tier.** Keyed on `[ -f "$BASE/split_ids.json" ]` where `BASE=$REPO/ci/benchmarks/$BENCH/$TIER`. Any benchmark can adopt a held-out split by committing that file; no benchmark name appears in the logic. A test asserts no other tier ships one, so the branch is dead for them, and another test asserts the no-committed-split path still produces the byte-identical no-holdout split.

**2. `split_seed` in the generated spec — provable no-op.** The only spec line added is `split_seed: ${SPLIT_SEED:-0}`. Core reads it as `spec.get("split_seed", 0)`, so an explicit `0` is identical to omitting it:

```
spec={}                  -> seed used = 0
spec={'split_seed': 0}   -> seed used = 0
```

**3. `MAX_TURNS` — confined to the `spreadsheetbench)` case arm.** Every occurrence classified by region; all five are inside that arm. The only shared-tail hit is the pre-existing `optimizer_max_turns`.

**4. Workflow — 10 comment lines plus one generic line**, `SPLIT_SEED: ${{ vars.BENCH_SPLIT_SEED || '0' }}`. Applies to all benchmarks, no benchmark name in the code, and unset ⇒ `0` ⇒ prior behaviour everywhere.

**5. `core/cap_evolve/` — untouched.** Only `core/tests/` gains new files.

### Benchmark names in generic regions

Three exist, all **pre-existing and untouched** by this PR, and all enumerate the four benchmarks equally: the `BENCH` argument validation, the `benchmark` input's `options:` list, and the planner's `BENCHES` list.

Every SpreadsheetBench-specific *executable* line added by this PR is inside the `spreadsheetbench)` arm or under `ci/benchmarks/spreadsheetbench/` / `templates/adapters/spreadsheetbench/`. Remaining mentions in shared files are comments — including the pointer to `make_split.py` as an example generator for the generic hook.

`234 passed` (CI-equivalent) / `239` with pandas; actionlint and `bash -n` clean.

The one item still not verifiable off-runner is unchanged: **recalc equivalence on real workbooks.** A smoke run before merge is still the right gate.

PR #290, opened 2026-08-03
feat(telemetry): add MLflow and OpenTelemetry observability

AIMikav:
## Summary

- Add optional `RunObserver` protocol to core and `capevolve_telemetry` package (follows `capevolve_harbor` pattern)
- **MLflow observer**: metrics, params, tags, candidate diffs, JOURNAL.md, optimizer reports as artifacts; `autolog: true` for per-LLM-call tracing
- **OTel observer**: spans exported via OTLP HTTP to MLflow's `/v1/traces` or any standalone collector
- Wired into all skill scripts (baseline, hill-climb, gepa, skillopt, finalize) — ~10 lines each, all `try/except` guarded
- Off by default — enable by adding `observers:` section to `capevolve.yaml`
- `run_name` config key for custom MLflow run naming; spec metadata logged as tags for filtering

## Test plan

- [x] 29 unit tests (9 core dispatch + 20 telemetry package, all mocked)
- [x] 407 existing core tests pass (1 pre-existing failure unrelated)
- [x] toy_calc end-to-end: metrics in Runs tab, spans in Traces tab, diffs in Artifacts tab
- [x] tau2-bench smoke run with 14B Qwen on OpenShift cluster


OsherElhadad:
## Review: `feat(telemetry): add MLflow and OpenTelemetry observability`

Reviewed the design, code, tests, docs and security surface, then **ran it end-to-end twice on real τ²-bench airline** — once without observers (control) and once with `mlflow` + `otel` enabled — using a real agent (`aws/gpt-oss-120b` via the IBM litellm gateway) and a real optimizer (`claude-opus-4-8` through Claude Code). Screenshots and findings below.

**Verdict:** the architecture is right and I want this in. The `RunObserver` protocol is the correct seam and the "core stays zero-dep, backends live outside" split matches the `capevolve_harbor` precedent exactly. But I hit **three bugs that silently disable telemetry** — including one I tripped over on my very first real run — plus coverage gaps. All are small fixes.

---

### It works — real run evidence

Both runs completed the full `baseline → step → finalize` lifecycle. The control run behaved identically to before the PR, and the observed run produced 27 metrics, 30 traces and 2 artifacts.

**Runs tab** — one clean run per `cap-evolve run`, named from `run_name`:
runs

**Run detail** — metrics, split params, and the spec metadata as tags (`algorithm_skill: hill-climb`, `optimizer_model: claude-opus-4-8`, `capabilities: system-prompt+tools`, `gate_mode: paired`). The tag-based filtering story genuinely works:
run metrics

**Model metrics** — all 27 series charted against `Step`:
model metrics

**Traces tab** — `autolog: true` captured every real agent + user-simulator LLM call with tokens and latency (these are actual τ²-bench airline turns: `YOU ARE BEING TRANSFERRED TO A HUMAN AGENT`, `###TRANSFER###`), and the OTel `capevolve.*` spans arrive over OTLP HTTP to `/v1/traces` in the same tab:
traces
otel span

**Artifacts tab** — the candidate diff (31.57 KB of real tool-code edits) and the optimizer error text:
artifacts

Cross-process resume works as designed — `observer_state.json` carried the MLflow `run_id` **and** the OTel `trace_id` from `baseline` into `hill-climb` and `finalize`, and `set_terminated` correctly fired only on `finalize` (`STATUS: FINISHED`, `best_id`, `test_reward`/`test_delta` all present).

Also confirmed **no regression**: `core/tests` has the same 5 pre-existing failures on `main` and on this branch (383 vs 393 passed — the delta is skips, not breakage), and the 29 new tests pass.

---

### Blocking bugs

**1. Setting `run_name` silently disables all telemetry.** `caps` is assigned inside `if not run_name:` but read unconditionally 13 lines later, so the documented `run_name` key raises `NameError` — swallowed by the bare `except`. This is not theoretical: it is exactly what happened on my first real run. No experiment, no run, no `observer_state.json`, no error message.

```python
# skills/phases/baseline/scripts/run.py:69 — inside the `if not run_name:` block
            caps = full_spec.get("capabilities")
...
# :82 — read unconditionally → NameError whenever run_name IS set
        if caps:
```

Fix: hoist `caps = full_spec.get("capabilities")` above the `if not run_name:`. Your own README documents `run_name` as the headline ergonomic feature, so this path is the common one.

**2. Without PyYAML, `observers:` parses to a dict and is silently ignored.** `core` is deliberately zero-dep and `specfile.read_yaml` falls back to a minimal reader that cannot do lists-of-dicts. I verified the fallback returns a single **merged dict** with `'- backend'` as a key:

```
observers -> {'- backend': 'otel', 'tracking_uri': ..., 'autolog': True, 'service_name': ...}
```

`load_observers` then iterates dict *keys* (strings), matches no backend, returns `[]`. The template comment says "Requires PyYAML (pulled in by mlflow)" — but that's a load-bearing hard requirement, not a note, and the failure is silent. Either add `pyyaml` to the `mlflow`/`otel` extras explicitly, or detect the dict shape and emit a real warning.

**3. Two `return 0` paths in baseline skip `close_observers()`.** The `--resume` fast-path (:65) and `--reuse-baseline` (:80) both return before the `close_observers()` at :147. Since `step_counter` and `run_id` are only persisted *in* `close_observers`, `observer_state.json` is never written — so the MLflow run created moments earlier is orphaned and every later phase logs nothing. `--resume` is a common path on expensive runs, which is exactly when you most want telemetry.

I applied all three fixes locally to get my run working; happy to push them as a follow-up commit if you prefer.

---

### Generalizability — the main gap

The `RunObserver` seam itself is genuinely generic: it hooks `RunDir.log_event`, so it is agnostic to benchmark, agent, capability and optimizer, and my tau2 run needed zero telemetry-specific code. That part of the design is right. But the **wiring** is not general:

- **Only 3 of 5 algorithms are wired.** `agent-optimize` and `evograph` have no observer block, and since `_resolve_algorithm` passes unknown names straight through, `algorithm_skill: evograph` runs via `cap-evolve run` with telemetry completely silent — no error, no hint.
- **`agent-optimize` emits `accept`/`reject`, which map to nothing.** The MLflow observer's `on_event` has no branch for them, so the agent-driven orchestration mode records no acceptance signal even if it were wired.
- **Standalone phases are unwired.** `evaluate`, `report`, `diagnose`, `gate` have no observer block, so the `/cap-evolve:<phase>` chain that `RUN.md` advertises as a first-class path produces no telemetry at all. Only `cap-evolve run` is instrumented.
- **Per-candidate metric names fragment the curve.** `evaluate` builds `f"{split}_{tag}_"` where `tag` is the candidate id, so a 20-iteration run yields `val_cand_0001_reward`, `val_cand_0002_reward`, … — twenty unplottable one-point series instead of one comparable curve. Verified in my run: `val_seed_reward` and `val_cand_0001_reward` are separate metrics. Better to log a stable `val_candidate_reward` at `step=N` and keep the candidate id as a tag (which you already do).

Since the stated goal is "follows the `capevolve_harbor` pattern", consider moving the load/close block into one shared helper (e.g. `RunDir.attach_observers(spec_path)`) instead of a copy-pasted ~10-line `try/except` in each of 5 scripts. That is the difference between "wired where someone remembered" and "wired by construction" — and it's the same amount of code.

---

### Correctness / quality

- **`_link_traces_to_run()` claims traces it does not own.** It tags *every* unlinked trace in the experiment with this run's id. Two runs sharing `experiment_name: cap-evolve` — the documented default — will steal each other's traces. I have direct evidence: two spans in my experiment are still attributed to a probe run I had **already deleted**. It also runs on every phase close, not once.
- **Unconditional `time.sleep(3)` in `close()`.** Paid on every phase close (3 phases minimum) even when tracing is off, and it happens before the `if not self._tracking_uri: return` guard, so even the no-op path sleeps. Move the guard first.
- **`step_counter` persists only at `close_observers()`.** A crash or `Ctrl-C` mid-loop loses it; the resumed run restarts at `step=1` and overwrites the existing curve. Consider saving state after each `step` event.
- **`trace.set_tracer_provider()` is a global one-shot.** It hijacks any provider the host process already configured, and a second call is ignored with a warning. Worth a comment at minimum, since adapters legitimately run inside larger apps.
- **`JOURNAL.md` is never actually uploaded.** It is read during `log_event("step")`, but the harness writes it at the same instant — my run's journal has mtime *equal* to the step timestamp, and the artifact list contains only `cand_0001.diff` and `cand_0001_error.txt`. The README advertises `JOURNAL.md` as a logged artifact. Read it on the *next* event or at close.
- **`.diff` is not previewable in MLflow.** Minor, but naming it `cand_0001.diff.txt` would make it readable in the UI instead of download-only.
- **Broad `except Exception: pass` everywhere.** Correct instinct — telemetry must never break a run — but currently indistinguishable from success. Bugs 1–3 were all invisible for exactly this reason. One `sys.stderr.write` or a `telemetry.log` in the run dir behind a debug flag would have saved me an hour.

### Security

No real concerns. No credentials are read or logged; `subprocess.run(["diff", ...])` uses a list form with no shell and a timeout; temp dirs are cleaned in `finally`. Two things worth noting in docs rather than code: candidate diffs and optimizer reports are **uploaded to the tracking server**, so a private prompt/tool-code capability leaves the machine the moment `observers:` is enabled; and `tracking_uri` over plain `http://` sends it in the clear. Worth one line in the README.

### Tests

29 tests, well-targeted at the event→metric mapping and the crash-isolation contract, and correctly mocked. Two gaps:
- **They never run in CI.** `ci.yml` runs only `python -m pytest core/tests -q`, so the 20 `capevolve_telemetry/tests` are dead weight in automation. The 9 core dispatch tests do run.
- **`test_otel_observer.py::test_state_round_trip` fails with `ModuleNotFoundError: No module named 'opentelemetry'`** when otel isn't installed, because `from_state` calls `_setup()` eagerly. Needs an `importorskip` or a lazier `from_state`.

Also `pip install -e ./capevolve_telemetry` **fails** — `does not appear to be a Python project` — because of the flat `package-dir = {"capevolve_telemetry" = "."}` layout. Non-editable install works. Worth noting in the README since editable is the natural dev instruction.

### Docs

Package README is genuinely good — the config tables and metric map are the right level of detail. But the feature is invisible from everywhere a user would actually look: **zero mentions** of MLflow/OTel/observers in `README.md`, `docs/INSTALL.md`, `docs/ARCHITECTURE.md`, `docs/EXTENDING.md` or `docs/TROUBLESHOOTING.md`. Given the three silent-failure modes above, `TROUBLESHOOTING.md` needs a "telemetry isn't showing up" entry more than most features do.

---

### Summary

| Area | Assessment |
|---|---|
| Architecture | Right seam, right dependency direction, matches harbor precedent |
| Works end-to-end | Yes — verified on real τ²-bench with real agent + real optimizer |
| Blocking bugs | 3, all silent, all small fixes |
| Generalizability | Protocol yes; wiring covers only `cap-evolve run` + 3/5 algorithms |
| Security | Fine; two doc notes on data egress |
| Tests | Good coverage, but not in CI + 1 failure without otel installed |
| Docs | Package README strong; absent from all main docs |

Fix the three silent failures and add the CI line, and this is ready. The wiring generalization and `_link_traces_to_run` scoping I would want before anyone runs two experiments concurrently.

<sub>Run details: τ²-bench airline (`sierra-research/tau2-bench`), agent + user-sim `aws/gpt-oss-120b`, optimizer `claude-opus-4-8` via Claude Code, `hill-climb`/`focus=all`, `capabilities: [system-prompt, tools]`, MLflow 3.15.1 with sqlite backend. Optimizer spent \$3.05 / 32,870 tokens on the observed iteration. Both runs used a 2-task smoke split, so the reward numbers (val 1.0 → test 1.0, gate-rejected) are a plumbing check, not a quality result.</sub>

OsherElhadad (approved):

Issue #276, opened 2026-08-01
The optimizer can edit the harness that scores it, and the gain is recorded as a capability improvement


eranra:
Split out of #272 so it is not closed with the SpreadsheetBench fixes — those address the symptom by instruction; this is the missing enforcement.

## What happened

In [run 30691123806](https://github.com/skillberry-ai/cap-evolve/actions/runs/30691123806), `cand_0002` raised val reward from 0.000 to 0.567. Its own `PROCESS.md`:

> | 1 | in-body guard (permission normalization) | `adapter.py::run_target` (BOTH the template `templates/adapters/spreadsheetbench/adapter.py` and the runtime copy `.capevolve/project/adapters/adapter.py`) | After `host_out_dir` is computed, a guarded chmod ORs `0o055` onto `data_dir` … |
>
> `prompt.md`: **left unchanged** (identical to champion/seed).

The capability under optimization was `[system-prompt]` — a single `prompt.md`. The optimizer diagnosed a genuine environment fault correctly (better than we had), but it fixed it by **editing the adapter that runs and scores the benchmark**, including the repo template outside the candidate directory. The harness then recorded the resulting jump as a capability gain, and the run's headline `0.000 → 0.567` is entirely infrastructure repair.

## Why nothing caught it

`harness._CAP_DIFF_SKIP` / `_SNAPSHOT_IGNORE` only decide what is *hidden from diffs* and *excluded from snapshots*. There is no check that an iteration's edits stayed inside the capability, so:

- an out-of-capability edit is invisible in the iteration diff the reviewer reads,
- the reward delta it produces is attributed to the capability,
- and the gate accepts it, because the gate only sees the number.

This is the honesty-critical case the acceptance gate exists for. It is adjacent to #239 (`run_step` swallowing `TamperError`) — the machinery for "this run is not trustworthy" partly exists but is not wired to this class of edit.

## Suggested direction

Detect and refuse, rather than trust the prompt. After each iteration, diff the candidate working directory against what the optimizer was permitted to touch (the capability path) and treat any modification outside it as a tamper: fail the iteration loudly instead of scoring it. The optimizer's working copy of `adapters/` is already known to the harness, so the permitted set is derivable, not guesswork.

[...]

PR #293, opened 2026-08-04
fix(spreadsheetbench): stop miseducating the optimizer, stop forgetting between runs, warn when the gate is too strict

eranra:
Three defects in the optimization **process**, found by cross-analysing pilots [30799393875](https://github.com/skillberry-ai/cap-evolve/actions/runs/30799393875) and [30890657732](https://github.com/skillberry-ai/cap-evolve/actions/runs/30890657732) — which between them give **four capability states scored on the same 50 val tasks**.

That 2×2 sets the context for everything below:

| category | tasks |
|---|---|
| **Always** pass (all 4 states) | 22 |
| **Never** pass (all 4 states) | **18** |
| Volatile | **10** |

Per-state 3/3 counts: **25, 28, 26, 29**. The entire observed range across four very different capability texts is **4 tasks out of 50**.

## 1. The TYPE diagnostic's advice was backwards half the time

PR #289 appended one **static** clause to every mismatch, whichever way it went:

```
 — write real numbers/dates, not their text form.
```

It fired on 6 tasks and was wrong on two:

| task | mismatch | old advice |
|---|---|---|
| 59259, 55931, 56427, 51958 | `str` where number/date expected | ✅ correct |
| **57232** | `float` where **`str`** expected | ❌ backwards |
| **50630** | `datetime` where **`str`** expected | ❌ backwards |

Pilot 30799393875's own `PROCESS.md` had already root-caused `50630` correctly — *"GT keeps the fragment as the original text string"* — so the optimizer was reading its correct diagnosis and our contradictory advice at the same time.

Feedback that points the wrong way is worse than none: the optimizer writes capability rules from it, and a rule pushing the wrong direction can **regress a task that currently passes**. Advice is now direction-aware, and two non-textual types (`int` vs `float`) get no directional claim at all.

## 2. Learning was not cumulative across runs

Every run started from the pristine seed, so each explored a different subset of rules and forgot the rest:

| token | run 1 `cand_0001` | run 2 `cand_0002` |
|---|---|---|
| `_xlfn` | 4 | **0** |
| `TEXTJOIN` | 3 | **0** |
| `volatile` | 3 | **0** |

Run 1 learned *"spill/volatile functions do not survive LibreOffice recalculation — write the literal"* and fixed `47741` and `51958`. Run 2 never rediscovered it and **both regressed** — the `0,3,0,0` pattern in the volatility matrix. So 2 tasks (0.04 val) were lost to *forgetting*, and part of what looked like run-to-run noise was lost knowledge rather than variance.

`SB_WARM_SEED=1` now starts from `seed_capability_warm/` — a **verbatim optimizer artifact, never hand-edited** (see its `PROVENANCE.md`). Hand-authoring rules there would make the next measured "optimizer gain" partly ours, which is exactly #276.

**Honesty controls**, because a warm-started `base→opt` delta is *not* comparable to a pristine run's (absolute higher, measured gain smaller):

- opt-in only; default stays pristine
- mutually exclusive with `SB_EMPTY_SEED` (the no-skill control) — `run_suite.sh` refuses both
- loud log disclosure, and `runmeta.json` records `"warm_seed"` so it travels with the number
- enabled for `pilot`; **`full` stays pristine, pinned by a test**, so the headline stays from-scratch

## 3. The gate has been too strict for hard scoring — silently, in every pilot

The committed overrides already said this:

> *"hard scoring makes per-task reward Bernoulli, so the gate's SE is WIDER than under soft. Pair it with a lower `gate_k_se` dispatch input (0.2) or almost nothing will be accepted"*

…but the workflow **always** sets `GATE_K_SE`, so `overrides.env` cannot enforce it and nothing warned. Both pilots ran `k_se=1.0` against `SB_SCORING=hard`, and run 30890657732's `cand_0003` scored **0.600 — above its accepted champion's 0.580 — and was rejected** on a delta of 0.020.

`run_suite.sh` now emits a `::warning::` when `SB_SCORING=hard` meets `gate_k_se >= 0.5`, and `runmeta.json` records `gate_k_se`.

## Also recorded, for scope realism

- **Coverage is essentially solved.** In the champion's 21 failures nearly all have `span == filled` (1/1, 50/50, 96/96, 146/146, 190/190); only `56427`, `30709` and `325_44` still short-fill. Further coverage work is not worth it.
- **LibreOffice recalc is not broadly failing** — only `73_45` mentions it — so that understatement risk isn't currently biting.
- **18 tasks have never passed under any condition**, several asking for VBA macros openpyxl can't write. Even a perfect prompt caps around ~0.64 on this val set.

## Tests

`422 passed, 1 skipped`. `actionlint` clean on `benchmarks.yml`. An existing guard on committed overrides caught the new key and was updated deliberately with its reason, as its own assert message asks.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -395,6 +395,12 @@ jobs:
             SOURCE_TEXT="manual ($BRANCH_NAME)"; pr="null"
           fi
           export SOURCE_TEXT
+          # SB_* flags live in the tier's committed overrides.env, which only run_suite.sh
+          # sources — so resolve them here too, or the disclosure field below would always read
+          # as "not warm" even on a warm-started run. An env var still wins over the file.
+          ov="$GITHUB_WORKSPACE/ci/benchmarks/${{ matrix.bench }}/${{ matrix.tier }}/overrides.env"
+          warm_file=$(sed -n 's/^[[:space:]]*SB_WARM_SEED=\([01]\).*/\1/p' "$ov" 2>/dev/null | tail -1)
+          if [ "${SB_WARM_SEED:-${warm_file:-0}}" = "1" ]; then warm_json=true; else warm_json=false; fi
           # Emit both free-form fields as JSON *string literals*. Shell quoting alone is not
           # enough: a branch name containing a double quote would close the string early and
           # produce a runmeta.json the aggregate job cannot parse (verified locally).
@@ -416,6 +422,8 @@ jobs:
             "trials": ${NUM_TRIALS:-10},
             "agent_model": "${AGENT_MODEL:-aws/gpt-oss-120b}",
             "optimizer_model": "${OPTIMIZER_MODEL:-claude-opus-4-8}",
+            "gate_k_se": ${GATE_K_SE:-1.0},
+            "warm_seed": $warm_json,
             "conclusion": "${{ job.status }}"
           }
           JSON
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,41 @@ All notable changes to cap-evolve are documented here. The format follows
 [Semantic Versioning](https://semver.org/) (currently `0.x` — anything may change).
 
 ## [Unreleased]
+### Fixed
+- **The TYPE diagnostic's advice was unconditionally backwards half the time.** PR #289 appended
+  one static clause to every type mismatch — *"write real numbers/dates, not their text form"* —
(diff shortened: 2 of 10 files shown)


eranra:
Landed on `main` as 6dfd3e7d.

Merged via SSH rather than `gh pr merge`: the OAuth token lacks the `workflow` scope required to write `.github/workflows/benchmarks.yml` to the default branch, so the API-side squash was refused. Content is identical to this branch's commit, rebased onto e1aa421b (#292 had landed meanwhile) and re-verified: 422 passed, 1 skipped.

PR #294, opened 2026-08-04
ci: fix aggregate history failure when no benchmark artifacts exist

eranra:
## Problem

When a PR has no `benchmark-smoke` or `benchmark-full` label, all `bench` matrix jobs are skipped entirely (the job-level `if` is false). The `aggregate history` job still runs because of `if: always()`, but `actions/download-artifact@v4` exits non-zero when its `pattern: benchmarks-*` matches **zero** artifacts — causing the aggregate job to fail with no actual benchmark work having been done.

## Fix

Add `if-no-files-found: ignore` to the **Download all benchmark artifacts** step. The **Build records** step already handles an empty `_artifacts/` directory gracefully (`have=0`, loop body never executes, no `exit 1`).

## Change

```diff
-        uses: actions/download-artifact@v4
-        with:
-          path: _artifacts
-          pattern: benchmarks-*
+        uses: actions/download-artifact@v4
+        with:
+          path: _artifacts
+          pattern: benchmarks-*
+          merge-multiple: false
+          if-no-files-found: ignore
```

No logic changes — only the download step's error-on-empty behaviour is suppressed.


eranra:
Closing — branch was based off feat/spreadsheetbench-adapter instead of main. Recreating correctly.

PR #309, opened 2026-08-07
fix(harness): an UNGRADEABLE trial is missing data, not a reward of 0.0

eranra:
Sibling of #307. That PR closed this hole on the **rollout** side; this is its **scoring** side.

## The bug

`errored` was synthesised only from `rollout.error` ([harness.py:241](https://github.com/skillberry-ai/cap-evolve/blob/main/core/cap_evolve/harness.py#L241)). But a rollout can **succeed** — target ran, tokens spent, artifact produced — and then *scoring* can fail. There's no `rollout.error` in that case, so the scorer's 0.0 was averaged in as a real measurement.

[Run 31161200250](https://github.com/skillberry-ai/cap-evolve/actions/runs/31161200250) (`smoke / swebench`) is the proof. The agent really ran — **$0.075 eval spend, 43,368 tokens, five diff-shaped patches** — then `run_evaluation` died for every instance:

```
ValueError: Some instance IDs not found in dataset!
```

Nothing was graded. But `raw` came back `{"errored": false, "valid_trials": 1}`, so:

- the 0.0s were averaged into a confident `mean reward 0.000 → 0.000`
- every task rendered `0.000 → 0.000` with **no** infra-error marker
- `assert_run.py`'s `measured` check saw nothing wrong — **the leg went green**
- it was **published to `benchmark-history`** as a real swebench measurement (since reverted in `0451adcb`)
- the optimizer burned **$3.44** across 3 iterations against feedback containing no signal

Strictly worse than the entitlement failure it replaced — that one at least rendered `⚠️ infra-error` and failed the job:

| | Entitlement failure (#308's case) | This scoring failure |
|---|---|---|
| Report row | `0.000 → ⚠️ infra-error` | `0.000 → 0.000` |
| Job | ❌ failed loudly | ✅ **green** |
| Published | no | **yes** |

## Commit 1 — close the hole

- **`harness.py`**: after `adapter.score()`, honour `score.raw["errored"]` — exactly as `_rescore_from_rollouts` already did on resume ([harness.py:371](https://github.com/skillberry-ai/cap-evolve/blob/main/core/cap_evolve/harness.py#L371)). The live and resume paths were inconsistent and only the resume path was right.
- **`swe_bench` adapter**: actually signal it. `_score_from_report` / `_evaluate_patches_batch` now return `(reward, feedback, ungradeable)`; new `_ungradeable_score()` sets `raw={"errored": True}`. Flagged: harness timeout, no report produced, harness exception, instance absent from an otherwise-valid report.
- A patch that ran and **genuinely failed its tests** stays a real `0.0` and stays in the mean. That distinction is the whole point.

## Commit 2 — fix the cause

Root-caused as far as possible without runner access. Reproduced locally with the *same* unpinned latest versions the runner installs (`swebench 4.1.0`, `datasets 5.0.1`) and the same five ids:

```
OK   princeton-nlp/SWE-bench_Lite -> loaded 5 instances
OK   SWE-bench/SWE-bench_Lite     -> loaded 5 instances
```

So the ids, dataset name and split are all correct — **this is environmental on the runner, not a repo defect.** `princeton-nlp/*` is a 307 redirect to `SWE-bench/*`, so every load needs live Hub access, and the runner prints on every run:

```
Warning: You are sending unauthenticated requests to the HF Hub.
Please set a HF_TOKEN to enable higher rate limits and faster downloads.
```

- **Pin** `swebench==4.1.0` / `datasets==5.0.1` — these were unpinned, so the grading harness could change under us with nothing in the log to say it had.
- **`HF_TOKEN`** passed through to the bench job (optional secret; absent → warning naming this failure mode, not a hard stop).
- **Dataset preflight** in `ci_setup.sh`: resolve the tier's ids through `load_swebench_dataset` *before* any spend. On failure it prints the exception, states that scoring would have failed for every task after the agent had already run, and names the two things to check. It also logs resolved versions + id count, so the next occurrence is self-diagnosing instead of needing SSH.

Defence in depth, not redundancy: the preflight stops the common cause up front; the guard means anything that still slips through fails loudly instead of publishing a zero.

## Verification

The three new tests **fail without** the harness change — including:

```
assert len(infra) == 4, "assert_run would still treat an ungraded run as measured"
E  AssertionError: assert 0 == 4
```

That is the production bug reproduced in a test: `assert_run`'s `measured` gate seeing **zero** infra tasks in a fully-ungraded run.

- ✅ **532 core tests pass** (1 skipped) with the fix
- ✅ `swe_bench` adapter self-check passes
- ✅ `ci_setup.sh` syntax-clean; dataset preflight verified live against the real dataset (`5/5 instances resolvable`)

## Caveat

The stale-cache half of the diagnosis is **inferred, not observed** — SSH to `skillberry-1` rejects my key, so I could not inspect the runner's HF cache. What is confirmed: the repo-side inputs are all correct, and the same code path succeeds off-runner. The preflight's diagnostics will settle it on the next run.

Independent of #308 (model entitlement preflight); both touch `benchmarks.yml` but in different hunks.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -285,6 +285,12 @@ jobs:
     env:
       ANTHROPIC_BASE_URL: ${{ secrets.ANTHROPIC_BASE_URL }}
       ANTHROPIC_AUTH_TOKEN: ${{ secrets.ANTHROPIC_AUTH_TOKEN }}
+      # swebench resolves its dataset over the HF Hub on every `run_evaluation` call.
+      # Unauthenticated requests are rate-limited per source IP, and when that resolution
+      # degrades the harness fails with "Some instance IDs not found in dataset!" only
+      # AFTER every patch has been generated. Optional: if the secret is unset, ci_setup.sh
+      # warns and continues rather than blocking the run.
+      HF_TOKEN: ${{ secrets.HF_TOKEN }}
       # Smoke always runs a fixed 3 iterations (cheap, fast regression) — decoupled
       # from the `iterations` input, which only controls the (expensive) full tier
       # (default 10 there — see the `iterations` input above).
--- a/ci/benchmarks/lib/ci_setup.sh
+++ b/ci/benchmarks/lib/ci_setup.sh
@@ -24,9 +24,12 @@ uv pip install -p "$CAPEVOLVE_PY" -q $IDX "$REPO/core" litellm
 case "$BENCH" in
   swebench)
     # PINNED. These were unpinned, so the grading harness could change under us between
-    # runs with nothing in the log to say so. Verified working together against
-    # princeton-nlp/SWE-bench_Lite (and its SWE-bench/* redirect) on 2026-08-07.
-    uv pip install -p "$CAPEVOLVE_PY" -q $IDX "swebench==4.1.0" "datasets==5.0.1"
+    # runs with nothing in the log to say it had. Pinned to what skillberry-1 was actually
+    # running on 2026-08-07, verified ON the runner to resolve all 5 smoke ids from
+    # princeton-nlp/SWE-bench_Lite (300-row test split, 0 missing). datasets 5.0.1 was
+    # verified equivalent off-runner; 5.0.0 is pinned because it is the observed-good state
+    # here and the point of a pin is reproducibility, not an untested upgrade.
+    uv pip install -p "$CAPEVOLVE_PY" -q $IDX "swebench==4.1.0" "datasets==5.0.0"
     command -v harbor >/dev/null 2>&1 || uv tool install $IDX harbor >/dev/null 2>&1 || true
 
     # Dataset preflight. `run_evaluation` resolves the dataset over the HF Hub on every
--- a/ci/benchmarks/lib/ci_setup.sh
+++ b/ci/benchmarks/lib/ci_setup.sh
@@ -23,8 +23,53 @@ uv pip install -p "$CAPEVOLVE_PY" -q $IDX "$REPO/core" litellm
 
 case "$BENCH" in
   swebench)
-    uv pip install -p "$CAPEVOLVE_PY" -q $IDX swebench datasets
-    command -v harbor >/dev/null 2>&1 || uv tool install $IDX harbor >/dev/null 2>&1 || true ;;
+    # PINNED. These were unpinned, so the grading harness could change under us between
+    # runs with nothing in the log to say so. Verified working together against
+    # princeton-nlp/SWE-bench_Lite (and its SWE-bench/* redirect) on 2026-08-07.
+    uv pip install -p "$CAPEVOLVE_PY" -q $IDX "swebench==4.1.0" "datasets==5.0.1"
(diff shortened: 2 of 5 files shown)


eranra:
## Correction: I got SSH to the runner, and the stale-cache hypothesis is wrong

The PR description above (and my first commit message) inferred a stale/corrupt HF dataset cache from an off-runner reproduction. I've since been able to log into `skillberry-1` and check directly. **That inference was wrong** — the cache is present and healthy:

```
versions: swebench 4.1.0 | datasets 5.0.0
LOAD OK -> 5 instances
raw rows = 300 | missing = []
```

All five smoke ids resolve on the runner *right now*, from a 36M `~/.cache/huggingface` containing both `datasets--princeton-nlp--SWE-bench_Lite` and `..._oracle`.

**So the failure was transient and is no longer reproducible.** Stating that plainly because it has a design consequence.

### What is confirmed

- **`HF_TOKEN` is absent everywhere on the runner** — not in the login environment, not in any runner env file. So every dataset resolution genuinely is unauthenticated and rate-limited per source IP.
- `datasets` on the runner is **5.0.0**, not the 5.0.1 I originally pinned. Pinning 5.0.1 would have been an untested upgrade of a currently-working component, so the pin is now **5.0.0** — the observed-good state. (5.0.1 verified equivalent off-runner if a bump is ever wanted.)

### The design consequence

A preflight that merely *verifies* would not have saved run 31161200250. Re-checked afterwards, the ids resolve fine — so a verify-only preflight would have passed, and the run would still have died mid-flight. Verification alone is the wrong shape of fix for a transient fault.

So the preflight now **warms, then goes offline**:

1. Resolve every dataset the run needs once, up front — the base dataset *and* the oracle dataset (`SWEBENCH_ORACLE` defaults on in `run_suite.sh`, and it is a second Hub-resolved dataset; missing it would break `tasks()` under offline mode).
2. Fail loudly and cheaply if that cannot be done.
3. Then export `HF_HUB_OFFLINE=1` / `HF_DATASETS_OFFLINE=1` for the rest of the job, so **no later call can touch the Hub at all**.

Verified on `skillberry-1` that both datasets load from the warm cache with offline mode enabled:

```
OFFLINE OK   princeton-nlp/SWE-bench_Lite        -> 5 rows
OFFLINE OK   princeton-nlp/SWE-bench_Lite_oracle -> 300 rows
```

This closes the window structurally rather than probabilistically, and it makes the dataset immutable for the duration of a run — which is what a benchmark wants anyway, and is consistent with pinning the harness.

### Unchanged

The harness guard in the first commit is still the load-bearing part. The cause is *not* confirmed, so the guarantee that matters is the one that does not depend on knowing it: any recurrence now fails loudly instead of publishing a zero.

### Unrelated hazard found on the runner

There is a stray `/tmp/inspect.py` on `skillberry-1` that shadows the stdlib `inspect` module for any Python process whose cwd is `/tmp`, producing a baffling `IndexError` from inside `importlib.metadata`. It bit my first diagnostic attempt. It does *not* affect the swebench harness (which runs in `tempfile.TemporaryDirectory()`, so its `sys.path[0]` is `/tmp/tmpXXXX`, not `/tmp`) — flagging it because it will waste someone's afternoon eventually.

Issue #349, opened 2026-08-22
Tests can import two cap_evolve trees at once (sys.path.insert vs the editable install) — presents as flaky tests, misdiagnosed twice today


OsherElhadad:
## Summary

`pip install ./core` can install a **stale copy** of the package taken from the untracked `core/build/lib/` tree that a previous build left behind. The installed `cap_evolve` then shadows the real source, so tests and runs exercise code that is not in the working tree. Three independent agents hit this today while validating unrelated changes, and two of them initially reported the resulting failures as "pre-existing failures on main".

That is the real cost: the failure mode does not look like a stale build, it looks like a flaky or order-dependent test suite. It sends people to debug the wrong thing, and it can make a genuinely broken change appear green (or a correct one appear broken).

## Evidence

On a checkout that had a stale `core/build/`:

```
$ wc -l core/build/lib/cap_evolve/harness.py core/cap_evolve/harness.py
2248 core/build/lib/cap_evolve/harness.py     <- installed
2287 core/cap_evolve/harness.py               <- source
```

A 39-line drift in `harness.py` alone. With that tree present, `pip install ./core` then `pytest core/tests -q` reports 3 failures on **unmodified main**. After `rm -rf core/build` and reinstalling, the same unmodified main is green:

```
789 passed, 12 skipped in 113.52s
core/tests/test_eventstream.py in isolation: 38 passed in 8.26s
```

Two independent confirmations of the same cause, from agents working in separate worktrees on unrelated changes:
- one reported `2 failed, 787 passed` and attributed it to order-dependent tests in `test_eventstream.py`
- one reported `3 failed, 861 passed` and "proved" it pre-existing by reverting its own changes — the stale build was constant across both halves of that comparison, so the proof held while the conclusion was wrong

`core/build/` is untracked and covered by `.gitignore:6` (`build/`), so this is a local-environment hazard rather than something committed. That is exactly why it is easy to miss: it never shows up in `git status`.

## Why it is worth fixing rather than documenting

The workaround people converge on is `PYTHONPATH=core pytest core/tests -q`, which works but silently diverges from how the package is actually consumed. And a contributor who does not know the trick will file a bug against the test suite instead.

## Fix

Any of these would close it; the first two are the cheap ones.

[...]

PR #355, opened 2026-08-22
fix(core): one shared optimizer-context assembler — gepa and skillopt stop running blind (#109)

### LINKED ISSUE #109: GEPA & SkillOpt run blind: wire optimizer-context injection + capability brief; un-gate CLI flags from hill-climb-only

OsherElhadad:
Closes #109. Supersedes #199 (that branch conflicts in 13 files against current main; this is a redesign against `d94fb336`).

Second commit addresses the merge review (`.review/merge-review-355.md`, verdict FIX-FIRST): the guard's scope claim, the undisclosed prompt-size increase, and three follow-up issues. See "Review fixes" below.

## The bug

`hill-climb` was the only algorithm that received the context `docs/ARCHITECTURE.md` promises ("What the optimizer receives each iteration"). GEPA never injected any of it; SkillOpt called the prompt renderer bare. Two of the three arms were handed a strictly worse prompt than the third, so any algorithm-vs-algorithm number on this repo's own benchmarks was confounded — "hill-climb beat gepa" was untestable. SkillOpt's `{{PARALLEL_NOTE}}` was worse than absent: `optimizer_name=None` made it *tell a parallel-capable optimizer it was sequential*.

## Parity table, re-verified against current main (`d94fb336`)

BEFORE (all line numbers from `main`, not from PR #199's stale branch):

| context piece | hill-climb | gepa | skillopt |
|---|---|---|---|
| `./trajectories/` (parent's verbatim rollouts) | YES — `harness.py:1461` via `run_step` | **NO** — no inject call anywhere; `gepa.py:571` only `copytree(parent_dir, workdir)` | YES — `skillopt.py:296` → `harness.py:1461` |
| `./guidance/<cap>/` capability skill | YES — `harness.py:2197` + `1461` `capabilities=` | **NO** | **NO** — `capabilities=` not passed (`skillopt.py:296`) |
| native skills dir (`.claude/skills/…`) | YES — same inject | **NO** | **NO** (same reason) |
| `./guidance/diagnose/` | YES | **NO** | YES (unconditional part of inject) |
| `./guidance/sources/` (`capability_sources`) | YES | **NO** | **NO** — param absent from `skillopt_loop` |
| `./guidance/optimizer/<name>.md` | YES | **NO** | **NO** |
| `{{CAP_BRIEF}}` | YES — `harness.py:2197` | **NO** — no `_focus_instructions` call at all | **EMPTY** — `skillopt.py:291` bare |
| `{{FAILURES}}` / `{{PASSING}}` | YES | **NO** | YES |
| intake-authored INSTRUCTIONS template | YES | **NO** — hand-rolled `gepa.py:289 _instructions` | **NO** — default template only |
| `{{BENCH_REPO}}` | YES | **NO** | **NO** |
| `{{PARALLEL_NOTE}}` | YES | **NO** | **WRONG** — always claims sequential |
| `{{TARGET_READER}}` | YES | **NO** | **NO** |
| `{{EMPTY_SEED}}` | YES | **NO** | **NO** (fell back to the reward heuristic → claimed an empty seed that wasn't) |
| prior impact / protect-set (LEDGER/JOURNAL/RUNMAP/`prior_iterations/`) | YES — `harness.py:1062` | YES — `gepa.py:582` | YES — via `run_step` |

Confirmed by grep on main: `_inject_optimizer_context` is defined at `harness.py:1166` and called from exactly one place, `harness.py:1461`. `skillopt.py:291` is `harness._focus_instructions(current_val, minibatch_ids, label)` — no other args. CLI gates on main: `cli.py:860` (`--capabilities`), `cli.py:865` (`--optimizer-name`), `cli.py:870` (`--instructions-file` / `--bench-repo` / `--capability-sources` / `--target-model` / `--target-profile-file`), all `algorithm_name == "hill-climb"`. And the flags did not exist downstream either — `skills/algorithms/{gepa,skillopt}/scripts/run.py` declared none of them, so un-gating the CLI alone would have crashed the run on unknown args.

AFTER (this branch) — every piece for each of the three algorithms that thread the context, from one call each:

| context piece | hill-climb | gepa | skillopt |
|---|---|---|---|
| all 14 rows above | `harness.py:2300` builds the ctx, `2335` `ctx.instructions(...)`, `1470` `ctx.inject(...)` via `run_step` | `gepa.py:501` ctx, `588` `ctx.inject(..., tag=f"mb_p_{n:04d}")`, `591` `ctx.instructions(..., extra=_instructions(...))` | `skillopt.py:234` ctx, `300` `ctx.instructions(...)`, `309` `ctx=ctx` → `harness.py:1470`; slow update at `484`/`491`/`498` |
| CLI flags | `cli.py:868/873/878` — `algorithm_name in OPTIMIZER_CONTEXT_ALGORITHMS` (`cli.py:387`) | same | same |
| run.py flag set | `harness.OptimizerContext.add_arguments(p)` | same | same |

`cli.py:681` (`algorithm_focus` gated to hill-climb) is left alone on purpose: `--focus` is genuinely a hill-climb schedule, not context.

**Scope.** `agent-optimize` and `evograph` also ship, declare none of the context flags, drive their own loops, and are in neither list — they were blind on main and remain outside this guarantee. Out of scope for #109; the real fix (discovering algorithms instead of enumerating them) is #375.

## The shared assembler, in three sentences

`harness.OptimizerContext` is a dataclass holding the read-context (capabilities, optimizer name, capability sources, project dir, instructions template, bench repo, target-reader brief) with exactly two methods: `inject()` writes it into an iteration workdir and `instructions()` renders that iteration's prompt from the shared template. All three loops accept `ctx=` and call those two methods, and `run_step`'s four threaded context params collapse into `ctx=`, so there is one assembly path rather than hill-climb's blocks copied into two more places. Nothing in it branches on the caller — the two genuinely per-algorithm inputs are parameters: `algorithm` (selects the algorithm brief) and `extra` (the caller's own block, appended verbatim: gepa's reflective-dataset pointer, skillopt's edit-budget/rejected buffer, the slow-update longitudinal block).

Three supporting changes the injection *requires*:

1. `_copy_step_trajectories`/`_inject_optimizer_context` take `tag=`. GEPA's parent is a frontier member, not the run's best, so it passes its own minibatch eval tag and the optimizer reads the untruncated traces behind its own `REFLECTION.md`. (Known limit, filed as #376: from iteration 2 the parent's minibatch eval is a cache hit that writes no rollout file, so the tag stops resolving and the fallback hands over the whole mixed `rollouts/train/`. Still strictly better than main's nothing.)
2. `gepa.py` snapshots accepted candidates with `harness._SNAPSHOT_IGNORE` (`gepa.py:684`, `848`) — the exclusion `run_step` already applied. This is what actually keeps injected context out of candidate dirs, and hence out of the merge's component lists and the next iteration's parent copy. **Duplicated by #350** (byte-compared: semantically identical); drop these three hunks here on rebase once #350 lands — see "Rebase" below.
3. `types.NON_CAPABILITY_DIRS`/`_FILES` (main's already-shared list, used by `cache.hash_candidate_dir`, `gepa._components`, `diffview`) gain the injected read-context: `trajectories`, `guidance`, the agent skills dot-dirs, and the agent instructions files. Required: `gepa.py:640` hashes the **workdir**, not a snapshot, so without this the injected ~20 KB folds into the child's hash and a cache entry written for a child can never be re-read when that child is later re-sampled as a parent from its clean candidate dir. For the component-selector path it is defence-in-depth — in-iteration ordering (`_components` runs before `inject`) plus the snapshot ignore already secure that.

## Cost: gepa's prompt grows ~15x — intentional, disclosed

Measured on the same zero-API toy_calc run at `d94fb336` and on this branch, comparing the rendered `work/<cid>/INSTRUCTIONS.md`:

| algorithm | main | this branch | delta |
|---|---|---|---|
| hill-climb | 20,936 B (it1) / 20,478 B | 20,936 B / 20,478 B | **0 — byte-identical** |
| gepa | **1,474 B / 16 lines** | **22,192 B / 279 lines** | **+20,718 B, ≈15x, ≈ +5k input tokens per iteration** |
| skillopt (normal step) | 20,648 B | 20,562 B | −86 B (a *false* EMPTY-SEED block removed) |
| skillopt (slow/meta update) | 1,825 B | 21,115 B | **+19,290 B** — the slow update now renders through the shared template too |

The gepa and skillopt-slow increases are the point of the PR (gepa was proposing off a 16-line prompt), and it is the same prompt hill-climb has always had — but on a **paid** optimizer this is a real per-iteration input-token cost increase, so it should be a conscious merge decision, not a surprise. hill-climb is unchanged to the byte.

The skillopt normal-step content change is strictly corrective: it drops a false `## EMPTY SEED — create the capability from scratch` block (main left `seed_empty=None`, so the reward heuristic claimed the capability dir was empty whenever baseline val was 0.0, though `examples/toy_calc/capability/prompt.txt` exists), fixes `Algorithm: hill-climb.` → `Algorithm: skillopt.`, and fixes the inverted parallel note.

## The guard

`core/tests/test_optimizer_context_parity.py` — parametrized over `("hill-climb", "gepa", "skillopt")`, one real zero-API iteration each (mock optimizer, `toy_calc`, `claude-code` as the resolved optimizer row so native placement and the parallel note are exercised):

- `test_every_algorithm_gets_the_same_optimizer_context` — asserts on the assembled workdir: `./trajectories/` non-empty, `./guidance/<cap>/SKILL.md`, `.claude/skills/<cap>/SKILL.md`, `./guidance/diagnose/SKILL.md`, `./guidance/sources/<file>`, `./guidance/optimizer/claude-code.md`, `LEDGER.md`/`JOURNAL.md`/`RUNMAP.md`; and on the prompt: no unrendered `{{`, the bench-repo path, the capability brief, "fan out".
- `test_injected_context_never_lands_in_a_candidate` — no snapshotted candidate may contain `trajectories/`, `guidance/` or `.claude/`. Asserted per algorithm because gepa snapshots from its own loop, not through `run_step`.
- `test_ignore_lists_cover_everything_injection_writes` — every injected path is declared non-capability, dirs and files.

**Scope is enumerated, not discovered, and the docstring says so.** `ALGORITHMS` is a literal, so an algorithm absent from it is not covered and can still run blind with this file green. That is why #375 exists.

Each arm verified RED independently. Dropping gepa's `ctx.inject(...)`:

```
E       AssertionError: gepa: no ./trajectories/
FAILED core/tests/test_optimizer_context_parity.py::test_every_algorithm_gets_the_same_optimizer_context[gepa]
1 failed, 3 passed in 3.16s
```

Reverting gepa's snapshot ignore (the change #350 duplicates) — this is what makes dropping that hunk on rebase safe in any merge order:

```
E               assert not True
E                +    where exists = (PosixPath('.../candidates/gepa_0001') / 'trajectories').exists
FAILED core/tests/test_optimizer_context_parity.py::test_injected_context_never_lands_in_a_candidate[gepa]
1 failed, 6 passed in 6.30s
```

The reviewer independently broke each of the other arms (`ctx=ctx` off skillopt's and hill-climb's `run_step` calls, skillopt reverted to main's bare renderer) and got a distinct, algorithm-named failure each time — see `.review/merge-review-355.md`.

## Review fixes (second commit)

- **B1 — overclaim removed.** The body and the test docstring said "a newly added algorithm that forgets to thread the context fails here". It does not: the algorithm list is a literal. Reworded to "each of the three algorithms that thread the context", with the scope limit and the two out-of-scope algorithms stated in the docstring, and #375 filed for the real fix.
- **B2 — prompt-size increase disclosed** with my own measurements, above (including the skillopt slow-update case the review did not cover).
- **B3 — rebase**, below.
- Reviewer nits taken: deleted the never-read `OptimizerContext.ARG_DESTS`; documented that a passed `ctx` wins over `hill_climb_loop`'s legacy per-piece kwargs; widened the ignore-list assertion to all injected dirs *and* files; corrected the commit message's claim about which change stops the component selector.
- Follow-ups filed instead of scope-creeping this PR: **#375** (parity test should discover algorithms), **#376** (gepa's `tag=` degrading from iteration 2, and the guard's blindness to it), **#378** (basename-anywhere ignore matching vs a capability that legitimately is `AGENTS.md` or lives under `.claude/`; pre-existing on main).

## Rebase (per `.review/MERGE_PLAN.md`: 350 → 346 → 355)

Neither #350 nor #346 is on `main` yet (`git log origin/main` = `d94fb336`), so this branch still carries its own copy of the gepa snapshot change — dropping it before #350 lands would regress candidate cleanliness and leave this branch unable to prove it. Mechanical instructions for the merger, verified against both branches:

1. From **#350**, drop these three hunks here (semantically identical, reviewer byte-compared): the `_SNAPSHOT_IGNORE` entry in gepa's `from .harness import (...)` block (already sorted to #350's position so the block auto-resolves), `gepa.py:684`, `gepa.py:848`. Keep this PR's two-line comment at `684`. `test_injected_context_never_lands_in_a_candidate` then proves the drop was safe. #350 owns the `_SNAPSHOT_IGNORE` *tuple*; this PR never edits it, and #350 never edits `types.py`, so there is no semantic conflict.
2. From **#346**, drop `rejected, history` from the `_augment_instructions(...)` call at `gepa.py:593` (arity change to `(instructions, workdir, run_dir)`). #346 does not touch `types.py`; it leaves `MEMORY.md` in `NON_CAPABILITY_FILES` as a harmless dead entry.

## Evidence — every run performed, verbatim

Pinned full suite per #349 (`PYTHONPATH="$PWD/core:$PWD/dashboard/backend"`); clean main is `789 passed, 12 skipped`:

```
796 passed, 12 skipped in 119.66s (0:01:59)
```

Three refreshed algorithm runs — zero-API, deterministic, real `cap-evolve run` through `cli.py` (the toy `run.sh`, with only `algorithm_skill` rewritten for the other two arms), fresh temp project each:

```
hill-climb  run_dir=.capevolve/run_demo   best_id=cand_0001   baseline_val=0.0  test_reward=1.0  test_delta=+1.0  iterations=3
gepa        run_dir=.capevolve/run_demo_gepa      best_id=gepa_0001   baseline_val=0.0  test_reward=1.0  test_delta=+1.0  iterations=3
skillopt    run_dir=.capevolve/run_demo_skillopt  best_id=so_e01s01   baseline_val=0.0  test_reward=1.0  test_delta=+1.0  iterations=3
```

Context actually on disk, gepa iteration 1 (`work/gepa_0001/`) — on main this dir has no `guidance/` and no `trajectories/` at all:

```
FOCUS.md   guidance   INSTRUCTIONS.md   JOURNAL.md   LEDGER.md
PROCESS.md   prompt.txt   REFLECTION.md   RUNMAP.md   trajectories
guidance/: diagnose  optimizer  system-prompt
```

skillopt iteration 1 (`work/so_e01s01/`):

```
guidance   INSTRUCTIONS.md   JOURNAL.md   LEDGER.md
PROCESS.md   prompt.txt   RUNMAP.md   trajectories
guidance/: diagnose  optimizer  system-prompt
```

And the accepted gepa candidate carries the capability only — no injected read-context:

```
$ ls -a run_demo_gepa/candidates/gepa_0001/
.  ..  FOCUS.md  INSTRUCTIONS.md  PROCESS.md  prompt.txt  REFLECTION.md
```

(`FOCUS.md`/`REFLECTION.md`/`INSTRUCTIONS.md` are gepa's own scratch; #350 adds the first two to `_SNAPSHOT_IGNORE`, which is another reason it lands first.)

## Principles

- **Generic, not specific**: the assembler names no benchmark, capability, optimizer or vendor and does not branch on the algorithm; per-algorithm content arrives as `algorithm=` / `extra=`. The four remaining algorithm-name literals are all in `cli.py`, which already resolves algorithms by name, and they *replace* three `== "hill-climb"` comparisons.
- **Zero deps**: `dataclasses` + `pathlib`.
- **Honesty invariants untouched**: no change to the gate, splits, seal or tamper guard. The one adjacent edit — the ignore-list — is required, and it extends main's *existing single* list rather than adding a fourth copy.
- **Deletion**: hill-climb's `run.py` loses 38 lines of flag boilerplate and its own profile-resolution block; `run_step` loses four params. No existing test was touched, relaxed or skipped (`git diff --name-status -- core/tests/` = one added file).
- Dropped from #199: `optimizer_context.py` as a new module (the assembly already lived in `harness.py`), `log_profile`, and the docstring pre-advertising itself as the extension point for seven future issues.

### DIFF
--- a/core/cap_evolve/cli.py
+++ b/core/cap_evolve/cli.py
@@ -378,6 +378,14 @@ _ALGO_FOCUS_ALIASES = {
 #: epoch schedule, so the spec key is inert for them — and we say so rather than drop it.
 CONVERGENCE_ALGORITHMS = frozenset({"hill-climb"})
 
+#: Algorithms whose ``run.py`` declares the optimizer-context flags (via
+#: ``harness.OptimizerContext.add_arguments``) and can therefore be handed the full
+#: read-context: capability skills, the instructions template, the benchmark repo, the
+#: optimizer name, the capability sources, and the target-reader profile. These flags used
+#: to be gated to hill-climb alone, which silently handed gepa and skillopt a strictly
+#: thinner prompt and made any cross-algorithm comparison meaningless.
+OPTIMIZER_CONTEXT_ALGORITHMS = frozenset({"hill-climb", "gepa", "skillopt"})
+
 
 def _resolve_algorithm(name: str) -> tuple[str, str | None]:
     """Map a spec ``algorithm_skill`` to (skill_name, focus).
@@ -851,23 +859,23 @@ def _cmd_run(argv):
     if algorithm_focus is not None:
         alg_cmd += ["--focus", algorithm_focus]
     # Surface the selected capability skills to the optimizer prompt so it knows the
-    # allowed edit space (e.g. tools → may add composite tools). hill-climb consumes
-    # --capabilities; algorithms without the flag ignore the extra arg via argparse error,
-    # so only pass it to those that accept it.
+    # allowed edit space (e.g. tools → may add composite tools). Passed to every
+    # algorithm that declares the optimizer-context flags, so each one prompts from the
+    # same read-context.
     caps = spec.get("capabilities") or []
     if isinstance(caps, str):
         caps = [c.strip() for c in caps.split(",") if c.strip()]
-    if caps and algorithm_name == "hill-climb":
+    if caps and algorithm_name in OPTIMIZER_CONTEXT_ALGORITHMS:
         alg_cmd += ["--capabilities", ",".join(str(c) for c in caps)]
     # Thread the resolved optimizer NAME so the harness can copy that optimizer's
     # features reference (parallel-subagent capabilities etc.) into each iteration's
-    # workdir. Only hill-climb accepts the flag; other algorithms ignore it.
-    if algorithm_name == "hill-climb":
+    # workdir and place the skills where that agent natively finds them.
+    if algorithm_name in OPTIMIZER_CONTEXT_ALGORITHMS:
         alg_cmd += ["--optimizer-name", str(optimizer_name)]
     # Optimizer-instructions template (intake-authored, per benchmark) + benchmark repo
     # as read-only optimizer context. Both are resolved project-relative if not absolute.
     # The instructions file defaults to the scaffolded project/optimizer/INSTRUCTIONS.md.
-    if algorithm_name == "hill-climb":
+    if algorithm_name in OPTIMIZER_CONTEXT_ALGORITHMS:
         instr = spec.get("optimizer_instructions_file") or "optimizer/INSTRUCTIONS.md"
         instr_p = Path(instr)
         if not instr_p.is_absolute() and not instr_p.exists():
--- a/core/cap_evolve/gepa.py
+++ b/core/cap_evolve/gepa.py
@@ -55,6 +55,8 @@ from . import integrity
 from . import selection
 from .cache import EvalCache, hash_candidate_dir
 from .harness import (
+    OptimizerContext,
+    _SNAPSHOT_IGNORE,
     _augment_instructions,
     _init_memory_store,
     _live,
@@ -473,6 +475,7 @@ def gepa_loop(
     store=None,
     resume: bool = False,
     protected_patterns=None,
+    ctx: OptimizerContext | None = None,
 ) -> dict:
     """Run GEPA's sample-efficient reflective Pareto loop.
 
@@ -495,6 +498,7 @@ def gepa_loop(
     highest-val pool member; the test split is never touched.
     """
     gate_kwargs = dict(gate_kwargs or {})
+    ctx = ctx or OptimizerContext()   # the SAME read-context assembly as its siblings
     rejected, history, store = _init_memory_store(run_dir, store)
     cache = EvalCache(run_dir.root / "eval_cache.json")
     rng = random.Random(seed)
@@ -576,9 +580,19 @@ def gepa_loop(
             comp_cursor += 1
         else:
             focus = None  # 'all'
+        # The shared optimizer read-context (capability skill(s) natively + under
+        # ./guidance/, the diagnose method, sources, the optimizer features ref). The
+        # trajectories are tagged with THIS parent's minibatch eval — gepa's parent is a
+        # frontier member, not the run's best, so the untruncated traces behind
+        # REFLECTION.md are the ones to hand over.
+        ctx.inject(adapter, run_dir, workdir, split="train", tag=f"mb_p_{n:04d}")
         refl_summary = _write_reflection(workdir, parent_mb)
         focus_label = _write_focus(workdir, comps, focus)
-        instructions = _instructions(refl_summary, focus_label, mb)
+        instructions = ctx.instructions(
+            parent_mb, mb, f"GEPA minibatch of {len(mb)} train tasks, component focus "
+                           f"{focus_label}",
+            algorithm="gepa", parent_dir=parent_dir,
+            extra=_instructions(refl_summary, focus_label, mb))
         instructions = _augment_instructions(instructions, workdir, run_dir, rejected, history)
 
         # Seal the eval surface for this child. GEPA runs the optimizer itself rather
@@ -665,7 +679,9 @@ def gepa_loop(
         summary = (f"candidate {cid} (val {cand_val.reward:.3f}, "
                    f"Δ {cand_val.reward - parent_result.reward:+.3f})")
         if accepted:
-            run_dir.snapshot(cid, workdir)
+            # ignore=: the injected read-context (trajectories/guidance/native skills) is
+            # NOT part of the candidate — same exclusion hill-climb's run_step applies.
+            run_dir.snapshot(cid, workdir, ignore=_SNAPSHOT_IGNORE)
             child_dir = run_dir.candidate_dir(cid)
             pool.append(_entry(cid, child_dir, cand_val, parent=parent["id"]))
             lineage[cid] = parent["id"]
@@ -829,7 +845,7 @@ def _try_merge(
     run_dir.update_spent(iterations=1, accepted=accepted)
     summary = f"merge {mid} of {a['id']}+{b['id']} (val {cand_val.reward:.3f})"
     if accepted:
-        run_dir.snapshot(mid, workdir)
+        run_dir.snapshot(mid, workdir, ignore=_SNAPSHOT_IGNORE)
         pool.append(_entry(mid, run_dir.candidate_dir(mid), cand_val, parent=base_parent["id"]))
         lineage[mid] = base_parent["id"]
         history.add(mid, summary, cand_val.reward)
--- a/core/cap_evolve/gepa.py
+++ b/core/cap_evolve/gepa.py
@@ -56,12 +56,12 @@ from . import selection
 from .cache import EvalCache, hash_candidate_dir
 from .harness import (
     OptimizerContext,
-    _SNAPSHOT_IGNORE,
     _augment_instructions,
     _init_memory_store,
     _live,
     _paired_deltas,
     _resolve_workers,
+    _SNAPSHOT_IGNORE,
     evaluate_candidate,
     split_result_from_rollouts,
 )
--- a/core/cap_evolve/harness.py
+++ b/core/cap_evolve/harness.py
@@ -2124,10 +2124,6 @@ class OptimizerContext:
 
     # ---- construction from an algorithm run.py --------------------------------
 
-    #: The context flags every algorithm's ``run.py`` declares via ``add_arguments``.
-    ARG_DESTS = ("capabilities", "instructions_file", "bench_repo", "optimizer_name",
-                 "capability_sources", "target_model", "target_profile_file")
-
     @staticmethod
     def add_arguments(p) -> None:
         """Declare the optimizer-context flags on an algorithm's argument parser.
@@ -2290,6 +2286,11 @@ def hill_climb_loop(
     ``protected_patterns`` seals the eval surface (see ``run_step``); ``convergence``
     turns on the graded plateau signal (warn → paradigm_shift → stop). Both default
     off, so a caller that passes neither runs exactly as before.
+
+    ``ctx`` is the shared optimizer read-context. When it is given it WINS: the legacy
+    per-piece kwargs below it (``capabilities`` … ``target_profile_file``) are then unused,
+    kept only for the callers/tests that predate ``OptimizerContext``. Pass one or the
+    other, not both.
     """
     gate_kwargs = dict(gate_kwargs or {})
     rejected, history, store = _init_memory_store(run_dir, store)
--- a/core/cap_evolve/harness.py
+++ b/core/cap_evolve/harness.py
@@ -20,6 +20,7 @@ import shutil
 import subprocess
 import sys
 import time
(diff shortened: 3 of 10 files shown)


OsherElhadad:
Superseded by #386, which is this change rebased onto the new main (post #350/#346/#347). The prepared drop-list was applied: all three duplicated gepa snapshot-ignore hunks removed in favour of #350's, and #346's `_augment_instructions` arity adopted. `test_injected_context_never_lands_in_a_candidate` survives and was re-verified RED by reverting the single line #350 owns. Closing in favour of #386.

PR #356, opened 2026-08-22
fix(core): route every algorithm through one iteration-record + journal step (#216, #224)

### LINKED ISSUE #216: GEPA runs never write a run-level JOURNAL.md — _reconcile_journal is only called from run_step, which GEPA bypasses
GEPA runs never produce a run-level `JOURNAL.md`. The reconciliation that writes it is called from exactly one place, and GEPA does not go through that place.

## Verified
```
$ grep -rn '_reconcile_journal' core/cap_evolve/*.py
harness.py:784:   ... and ``_reconcile_journal`` folds that back into the run-level file after the step."""
harness.py:801:   def _reconcile_journal(workdir, run_dir, cid, *, ...)
harness.py:1340:      _reconcile_journal(workdir, run_dir, cid, accepted=accepted, ...)
```
The single call site is inside `run_step`. GEPA has its own loop and **bypasses `run_step`**, so on a GEPA run the journal is never folded back and the run-level file is never written.

## Why it matters
This is the same structural bug as #109's, which was just fixed: `_parent_map` / `_build_ledger` / `_build_runmap` filtered `kind == "step"`, and because GEPA emits `gepa_val_gate` instead, its `LEDGER.md`, `RUNMAP.md` and `prior_iterations/` were **permanently empty while the prompt instructed the optimizer to read them**. PR #199 fixed those three by routing them through a shared `rundir.ITERATION_EVENT_KINDS` / `RunDir.iteration_events()`. The journal was not in that set, so it is the surviving instance of the pattern.

Consequences:
- The journal is part of the run's explainability record. On the flagship algorithm it silently doesn't exist, so anything a human or the dashboard expects to read there is empty for GEPA and populated for hill-climb — surfaces disagreeing about the same run.
- It's invisible: nothing errors, the file is simply absent.

## Fix
Do **not** add a second call site — that reintroduces the desync this repo keeps paying for. Options in preference order:

1. **Move the reconciliation to the shared path**, alongside how #199 unified the iteration-event consumers, so any algorithm that records an accepted iteration reconciles the journal by construction. Check whether it belongs next to `RunDir.iteration_events()`.
2. If GEPA genuinely has no per-candidate journal to fold (verify — does `_augment_instructions` write one into GEPA workdirs the way it writes `PROCESS.md`?), then the honest fix is to **document that `JOURNAL.md` is hill-climb/skillopt-only** and remove any prompt or doc text that tells a GEPA optimizer to read it. Silence is only acceptable if nothing claims otherwise.

Check `skillopt` too — it also has its own loop; confirm whether it reaches `run_step` or shares GEPA's gap.

## Acceptance
- Either a GEPA run produces a correct run-level `JOURNAL.md`, or no prompt/doc/dashboard surface claims it exists for GEPA.
- A test asserts the outcome **per algorithm** (parametrized over hill-climb / gepa / skillopt), so a future algorithm bypassing the shared path fails a test rather than silently producing a partial record. #211 uses exactly this parametrized-parity pattern for snapshots — mirror it.
- While here: audit for any *other* consumer still filtering by event kind or hanging off `run_step` that GEPA/skillopt bypass. #199 fixed three, #211 fixed the snapshot ignores, this is the fourth — a deliberate sweep is likely cheaper than finding the fifth in review.

## Provenance
Found by the reviewer of PR #211 (issue #110) as a side observation; call site independently confirmed by the epic #127 orchestrator.

OsherElhadad:
## Root cause (one sentence)

An iteration was only recorded by whichever algorithm happened to open-code the calls, so GEPA — which bypasses `run_step` by design (`gepa._full_val_gate`'s own docstring says so) — charged iterations against the budget while writing **no `step` event** and **never reconciling the run-level `JOURNAL.md`**.

Two symptoms, one cause: #216 (no `JOURNAL.md` on a GEPA run) and #109-as-restated-in-#224 (`_parent_map` / `_build_ledger` / `_build_runmap` filter `kind == "step"`, so GEPA's `LEDGER.md` / `RUNMAP.md` / `prior_iterations/` stayed empty for the whole run **while the optimizer prompt instructed it to read them**, and told it its best was `seed`).

## The shared seam

`harness.record_iteration(run_dir, workdir, cid, *, parent_id, accepted, reason, val=None, parent_val=None, indecisive=False, **extra)` — `harness.py:1005-1047`.

Three things must happen exactly once per iteration, and they were open-coded per algorithm:

1. charge the iteration (`update_spent(iterations=1, accepted=…)`);
2. write the canonical **`step`** event — the one iteration record every consumer enumerates (`_parent_map`, `_build_ledger`, `_build_runmap`, the dashboard graph, the event stream, the TUI);
3. reconcile the run-level append-only `JOURNAL.md` with the entry the optimizer appended in `workdir` — otherwise `_seed_journal` silently discards it on the next iteration.

**The enforcement is a test, not a convention:** `record_iteration` is the *only* place `update_spent(iterations=…)` is called anywhere in the tree, and `test_exactly_one_place_charges_an_iteration` asserts that on the source with `ast`:

```
core/cap_evolve/harness.py:1040:    run_dir.update_spent(iterations=1, accepted=None if indecisive else accepted)
```

Charging an iteration and recording it are now the same call, so an algorithm cannot do the first without the second. `len(step records) == spent.iterations` follows — which is also why the test needs no `step_indecisive` caveat: `record_iteration` writes a `step` on the indecisive path too, and the caller keeps its own `step_indecisive` detail event.

### Call sites routed through it (six)

| Site | Was |
|---|---|
| `harness.run_step` (hill-climb, skillopt) | `update_spent` + `log_event("step")` + `_reconcile_journal`, open-coded |
| `harness._tamper_step` | `update_spent` + `log_event("step")`, **no journal** |
| `gepa._full_val_gate` (main loop **and** `_try_merge`) | `log_event("gepa_val_gate")`, no journal; callers charged separately |
| `gepa` minibatch-local reject | `update_spent` only — the iteration was **invisible** |
| `gepa` indecisive tamper path | `update_spent` + `step_indecisive`, no `step`, no journal |
| `agent-optimize/scripts/commit.py` | `log_event("accept"/"reject")` + `update_spent`, no `step`, no journal |

`evograph` has no deterministic engine (its `run.py` refuses to run); its loop is prose, and §2.6 said only "log the round to the run dir event log". It now names the seam explicitly, so the one algorithm that *can only* forget is told exactly what to call.

Deliberately **not** in the seam: `snapshot` / `set_best`. Their correct arguments are genuinely algorithm-specific (`run_step` snapshots every candidate with `_SNAPSHOT_IGNORE`, gepa only accepted ones) and neither is silently droppable the way the three above were.

### Also: the duplicate direction of the same bug

`_STEP_KINDS` in `eventstream.py` and `dashboard.py` listed `gepa_val_gate` and `skillopt_step` alongside `step`. skillopt logs `skillopt_step` *after* `run_step` already logged `step` for the same candidate → **two rendered rows per iteration**. `skillopt_step` leaves both tuples and is added to `eventstream.BOOKKEEPING_KINDS` — which is what makes the duplicate row actually go away rather than fall through to the generic key dumper:

```
before : [16:40:24] ACCEPT  so_e01s01  val=1.0000 (parent 0.0000)  — paired Δ̄=+1.0000 > 0 …
         [16:40:24] ACCEPT e1s1  so_e01s01  val=1.0000 (parent ?)          <-- duplicate
after  : [16:40:24] ACCEPT  so_e01s01  val=1.0000 (parent 0.0000)  — paired Δ̄=+1.0000 > 0 …
         (skillopt_step → None, skipped in the terminal; the dashboard still shows it)
```

**Correction to an earlier version of this description:** it claimed the duplicate also double-summed `opt_cost_usd` in the cost ledger. That was wrong — `skillopt_step` carries only `epoch / step_in_epoch / global_step / edit_budget / requested_edits / applied_changes / accept / candidate / val`, and `gepa_val_gate` only `candidate / accept / reason / val / parent / parent_val`. Neither carries `opt_cost_usd`, so `eventstream._accumulate` added `float(None or 0.0)` = 0.0. There was never a cost error, only duplicate rows. The code change is unaffected.

### Why `gepa_val_gate` stays in the dashboard tuple but `skillopt_step` does not

Not symmetry — they fail differently.

`gepa_val_gate` is a **legacy iteration record**. Run dirs already on disk from before this fix have *no* `step` events at all (that is the bug), so `dashboard.reduce_run` is the only thing rendering them and `gepa_val_gate` is the only record it has. An earlier revision of this PR dropped it and silently regressed every historic gepa run's lineage graph from 2 nodes to 1. It is restored. Same reasoning keeps agent-mode `commit.py`'s `accept`/`reject` there.

`skillopt_step` is **not** a legacy record: it is epoch detail logged *alongside* a `step` for the same candidate, on current runs. It never carried a graph anyone needs, so it is the one kind that leaves.

Keeping the legacy kinds is free on a new run because the reducer keys one iteration per **candidate**, not per event — the extra kind lands on the node `step` already created:

```
$ # HISTORIC gepa run dir (pre-fix: gepa_val_gate only, ZERO step events)
without gepa_val_gate : graph nodes: 1 -> ['seed']                       <-- the regression
with    gepa_val_gate : graph nodes: 2 -> ['gepa_0001', 'seed']          <-- restored

$ # NEW runs — no duplicate nodes reintroduced
NEW gepa (post-fix)    : 4 nodes -> ['gepa_0001','gepa_0002','gepa_0003','seed']  (unique: True)
NEW skillopt (post-fix): 4 nodes -> ['seed','so_e01s01','so_e02_slow','so_e02s01'] (unique: True)
```

## Evidence: zero-API end-to-end, per algorithm

`examples/toy_calc` via `cap-evolve run` (deterministic, no API key).

### GEPA — the core before/after pair

**BEFORE** (`algorithm_skill: gepa`, 3 iterations charged):
```
$ ls -la .capevolve/run_demo/JOURNAL.md
ls: JOURNAL.md: No such file or directory

$ grep -o '"kind": "[a-z_]*"' events.jsonl | sort | uniq -c
   3 "kind": "gepa_local_gate"
   3 "kind": "gepa_select"
   1 "kind": "gepa_start"
   1 "kind": "gepa_val_gate"        <-- 1 record for 3 charged iterations
   ...                              <-- ZERO "step" events
```

**AFTER** (same spec):
```
$ ls -la .capevolve/run_demo/JOURNAL.md
-rw-r--r--  1 osherelhadad  wheel  2902 Aug 22 17:02 JOURNAL.md

$ grep -o '"kind": "[a-z_]*"' events.jsonl | sort | uniq -c
   3 "kind": "gepa_local_gate"
   3 "kind": "gepa_select"
   1 "kind": "gepa_start"
   3 "kind": "step"                 <-- 3 records == 3 charged iterations

$ tail -14 JOURNAL.md
## Iteration gepa_0001 — (no handover written by the optimizer)

> **RESULT (framework, objective):** ACCEPTED (new champion) · val=1.000 Δ=+1.000 · fixed={a1, a4} · broke={—}.
<!-- gepa_0001: ACCEPTED val=1.000 Δ=+1.000 -->

## Iteration gepa_0002 — (no handover written by the optimizer)

> **RESULT (framework, objective):** REJECTED (champion unchanged) · val=— Δ=— · fixed={—} · broke={—}. …
<!-- gepa_0002: rejected val=— Δ=— -->

## Iteration gepa_0003 — (no handover written by the optimizer)

> **RESULT (framework, objective):** REJECTED (champion unchanged) · val=— Δ=— · fixed={—} · broke={—}. …
<!-- gepa_0003: rejected val=— Δ=— -->
```
`gepa_0002`/`gepa_0003` are the minibatch-local rejects that were previously invisible. `val=—`, not a number: they never bought a full-val eval, and a minibatch reward must never be written into the field the ledger and the dashboard's val curve read. ("no handover written by the optimizer" is correct — the `mock` optimizer writes none.) `_reconcile_journal` now accepts `val=None`/`delta=None` for exactly this case.

### agent-optimize — the third recording path, same before/after

Agent mode, so `cap-evolve run` does baseline then hands off; its end-of-iteration is `commit.py`, driven here exactly as the agent would (`--decision accept` then `--decision reject`).

**BEFORE** (pre-fix `commit.py` from `HEAD`):
```
JOURNAL.md: No such file or directory
   1 "kind": "accept"
   1 "kind": "reject"                <-- ZERO "step" events
```
**AFTER**:
```
-rw-r--r--  1 osherelhadad  wheel  2526 Aug 22 17:04 JOURNAL.md
   1 "kind": "accept"
   1 "kind": "reject"
   2 "kind": "step"

{"kind": "step", "candidate": "cand_r1", "accept": true,  "val": 1.0, "parent": "seed",    …}
{"kind": "step", "candidate": "cand_r2", "accept": false, "val": 0.5, "parent": "cand_r1", …}
```
Note `cand_r2`'s parent is `cand_r1` — `commit.py` now reads `best_id` *before* `set_best` moves it, so the lineage edge is real rather than the dashboard's `last_accepted` guess.

### hill-climb and skillopt — unchanged behaviour, confirmed not regressed

| algorithm | JOURNAL.md before → after | `step` records / charged iterations |
|---|---|---|
| hill-climb | present → present (3 entries) | 3 / 3 |
| skillopt | present → present (3 entries) | 3 / 3, plus 2 `skillopt_step` now correctly auxiliary |
| gepa | **absent → present (3 entries)** | **1 → 3 / 3** |
| agent-optimize | **absent → present (2 entries)** | **0 → 2 / 2** |
| evograph | no engine (agent-prose loop) | SKILL.md §2.6 now names the seam |

All four runnable algorithms reach `test_reward 1.0` from `baseline_val 0.0` with the test split sealed — the fix changes bookkeeping, not the optimization or the gate.

## The parametrized regression test

`core/tests/test_iteration_record_parity.py` — parametrized over `hill-climb`, `gepa`, `skillopt`, `agent-optimize` (8 cases), offline and zero-API:

- `len(step records) == spent.iterations` — the single assertion the issue asked for, and the one that would have caught all of #109/#216/the skillopt double-count;
- exactly one record per candidate (the duplicate direction);
- every record carries `parent` and `accept` (the lineage edge `_parent_map`/LEDGER/RUNMAP read);
- a non-empty run-level `JOURNAL.md` naming every candidate;
- and a second test asserting the *consumers* are not left empty — `_parent_map` non-empty, `LEDGER.md` not `"(baseline only)"`, `RUNMAP.md` not `"(no prior iterations yet)"`, every candidate present in the ledger. That is #109's exact symptom, asserted per algorithm.

**These tests enumerate, and enumeration is exactly the failure mode this PR is about** — a sixth algorithm is invisible to them until someone remembers to add it to the dict, which is the same forgetting the seam exists to prevent. So the file also carries the invariant itself, asserted on the source rather than on a list of algorithms:

`test_exactly_one_place_charges_an_iteration` parses every `.py` under `core/cap_evolve/` and `skills/` with `ast`, collects every `update_spent(iterations=…)` call site with its enclosing function, and asserts there is exactly one — `harness.record_iteration`. It holds for an algorithm nobody has written yet. Verified by adding a rogue charge site to `skillopt.py`:

```
E  AssertionError: an iteration must be charged in exactly ONE place — harness.record_iteration,
E  which also writes the `step` record and reconciles JOURNAL.md. Found 2:
E  [('core/cap_evolve/harness.py', 1040, 'record_iteration'),
E   ('core/cap_evolve/skillopt.py', 506, '_fake_sixth_algorithm_loop')].
E  If one of these is a new algorithm's loop, call harness.record_iteration instead of
E  update_spent(iterations=...).
```

```
$ PYTHONPATH="$PWD/core:$PWD/dashboard/backend" python3 -m pytest core/tests/test_iteration_record_parity.py -q
9 passed in 14.12s
```

## Which of #224's four claimed bugs this fixes — and which it does not

| # | Claim | This PR |
|---|---|---|
| **#109** — consumers filter `kind == "step"`, so GEPA's LEDGER/RUNMAP/`prior_iterations/` are permanently empty | **FIXED** at the source: GEPA writes `step`. Asserted per algorithm by `test_no_consumer_is_left_empty`. |
| **#216** — `_reconcile_journal` only reachable from `run_step`, so GEPA never writes a run-level `JOURNAL.md` | **FIXED**. Before/after above; `Closes #216`. Also fixes it for `_tamper_step` and agent-optimize, which had the same gap. |
| **SkillOpt double-count** — two rows per iteration, second with `parent=None` | **FIXED** at the source: `skillopt_step` is out of both `_STEP_KINDS` tuples and into `eventstream.BOOKKEEPING_KINDS`, so one row per iteration. No cost effect — see the correction above; the event carries no `opt_cost_usd`. The severe half of the original claim — "poisoned `_parent_map` / premature plateau `stop`" — was **not real on main**: it described PR #199's shared reader, which was never merged. |
| **#110** — 4-5 copies of the optimizer-scratch name list | **NOT in this PR.** Already mostly fixed on main by `types.NON_CAPABILITY_FILES`. The residual (`FOCUS.md`/`REFLECTION.md` missing from `harness._SNAPSHOT_IGNORE`, and gepa's `snapshot` calls not passing `ignore=`) is snapshot hygiene, a different concern, and is owned by the #110/PR #211 work — deliberately left out to keep this diff to the one root cause. |

### Explicitly not done

**Iterations do not get a new stable identity.** #224's headline proposal — a new gated-iteration record type, a typed `RunDir.iteration_events()` as the only enumeration path, and making the kind constants private — is **not** implemented, and is not required for the routing fix to be correct. `step` already carries every field the proposed record would (`candidate`, `parent`, `accept`, `reason`, `val`, `parent_val`); the bug was one algorithm not writing it. Adding a second representation to fix "consumers re-derive iteration-ness" would leave *both* representations and double the surface a sixth algorithm can forget. With `update_spent(iterations=…)` now callable from exactly one function, the identity is enforced where it matters — at the write, not at every read.

Two further items left alone, both cosmetic and pre-existing:
- GEPA's optimizer spend is still recorded via `update_spent` only, not carried on its `step` event, so the dashboard cost ledger cannot attribute it per candidate. Threading `opt_cost_usd` into `_full_val_gate` is a separate change.
- `dashboard._ALGO_MARKERS` identifies `hill-climb` by a bare `step`. Now that agent-mode `commit.py` also writes `step`, a hand-built run dir with no `run_config` event and no `screen` event could be mislabelled `hill-climb` instead of "not recorded". Real runs are unaffected: `run_config` is emitted by the CLI and wins over kind inference.

## Full suite

```
$ PYTHONPATH="$PWD/core:$PWD/dashboard/backend" python3 -m pytest core/tests -q
798 passed, 12 skipped in 130.73s (0:02:10)
```
(clean main is `789 passed, 12 skipped`; +9 are this PR's new tests.) Zero existing test changes were needed — including `test_gepa.py` (all cases green: the local gate still short-circuits before any full-val spend, the metric-call budget still caps the run, the merge path still works) and `test_dashboard_status_cost_log.py`'s skillopt dedup case.

## Merge coordination

Per `.review/MERGE_PLAN.md` the order is **350 → 346 → 355 → this PR**. #355 restructures the same iteration path, so expect to rebase this on top of it.

`harness.py` hunks touch only two regions, both away from the other in-flight work:
- `963-1047` — `_reconcile_journal` signature/format + the new `record_iteration` (inserted after it);
- `1433` (`_tamper_step`) and `1611-1635` (`run_step`'s tail).

No overlap with the optimizer-stderr path (~line 600) or `_SNAPSHOT_IGNORE` (~line 1842).

### DIFF
--- a/core/cap_evolve/dashboard.py
+++ b/core/cap_evolve/dashboard.py
@@ -324,12 +324,19 @@ _ALGO_MARKERS = (
 )
 
 #: Every event kind that means "one candidate was proposed and judged". Every algorithm
-#: now writes ``step`` via ``harness.record_iteration`` (#216/#224) — including gepa,
-#: which used to write only ``gepa_val_gate``, and agent-mode ``commit.py``, whose
-#: ``accept``/``reject`` are kept here so run dirs recorded BEFORE that fix still render.
-#: ``skillopt_step`` / ``gepa_val_gate`` are deliberately absent: they duplicated the
-#: same candidate's iteration (see the per-candidate dedup below).
-_STEP_KINDS = ("step", "accept", "reject")
+#: now writes ``step`` via ``harness.record_iteration`` (#216/#224).
+#:
+#: The legacy kinds STAY here — ``gepa_val_gate`` and agent-mode ``commit.py``'s
+#: ``accept``/``reject`` — because run dirs recorded BEFORE that fix have no ``step``
+#: events at all (that was the bug), and this reducer is what renders them. Dropping
+#: ``gepa_val_gate`` took a historic gepa run's graph from 2 nodes to 1, retroactively.
+#: They cost nothing on a new run: the loop below keys one iteration per CANDIDATE, so
+#: the extra kind lands on the node ``step`` already created.
+#:
+#: ``skillopt_step`` is the one kind deliberately absent, and for a different reason —
+#: it is not a legacy record but epoch DETAIL logged alongside a ``step`` for the same
+#: candidate on the same (current) runs, so it never carried a graph anyone needs.
+_STEP_KINDS = ("step", "gepa_val_gate", "accept", "reject")
 
 #: Kinds whose presence means "this candidate was accepted" without an ``accept`` field.
 _ACCEPT_KINDS = ("accept",)
--- a/core/cap_evolve/dashboard.py
+++ b/core/cap_evolve/dashboard.py
@@ -323,12 +323,13 @@ _ALGO_MARKERS = (
     ("hill-climb", ("convergence", "step")),
 )
 
-#: Every event kind that means "one candidate was proposed and judged". The
-#: deterministic loops log ``step``/``skillopt_step``/``gepa_val_gate``; the AGENT-mode
-#: loops (agent-optimize, evograph) log ``accept``/``reject`` via the algorithm's
-#: ``commit.py``. The reducer used to know only the first three, so every free-form
-#: agentic run rendered as a graph with nothing but a seed node.
-_STEP_KINDS = ("step", "skillopt_step", "gepa_val_gate", "accept", "reject")
+#: Every event kind that means "one candidate was proposed and judged". Every algorithm
+#: now writes ``step`` via ``harness.record_iteration`` (#216/#224) — including gepa,
+#: which used to write only ``gepa_val_gate``, and agent-mode ``commit.py``, whose
+#: ``accept``/``reject`` are kept here so run dirs recorded BEFORE that fix still render.
+#: ``skillopt_step`` / ``gepa_val_gate`` are deliberately absent: they duplicated the
+#: same candidate's iteration (see the per-candidate dedup below).
+_STEP_KINDS = ("step", "accept", "reject")
 
 #: Kinds whose presence means "this candidate was accepted" without an ``accept`` field.
 _ACCEPT_KINDS = ("accept",)
--- a/core/cap_evolve/eventstream.py
+++ b/core/cap_evolve/eventstream.py
@@ -73,8 +73,11 @@ BOOKKEEPING_KINDS = ("minibatch", "optimizer_context_warning", "target_profile",
                      "gepa_merge_skip", "gepa_merge_local", "skillopt_slow_eval",
                      "skillopt_slow_update")
 
-# The one-iteration-finished events, each carrying the optimizer's own spend.
-_STEP_KINDS = ("step", "gepa_val_gate", "skillopt_step")
+# The one-iteration-finished event, carrying the optimizer's own spend. Exactly ONE
+# kind: every algorithm writes it via ``harness.record_iteration`` (#216/#224).
+# ``gepa_val_gate`` is gone and ``skillopt_step`` is auxiliary epoch detail — listing
+# either here rendered two rows and summed ``opt_cost_usd`` twice for one iteration.
+_STEP_KINDS = ("step",)
 
 
 def read_new_events(path: Path, offset: int) -> tuple[list[dict], int]:
@@ -356,10 +359,7 @@ def format_event(ev: dict, totals: dict | None = None, *,
     elif kind in _STEP_KINDS:
         ok = bool(ev.get("accept"))
         verdict = "ACCEPT" if ok else "reject"
-        where = ""
-        if kind == "skillopt_step":
-            where = f" e{ev.get('epoch')}s{ev.get('step_in_epoch')}"
-        body = (f"{verdict}{where}  {ev.get('candidate')}  val={_num(ev.get('val'))}"
+        body = (f"{verdict}  {ev.get('candidate')}  val={_num(ev.get('val'))}"
                 f" (parent {_num(ev.get('parent_val'))})")
         if ev.get("reason"):
             body += f"  — {ev['reason']}"
--- a/core/cap_evolve/eventstream.py
+++ b/core/cap_evolve/eventstream.py
@@ -71,12 +71,13 @@ FOLLOW_END = "_follow_end"
 BOOKKEEPING_KINDS = ("minibatch", "optimizer_context_warning", "target_profile",
                      "seed_dir_created", "splits_warning", "gepa_resume",
                      "gepa_merge_skip", "gepa_merge_local", "skillopt_slow_eval",
-                     "skillopt_slow_update")
+                     "skillopt_slow_update", "skillopt_step")
 
 # The one-iteration-finished event, carrying the optimizer's own spend. Exactly ONE
 # kind: every algorithm writes it via ``harness.record_iteration`` (#216/#224).
-# ``gepa_val_gate`` is gone and ``skillopt_step`` is auxiliary epoch detail — listing
-# either here rendered two rows and summed ``opt_cost_usd`` twice for one iteration.
+# ``gepa_val_gate`` is no longer emitted at all, and ``skillopt_step`` is auxiliary
+# epoch detail that duplicated the same candidate's row — it is skipped in the terminal
+# via BOOKKEEPING_KINDS above (the dashboard still shows it).
 _STEP_KINDS = ("step",)
 
 
--- a/core/cap_evolve/gepa.py
+++ b/core/cap_evolve/gepa.py
@@ -61,6 +61,7 @@ from .harness import (
     _paired_deltas,
     _resolve_workers,
     evaluate_candidate,
+    record_iteration,
     split_result_from_rollouts,
 )
 from .loop import SplitResult, aggregate_scores, has_valid_trials
@@ -611,9 +612,10 @@ def gepa_loop(
             if not report.ok:
                 run_dir.log_event("tamper_detected", candidate=cid, parent=parent["id"],
                                   report=report.to_dict(), reason=report.reason)
-                run_dir.update_spent(iterations=1, accepted=None)
-                run_dir.log_event("step_indecisive", candidate=cid,
-                                  reason="indecisive (integrity): " + report.reason)
+                _reason = "indecisive (integrity): " + report.reason
+                record_iteration(run_dir, workdir, cid, parent_id=parent["id"],
+                                 accepted=False, reason=_reason, indecisive=True)
+                run_dir.log_event("step_indecisive", candidate=cid, reason=_reason)
                 steps.append({"candidate_id": cid, "parent_id": parent["id"], "minibatch": mb,
                               "accepted": False, "candidate_val": None,
                               "tamper": report.to_dict(), "optimizer_error": opt_error,
@@ -638,8 +640,12 @@ def gepa_loop(
         }
 
         if not local_pass:
-            # Cheap rejection — no full-val spend. Record it for memory + audit.
-            run_dir.update_spent(iterations=1, accepted=False)
+            # Cheap rejection — no full-val spend, so ``val`` stays None: the minibatch
+            # reward is NOT a val number and must never be written into that field.
+            record_iteration(run_dir, workdir, cid, parent_id=parent["id"], accepted=False,
+                             reason="local minibatch gate: sum(child) <= sum(parent)",
+                             parent_val=parent["result"].reward, focus=focus_label,
+                             mb_child=child_mb.reward, mb_parent=parent_mb.reward)
             rejected.add(cid, f"candidate {cid} (mb {child_mb.reward:.3f} vs parent "
                               f"{parent_mb.reward:.3f})",
                          "local minibatch gate: sum(child) <= sum(parent)", child_mb.reward)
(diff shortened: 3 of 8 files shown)


OsherElhadad:
Superseded by #396 — rebased onto current `main` (23 PRs later). This branch is untouched; please merge #396 instead.

Issue #252, opened 2026-07-31
A relative optimizer_instructions_file resolves differently in check and run — run silently falls back to a generic template


OsherElhadad:
A relative `optimizer_instructions_file` in `capevolve.yaml` is resolved against **two different base directories** by `cap-evolve check` and `cap-evolve run`. `check` passes; `run` silently drops the flag and the optimizer receives a **generic template telling it to edit code the project does not have**.

## The asymmetry, verified on `main`
`skills/phases/implement-and-check/scripts/pipeline_selftest.py:73-77` resolves **project-relative** and *reports a problem* when missing:
```python
instr_rel = str(spec.get("optimizer_instructions_file") or "optimizer/INSTRUCTIONS.md")
instr_path = project / instr_rel
if not instr_path.exists():
    problems.append(f"optimizer_instructions_file points at a missing file: {instr_rel} ...")
```
`core/cap_evolve/cli.py` resolves the same key against the **caller's cwd** (`project` is itself the relative `".capevolve/project"`, cli.py:194-196), and on a miss simply **omits the flag** — no warning, no problem, exit 0. `harness.py:1790` then falls back to the generic template, and `harness.py:1792` has a bare `except` that also lands there.

So the hard gate greenlights a spec that `run` silently ignores. `check` is supposed to be the thing that catches a half-wired project.

## Observed consequence
Found while building #124's cheap-real example (PR #249): the optimizer was handed the **tau2 "edit the tool code" template** for a project with **no tools**, and spent real money optimizing the wrong task. The run completed with exit 0 and a plausible-looking number. From the outside it was indistinguishable from an honest result.

This is the epic's signature failure at its most expensive: the fallback is silent, the wrong instructions are *plausible*, and the only symptom is a bad score that reads as a hard benchmark.

## Fix
1. **One resolution rule, in one place.** Resolve `optimizer_instructions_file` (and every sibling path key — `target_profile_file` has the same shape at `cli.py:390-395`) **project-relative**, matching what the selftest already validates. A spec path should mean the same thing to every consumer.
2. **Never fall back silently.** If the file is named and missing, that is a spec error: fail, or at minimum warn on stderr naming the resolved path tried. Compare **#251** — three layers were discarding optimizer stderr on success, so even an existing warning would not have surfaced. Fix this one after or alongside that.
3. **Remove the bare `except` at `harness.py:1792`** that routes any failure to the generic template. A parse error and "no file configured" are different facts and must not share an outcome.
[...]

Issue #251, opened 2026-07-30
Optimizer stderr is discarded on success by three separate capture_output layers — agent CLI diagnostics never reach the user


OsherElhadad:
Three independent layers capture the optimizer subprocess's stderr and discard it **on success**, so anything an agent CLI writes to stderr during a normal run is invisible to the user.

## Verified on `main`
```
$ grep -c 'capture_output=True' core/cap_evolve/harness.py core/cap_evolve/cli.py \
      skills/optimizers/run-optimizer/scripts/run.py
core/cap_evolve/harness.py:1
core/cap_evolve/cli.py:1
skills/optimizers/run-optimizer/scripts/run.py:1
```
`run-optimizer/scripts/run.py:298` → `harness.py:450` → `cli.py:436`. Each wraps the next; each keeps stderr only to attach it to an *error*. On a zero-exit run all three drop it.

## How it was found
While fixing #133 (PR #248), a `_mock_apply.py` warning was added for the case where no mock script is findable — the defect being that the documented quickstart path silently produced `test_reward 0.0` with exit 0 and no diagnostic. **Adding the warning was not enough: it never reached the user**, because all three layers swallowed it on the successful path. Fixing one or two layers would have looked correct and changed nothing.

## Why it matters beyond that one warning
- **Any** optimizer CLI's diagnostics are lost: rate-limit notices, deprecation warnings, partial-failure messages, "model not found, falling back to…", auth warnings. All are exactly the information a user needs when a run completes but the numbers look wrong.
- It compounds the epic's most repeated defect — silence that looks like success. A run that finishes with exit 0 and a bad number, where the subprocess *did* explain why, is the worst version of that: the explanation existed and was thrown away.
- It also hid a real bug during this batch: a relative `optimizer_instructions_file` silently falls back to the generic template (found in #124/PR #249), so the optimizer received tau2 "edit the tool code" advice for a project with no tools and **optimized the wrong task**. Any stderr from that path would have gone nowhere.

## Status
**Fixed in PR #248** — all three layers now relay stderr on success, verified end to end (the warning appears, exit stays 0, and stdout remains exactly one JSON object per #217). 213 tests pass with it.

This issue exists because that fix is **cross-cutting plumbing landing inside a quickstart PR**. Its author flagged the scope honestly: `harness.py`, `cli.py` and `run-optimizer` are shared, so #248 now carries a small behavioural change for *every* optimizer, not just the quickstart path. Filing it so:
[...]

Issue #217, opened 2026-07-29
cap-evolve run prints two JSON objects to stdout — the machine-readable contract has never been parseable


OsherElhadad:
`cap-evolve run` prints **two** concatenated JSON objects to stdout, so the documented machine-readable contract has never actually worked. Any script doing `cap-evolve run ... | jq` or `json.loads(...)` fails.

## Verified on `main`
`core/cap_evolve/cli.py:177-180`
```python
if dash_mode == "auto":
    status = dashboard_launch.maybe_launch(
        proj_abs.parent, mode=dash_mode, port=dash_port, open_browser=True)
    print(json.dumps(status))          # ← object #1 on stdout
```
…then `_cmd_run` prints the final report object at the end — object #2. Result:
```
$ cap-evolve run --spec ... | python -c 'import json,sys; json.load(sys.stdin)'
json.decoder.JSONDecodeError: Extra data: line 2 column 1 (char 39)
```

`dash_mode` defaults to `auto`, and `maybe_launch` prints its status whether the dashboard **launches or is skipped** — so this fires on essentially every default run, not an edge case.

## Why it matters
- `RUN.md` and the docs present the final object as the machine-readable result. It isn't parseable, so any automation built on it breaks on the first run.
- It's silent: both objects are individually valid JSON, so a reader eyeballing the output sees nothing wrong. Only a parser notices.
- It defeats the separation the project already decided on: #116 (PR #191) deliberately routes human-facing progress to **stderr** precisely so stdout stays a single parseable object. This line predates that convention and contradicts it.

## Fix
Move the dashboard-launch status to **stderr** (it is progress/diagnostic information, not the run's result), leaving exactly one JSON object on stdout. PR #214 (issue #137) already does this and proves it:
```
json.loads OK — keys: ['baseline_val', 'best_id', 'dashboard', 'dashboard_server',
 'iterations', 'run_dir', 'test_baseline_reward', 'test_delta', 'test_pass_k', 'test_reward']
```
Note the final object already carries `dashboard` and `dashboard_server` keys, so no information is lost by moving the early line off stdout.

`examples/toy_calc/run.sh` has the same defect in shell form (a `Working directory:` echo on stdout) — also fixed in #214.

## Acceptance
- `cap-evolve run` stdout parses as exactly one JSON object, with the dashboard launched **and** skipped (test both — the skip branch also printed).
[...]

PR #357, opened 2026-08-22
fix(optimizer): a step that never got its prompt must not look like a step that proposed nothing (#251, #252)

### LINKED ISSUE #354: run-optimizer silently substitutes an EMPTY prompt when --prompt names a missing file, and exits 0
`skills/optimizers/run-optimizer/scripts/run.py` substitutes an **empty prompt** when
`--prompt` names a file that does not exist, and exits 0.

## Verified on `main`

`run.py:249`:

```python
prompt_path = Path(args.prompt).resolve()
prompt_text = prompt_path.read_text(encoding="utf-8") if prompt_path.exists() else ""
```

Reproduced:

```
$ python skills/optimizers/run-optimizer/scripts/run.py --name mock \
    --workdir /tmp/rotestA --prompt /tmp/rotestA/NOPE.md
{"optimizer": "mock", "cli_present": true, "returncode": 0, "auth_present": [],
 "stdout_tail": "{\"optimizer\": \"mock\", \"applied\": [], \"note\": \"no mock_script.json found; no edits made\"}\n",
 "stderr_tail": ""}
EXIT=0
```

No error, no warning, exit 0.

## Why it matters

`{prompt_text}` is inlined by **10 of the 14 registry rows** (`claude-code`, `codex`,
`gemini-cli`, `opencode`, `ibm-bob`, `cursor`, `droid`, `copilot`, `kimi`, `pi`), so the
expansion becomes e.g. `claude -p '' --permission-mode acceptEdits …` — a real, billed
agent invocation with **no instructions at all**. The iteration then produces no edit, the
gate rejects the candidate as "the optimizer had nothing to propose", and the run reports a
plausible number. From the outside it is indistinguishable from an honest iteration where
the optimizer genuinely had no ideas — which corrupts the run's evidence, not just one step.

`--prompt` is already mandatory (`run.py:233-234`). A mandatory path that silently resolves
to `""` is strictly worse than a missing flag.

## Fix

Make a named-but-missing prompt a hard error, matching the existing `cli_present: false`
shape at `run.py:288-292`: print `{"optimizer": …, "error": "--prompt file does not exist: <resolved path>"}`
and return 2.

## Acceptance

- `run.py --prompt <missing>` exits non-zero with an error naming the **resolved** path.
- No code path sends `prompt_text=""` to a non-`mock` row.
- Pinned by a regression test.

## Family

Same signature as #252 (silent plausible fallback → wrong work → exit 0) and #251 (the
diagnostic that would have explained it is discarded). All three are being fixed together:
an optimizer step that did not get what it needed must not look like a successful step that
had nothing to propose.

Found while re-verifying #251/#252 against current `main` during the review of
`skills/optimizers/run-optimizer`.

OsherElhadad:
Three verified bugs in the optimizer invocation path. One PR, because they are all the
same failure: **the optimizer subprocess doesn't get what it needed, and nobody finds out.**

The unifying rule this PR enforces: *an optimizer step that did not get what it needed must
not look like a successful step that had nothing to propose.* That confusion does not waste
one iteration — it corrupts the run's evidence, because the gate's "rejected: no
improvement" becomes indistinguishable from "the optimizer was never given a chance".

Fixes #354 (opened for bug A, which had no issue), #251, #252.

---

## A — silent empty prompt (#354)

`run.py:249` substituted `prompt_text = ""` when `--prompt` named a missing file, and exited 0.
For the **10 rows that inline `{prompt_text}`** that expands to e.g. `claude -p '' …` — a real,
billed agent invocation with no instructions.

**Before**
```
$ python skills/optimizers/run-optimizer/scripts/run.py --name mock \
    --workdir /tmp/rotestA --prompt /tmp/rotestA/NOPE.md
{"optimizer": "mock", "cli_present": true, "returncode": 0, "auth_present": [],
 "stdout_tail": "{\"optimizer\": \"mock\", \"applied\": [], \"note\": \"no mock_script.json found; no edits made\"}\n",
 "stderr_tail": ""}
EXIT=0
```

**After**
```
$ python skills/optimizers/run-optimizer/scripts/run.py --name mock \
    --workdir /tmp/rotestA --prompt /tmp/rotestA/NOPE.md
{"optimizer": "mock", "error": "--prompt file does not exist: /private/tmp/rotestA/NOPE.md"}
EXIT=2

$ ... --prompt /tmp/rotestA/INSTRUCTIONS.md          # existing file, unchanged
{"optimizer": "mock", "cli_present": true, "returncode": 0, ...}
EXIT=0
```

The runner never reaches the agent CLI, so nothing is billed. `--prompt` was already
mandatory; a mandatory path that resolves to `""` is strictly worse than a missing flag.

---

## B — discarded stderr (#251)

All three `capture_output=True` layers still dropped the optimizer's stderr on the **zero-exit**
path, confirmed on `main` (PR #248's fix is not present).

**Repro: a stub "optimizer" that writes a real diagnostic and exits 0**
```sh
$ cat /tmp/rotestB/stub_agent.sh
#!/bin/sh
echo "STUB-DIAGNOSTIC: rate limit hit, model fell back to a cheaper one" >&2
exit 0
$ export CAPEVOLVE_OPTIMIZER_CMD="/tmp/rotestB/stub_agent.sh --workdir {workdir} --prompt {prompt}"
```

**Before**
```
# layer 1 — run.py
$ python .../run.py --name generic --workdir /tmp/rotestB --prompt /tmp/rotestB/INSTRUCTIONS.md 2>err
{"optimizer": "generic", ..., "returncode": 0, "stderr_tail": "STUB-DIAGNOSTIC: ...\n"}
EXIT=0
run.py OWN stderr: []                        <- buried inside the JSON, never relayed

# layer 2 — harness.optimizer_from_command
optimizer_from_command returned: None
harness caller's stderr: []                  <- the whole payload dropped on returncode 0

# layer 3 — cli.py `run()` reads proc.stderr only on the failure branches
```

**After**
```
# layer 1
run.py OWN stderr: [STUB-DIAGNOSTIC: rate limit hit, model fell back to a cheaper one]
EXIT=0
stdout json objects: 1
stdout parses as exactly one JSON object: OK          <- #217 intact

# layer 2
optimizer_from_command returned: {"stderr": "STUB-DIAGNOSTIC: rate limit hit, ...\n"}
```

**End to end** (the toy wiring with no `mock_script.json`, i.e. #251's own founding example —
a real optimizer that explains itself and exits 0):
```
--- persisted in the run dir ---
$D/.capevolve/run_e2e/work/cand_0001.optimizer.stderr
$D/.capevolve/run_e2e/work/cand_0002.optimizer.stderr
--- content ---
mock optimizer: no mock_script.json found; no edits made (looked under .../cand_0001 and $CAPEVOLVE_MOCK_SCRIPT)
--- event stream ---
"kind": "optimizer_stderr", "candidate": "cand_0001", "path": "work/cand_0001.optimizer.stderr", "chars": 212, "tail": "mock optimizer: no mock_script.json found; ..."
```

### Volume: what goes where, and why

| stream | choice | reason |
|---|---|---|
| `run.py` → its own stderr | **always relay**, verbatim | the runner is the only place that has the child's stderr as a stream; `stdout` is the one-JSON contract (#217) so `stderr` is the free channel, and a stream survives the layers above |
| harness → run dir + event stream | **capture, persist to `work/<cand>.optimizer.stderr`, log `optimizer_stderr` with a 2 KB tail** | an agent CLI can be extremely chatty over a 50-iteration run. Dumping it to the terminal on success would bury the progress output the user is actually watching, and the interesting case is usually read *after* the run, next to the candidate that produced it. Durable + attributable + greppable beats transient |
| cli.py phase steps → terminal stderr | **relay on success**, tail-clipped to 8 KB, guarded by the existing `_stderr_is_usable()` | there are only four phase steps per run and their stderr is normally empty, so volume is a non-issue; this is the half that makes a diagnostic actually reach a human in real time — and it is what carries bug C's new warning to the terminal |
| `stderr_tail` in the JSON | cap raised 500 → 800 to match `stdout_tail` | once stderr is a relayed stream, the tail is only a convenience field; the asymmetric 500 was just noise |

`_mock_apply.py`'s "no mock_script.json found" note also moved from **stdout to stderr** — it was
on the wrong stream *and* discarded by both layers above, which is why #251 says adding that
warning changed nothing.

Contract now stated in `CONTRIBUTING.md` next to the stdout half: **stdout is captured and
parsed, stderr is relayed** — on the zero-exit path too.

---

## C — instructions-path resolution (#252)

One relative `optimizer_instructions_file`, **three different answers**. Repro fixture: a
project-scoped template at `$D/.capevolve/project/optimizer/INSTRUCTIONS.md` and a cwd-relative
decoy at the same relative path, each marked, then read back from what the optimizer *actually
received* (`candidates/*/INSTRUCTIONS.md`).

**Before**
```
--- (1) check / pipeline_selftest resolution (project-relative) ---
  "ok": true,
    "optimizer template OK with intact placeholders: optimizer/INSTRUCTIONS.md"
--- (2) run, invoked from $D (cwd == workdir) ---
    optimizer actually received: CWD-DECOY-MARKER
--- (3) run, invoked from /tmp (different cwd, absolute --project) ---
    optimizer actually received: NEITHER (silent generic-template fallback)
```

**After**
```
--- (1) check / pipeline_selftest resolution (project-relative) ---
  "ok": true, ...
--- (2) run, invoked from $D (cwd == workdir) ---
    optimizer actually received: PROJECT-SCOPED-MARKER
--- (3) run, invoked from /tmp (different cwd, absolute --project) ---
    optimizer actually received: PROJECT-SCOPED-MARKER
```

- `specfile.resolve_project_path(project, value)` is the **single rule**: a relative spec path
  is project-relative; absolute is returned unchanged. Used by `cap-evolve run` *and*
  `implement-and-check`'s `pipeline_selftest`, so a spec the gate passes cannot have a path the
  run resolves to a different file. A test asserts both call it.
- It resolves against the **absolute** project dir. Resolving against the relative
  `".capevolve/project"` was case 3: "project-relative" was still cwd-dependent. Also applied to
  the sibling `runner_repo_path`, which had the identical two-probe shape.
- A miss now **warns**, naming the path tried:
  ```
  warning: optimizer_instructions_file 'optimizer/INSTRUCTIONS.md' does not exist (resolved
  project-relative to /.../.capevolve/project/optimizer/INSTRUCTIONS.md) — the optimizer will
  get cap-evolve's GENERIC template, not your capability-scoped one; `cap-evolve check` and the
  implement-and-check self-test catch this
  ```
- `_focus_instructions`' silent fallbacks are gone: the bare `except` is now `except OSError` with
  a reason, and the third silent path (a named, readable template that lacks `{{FOCUS_SUMMARY}}`)
  also announces itself. A template that was never *named* stays quiet — there is nothing to report.
- **Not made fatal, deliberately.** Missing capability-scoped instructions ≠ no instructions:
  there is a complete built-in template, so the proportionate response is a loud diagnostic, not
  killing a long run. The empty *prompt* (bug A) is the one that must be fatal, because there the
  optimizer gets literally nothing. This satisfies #252's "fails **or warns visibly**".
- `examples/toy_calc/run.sh` now copies `templates/project/optimizer/` — its spec names
  `optimizer/INSTRUCTIONS.md` and the example never shipped it, so the toy was silently running on
  the generic fallback (and would now warn on every run). One line; the shipped example is
  correctly wired and quiet.

Note how B and C interlock: C's warning reaches a terminal **only because** of B's layer-3 fix.
That is exactly #251's point — a warning nobody can see is not a fix.

---

## Evidence

```
$ python -m pytest core/tests -q 2>&1 | tail -25
797 passed, 12 skipped in 114.02s (0:01:54)

$ python skills/optimizers/run-optimizer/scripts/check.py
{"notes":[...],"ok":true,"problems":[],"skill":"run-optimizer"}

$ bash examples/toy_calc/run.sh 2>&1 | tail -15
{
  "best_id": "cand_0001",
  "baseline_val": 0.0,
  "test_reward": 1.0,
  "test_baseline_reward": 0.0,
  "test_delta": 1.0,
  "test_pass_k": {"1": 1.0},
  "iterations": 3,
  ...
}
```

### Regression tests

New file `core/tests/test_optimizer_subprocess_silence.py` (8 tests) — deliberately a **new
file**, to avoid colliding with the agents concurrently editing `core/tests/test_core.py`.
Verified to fail before the fix and pass after:

```
# with the source changes reverted, the new tests only:
7 failed, 1 passed          (the 1 pass is the "an existing prompt still runs" control)
# with the fix:
8 passed
```

| test | bug |
|---|---|
| `test_missing_prompt_file_is_a_hard_error_not_an_empty_prompt` | A |
| `test_existing_prompt_file_still_runs` | A (control) |
| `test_run_py_relays_child_stderr_on_success_and_keeps_stdout_one_json` | B, layer 1 (+ #217) |
| `test_harness_keeps_optimizer_stderr_on_the_zero_exit_path` | B, layer 2 |
| `test_recorded_optimizer_stderr_is_persisted_and_logged` | B, persistence + event |
| `test_relative_spec_path_is_project_relative_regardless_of_cwd` | C (asserts a cwd decoy never wins) |
| `test_check_and_run_share_the_resolver` | C (`check` and `run` cannot diverge again) |
| `test_named_but_unrenderable_template_warns_instead_of_falling_back_silently` | C (all three fallbacks) |

---

## Merger notes — exact lines touched in shared files

`harness.py` is the highest-collision file in flight, so the diff there is surgical.
**Post-merge line numbers:**

| range | what | collides with |
|---|---|---|
| 612–621 | `optimizer_from_command._run`: keep `proc.stderr` on the zero-exit path in the returned report | nothing — no other agent is on the stderr path |
| 626–647 | **new** `_record_optimizer_stderr()` helper, inserted immediately before `_optimizer_failure_detail` | nothing; pure addition |
| 1516 | **one added line** in `run_step`'s propose block: `_record_optimizer_stderr(run_dir, cid, opt_report.get("stderr"))` | nothing |
| 2076–2096 | `_focus_instructions`' template-resolution block: `except OSError` + a reason instead of a bare `except`, and a warning before the generic fallback | ⚠️ **this is inside the optimizer-context assembly.** Contained to the ~10 lines around `tmpl_path = …` / the fallback boundary; the `repl` dict, every placeholder, and the whole fallback prose body are untouched. Whoever owns the context-assembly change should be able to take both. |

`_SNAPSHOT_IGNORE` (~1842) is **not touched** — that is the other agent's change, and none of
these three bugs need it.

Other files: `cli.py` (the `run()` helper + the hill-climb `--instructions-file` /
`--bench-repo` block), `specfile.py` (pure addition), `run.py`, `_mock_apply.py`,
`pipeline_selftest.py` (one line), `CONTRIBUTING.md`, `SKILL.md`,
`examples/toy_calc/run.sh` (one line).

Zero new runtime deps; no vendor branch anywhere — all 14 optimizers keep living as registry
rows (CONTEXT.md principle 1).

### Deliberately out of scope
- The rest of `run-optimizer`'s review (`_LEGACY_ALIASES` dead code, `_SNAPSHOT_IGNORE`
  derivation, `build_command`'s model-drop peek, SKILL.md's registry-field table, checked-in
  `__pycache__`) — real, but unrelated to "the subprocess failed and nobody found out".
- #252's wider audit of `dataset_source` / `split_ids_file` / `capability_path`. Fixed here:
  `optimizer_instructions_file` and `runner_repo_path`, the two that shared the broken two-probe
  shape at the same call site. The others are consumed elsewhere and deserve their own pass.
- Auditing every remaining `capture_output=True` on a scoring path (#251's own follow-up).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,6 +29,10 @@ See [docs/EXTENDING.md](docs/EXTENDING.md) for the token vocabulary and wiring.
   into a skill. Gate on val, seal test, report variance.
 - **Skills stay host-agnostic.** `scripts/run.py` must print a single JSON object
   to stdout (the contract), and must not depend on a specific agent host.
+- **Subprocess streams: stdout is captured and parsed, stderr is RELAYED** — on the
+  zero-exit path too. A step that exits 0 after explaining a problem must not look
+  like a step that had nothing to say. (Optimizer stderr is additionally persisted to
+  `work/<cand>.optimizer.stderr` and logged as an `optimizer_stderr` event.)
 - **Zero runtime deps in core.** Optional features go behind extras.
 - Add a test for any core change (`core/tests/`). Run `python -m compileall core skills`.
 
--- a/core/cap_evolve/cli.py
+++ b/core/cap_evolve/cli.py
@@ -687,7 +687,15 @@ def _cmd_run(argv):
     py = sys.executable
 
     def run(cmd):
-        return subprocess.run(cmd, capture_output=True, text=True, cwd=str(workdir))
+        proc = subprocess.run(cmd, capture_output=True, text=True, cwd=str(workdir))
+        # Relay a SUCCESSFUL step's stderr too. Only the failure paths below read
+        # proc.stderr, so a step that printed a real diagnostic and exited 0 (the
+        # optimizer's relayed warnings among them) was silently discarded here.
+        # Guarded by _stderr_is_usable(): under `2>&-` this would land in stdout and
+        # break the one-JSON-document contract.
+        if proc.returncode == 0 and proc.stderr and proc.stderr.strip() and _stderr_is_usable():
+            print(_clip(proc.stderr, head=0, tail=8000), file=sys.stderr, flush=True)
+        return proc
 
     # The run sequence is built from the manifest + spec (orchestrate validates the
     # needs/provides DAG); it now includes intake + the check gate before baseline.
@@ -868,18 +876,29 @@ def _cmd_run(argv):
     # as read-only optimizer context. Both are resolved project-relative if not absolute.
     # The instructions file defaults to the scaffolded project/optimizer/INSTRUCTIONS.md.
     if algorithm_name == "hill-climb":
+        from .specfile import resolve_project_path
         instr = spec.get("optimizer_instructions_file") or "optimizer/INSTRUCTIONS.md"
-        instr_p = Path(instr)
-        if not instr_p.is_absolute() and not instr_p.exists():
-            instr_p = Path(project) / instr
+        # ONE resolution rule, shared with implement-and-check's pipeline_selftest:
+        # a relative spec path is PROJECT-relative. It used to be probed against the
+        # caller's cwd first, so the same key resolved to a different file depending on
+        # where `cap-evolve run` was invoked from — and on a miss the flag was silently
+        # dropped and the optimizer got the generic template instead of the
+        # capability-scoped one intake authored (#252).
+        # proj_abs, not the relative `project`: resolving against a relative project
+        # dir is still cwd-dependent, which is the bug (#252). An absolute path is also
+        # safer for the subprocess, which runs with cwd=workdir.
+        instr_p = resolve_project_path(proj_abs, instr)
         if instr_p.exists():
             alg_cmd += ["--instructions-file", str(instr_p)]
+        elif _stderr_is_usable():
+            print(f"warning: optimizer_instructions_file {instr!r} does not exist "
+                  f"(resolved project-relative to {instr_p}) — the optimizer will get "
+                  f"cap-evolve's GENERIC template, not your capability-scoped one; "
+                  f"`cap-evolve check` and the implement-and-check self-test catch this",
+                  file=sys.stderr, flush=True)
         repo = spec.get("runner_repo_path")
         if repo:
-            repo_p = Path(str(repo))
-            if not repo_p.is_absolute() and not repo_p.exists():
-                repo_p = Path(project) / str(repo)
-            alg_cmd += ["--bench-repo", str(repo_p)]
+            alg_cmd += ["--bench-repo", str(resolve_project_path(proj_abs, str(repo)))]
         # Supporting source files (data models / types the tools import) copied verbatim
         # into the optimizer's ./guidance/sources/ so it can write correct code. Resolved
         # project-relative by the harness; we pass them through as given.
--- a/core/cap_evolve/harness.py
+++ b/core/cap_evolve/harness.py
@@ -609,10 +609,40 @@ def optimizer_from_command(cmd_template: list[str]) -> OptimizerFn:
             err.cost = _parse_optimizer_cost(proc.stdout)  # type: ignore[attr-defined]
             raise err
         # Capture optimizer spend (cost_usd/tokens) from run-optimizer's JSON payload
-        # so it counts against the budget and shows in the dashboard. Returns None
+        # so it counts against the budget and shows in the dashboard. Cost is None
         # when the agent CLI emitted no structured cost (spend stays unmeasured).
-        return _parse_optimizer_cost(proc.stdout)
+        report = _parse_optimizer_cost(proc.stdout) or {}
+        # Optimizer stderr on the SUCCESS path: keep it. A CLI that explains itself and
+        # then exits 0 (rate limit, model fallback, auth warning, "no script found") is
+        # the case where discarding stderr turns a step that never got what it needed
+        # into a step that merely proposed nothing. The caller persists it into the run
+        # dir and logs it; it is deliberately not dumped to the terminal on success.
+        if proc.stderr and proc.stderr.strip():
+            report["stderr"] = proc.stderr
+        return report or None
     return _run
+
+
+def _record_optimizer_stderr(run_dir, cid: str, text) -> None:
+    """Persist a SUCCESSFUL optimizer step's stderr and announce it in the event log.
+
+    Always captured, always written to ``work/<cid>.optimizer.stderr`` (durable and
+    reviewable next to the candidate that produced it), and surfaced as an
+    ``optimizer_stderr`` event with a tail — rather than dumped to the terminal,
+    which on a long run would bury the progress output the user is watching.
+    """
+    if not text or not str(text).strip():
+        return
+    text = str(text)
+    rel = f"work/{cid}.optimizer.stderr"
+    try:
+        dst = run_dir.root / rel
+        dst.parent.mkdir(parents=True, exist_ok=True)
+        dst.write_text(text, encoding="utf-8")
+    except OSError as e:
+        rel = f"<unwritable: {e}>"
+    run_dir.log_event("optimizer_stderr", candidate=cid, path=rel,
+                      chars=len(text), tail=text[-2000:])
 
 
 def _optimizer_failure_detail(proc: "subprocess.CompletedProcess") -> str:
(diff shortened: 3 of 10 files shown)


OsherElhadad:
Superseded by #393 — same three fixes rebased onto current main (22 PRs merged since this opened; #386/#391/#370 reworked the `_focus_instructions` region and #388/#391 replaced the `hill-climb` predicate in `cli.py` with `OPTIMIZER_CONTEXT_ALGORITHMS`). All three bugs re-verified fixed after the rebase, 820 passed / 12 skipped, and the 8 regression tests re-confirmed RED-able (7 failed, 1 control passed). Please close this in favour of #393.

PR #345, opened 2026-08-22
feat(skills): dogfood the skill authoring lint in CI (report-only) + validator fixes

OsherElhadad:
Closes #107 (redesign of #213 against current main).

Scope is deliberately **the lint and its CI wiring only**. The skills themselves are being
rewritten concurrently under #322-#339, so this PR does not touch a single `SKILL.md` — it
produces the authoritative violation checklist those rewrites work against, and the gate that
stops the next regression.

## Extend, not new

`skill-package/scripts/abstract.py` already **is** the authoring validator (`MAX_BODY_LINES = 500`,
frontmatter/reference/TOC rules) shipped to user skills, so it is extended in place and pointed at
our own skills; `core/cap_evolve/skillcheck.py` was read first and rejected as the wrong module — it
is the shared harness for each skill's *behavioral* `scripts/check.py` (`Checker`, `SyntheticAdapter`,
`temp_run_dir`) and contains no authoring rule at all.

## Rule table

Every rule below is enforced by the shipped validator and reported by the lint.

| Rule | Budget / condition | Severity in lint |
|---|---|---|
| `name` present | non-empty frontmatter key | hard error |
| `name` length | ≤ 64 chars | hard error |
| `name` charset | `[a-z0-9-]` only | hard error |
| `name` no XML | no `<tag>` | hard error |
| `description` present | non-empty | hard error |
| `description` length | ≤ 1024 chars | hard error |
| `description` no XML | no `<tag>` | hard error |
| `description` when-to-use | states WHEN (conditional *or* positional) | advisory |
| body length | ≤ 500 lines | error (promoted) |
| body tokens | ≤ ~5000 (~chars/4) | error (promoted) |
| references one level deep | no `references/<dir>/`, **and no reference linking another reference** | error (promoted) |
| long-reference TOC | > 300 lines needs ≥ 3 `[section](#anchor)` links up front | error (promoted) |
| relative links resolve | from the body **and** from every reference | error (promoted) |

Severity split: structural findings are measurable, so the lint promotes them to errors;
`validate()` keeps them as warnings because an optimizer mid-run must not be hard-blocked by a
style budget. Description *style* heuristics (POV, all-caps, truncation risk, "say WHEN") stay
advisory — they are judgement calls. `test_the_when_clause_advisory_stays_advisory` pins that.

### Validator bugs fixed (all live on main, all user-facing)

1. **`description: >-` bypassed every description check.** A YAML block scalar parsed as the
[...]

Issue #340, opened 2026-08-22
All 20 skills: cross-skill duplication, frontmatter drift, and the always-in-context token budget


OsherElhadad:
# Cross-skill review: duplication, coherence, and token budget across all 20 skills

## Summary

The 20 skills total **3,829 lines / ~60.8k tokens** of triggered body. Roughly **1,100 of
those lines (≈29%) are restatements of six invariants that already have an owner skill**, and
the two largest files (`agent-optimize` 922, `tools` 749) hold **44% of the corpus in 2 of 20
files** while each already ships a `references/` directory that is under-used.

Three findings dominate:

1. **The failure taxonomy** (KNOWLEDGE / BEHAVIORAL / DECISION-PERMISSION / CAPABILITY-GAP +
   blast radius) is stated **five times across three skills — twice inside `diagnose`
   itself** — for a total of **~166 duplicated lines**. This is not merely redundant on
   disk: `core/cap_evolve/harness.py:1193-1243` copies the selected capability skill **and**
   the `diagnose` skill into the *same* optimizer workdir, so the same reader receives the
   same taxonomy 2-3× in one prompt.
2. **The honesty invariants** (split frozen once → val-only gate `Δ > k·SE` → sealed test →
   never edit `splits.json`/`rollouts/test/*`) are restated in **10 skills beyond their three
   owners** (`gate`, `finalize`, `report`), ≈**96 lines**. Core enforces all of them
   (`rundir.py`, `gate.py`, `splits.py`); the prose is documentation of code, repeated.
3. **The "Agent-mode loop" block** appears in 6 skills (≈**55 lines**), each a slightly
   different paraphrase of the same five rules — and the paraphrases have already **drifted**
   (only `hill-climb` and `skillopt` mention stall; only `evograph` and `agent-optimize`
   mention worktrees/tags).

Structurally, the SKILL.md frontmatter carries `provides`/`needs`/`sources` that **nothing
reads** — the real dependency DAG is validated from `meta.yaml` → `manifest.json` by
`skills/orchestrate/orchestrate/scripts/run.py:60-81`, and `build_manifest.py:69-76` only
compares `name` and `component` between the two files. The duplicate has already drifted
(`baseline`). Nine skills additionally restate their own frontmatter tokens as prose under
`## Inputs / outputs (manifest tokens)` (53 lines).

**Highest-leverage single fix:** make `diagnose` the sole owner of the failure taxonomy and
[...]

PR #359, opened 2026-08-22
using-cap-evolve: rewrite the router's description and fix three misroutes (#339)

### LINKED ISSUE #339: using-cap-evolve: the router's description is its whole job — rewrite it and cover the missing states
# Skill review: `skills/orchestrate/using-cap-evolve` (entry-point router)

## Summary
The router's mechanism is sound — `resolve_state()` is a pure read-only function of the project
dir, it conservatively degrades to `scaffolded` when core is not importable rather than claiming
`ready` it cannot prove (`scripts/run.py:91-94`), and its own `check.py` passes. But its two
load-bearing jobs are both weak. The **description** spends over half its ~85 words re-describing
the body (two run modes, honesty rules, routing internals) instead of buying trigger coverage,
uses non-typed placeholders (`<X>`, `<benchmark>`), and names zero near-miss negatives even though
19 sibling skills compete for the word "optimize". The **state machine has three demonstrated
misroutes and one terminal sink**: a run dir without `state.json` is invisible, so a live run is
reported as `finalized`; `ready` becomes unreachable forever after the first run dir appears; and
`running` routes to `report`, which continues nothing — `cap-evolve run --resume`
(`core/cap_evolve/cli.py:553-557`) and `--reuse-baseline` (`:558`) are never mentioned anywhere in
the skill. Two factual claims about `cap-evolve run` are wrong. Separately, the 60-line body is
~35 lines of ceremony that duplicates `orchestrate`, the SessionStart hook, and `run.py`'s own
output. Recommendation: **keep the two skills separate**, shrink this one to ~22 lines, delete
`references/routing.md`, and fix the state machine.

### 1. The description spends its budget on the body's content, not on triggering
`SKILL.md:3`. Of ~85 words, the clauses "Explains the two ways to run (standalone
`/cap-evolve:<phase>` chain vs the fully-automatic `cap-evolve run`) and restates the
non-negotiable honesty rules" and "Decides whether a project already exists and routes
accordingly: to intake (Phase 1)…" describe *how the body works*. skill-creator:89-90 says the
metadata level is "~100 words" and is the only thing that decides firing; skill-creator:67 says
all "when to use" info goes in the description. Internals in the description crowd out the
keywords that decide the firing.

**Fix:** delete both clauses; spend the words on trigger phrases and negative boundaries. See
*Proposed description*.

### 2. The trigger examples are placeholders no user types
`SKILL.md:3` offers `"optimize <X>"`, `"make <X> score higher on <benchmark>"`. skill-creator:348-352
is explicit that eval/trigger phrasing must be concrete, not abstract (`Bad: "Create a chart"` /
`Good:` a long, specific, realistic sentence). `<X>` matches no real utterance. Missing high-value
keywords a real user does type: **eval, evals, pass rate, benchmark score, accuracy, task suite,
regression, `.capevolve`, capevolve.yaml, resume the run**. The `.capevolve/` project marker is the
single strongest signal for the returning-user case and appears only buried mid-sentence as a path.

**Fix:** replace placeholders with real utterances ("raise the pass rate on these tasks", "my agent
keeps failing these tasks") and promote `.capevolve/`-in-the-tree to its own trigger clause.

### 3. No near-miss negatives, in the one skill that most needs them
`SKILL.md:3` names nothing it should *not* fire on. skill-creator:356-358 calls near-misses "the
most valuable" cases and warns against negatives that are "obviously irrelevant". Real near-misses
here: (a) "optimize this SQL query / make this function faster" — performance work, not eval-driven
capability optimization; (b) "improve the wording of this prompt" with no eval to score against;
(c) "write me a skill for X" (that is skill-creator); (d) the user naming a phase or algorithm
outright — "run hill-climb", "score this candidate on val" should hit `hill-climb` / `evaluate`
directly, not bounce through a router.

**Fix:** add one negative clause covering performance-optimization, no-eval prompt tweaks, and
direct phase/algorithm requests.

### 4. All-caps `OPTIMIZE` is emphasis noise in a description
`SKILL.md:3`, and the same word in the hook text at
`plugins/cap-evolve/hooks/inject_router.py:34`. The known failure mode is that imperative caps in a
*description* over-trigger current models. The router should be pushy via coverage ("even when they
never say 'optimize'"), which is the pattern skill-creator:67 endorses — not via shouting.

**Fix:** lowercase it in both places.

### 5. A live run is invisible, and reported as finalized — worst misroute in the skill
`scripts/run.py:36` filters candidate runs on `(r / "state.json").exists()`. Reproduced against the
real function:

| on-disk state | router says | should say |
|---|---|---|
| A: `run_.../` exists, no `state.json` yet | `scaffolded` → implement-and-check | a run is starting/half-created |
| C: older sealed run + newer live run with no `state.json` | `finalized` → report | `running` (the newer run) |

Case A tells the agent to go implement an adapter that is already implemented. Case C tells the
agent the run is **done and sealed while an optimization is mid-flight** — it will report a stale
headline number. `state.json` is written by the run, so any window before that write (and any
crash inside it) produces this.

**Fix:** in `_latest_run`, glob `run_*` directories unconditionally
(`sorted(r for r in base.glob("run_*") if r.is_dir())`); treat a run dir with neither `state.json`
nor `splits.json` as its own state (`run_started`) whose `next` is `cap-evolve run --resume`.

### 6. `ready` is unreachable after the first run, so a finished project is a terminal sink
`scripts/run.py:62-76` returns from the `run is not None` branch **before** the `green`/`ready`
branch at `:82-86`. Reproduced: a project with one sealed run resolves to
`{"state": "finalized", "next": "/cap-evolve:report"}` forever. A user who says "that worked, now
run it again with a bigger budget" gets pointed at a report. This is the natural second request and
the router has no answer for it — even though `cap-evolve run --reuse-baseline <prior run dir>`
(`core/cap_evolve/cli.py:558-560`) is exactly the primitive for it.

**Fix:** in the `finalized` branch, set `next` to a new run
(`cap-evolve run --spec … --reuse-baseline <run>`) and mention report as the read-only inspect
option. `ready` should also be reachable when every existing run is sealed.

### 7. `running` routes to `report`, which continues nothing; `--resume` appears nowhere
`scripts/run.py:74-76` sets `next: "/cap-evolve:report"` while its own `reason` string says
"continue the algorithm loop" — the two disagree, and neither the SKILL.md nor
`references/routing.md` ever mentions `cap-evolve run --resume`, documented at
`core/cap_evolve/cli.py:553-557` as the supported way to pick an interrupted run up at iteration
N+1. An interrupted run is the single most likely reason a user re-enters this router mid-project,
and the router's answer is "go read a report".

**Fix:** `running` → `next: "cap-evolve run --resume"`, with report as the inspect option. Add one
line to the body: interrupted run → `--resume`; sealed run, new attempt → `--reuse-baseline`.

### 8. Two factual errors about `cap-evolve run` (also in the hook text)
`SKILL.md:41-46` says the automatic path "Sequences every phase, enforces the hard gate before
spending budget, decides when to stop (budget/stall), and ends with the sealed-test number."

- `cap-evolve run` does **not** run intake. `core/cap_evolve/cli.py:722-724`: *"Hard gate:
  cap-evolve check must pass before any budget is spent (intake is the user's job before `run`;
  here we enforce the check half of implement-and-check)."* The router lists intake as part of what
  the automatic path sequences, which will send a fresh user straight into a failing `run`.
- Under `orchestration_mode: agent` it does **not** end with a sealed-test number.
  `core/cap_evolve/cli.py:696-698` sets the sequence to
  `[intake, implement-and-check, baseline, "<handoff: agent drives the loop, then cap-evolve finalize>"]`
  — the loop is handed back to the agent (see `skills/orchestrate/orchestrate/SKILL.md:44-47`).
  The router never mentions agent mode, and `SKILL.md:38-39` compounds the confusion by writing
  "`agent-optimize` in agent mode", conflating an algorithm name with `orchestration_mode`.

**Fix:** state that `run` presumes intake already happened, and that there are **three** paths
(phase chain / deterministic `run` / `orchestration_mode: agent` handoff) — one line each, pointing
at `orchestrate` for detail. Fix the same two claims in `inject_router.py:34-44`.

### 9. `needs: [project]` is backwards for a front door
`SKILL.md:9` and `meta.yaml:7-8` declare `needs: [project]` / `provides: [report]`. The router's
*primary* state is `fresh` — no project exists (`run.py:58-60`) — and it never writes a report.
`references/routing.md:31-32` defends this as "pass manifest DAG validation", i.e. the tokens exist
to satisfy a validator, not to describe reality. Verified: setting both to `[]` and rebuilding
leaves `skills/_registry/manifest.json` with `errors: []`.

**Fix:** `needs: []`, `provides: []` in both `SKILL.md` frontmatter and `meta.yaml`.

### 10. ~35 of 60 body lines are ceremony
- `SKILL.md:13-17` — "This is the front door… gets out of the way" restates the description. → 1 line.
- `SKILL.md:24-33` — the state table is the **third** copy of the same information: `run.py:8-15`
  (docstring), `references/routing.md:12-18` (table), and here. The script already prints `state`,
  `next`, and a human-readable `reason` per state, so the body does not need to enumerate them —
  it needs to say "run it and follow `next`" plus the two rules the JSON cannot convey (never skip
  intake on a fresh request; never fabricate a NEEDED input).
- `SKILL.md:51-56` — the four honesty rules. A read-only router cannot violate them, they are
  enforced by core and hooks (as the section itself says), and they are restated in 6 other skill
  bodies. This is exactly the per-trigger duplication tax the brief warns about. → delete, or one
  pointer line.
- `SKILL.md:35-49` — "Two ways to run" duplicates `orchestrate/SKILL.md:57-64` + `RUN.md`. → 3
  lines pointing there (plus the agent-mode path from finding 8).

**Fix:** target ~22 lines: one-line framing, the resolver command, the two non-obvious rules, three
one-line run paths pointing at `orchestrate`/`RUN.md`, resume/reuse-baseline note. Delete the
`## References` section with `routing.md`.

### 11. Delete `references/routing.md`
All three of its sections fail to pull their weight. "The state machine" (`:8-22`) is the third
copy (finding 10). "Relationship to `orchestrate`" (`:24-32`) exists only to explain why two skills
are not the same skill — that need disappears once the router is thin, and its DAG claim is the
fiction from finding 9. "Session-start injection" (`:34-51`) reproduces the JSON envelope already
documented verbatim in `plugins/cap-evolve/hooks/inject_router.py:11-19`; the hook's contract is
not something an agent using the router needs in context.

**Fix:** delete the file and the `## References` section. If the state table is wanted for humans,
it belongs in the `run.py` docstring, which already has it.

### 12. Minor: nonstandard `arguments:` field
`SKILL.md:6` sets `arguments: "$ARGUMENTS"`; 1 of 20 skills does this, while 20 of 20 use
`argument-hint`. `$ARGUMENTS` is interpolated for commands, not skill frontmatter, so this line is
inert.

**Fix:** delete `SKILL.md:6`.

### 13. Minor (repo-wide, worst here): the first command in the session uses a relative path
`SKILL.md:22` says `python scripts/run.py "$ARGUMENTS" --base .capevolve`. The agent's cwd is the
user's project, not the skill dir, so this fails as written — verified. 18 of 20 skills share the
convention, so this is not this skill's invention, but it is the *first* command a cap-evolve
session runs, which makes it the worst place for it. `skills/algorithms/agent-optimize/SKILL.md:31`
already shows the fix in-repo: `S="${CAPEVOLVE_SKILLS_DIR:?…}"`.

**Fix:** `python "${CAPEVOLVE_SKILLS_DIR:?}"/orchestrate/using-cap-evolve/scripts/run.py …`.

### What is already right (no work needed)
`resolve_state` is pure and non-mutating, and the `green is None` fallback at `run.py:91-94`
returns `scaffolded` rather than a `ready` it cannot prove — the honest choice, and
`routing.md:20-22` documents why. `check.py` is a real behavioral check (asserts fresh→intake and
that a scaffolded base is not `fresh`), not a stub, and it passes. `allowed-tools: Read, Bash` is
correctly minimal for a read-only router. The skill is generic: no benchmark, vendor, algorithm, or
capability is hardcoded anywhere.

## Proposed description

**Before** (`SKILL.md:3`, ~85 words):
> Entry-point router for cap-evolve. Use the moment a user asks to OPTIMIZE an agent capability
> against an eval — "optimize \<X\>", "make \<X\> score higher on \<benchmark\>", "improve this
> skill/tool/prompt". Decides whether a project already exists and routes accordingly: to intake
> (Phase 1) for a fresh request, or straight into the phase chain / `cap-evolve run` when
> `.capevolve/project/` is already scaffolded. Explains the two ways to run (standalone
> `/cap-evolve:<phase>` chain vs the fully-automatic `cap-evolve run`) and restates the
> non-negotiable honesty rules. Does no optimization itself.

**After** (~110 words):
> Front door for cap-evolve: routes an optimization request to the right pipeline phase. Use
> whenever someone wants an agent, skill, system prompt, tool surface, or MCP toolset to score
> higher on an eval, benchmark, or task suite — "optimize my skill", "raise the pass rate on these
> tasks" — even when they never say "optimize", and whenever a `.capevolve/` project or unfinished
> run is in the tree. Hands off to intake, the check gate, a resumed run, or the report; optimizes
> nothing itself. Not for making code faster or rewording one prompt with no eval; if the user
> names a phase or algorithm outright (baseline, gate, hill-climb, gepa), use that skill.

Change by change:
1. **"Entry-point router" → "Front door … routes an optimization request to the right pipeline
   phase."** Same claim, but says what a *user* gets rather than naming an architectural role.
2. **Dropped the "Decides whether a project already exists and routes accordingly: to intake
   (Phase 1) …" clause.** Internals; the body and `run.py`'s output own this. Frees ~30 words.
3. **Dropped "Explains the two ways to run … and restates the non-negotiable honesty rules."**
   Pure body description, and after findings 8/10 both statements change anyway.
4. **Placeholders → real utterances.** `<X>`/`<benchmark>` match nothing a user types
   (skill-creator:348-352).
5. **Named the four capability surfaces (agent, skill, system prompt, tool surface, MCP toolset)
   and the three eval nouns (eval, benchmark, task suite).** These are the actual retrieval
   keywords; the old text had "skill/tool/prompt" once, compressed into a slash-list.
6. **Added `.capevolve/` project or unfinished run in the tree as its own trigger.** The
   returning-user case, which currently depends on the user re-saying "optimize".
7. **Added "even when they never say 'optimize'".** The under-trigger nudge in the form
   skill-creator:67 endorses, replacing the caps.
8. **Lowercased OPTIMIZE.** Caps in a description over-trigger current models.
9. **Added the negative clause** (performance work, no-eval prompt rewording, direct
   phase/algorithm requests). skill-creator:356-358: near-misses are the highest-value boundary,
   and 19 siblings compete for "optimize".
10. **Kept "optimizes nothing itself"** — it is the one internal fact that changes routing
    behavior (the router must hand off rather than start editing).

Honest trade-off: this is ~25 words longer. The added words all buy trigger coverage or a negative
boundary; the deleted words described the body. If a hard ~85-word cap is wanted, drop the
parenthesized algorithm list in the last clause.

## Uncovered routing states
1. **Run dir created, `state.json` not yet written** — invisible to `_latest_run`
   (`run.py:36`); resolves to `scaffolded` → "implement the adapter". *(finding 5, case A)*
2. **Newer live run with no `state.json` alongside an older sealed run** — resolves to
   `finalized` → report, i.e. claims a run is sealed while it is mid-flight. *(finding 5, case C)*
3. **Sealed run, user wants another run** — permanently `finalized` → report; `ready` is
   unreachable once any run dir exists (`run.py:62-76` precedes `:82-86`). No path to
   `--reuse-baseline`. *(finding 6)*
4. **Interrupted / stale run** — indistinguishable from a healthy in-flight run; both get
   `next: report`, never `cap-evolve run --resume`. *(finding 7)*
5. **`orchestration_mode: agent`, loop handed back to the agent** — the router has no state for
   "cap-evolve did baseline and handed you the loop", and its body does not mention agent mode at
   all. *(finding 8)*
6. **Project dir with an implemented adapter but no `capevolve.yaml`** — resolves to `fresh` →
   intake (`run.py:78-80`), which will re-scaffold over existing work. At minimum the `reason`
   should warn that intake writes into a non-empty project dir.
7. **`run_*` dir whose `splits.json` is unreadable/corrupt** — `run.py:66-69` swallows the
   exception to `{}` and reports `running`, silently treating a corrupt split as an unsealed one.
   Should surface the corruption; the split file is a honesty-critical artifact.

## Acceptance criteria
- [ ] `SKILL.md:3` replaced with the proposed description (or an equivalent that: is third person,
      contains ≥2 realistic non-placeholder trigger phrases, names `.capevolve/` as a trigger,
      contains an explicit negative clause, and contains no all-caps imperatives).
- [ ] `SKILL.md` body ≤ 25 lines; the state enumeration, the honesty-rules section, and the
      `## References` section are gone; three run paths each get one line pointing at
      `orchestrate` / `RUN.md`.
- [ ] `references/routing.md` deleted.
- [ ] `SKILL.md:6` (`arguments:`) deleted; `needs: []` / `provides: []` in both `SKILL.md` and
      `meta.yaml`; `python skills/_registry/build_manifest.py` still reports `errors: []`.
- [ ] `_latest_run` no longer requires `state.json`; a run dir with no `state.json` resolves to a
      non-`scaffolded` state, and a newer live run is never masked by an older sealed one.
      Regression test in `scripts/check.py` for both cases (A and C in finding 5).
- [ ] `finalized` → `next` names a new run with `--reuse-baseline`; `running` → `next` names
      `cap-evolve run --resume`. `check.py` asserts both strings.
- [ ] The claims "sequences intake" and "ends with the sealed-test number" are corrected in
      `SKILL.md` and in `plugins/cap-evolve/hooks/inject_router.py:34-44`; agent mode is named as a
      third path.
- [ ] `SKILL.md:22` invokes `run.py` via `${CAPEVOLVE_SKILLS_DIR}` so it works from a project cwd.
- [ ] `python scripts/check.py` in the skill dir still exits 0.

## Out of scope
- Merging `using-cap-evolve` into `orchestrate`. **Keep them separate.** The case *for* merging is
  real: identical `component`/`needs`/`provides`, both own a `scripts/run.py`, the router's
  "two ways to run" duplicates `orchestrate:57-64`, and `routing.md:24-29` has to exist purely to
  explain why they differ — a documentation smell. But the two fire on different surfaces (a cold,
  vague "optimize X" vs. "drive the whole run autonomously"), and merging forces one description to
  carry both, which is precisely how routers stop firing. Progressive disclosure decides it: the
  router must be a ~20-line body that is cheap to load on every session start, while
  `orchestrate`'s 85-line driver body should load only when a run is actually being driven. Merged,
  every "optimize X" pays the driver's cost. Keep separate, shrink the router, and delete
  `routing.md` — after which the "why are these different" section is no longer needed.
- Repo-wide fix of the relative `python scripts/run.py` convention across the other 17 skills
  (finding 13 fixes it only here, where it runs first).
- Changing the SessionStart hook's *trigger* heuristic (`_looks_like_capevolve`); only its two
  factually wrong sentences are in scope.
- Any change to `core/cap_evolve/cli.py`. `--resume` / `--reuse-baseline` already exist; the router
  simply never mentions them.

OsherElhadad:
Closes #339.

The description is the only thing that decides whether a skill fires, and this is the
entry-point router — a router that does not fire makes the other 19 skills unreachable.
Both of its load-bearing jobs (the description, the state machine) were weak.

### `wc -l`

| file | before | after |
|---|---|---|
| `SKILL.md` | 60 | **37** (frontmatter 10 + 26 body lines) |
| `references/routing.md` | 51 | **deleted** |
| `scripts/run.py` | 116 | 145 |
| `scripts/check.py` | 52 | 88 |

Lines removed under the ownership contract (#340): **16** — the four honesty rules
(`SKILL.md:51-56`, owners `gate`/`finalize`/`report`) and the 10-line state enumeration
(`SKILL.md:24-33`), which was the third copy of the same table (`run.py`'s docstring and
`routing.md:12-18` being the other two) and duplicated the JSON the script already prints.
The ask-don't-fabricate rule survives as a 1-line pointer to its owner (`intake`), the form
the contract explicitly blesses. No agent-mode loop text and no failure-taxonomy text was
present in this skill, so nothing was owed to `orchestrate/orchestrate` or `diagnose`.

### Authoring lint (PR #345's script, run against this worktree)

```
$ git show origin/feat/issue-107-skill-authoring-lint:skills/_registry/lint_skills.py > /tmp/lint_skills_339.py

# before
skill authoring lint — 20 skill package(s)
  ERROR  orchestrate/using-cap-evolve: description must not contain XML tags
  advise orchestrate/using-cap-evolve: description should say WHEN to use the skill …
FAIL — 6 authoring violation(s) not in the baseline (0 known, 11 skill(s) with advisories)

# after
skill authoring lint — 20 skill package(s)
FAIL — 5 authoring violation(s) not in the baseline (0 known, 10 skill(s) with advisories)
```

`using-cap-evolve` no longer appears in either list. **The XML violation is cleared** — it
was `<X>`, `<benchmark>` and `<phase>`, all matched by `abstract.py:34`'s
`XML_TAG_RE = </?[a-zA-Z][^>]*>`. The remaining 5 errors belong to other skills (other
agents' PRs). The residual `FAIL` is the pre-existing baseline state of the repo, not this
change.

### Manifest

`python skills/_registry/build_manifest.py` → 20 skills, `errors: []` with
`needs: []` / `provides: []`.

### Tests

```
789 passed, 12 skipped in 116.46s
```

Exactly clean main (see #349 for why the `PYTHONPATH` pin is required in a worktree).

### Toy example

```
{"run_dir": ".capevolve/run_demo", "best_id": "cand_0001", "baseline_val": 0.0,
 "test_reward": 1.0, "test_baseline_reward": 0.0, "test_delta": 1.0, "iterations": 3}
```

Gate-accepted, test 1.0 from a 0.0 baseline.

### Skill check

`python scripts/check.py` → `{"ok": true, "problems": []}`, now covering the two
reproduced misroutes as regressions.

---

## Description: before → after

**Before** (85 words, contains XML):

> Entry-point router for cap-evolve. Use the moment a user asks to OPTIMIZE an agent
> capability against an eval — "optimize \<X\>", "make \<X\> score higher on \<benchmark\>",
> "improve this skill/tool/prompt". Decides whether a project already exists and routes
> accordingly: to intake (Phase 1) for a fresh request, or straight into the phase chain /
> `cap-evolve run` when `.capevolve/project/` is already scaffolded. Explains the two ways
> to run (standalone `/cap-evolve:<phase>` chain vs the fully-automatic `cap-evolve run`)
> and restates the non-negotiable honesty rules. Does no optimization itself.

**After** (115 words, 872 chars — under the 1024 cap):

> Front door for cap-evolve: routes an optimization request to the right pipeline phase.
> Use when someone wants an agent, skill, system prompt, tool surface, or MCP toolset to
> score higher on an eval, benchmark, or task suite — "optimize my skill", "raise the pass
> rate on these tasks", "my agent keeps failing these cases", "get this prompt's accuracy up
> on my evals" — even when they never say "optimize", and whenever a .capevolve/ project or
> an unfinished run is in the tree. Routes to intake, the check gate, a resumed run, or the
> report; optimizes nothing itself. Not for making code or a query faster, and not for
> rewording one prompt with no eval to score it against. When the user names a phase or
> algorithm outright (baseline, gate, hill-climb, gepa), use that skill directly.

Change by change:

1. **Removed `<X>` / `<benchmark>` / `<phase>`.** Hard validity failure (`abstract.py:186-196`),
   and skill-creator:348-352 is explicit that trigger phrasing must be concrete, not abstract —
   `<X>` matches no real utterance either way, so the fix and the improvement are the same edit.
2. **Dropped "Decides whether a project already exists and routes accordingly…".** Internals.
   `run.py` prints exactly this, per state, with a human-readable `reason`. ~30 words freed.
3. **Dropped "Explains the two ways to run … and restates the non-negotiable honesty rules."**
   Pure body description — and both statements are now false anyway: there are three run paths,
   and the honesty restatement is gone as a duplicate of `gate`/`finalize`/`report`.
4. **Named the capability surfaces and eval nouns** — agent, skill, system prompt, tool surface,
   MCP toolset × eval, benchmark, task suite, pass rate, accuracy. These are the actual retrieval
   keywords; the old text compressed them into one slash-list, "skill/tool/prompt".
5. **Four real utterances** in place of the two placeholders, including "my agent keeps failing
   these cases", which never says "optimize" at all.
6. **`.capevolve/` project or unfinished run in the tree is now its own trigger clause.** The
   returning-user case previously depended on the user re-saying "optimize" from scratch.
7. **"even when they never say 'optimize'"** — the under-trigger nudge in the form
   skill-creator:67 endorses, replacing the caps.
8. **Lowercased `OPTIMIZE`** (also in the hook). All-caps imperatives in a *description*
   over-trigger current models; `abstract.py:204-206` warns on it directly.
9. **Added the negative clause.** skill-creator:356-358 calls near-misses the highest-value
   boundary, and 19 siblings compete for "optimize". Three near-miss families: performance work,
   no-eval prompt rewording, and a directly-named phase/algorithm.
10. **Kept "optimizes nothing itself"** — the one internal fact that changes behaviour: the
    router must hand off, not start editing.

Honest trade-off: +30 words. All of them buy trigger coverage or a boundary; every deleted
word described the body.

## Trigger sanity check

`claude -p` **was not run.** A skill enters `available_skills` only when installed — as the
`cap-evolve` plugin or under `~/.claude/skills`. This worktree's edit is neither, so `claude -p`
would have measured the *shipped* description, not this one, and installing it would have
mutated shared state that ~60 other agents in this session read. I reasoned about separation
instead, below.

### Should trigger (10)

1. "our support triage skill is at like 62% on the 180-case suite in `evals/triage.jsonl` and product wants 80 before launch — can you actually push that number up?"
2. "my agent keeps picking the wrong tool on the refund flow, roughly 1 in 4 of the airline tasks die on it. fix the tool descriptions so it stops"
3. "raise the pass rate on `tests/agent_cases/*.yaml`. currently 0.41. i don't care how, just make it higher and prove it wasn't noise"
4. "there's a `.capevolve/` dir in this repo from last week and i think the run died halfway through. what do i do with it"
5. "i wrote a SKILL.md for invoice extraction and it works maybe half the time. i have 40 labeled invoices in `fixtures/`. make the skill better against those"
6. "can you tune `prompts/agent.md` so accuracy on our regression suite goes back up? we went 0.78 → 0.71 last release"
7. "we need the MCP toolset for the billing server to stop misfilling arguments — there's a scored harness in `bench/`"
8. "improve the accuracy of my classifier prompt on the eval set in `data/eval.jsonl`, target 90%"
9. "score-wise this thing is stuck at chance. optimize the agent against the benchmark and show me the honest number"
10. "the coding agent's skill package for db migrations is too vague and it wanders off. we have 25 graded tasks — get it up"

### Should NOT trigger — near-misses (10)

1. "optimize this SQL query, it's doing a full table scan on the 40M-row `orders` table"
2. "make this python function faster, it's O(n²) and we call it on every request"
3. "reword the system prompt in `prompts/agent.md` so it sounds less robotic and stops apologising"
4. "write me a skill for converting our Figma tokens into a tailwind config"
5. "run hill-climb on the project in `.capevolve`, budget 20 iterations, stall 5"
6. "score candidate `cand_0007` on val and tell me the standard error"
7. "our eval suite takes 40 minutes end to end — can you parallelise the harness?"
8. "pass rate dropped 0.80 → 0.60 after yesterday's merge, find which commit did it"
9. "add 50 more test cases to `evals/triage.jsonl` covering refund edge cases"
10. "which benchmark should we measure our agent on, swe-bench or tau-bench?"

### Does the description separate them?

Every positive carries three things at once: a **capability surface** (skill / prompt / tool
surface / MCP toolset / agent), a **scored artifact** (eval, benchmark, task suite, pass rate,
accuracy), and a **directional want** (higher, up, better). The description names all three
lexical families explicitly. #4 carries none of them and is caught only by the `.capevolve/`
clause — which is precisely why that clause was added. #2, #9 and #10 never use the word
"optimize" in the sense the old description keyed on, which is what "even when they never say
'optimize'" is for.

The negatives split into three groups, matched by the three parts of the negative clause:

- **1, 2, 7** have the directional want but the object is code or runtime latency.
  "Not for making code or a query faster" is a near-literal keyword match on "optimize this SQL
  query" and "make this function faster". #7 is the subtler one — it shares the word *eval* — but
  it wants the harness faster, not the capability better, and the description ties every eval
  noun to *score higher*, never to *run faster*.
- **3** has the capability surface and no scored artifact. "not for rewording one prompt with no
  eval to score it against" makes the presence of an eval the discriminator, which is the actual
  structural difference between #3 and positive #6 — the same file, the same verb, and only the
  eval separates them. This is the pair I'd most want a real eval on.
- **5, 6** are covered by the last sentence naming baseline, gate, hill-climb, gepa outright.

**Two boundaries I do not claim are clean, honestly:**

- **#4 ("write me a skill for…")** — creation, not improvement against a score. The description
  keeps every trigger phrase comparative ("score higher", "raise", "up") and never pairs
  "improve" with an object that has no score, so the lexical signal is there, but it is
  inferential rather than stated. I chose not to spend another clause on it: the failure mode is
  the router firing, printing `fresh`, and offering intake — visible and cheap to correct — where
  an over-tightened description that stops firing is not.
- **#8 (bisect a regression)** — shares *pass rate* and *dropped*, and wants a cause, not a
  higher score. Genuinely ambiguous; a human would ask. Same reasoning: a false fire here is
  recoverable in one turn.

Both are listed so a later description-optimizer run has them as known-hard cases rather than
rediscovering them.

## State machine: the misroutes, verified against core

Verified against `core/cap_evolve/cli.py`'s `run` subcommand and `core/cap_evolve/rundir.py`
before writing any rule — none of these states is invented.

| on-disk state | before | after | why the state is real |
|---|---|---|---|
| `run_*` exists, no `state.json` | `scaffolded` → implement the adapter | `run_started` → `cap-evolve run --resume --run-ts <ts>` | `rundir.py:186-200` documents exactly this as "a torn/partial create" in `RunDir.create`'s `exist_ok` contract |
| newer live run + older sealed run | `finalized` → report | `running` → `--resume` | `_latest_run` filtered on `state.json` (`run.py:36`), so the newer dir was skipped entirely |
| sealed run, user wants another attempt | `finalized` → report, forever | `finalized` → `cap-evolve run --spec … --reuse-baseline <run>` | `cli.py:558-560` — `--reuse-baseline` is exactly this primitive; nothing in the skill ever mentioned it |
| interrupted run | `running` → report (which continues nothing) | `running` → `cap-evolve run --resume` | `cli.py:553-557` documents `--resume` as picking the loop up at iteration N+1 |
| `splits.json` unreadable | swallowed to `{}`, reported as unsealed | surfaced in `reason`, seal declared unconfirmable | the split is an honesty-critical artifact; silently reading a corrupt one as "not sealed" is the wrong direction to fail in |
| project dir, no `capevolve.yaml` | `fresh` → intake, silently | `fresh` → intake, `reason` warns intake scaffolds into a non-empty dir | `run.py:78-80` |

`--run-ts` is on the `run_started` and `running` next-commands deliberately: `RunDir.latest`
(`rundir.py:221-226`) also filters on `state.json`, so a bare `--resume` would raise
`FileNotFoundError` on exactly the dir we are routing to (`cli.py:747-753`).

Cases A and C are asserted in `scripts/check.py`, along with the `--resume` / `--reuse-baseline`
strings, so a regression fails the skill's own check.

## Two factual errors corrected

Both in `SKILL.md:41-46` and in the SessionStart hook text (`inject_router.py:34-44`):

- **`cap-evolve run` does not run intake.** `cli.py:719-720`: *"Hard gate: cap-evolve check must
  pass before any budget is spent (intake is the user's job before `run`; here we enforce the
  check half of implement-and-check)."* The `sequence` list at `cli.py:697-700` is display-only —
  grepping `intake` through `cli.py` shows no invocation, only a best-effort read of
  `intake.json` for cost accounting. The old text sent a fresh user straight into a failing `run`.
- **Under `orchestration_mode: agent` it does not end with a sealed-test number.** `cli.py:696-698`
  sets the sequence to `[intake, implement-and-check, baseline, "<handoff: agent drives the loop,
  then cap-evolve finalize>"]`. Agent mode is now named as the third run path, and
  `SKILL.md:38-39`'s "`agent-optimize` in agent mode" — which conflated an algorithm name with
  `orchestration_mode` — is gone.

## Keep separate, not merged

Kept separate. The case for merging is real (identical `component`, both own a `scripts/run.py`,
the "two ways to run" section duplicated `orchestrate:57-64`, and `routing.md:24-29` existed
purely to explain why they differ — a documentation smell). Progressive disclosure decides it:
the router must be a ~26-line body that is cheap to load at every session start, while
`orchestrate`'s 85-line driver body should load only when a run is actually being driven. Merged,
every "optimize X" pays the driver's cost, and one description would have to fire on both a cold
vague request and "drive the whole run" — which is how routers stop firing.

The boundary is now stated in one line in this skill: *"this router picks the door,
`orchestrate` drives the run."* Once the router is thin, `routing.md`'s "why are these different"
section has nothing left to explain, which is why it could be deleted.

## What `orchestrate/orchestrate` should drop

Per the ownership contract I did not touch another agent's skill dir. For the owning agent:

- `orchestrate/SKILL.md:21-25` — the `## Inputs / outputs (manifest tokens)` section restating
  its own frontmatter as prose. One of the nine instances the contract flags for deletion.
- `orchestrate/SKILL.md:34` — "the loop named in `capevolve.yaml` (default `all-at-once`)".
  `all-at-once` no longer exists as an algorithm; it is now `hill-climb --focus all`.
- The matching boundary line, so the pair reads symmetrically: *"`using-cap-evolve` picks the
  door; this skill drives the run."*

## Other notes

- `arguments: "$ARGUMENTS"` deleted from `SKILL.md:6` — inert (`$ARGUMENTS` is interpolated for
  commands, not skill frontmatter), and 19 of 20 skills use `argument-hint` alone.
- `needs: []` / `provides: []` in both `SKILL.md` and `meta.yaml`. A front door's primary state
  is "no project exists" and it never writes a report; `routing.md:31-32` defended the old tokens
  as existing to satisfy a validator, and the validator is satisfied without them (`errors: []`).
- `sources: [evo, superpowers]` left alone. The contract says do not add more consumers for these
  fields; removing them is a separate, repo-wide call.
- The resolver is now invoked as
  `S=<this skill's own directory>; python "$S/scripts/run.py" --base .capevolve`, where `S` is the
  absolute path the agent loaded this SKILL.md from. `scripts/_bootstrap.py` already anchors its
  core lookup at `Path(__file__).resolve()`, so the script runs from any cwd with no env var and
  no assumption about where the user is standing. The relative `python scripts/run.py` that 17
  skills share failed from a user's cwd — bad anywhere, worst as the first command of a session —
  and fixing it for those 17 is out of scope here, but this shows the pattern. (An earlier revision
  of this branch used `${CAPEVOLVE_SKILLS_DIR:?…}`; nothing sets that variable for a plugin
  install, so bash aborted. Replaced in 6aabef5c — see the review comment below for the
  outside-the-repo verification.)


OsherElhadad:
Fixed in 6aabef5c — took **option 1**, resolving the script path from the skill file's own location so the env var disappears entirely.

You were right that the combination is the bug, not either half. `${CAPEVOLVE_SKILLS_DIR:?…}` aborts bash with exit 127 when unset, and the same commit had deleted both fallbacks. The router's one actionable command was unusable for a plugin install.

The body's command is now:

```bash
S=<this skill's own directory>; python "$S/scripts/run.py" --base .capevolve
```

with the two preceding lines saying `S` is the absolute path the agent loaded this SKILL.md from — the one location always known at that point, requiring no env var and no assumption about cwd. `scripts/_bootstrap.py` already anchors its core lookup at `Path(__file__).resolve()`, so the script itself needs nothing else to run from an arbitrary directory.

### Verified from outside the repo, `CAPEVOLVE_SKILLS_DIR` unset

```
$ python -c "import os; print('CAPEVOLVE_SKILLS_DIR:', os.environ.get('CAPEVOLVE_SKILLS_DIR','UNSET'))"
CAPEVOLVE_SKILLS_DIR: UNSET

$ cd /tmp/router339            # a bare project: .capevolve/project/capevolve.yaml + one run dir
$ python /…/skills/orchestrate/using-cap-evolve/scripts/run.py --base .capevolve
{
  "state": "run_started",
  "next": "cap-evolve run --resume --run-ts 20260101_000000",
  "run": ".capevolve/run_20260101_000000",
  "reason": "a run dir exists but has no state.json — a torn/partial create. Resume it by ts …"
  …
}
```

Correct state, correct next command, from a cwd with no relationship to the checkout.

The previous form in the same shell:

```
$ bash -c 'python "${CAPEVOLVE_SKILLS_DIR:?point at the cap-evolve skills dir}"/…/run.py --base .capevolve'
bash: CAPEVOLVE_SKILLS_DIR: point at the cap-evolve skills dir
exit=127
```

### Re-run

```
789 passed, 12 skipped in 111.05s
```

`python scripts/check.py` → `{"ok": true, "problems": []}`. Authoring lint unchanged: `using-cap-evolve` still absent from both the error and advisory lists.

### The shrink stands

60 → 37 lines (+2 for the invocation note). No state table came back, `references/routing.md` stays deleted, and the description is untouched. The router is still the only skill that will work from an arbitrary cwd — noted in the PR body that the other 17 relative-path invocations are out of scope here, but this now demonstrates the pattern for whoever picks that up.

PR #360, opened 2026-08-22
feat(skill-package): optimize the whole package — references and bundled scripts, not just SKILL.md (#322)

### LINKED ISSUE #322: skill-package capability: the optimizer can only edit SKILL.md — make the whole package (references + scripts) optimizable
# skill-package capability: the optimizer cannot optimize a package, and nothing it writes is validated

## Summary

`skill-package` claims in its own description to "optimize an Agent Skill package itself — its
SKILL.md (frontmatter + body), references, and bundled scripts" and to "enforce the skill-creator
authoring rules … so edits stay valid skills". Both halves are false as shipped. `materialize()`
returns only `SKILL.md` + `references/*.md`, so `scripts/` and `assets/` are not components of the
artifact it declares itself to own; and `validate()` — the entire enforcement story — is **never
called by the optimization loop** (no reference in `harness.py`, `gepa.py`, `skillopt.py`, or
`adapter.py`), so every rule in it is enforced on nothing during a real run. The determinism goal
is further out of reach than the scope gap suggests: `validate()` has zero coverage of script
content, so an optimizer that writes a bundled script can ship a syntax error and still be gated
as valid, and the repo's own `examples/skillsbench/optimizer/INSTRUCTIONS.md` asks the optimizer to
"re-validate every touched skill" *in prose* — the exact model-can-skip-it failure this capability
exists to eliminate. On top of that, roughly 45 of SKILL.md's 122 lines are third copies of rules
stated in `references/concepts.md`, in `harness._algorithm_brief`, or in the three sibling
capability skills, and four separate validate() rules check the wrong thing (probes below).

## Findings

**F1. Package-scope gap: `materialize()` exposes SKILL.md + references only — never scripts or assets.**

`scripts/abstract.py:73-83` is the whole component view:

```python
def _materialize_one(skill_dir: Path, prefix: str = "") -> dict:
    """Flatten ONE skill package's SKILL.md + references into named components."""
    parts = {}
    skill_md = skill_dir / "SKILL.md"
    if skill_md.exists():
        parts[f"{prefix}SKILL.md"] = skill_md.read_text(encoding="utf-8")
    refs = skill_dir / "references"
    if refs.is_dir():
        for f in sorted(refs.glob("*.md")):
            parts[f"{prefix}references/{f.name}"] = f.read_text(encoding="utf-8")
    return parts
```

No `scripts/` branch, no `assets/` branch. Probe (temp package with `scripts/x.py` and
`assets/tpl.html`): `materialize()` → `['SKILL.md']` only.

This directly contradicts the module's own docstring at `scripts/abstract.py:1-6` ("optimize an
Agent Skill package (SKILL.md + refs + scripts)"), `SKILL.md:14-15` ("This capability treats that
whole package as the editable artifact"), and the frontmatter description at `SKILL.md:3`.

Honest scoping of the blast radius, because the brief overstates it: `gepa._components`
(`core/cap_evolve/gepa.py:81-100`) builds its component list by `rglob` over the candidate dir, not
by calling `materialize()`, so in a GEPA run `scripts/*.py` *are* reachable as focusable components
and the coding-agent optimizer edits files in place in `work/<cid>/` (`harness.run_step:1456`
`shutil.copytree(parent_dir, workdir)`). So scripts are physically editable today. What is broken
is (a) the capability's declared contract, (b) `Candidate.text_parts` built in `scripts/run.py:21-23`
(the inspection surface a human or another algorithm reads), and (c) the fact that nothing the
capability owns knows a script exists.

**Fix:** extend `_materialize_one` to include `scripts/*` and `assets/*` as namespaced components
(`scripts/<name>`, `assets/<name>`). For text/code files include the content; for binary assets
include a stub line (`<binary asset, N bytes>`) so the optimizer sees the inventory without blowing
the prompt. Cap any single component and note in the component name that a script is *executed*,
not read into the downstream agent's context (that distinction is the whole point of lever 4 at
`SKILL.md:43-47`).

---

**F2. `validate()` is never called by the optimization loop, so nothing it checks is enforced.**

```
$ grep -rn "validate" core/cap_evolve/{harness,adapter,gepa,skillopt,check}.py
core/cap_evolve/harness.py:1706:  "primitive — validate/normalize inputs, enforce a GENERAL rule, …"   # prose in a prompt string
```

That single hit is text inside `_CAP_EDIT_SPACE`, not a call. `abstract.validate()` is reachable
only from `scripts/run.py:22` (a manual `--path` inspection command) and `scripts/check.py:28` (the
self-test, which runs against a synthetic temp package, never the real candidate). No accept/reject
path consults it.

Consequence: `SKILL.md:3` — "Enforces the skill-creator authoring rules (progressive disclosure,
valid frontmatter, body budget, one-level references) so edits stay valid skills" — is false. An
iteration that corrupts the frontmatter is scored, gated, and can be accepted purely on reward.

The repo already knows this and papers over it with prose: `examples/skillsbench/optimizer/INSTRUCTIONS.md:91`
("Re-validate every touched skill against the authoring rules"), `:135`, `:171`, and
`examples/skillsbench/PROMPT.md:226` ("edit must keep the package a VALID skill (validate via the
capability's run.py/check)"). Asking the model to remember to validate is precisely the failure mode
finding F3 is about.

**Fix:** call the capability's `validate()` inside `harness.run_step` after the optimizer returns
and **before** any rollout is paid for. On hard `problems`, follow the pattern the integrity guard
already established (`harness.run_step` docstring, `protected_patterns`): make the step
**indecisive** — no reward recorded, stall counter untouched, best unchanged — since an invalid
package's score measures a broken artifact, not an edit. Surface `warnings` into the rejected/memory
feedback so the next iteration sees them. This needs a generic hook (resolve
`skills/capabilities/<cap>/scripts/abstract.py::validate` if present), not a `skill-package`
special case, or it violates the generic-not-specific rule.

---

**F3. Determinism gap: `validate()` has zero coverage of script content, so a bundled script can be
broken and still pass.**

`_validate_one` (`scripts/abstract.py:166-240`) touches `SKILL.md` and `references/`. The word
`scripts` appears exactly once, at line 236, inside the broken-link regex
`r"\(((?:references|scripts|assets)/[^)\s]+)\)"` — i.e. it checks that a linked script *path exists*
and never that the file is valid Python, importable, or runnable.

Probe: package with `SKILL.md` linking `scripts/x.py`, where `x.py` contains `def broken(:` →
`validate()` returns `{'ok': True, 'name': 'demo', 'problems': [], 'warnings': []}`.

So the answer to "can the optimizer create or modify a script and can `validate()` check it runs?"
is: it can create one (F1's nuance), and nothing checks anything about it. There is no mechanism by
which a script edit becomes *more* reliable than the prose it replaces — which nullifies the stated
major goal.

Compounding it, the string the optimizer actually reads every iteration is a bland one-liner.
`core/cap_evolve/harness.py:1720-1723`:

```python
"skill-package": "Edit the SKILL.md (frontmatter + body), its references, and bundled "
                 "scripts, staying within skill-creator rules (valid frontmatter, "
                 "progressive disclosure, one-level references, concise body).",
```

Compare the `tools` entry at `harness.py:1699-1713`, which spends 15 lines pushing determinism
("HIGHEST-LEVERAGE EDIT: WRITE A NEW CODE-BEARING TOOL … because a deterministic tool can't be
'forgotten' the way a prompt rule can", plus two named patterns and "The body must be real code,
never '...' or docstring-only"). `skill-package` gets one subordinate clause and no push at all.

**Fix, three parts:**
1. Add script validation to `_validate_one`: for each `scripts/**/*.py`, `ast.parse()` the source
   (syntax error → hard `problem`); flag a body that is only a docstring/`pass`/`...` (the
   `tools` brief's "never `...` or docstring-only" rule, which belongs here too).
2. Add an executability check: require every bundled script to support a self-check entry point
   (`python <script> --self-check`, or an `if __name__ == "__main__":` demo) and run it in the
   candidate dir with a timeout and no network, recording per-script `{ok, exit_code, stderr_tail}`.
   A script that cannot be run is not deterministic code, it is a file. This is also the only way
   to satisfy the "principle of lack of surprise" (F8e) now that scripts are an edit target — a
   bundled script is executable context.
3. Rewrite `_CAP_EDIT_SPACE["skill-package"]` to name the determinism lever explicitly and
   mirror the `tools` framing: prefer converting a step the traces show the agent skipping into a
   bundled script the SKILL.md invokes; state execute-vs-read intent; the script must be real code
   with a runnable self-check. Keep it generic (no benchmark or algorithm names).

---

**F4. `apply()` is an unpoliced text-blob writer with a path-traversal hole.**

`scripts/abstract.py:103-123` accepts `{"file", "op": set|append|ensure_contains, "text"}` and does
`target = capability_dir / e["file"]` with no containment check. Probe:
`apply(cap, [{"file": "../escaped.txt", "op": "set", "text": "pwned"}])` writes outside the
capability dir. Given F1's premise that the optimizer will be writing code, an unbounded write
target in the handler that other tooling and the mock optimizer share is not acceptable.

There is also no action policy, unlike the sibling capabilities: `core/cap_evolve/tool_surface.py:11-16`
and `:53-91` define an allow-list of edit kinds (`description|params|examples|schema|code|add|remove|compose`)
so `mcp-tool` can lock off code edits while `tools` permits them, with refusals reported. Once
`skill-package` can write scripts, it needs the same knob — a run should be able to allow
description/body/reference edits while forbidding new executable code.

**Fix:** resolve both paths and reject any edit whose `target.resolve()` is not inside
`capability_dir.resolve()` (report it in a `refused` list rather than raising, matching
`tool_surface.apply`'s `{"changed", "refused"}` contract). Add a `DEFAULT_POLICY` with kinds
`frontmatter|body|reference|script|asset|add|remove` and an `inputs/policy.json` override, reusing
`tool_surface.load_policy`'s pattern so the two capabilities behave the same way.

---

**F5. Folded/multi-line YAML in the frontmatter silently bypasses every description lint.**

`_parse_frontmatter` (`scripts/abstract.py:43-53`) is a hand-rolled line splitter:

```python
for line in m.group(1).splitlines():
    if ":" in line and not line.startswith(" "):
        k, _, v = line.partition(":")
        fm[k.strip()] = v.strip().strip('"').strip("'")
```

A block scalar (`description: >` / `description: |`) — a completely ordinary way to write a long
description, and long descriptions are exactly what this capability produces — yields the literal
value `'>'`. Probe with `description: >` followed by `I can help. ALWAYS use it.`:

- parsed description: `'>'`
- problems: `[]`
- warnings: only the generic "should say WHEN to use the skill"

The length cap, the XML-tag check, the first-person POV lint, the all-caps over-trigger lint and the
1,536-char front-load advisory **all silently pass** on a description that violates two of them.
This is worse than a false negative: it is a silent bypass of the highest-leverage lever's entire
validation.

**Fix:** handle `>`/`|`/`>-`/`|-` block scalars and quoted continuations by joining indented
following lines, or — if a stdlib-only minimal parser is preferred — treat a bare `>`/`|` value as
unparseable and raise it as a hard `problem` ("frontmatter `description` uses a block scalar this
validator cannot read; write it on one line") so it fails loudly instead of passing blind. Add both
cases to `check.py`.

---

**F6. Four reference-structure rules check the wrong thing.**

All from `scripts/abstract.py:223-238`, each with a probe:

- **"one level deep" checks subdirectories, not pointer nesting.** Lines 225-228 warn only when
  `references/<subdir>/` exists. The actual skill-creator rule (`~/.agents/skills/skill-creator/SKILL.md:96-97`:
  "a ref must not point to another ref"; restated at `references/anti-patterns.md:31-34`) is about
  *pointers*. Probe: `references/a.md` containing `[b](references/b.md)` → `warnings: []`. The rule
  advertised at `SKILL.md:41-42` and `concepts.md:81` is not the rule implemented.
  **Fix:** scan each `references/*.md` for `(references/…)` links and warn on every hit.
- **Orphan references are undetected.** `anti-patterns.md:35-36` calls "a reference with no
  pointer" an anti-pattern; nothing checks it. Probe: a `references/orphan.md` that SKILL.md never
  links → no warning. **Fix:** warn for any `references/*.md` not linked from the body.
- **The TOC check is a `"## " in text[:1500]` proxy.** Line 231 passes any long reference whose
  first 1500 chars contain any h2. Probe: a 402-line ref whose only early h2 is `## Intro` → no
  warning. **Fix:** require an early list of links/anchors, or at minimum ≥3 h2 headings in the
  first 1500 chars, and say in the warning what a TOC is.
- **The file is read twice.** Line 231 calls `f.read_text()` a second time after line 230 already
  read it. Trivial; fix while touching the block.

---

**F7. The body budget is advisory only, so the "enforces" claim overstates it — and the 5k-token
number is invented.**

`scripts/abstract.py:214-221` appends body overruns to `warnings`, never `problems`. Probe: a
600-line body → `ok: True`, one warning. Since `ok` is what any caller would gate on (and per F2
nobody calls it at all), the "body budget" in `SKILL.md:3` and the `MAX_BODY_LINES`/`MAX_BODY_TOKENS`
constants are advice, not enforcement.

Separately, `MAX_BODY_TOKENS = 5000` (`abstract.py:36`) has no source. skill-creator says
"<500 lines ideal" and then explicitly "These word counts are approximate and you can feel free to
go longer if needed" (`~/.agents/skills/skill-creator/SKILL.md:93`). `SKILL.md:35-36` presents
"**<500 lines AND ~5k tokens**" as a skill-creator rule and `concepts.md:27` repeats it as
"Target **<500 lines / under ~5k tokens**"; `concepts.md:80` lists it under "Validity rules".
Sibling irony worth noting: `skills/capabilities/tools/SKILL.md` is 749 lines, so cap-evolve's own
largest capability skill violates the rule this one polices.

**Fix:** either promote >500 lines to a hard `problem` (and say so), or reword `SKILL.md:3` and
`concepts.md:75-82` to "checks and warns" rather than "enforces". Keep the 5k-token figure only if
it is labelled as this repo's own heuristic, not attributed to skill-creator.

---

**F8. Fidelity to skill-creator: five real gaps, one thing correctly delegated.**

- **(a) `assets/` has no lever.** `SKILL.md:15` and `concepts.md:16` list `assets/` as part of the
  package, then the four levers at `SKILL.md:20-47` never mention it and neither
  `materialize()` nor `validate()` touches it. skill-creator's anatomy block
  (`skill-creator/SKILL.md:75-83`) names it as one of three bundled-resource kinds ("files used in
  output (templates, icons, fonts)"). **Fix:** add it as a fifth lever in one line (templates the
  skill emits, not context the agent reads) or delete the mentions; a listed-but-unreachable part
  of the package is worse than an omission.
- **(b) The description-optimization loop ships as prose while skill-creator ships it as code.**
  skill-creator bundles `scripts/run_loop.py` (11.5 KB) and `scripts/improve_description.py`
  (11.1 KB) and its SKILL.md gives the exact mechanic (`skill-creator/SKILL.md:375-395`): 60/40
  train/held-out split, **each query run 3 times** for a reliable trigger rate, propose, re-evaluate
  on both, return `best_description` **selected by test score**. `references/description-optimization.md:32-41`
  renders that as four narrative bullets ("Assemble a held-out set… Propose candidate
  descriptions… Measure trigger-rate… Keep the candidate that scores best") with no 3-run detail,
  no split ratio, and **no script**. This is the "bundle repeated work as a script" rule
  (`skill-creator/SKILL.md:301`) violated by the very skill that teaches it. **Fix:** ship
  `scripts/trigger_eval.py` that takes a JSON eval set (`[{"query", "should_trigger"}]`), splits it
  deterministically by seed, runs each query N times against a candidate description, and prints
  `{train_score, heldout_score, per_query}` — then make `description-optimization.md` point at it
  instead of describing it. This is the single highest-value deterministic-code addition available
  to this skill and it doubles as the worked example the determinism goal needs.
- **(c) Domain organization by variant is missing from the body.** skill-creator devotes a block to
  it (`skill-creator/SKILL.md:100-109`: `references/aws.md`, `gcp.md`, `azure.md`, "Claude reads
  only the relevant reference file"). It appears only at `concepts.md:63-65`. Given the body already
  spends 4 lines on references (`SKILL.md:38-42`), one clause naming the variant-split pattern is
  the cheapest way to make lever 3 actionable.
- **(d) Eval/assertion discipline is correctly delegated — no work needed.** skill-creator's
  with-skill/without-skill baseline, `evals/evals.json`, assertions, and `agents/grader.md` are
  replaced in cap-evolve by the val split + significance gate. Do **not** import them; that would
  duplicate core-owned machinery. Worth one sentence in the skill saying so, so a future editor
  does not "helpfully" add them.
- **(e) The lack-of-surprise principle is one clause and is about to become load-bearing.**
  `SKILL.md:51-52` ("Don't introduce unaudited/exfiltrating content — a skill body is executable
  context") is an adequate summary of `skill-creator/SKILL.md:111` today, but the determinism goal
  makes the optimizer an author of executable code that will run on the user's machine. **Fix:**
  once F3's script execution lands, the self-check must run with a timeout and without network
  access, and `validate()` should flag network/subprocess/`eval` usage in a newly added script as a
  warning the human sees in the diff.

Correctly present, no action: the "bundle repeated work seen across transcripts" heuristic is in all
three places it should be (`SKILL.md:43-44`, `concepts.md:60-62`, `anti-patterns.md:43-46`) and is
faithful to `skill-creator/SKILL.md:301`. The reader-tier block (`SKILL.md:64-71`) is genuinely
specific to this capability and earns its lines.

---

**F9. Conciseness: ~45 of 122 SKILL.md lines are a second or third copy of something already in
context.**

- `SKILL.md:83-93` ("## Measure every edit against the objective") restates the honesty invariants
  that CONTEXT.md declares core-owned, and `harness._algorithm_brief` (`core/cap_evolve/harness.py:1770-1780`)
  already injects them into the optimizer prompt **every iteration**: "It is ACCEPTED only if the
  per-task improvement clears a significance bar … Aim for a real, generalizing gain across a CLASS
  of failures — not a one-off patch to a single task (that overfits …)". `concepts.md:67-73` is a
  third copy. **Fix:** delete `SKILL.md:83-89` outright; keep only the *skill-specific* half —
  lines 90-93, the held-out trigger-rate objective, which nothing else says.
- `SKILL.md:101-107` ("## Optimizing it each iteration (analyze → ideate → edit)") is a paraphrased
  4-way duplicate: `skills/capabilities/mcp-tool/SKILL.md:216-224`, `system-prompt/SKILL.md:182-190`,
  and `tools/SKILL.md` all carry the same "analyze before editing: (a) recurring failures clustered
  by root cause and (b) good behavior seen only on some trials … Be economical: one good edit, then
  stop." It is also the `diagnose` skill's entire job, and `harness._inject_optimizer_context`
  copies `diagnose` into `guidance/diagnose/` for every run (`harness.py:1231-1243`). **Fix:** cut
  to one line pointing at `guidance/diagnose/`, or delete and let the shared brief carry it.
- `SKILL.md:54-62` ("## How agents use a skill (progressive disclosure)") duplicates
  `concepts.md:19-31` in substance, and lever 2 at `SKILL.md:34-36` has already told the reader the
  body is a recurring cost. **Fix:** collapse to the two sentences at `SKILL.md:61-62` (the
  consequences), which are the only non-duplicated part, and let `concepts.md` hold the model.
- `SKILL.md:73-81` ("## The description is the trigger") is the **third** statement of
  under/over-trigger, after lever 1 (`SKILL.md:25-32`) and `description-optimization.md:21-30`.
  **Fix:** delete the section; lever 1 already links the playbook.
- `SKILL.md:95-99` ("## Handlers") enumerates validate's rules, which `concepts.md:75-82` also
  enumerates and `abstract.py:8-22` documents a third time. **Fix:** reduce to the three handler
  signatures with no rule list.

Net: SKILL.md should land near 70 lines with more actual mechanism in it (the script lever and the
new `trigger_eval.py` pointer) than it has now.

Reference-level duplication: `anti-patterns.md:48-52` ("## Process": don't select on iteration
examples, one edit at a time) repeats `SKILL.md:83-93` *and* `concepts.md:67-73` *and* the harness
brief — a fourth copy. Delete it; the file's job is skill *smells*, not run discipline.

---

**F10. Generic-not-specific: mostly clean, three items to fix.**

- `SKILL.md:9` `sources: [agentskills, skillgrad, trace2skill, skillopt]` — `skillopt` is a
  cap-evolve *algorithm* (`skills/algorithms/skillopt`), so a capability skill's frontmatter names a
  specific algorithm. Drop it; the capability must not imply an algorithm.
- `concepts.md:72` "This mirrors ACE's generation → reflection → curation of an evolving playbook"
  (and `SKILL.md:102`'s "treat the skill like an evolving playbook you curate") is paper
  name-dropping that changes no action. Cut the ACE clause; `docs/sources.bib` is where citations
  belong.
- `concepts.md:40-44` presents the 1,536-char listing truncation and `maxSkillDescriptionChars` as
  a universal rule; it is Claude-Code-host-specific and configurable. A skill optimized for another
  host has no such cap. **Fix:** label it "Claude Code host default (configurable)". Same for
  `LISTING_CAP_CHARS` (`abstract.py:39`) — the warning text at `:210-212` should say which host it
  applies to. Also `SKILL.md:86` "the intake benchmark score" hardwires a phase name where "the
  benchmark score being optimized" would do.

Nothing here is overfitted to a benchmark, dataset, or vendor, and the four levers are stated in
terms of the artifact rather than any example skill — that part is right.

---

**F11. Stray artifacts: the brief's premise is wrong; `__pycache__` is not committed.**

```
$ git ls-files skills/capabilities/skill-package
  .../SKILL.md, meta.yaml, references/{anti-patterns,concepts,description-optimization}.md,
  scripts/{_bootstrap,abstract,check,run,token_report}.py
$ git status --short skills/capabilities/skill-package     # (empty)
$ grep -n pycache .gitignore
2:__pycache__/
```

`scripts/__pycache__/` exists on disk (6 `.pyc` files, 3.12 and 3.14) but is untracked and ignored.
No repo change needed. Reported so this is not "fixed" twice.

One real drift, out of this repo's control but worth knowing: the installed copy at
`~/.claude/skills/skill-package/SKILL.md` is dated 2026-06-18 against the repo's 2026-08-15, is 49
lines vs 122, and still carries `(the SkillGrad / Trace2Skill / SkillOpt case)` in its description —
which the repo version has already correctly removed. Whatever installs skills locally is stale.

---

**F12. `token_report.py` measures the one thing that is free and ignores the levers.**

`scripts/token_report.py:47-51` counts `references/*.md` tokens — content that costs **zero**
context until read, as its own docstring says at `:5-6`. It never inventories `scripts/`, which is
the lever this capability is supposed to push, nor `assets/`. So the report tells the optimizer the
size of the cheap thing and nothing about the deterministic-code surface.

**Fix:** add a `scripts` section listing each script with its size, whether it has a runnable entry
point, and its self-check status (once F3 lands), plus an explicit `"context_cost": 0` note on both
`references` and `scripts` so the optimizer reads the budget correctly instead of inferring that a
big reference is expensive.

---

**F13. `check.py` never exercises the things most likely to break.**

`scripts/check.py:19-64` covers: materialize finds SKILL.md, a valid skill passes, a bad name is
rejected, an XML tag in a name is rejected, POV/all-caps lints fire, `apply` writes a reference,
`token_report` returns budget fields. Every probe in this review that found a defect (F1 scripts/assets,
F3 broken script, F4 traversal, F5 block scalar, F6 nested ref / orphan ref / fake TOC, F7 oversize
body) is untested.

**Fix:** add one assertion per finding above, and — for the determinism goal — one end-to-end case:
`apply` a new `scripts/helper.py`, then assert `validate()` reports it as a component **and** that
its self-check ran and passed. That single test is what makes "the optimizer can add deterministic
code" a checked property instead of an aspiration.

## Acceptance criteria

- [ ] `materialize()` returns `scripts/*` and `assets/*` as namespaced components; a probe package
      containing both shows them in `run.py --path` output (F1).
- [ ] `harness.run_step` calls the capability's `validate()` after the optimizer returns and before
      any rollout spend, via a generic per-capability hook (no `skill-package` special-casing); hard
      `problems` make the step **indecisive** (no reward, stall counter untouched) and `warnings`
      reach the optimizer's rejected/memory feedback (F2).
- [ ] `validate()` `ast.parse()`s every bundled script, reports a syntax error as a hard `problem`,
      flags docstring-only/`...` bodies, and runs each script's self-check with a timeout and no
      network, reporting `{ok, exit_code, stderr_tail}` per script (F3).
- [ ] `_CAP_EDIT_SPACE["skill-package"]` in `core/cap_evolve/harness.py` names the deterministic-code
      lever explicitly, at comparable force to the `tools` entry, with no benchmark/algorithm names (F3).
- [ ] `apply()` refuses any edit resolving outside the capability dir (reported in a `refused` list,
      matching `tool_surface.apply`'s contract) and honors an action policy with an
      `inputs/policy.json` override (F4).
- [ ] `_parse_frontmatter` handles `>`/`|` block scalars and quoted continuations, or raises a hard
      `problem` on an unreadable value — a folded `description: >` containing "I can help." and
      "ALWAYS" fires both the POV and all-caps lints (F5).
- [ ] `validate()` warns on a `references/*.md` that links another reference, on an orphan reference
      with no pointer from SKILL.md, and requires a real TOC (not any h2) in a >300-line reference (F6).
- [ ] The "enforces … body budget" claim is either made true (>500 lines = hard `problem`) or the
      wording in `SKILL.md:3` and `concepts.md:75-82` is corrected to "checks and warns"; the 5k-token
      figure is either dropped or labelled as this repo's heuristic, not skill-creator's (F7).
- [ ] `assets/` is either given a one-line lever or removed from the package description (F8a).
- [ ] `scripts/trigger_eval.py` exists, takes a `[{"query","should_trigger"}]` set, splits it by
      seed, runs each query N times, prints `{train_score, heldout_score, per_query}`, and is the
      thing `references/description-optimization.md` points at instead of narrating (F8b).
- [ ] SKILL.md is ≤ ~75 lines with `## Measure every edit against the objective` (83-89),
      `## The description is the trigger` (73-81), and `## Optimizing it each iteration` (101-107)
      deleted or reduced to pointers; `anti-patterns.md`'s `## Process` section removed (F9).
- [ ] `sources:` drops `skillopt`; the ACE clause is gone from `concepts.md:72`; the 1,536 listing
      cap is labelled host-specific in both `concepts.md` and the `validate()` warning text (F10).
- [ ] `token_report.py` inventories `scripts/` with self-check status and states `context_cost: 0`
      for references and scripts (F12).
- [ ] `python scripts/check.py` passes, with one new assertion per finding F1, F3, F4, F5, F6, F7,
      and an end-to-end case that applies a new `scripts/helper.py` and asserts it appears as a
      component **and** its self-check ran and passed (F13).
- [ ] A real run (toy example is fine — deterministic, no API key) demonstrates the optimizer
      proposing a **script-level** edit that is materialized as a component, validated including its
      self-check, and put through the val significance gate; the run dir's candidate diff shows the
      script change and the step's accept/reject reason references the validation result.
- [ ] No new runtime dependency in `core/` (stdlib only).

## Out of scope

- Rewriting `_bootstrap.py`. It is 82 lines of load-bearing path resolution with the rationale in
  comments; it is not this skill's problem and touching it risks the stale-core bug it documents.
- Importing skill-creator's eval machinery (`evals.json`, assertions, `agents/grader.md`,
  `aggregate_benchmark.py`, the eval viewer). cap-evolve's val split + significance gate replaces it;
  duplicating it would violate the core-owns-honesty invariant (F8d).
- Fixing `skills/capabilities/tools/SKILL.md`'s 749-line body, or the duplicated "analyze → ideate →
  edit" and reader-tier blocks in the three sibling capability skills. Noted here only as evidence
  for F7 and F9; each needs its own issue.
- Refreshing the stale installed copy at `~/.claude/skills/skill-package/` (F11) — an install/sync
  concern, not a repo change.
- `scripts/__pycache__/` — already gitignored and untracked (F11). No action.
- Whether `name` may contain "anthropic"/"claude" (`abstract.py:184-185`, `concepts.md:35-36`).
  Neither `skill-creator/SKILL.md` nor `references/schemas.md` states this rule, so it cannot be
  verified against the authority this review is graded on; it may well be correct from the platform
  docs. Someone should attach the exact doc link to `concepts.md`, but not under this issue.

OsherElhadad:
Closes #322.

## Before → after: what the optimizer can edit

| part of the package | before | after |
|---|---|---|
| `SKILL.md` body | component, editable | unchanged |
| `SKILL.md` frontmatter | editable, but a `description: >` block scalar parsed as the literal `'>'`, silently bypassing every description lint | parsed and linted |
| `references/*.md` | component, editable | unchanged (plus nested-pointer / orphan / TOC checks) |
| `scripts/*` | **not a component** — the capability that claimed to own them didn't know they existed | component; `apply()` can create or rewrite one; `validate()` compiles it and runs its `--self-check` |
| `assets/*` | listed in the docs, unreachable in code | component (binaries as `<binary asset, N bytes>` stubs) |
| write containment | `capability_dir / e["file"]`, so `../escaped.txt` wrote outside the package | resolved and refused, reported in `refused` (matching `tool_surface.apply`'s contract) |
| action policy | none | `policy.json`: `frontmatter\|body\|reference\|script\|asset\|add\|remove` — a run can allow prose and forbid new executable code |
| enforcement | `validate()` was called by **nothing** in `harness`/`gepa`/`skillopt`/`adapter` — every rule enforced on nothing | `harness.run_step` calls each capability's own `validate()` after the optimizer returns and **before any rollout is paid for**, via a generic per-capability hook (no capability special-cased) |

An invalid candidate makes the step **indecisive** — no reward recorded, stall counter
untouched, best unchanged (the tamper path's discipline: a broken artifact's score measures
the breakage, not the edit). Unlike an infra outage the fault *is* the optimizer's, so the
reason is filed into the rejected memory and into the LEDGER's new **"Not scored — fix the
execution, keep the idea"** section; warnings are appended to the rejection feedback.
Problems the parent already had are excluded (digit-normalized compare), so a pre-existing
violation — a seed body already over budget — cannot wedge a run.

## `validate()` rule table

| rule | severity | note |
|---|---|---|
| no `SKILL.md` | problem | |
| `name` missing / >64 chars / not `[a-z0-9-]` / XML tag / reserved word | problem | |
| `description` empty / >1024 chars / XML tag | problem | now also reached through `>`/`|` block scalars |
| body >500 lines | **problem** (was warning) | so "body budget" is enforcement, not advice |
| body >~5k tokens | warning | labelled *this repo's* heuristic — skill-creator says "<500 lines ideal … feel free to go longer" |
| broken `references\|scripts\|assets/…` link from the body | problem | |
| bundled `.py` does not `ast.parse` | **problem** (new) | |
| bundled `.py` declares `--self-check` and it fails | **problem** (new) | run with a 30s timeout, cwd = package, stripped env (PATH/HOME/PYTHONPATH only, proxies blanked); `{ok, exit_code, stdout_tail, stderr_tail}` per script |
| bundled `.py` has no declared `--self-check` | warning (new) | nothing verifies it after an edit |
| bundled `.py` is docstring/`pass`/`...` only | warning (new) | the `tools` brief's "never `...` or docstring-only", which belongs here too |
| bundled `.py` imports network/subprocess or calls `eval`/`exec` | warning (new) | judged from the AST, not substrings — a skill is executable context, the human must see it in the diff |
| a `references/*.md` links another reference | warning (new) | the *pointer* rule, which the old subdirectory check never tested |
| a `references/*.md` nobody links from `SKILL.md` | warning (new) | orphan |
| reference >300 lines with no TOC in its first 1500 chars | warning (rule fixed) | was `"## " in text[:1500]`, which any single h2 satisfied; now ≥3 anchor links (or ≥3 section headings) **and positional** — a TOC behind a preamble is a TOC the head-reader never sees |
| POV drift / ALL-CAPS in the description / near the listing cap | warning | listing cap now labelled the Claude Code host default (configurable), not a universal rule |

## Test results

`python skills/capabilities/skill-package/scripts/check.py` — one case per rule, each with a
deliberately broken fixture:

```
{"self_check": "ok", "train_score": 1.0, "heldout_score": 1.0}
{
  "skill": "skill-package",
  "ok": true,
  "problems": [],
  "notes": [
    "materialize exposes SKILL.md + references + scripts + assets",
    "block-scalar (`description: >`) frontmatter is parsed and linted",
    "reference structure: nested pointers, orphans, missing TOC all warn",
    "apply() contains writes to the package and honors the action policy",
    "end-to-end: optimizer-created script -> component -> self-check ran and passed",
    "token_report inventories scripts; trigger_eval self-check passes"
  ]
}
```

Cases covered (each fails the check if the rule regresses): materialize sees
scripts + assets + a binary stub · a valid package passes · bad name · XML in name ·
POV/all-caps lints on a one-line **and** on a folded description · 600-line body is a hard
problem · nested pointer, orphan, and preamble-buried TOC all warn · broken link is a hard
problem · non-compiling script is a hard problem · stub body + missing self-check warn ·
failing self-check is a hard problem · `../escaped.txt` refused · a script edit refused by
`policy.json` · **end to end: `apply()` creates `scripts/helper.py`, `materialize()` returns
it as a component, `validate()` reports `self_check: "ran", ok: true`**.

Core suite, pinned imports per #349:

```
789 passed, 12 skipped in 109.84s (0:01:49)
```

(matches clean main; without the pin, 2 `test_eventstream.py` tests fail identically on a
clean tree — the dual-`cap_evolve`-tree artifact of #349, not this branch.)

Sibling capability checks (`tools`, `system-prompt`, `mcp-tool`) still pass.

This skill's own package, through its own validator: **0 problems, 0 authoring violations**
(the ref→ref link the lint flagged is gone — `concepts.md` no longer links its siblings; the
body is 98 lines). The 4 remaining warnings are informational about the capability's own
tooling (`_bootstrap.py`/`run.py` have no `--self-check`; `abstract.py`/`trigger_eval.py`
import `subprocess`, which is what running a self-check and shelling a judge require).

## Real-run evidence

`examples/toy_skill/` (new, zero-API): the capability IS a skill package, the tasks are
arithmetic as people write it (`3 plus 4`, `1,200 + 5`), the seed skill only *asks* for
normalization in prose, and the stand-in agent can only get those right by running a bundled
script the body names. **The score cannot rise without adding code**, which is the whole point.

**Mock optimizer** (`bash examples/toy_skill` recipe in its README): baseline val 0.333 →
candidate val 1.0, accepted `paired Δ̄=+0.6667 > 1.0·SE=0.3333 (n=3)`, sealed test **0.333 →
1.0**. `cap-evolve diff --best` (abridged):

```
seed → best   2 files  +53 -0
  SKILL.md  +7 -0
  + python scripts/normalize.py '<the expression>'    # prints the answer alone
  scripts/normalize.py  +46 -0
  + def normalize(expr): ...   # word ops + thousands separators
  + if __name__ == "__main__":
  +     if "--self-check" in sys.argv: _self_check()
```

Validation ran on it before any rollout: `capability_validation_warnings … scripts/normalize.py
uses eval() — a bundled script is executable context; confirm this is intended in the diff`.

**Real model** — `aws/gpt-oss-120b` through the OpenAI-compatible proxy, driven as the
`generic` optimizer by a ~100-line throwaway proposer (`/tmp/capreview/llm_optimizer.py`, not
committed) that materializes the package via `abstract.materialize`, asks for one edit, and
applies it via `abstract.apply`. 4 iterations, cents of spend:

```
cand_0001 | indecisive (validation): scripts/compute_arith.py --self-check failed (exit 1):
            FAIL 3 plus 4: got error: cannot parse '3 plus 4', expected 7
cand_0002 | indecisive (validation): scripts/compute_arith.py --self-check failed (exit 1): …
cand_0003 | indecisive (validation): scripts/compute_arith.py --self-check failed (exit 1): …
cand_0004 | paired Δ̄=+0.6667 > 1.0·SE=0.3333 (SE=0.3333, n=3)        ← ACCEPTED
val 0.333 → 1.0 · sealed test 1.0 (baseline 0.333, delta +0.667)
```

The three rejected candidates are the finding: a real model wrote a bundled script three times
and each time its own `--self-check` failed — one used `ast.Num`, removed in Python 3.12. On
`main` those would have been scored and could have been accepted on reward alone. Here they
were never scored, cost no rollouts, left `best` untouched, and the failure text was fed back.
The accepted candidate's diff is `scripts/compute_arith.py` (+81) plus the pointer line in
`SKILL.md`, i.e. **the accepted edit is code, not prose**.

## Conciseness

`SKILL.md` 122 → **98 lines** while *gaining* the script lever, the assets lever, the
variant-organization clause and the trigger-eval pointer. Deleted as restatements (per the
ownership contract): `## Measure every edit against the objective` (the honesty invariants,
which `harness._algorithm_brief` injects every iteration), `## The description is the trigger`
(third statement, after lever 1 and the reference), `## Optimizing it each iteration` (the
`diagnose` skill's job, and `diagnose` is copied into the same workdir), and the
progressive-disclosure model (kept in `concepts.md`). Also dropped: `anti-patterns.md`'s
`## Process` section (a fourth copy of run discipline), the ACE name-drop, and `skillopt` from
`sources:` (a capability must not name an algorithm). Kept, as the assigned owner: the
reader-capability-tier block. `references/` 196 → 199 lines net, with the description loop now
pointing at a script instead of narrating it.

## Not done

- **`skill-creator`'s eval machinery is deliberately not imported** (`evals.json`, assertions,
  `agents/grader.md`, the eval viewer). cap-evolve's val split + significance gate replaces it;
  duplicating it would fork the honesty invariants. `concepts.md` says so, so a future editor
  does not "helpfully" add it.
- **`trigger_eval.py` needs a judge you supply** (`--judge-cmd`). Shipping a built-in model
  client would put a vendor in a skill; the offline `--self-check` proves the split/scoring
  plumbing with a keyword judge, no network.
- **The self-check runner is not a sandbox.** It runs bundled code with a timeout, the package
  as cwd, no API keys and blanked proxy vars, and warns on network/subprocess/`eval` from the
  AST — but a determined script can still act. `CAPEVOLVE_NO_SCRIPT_EXEC=1` skips execution
  (compile checks stay). Real isolation is a container-level concern, not this handler's.
- **Only `.py` scripts are executed/compiled.** A `.sh` helper is materialized and link-checked
  but not run; no evidence yet that shell helpers matter here.
- **`validate()` is not called by `gepa`/`skillopt` directly** — they route through
  `harness.run_step`, which is where the hook lives. An algorithm that ever bypasses `run_step`
  would bypass validation too.
- **`assets/` gets one lever line and inventory, no asset-specific validation** (no image or
  template checks). Enough to stop it being listed-but-unreachable.
- **Out of scope, per the issue:** `_bootstrap.py`, `tools/SKILL.md`'s length (#348 handles it),
  the duplicated blocks in the three sibling capability skills, the stale installed copy at
  `~/.claude/skills/skill-package/`. `scripts/__pycache__/` was already gitignored and
  untracked — the issue's own F11 corrected that premise; nothing to remove.
- **What other skills should drop** (I did not touch their dirs): every skill except
  `capabilities/skill-package` should delete reader-capability-tier advice, and the copies of
  the honesty invariants in the 10 non-owner skills stay for their owners to remove.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,6 +81,27 @@ All notable changes to cap-evolve are documented here. The format follows
   `2,000 USD` into `$0.00`.
 
 ### Added
+- **`skill-package` optimizes the WHOLE package, and its rules are now enforced by the loop.**
+  `materialize()` exposed only `SKILL.md` + `references/*.md`, so `scripts/` and `assets/` were
+  not components of the artifact the capability claimed to own, and `validate()` — the entire
+  "edits stay valid skills" story — was never called from `harness`/`gepa`/`skillopt`. Now:
+  every file in the package is a component (binary assets as inventory stubs); `apply()` can
+  create or rewrite any of them (a NEW bundled script included), refuses writes escaping the
+  package, and honors an action policy (`policy.json`:
+  `frontmatter|body|reference|script|asset|add|remove`) so a run can allow prose but forbid new
+  code; `validate()` `ast.parse()`s every bundled script, rejects a stub body, and RUNS a
+  declared `--self-check` with a timeout and a stripped env. `harness.run_step` calls each
+  capability's own `validate()` after the optimizer returns and **before any rollout is paid
+  for** (generic per-capability hook, no capability special-cased): hard problems make the step
+  **indecisive** — no reward, stall counter untouched, best unchanged — with the reason filed in
+  the rejected memory and the LEDGER's new "Not scored" section, and warnings carried into the
+  optimizer's feedback. Problems the parent already had are excluded, so a pre-existing
+  violation cannot wedge a run. Also: block-scalar (`description: >`) frontmatter is parsed
+  instead of silently bypassing every description lint; body >500 lines is a hard problem;
+  nested/orphan references and fake TOCs warn; `scripts/trigger_eval.py` makes held-out
+  trigger-rate selection a deterministic script instead of prose; `examples/toy_skill/` is a
+  zero-API run whose capability IS a skill package and whose score can only rise by adding
+  bundled code. `skill-package/SKILL.md` shrank 122 → 100 lines while gaining the script lever.
 - **Four real τ²-bench airline runs** on `aws/gpt-oss-120b` (agent + user simulator) with
   `aws/claude-opus-5` proposing edits, committed with `events.jsonl` at
   `examples/tau2_airline/run_agentopt_v{2,3,4}/`. **All four are null results** — `best_id =
--- a/core/cap_evolve/harness.py
+++ b/core/cap_evolve/harness.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 import contextlib
 import json
 import os
+import re
 import shutil
 import subprocess
 import sys
@@ -906,10 +907,17 @@ def _build_ledger(workdir: Path, run_dir: RunDir, rejected, history) -> None:
 
     table = ["| iter | candidate | parent | outcome | val | Δ vs parent | broke (were passing) | fixed |",
              "| --- | --- | --- | --- | --- | --- | --- | --- |"]
+    void: list[str] = []
     for i, rec in enumerate(rows, 1):
         cid = str(rec.get("candidate"))
         par = str(rec.get("parent") or "seed")
         outcome = "ACCEPT" if rec.get("accept") else "reject"
+        # A step with no val was never measured (an invalid candidate, or an edit to a
+        # sealed file). Calling that "reject" teaches the wrong lesson — the edit's IDEA
+        # was never tested, only its execution was broken — so name it and say why.
+        if rec.get("val") is None and str(rec.get("reason", "")).startswith("indecisive"):
+            outcome = "not scored"
+            void.append(f"- **{cid}**: {str(rec.get('reason'))[:600]}")
         val = rec.get("val")
         pval = rec.get("parent_val")
         d = (f"{val - pval:+.3f}" if isinstance(val, (int, float))
@@ -931,7 +939,11 @@ def _build_ledger(workdir: Path, run_dir: RunDir, rejected, history) -> None:
         "this to never re-introduce a change that broke a task, and to see which approaches "
         "the gate accepted vs rejected.\n\n"
         "## Iterations\n" + "\n".join(table) + "\n\n"
-        f"## Current best: {best}\n"
+        + ("## Not scored — fix the execution, keep the idea\n"
+           "These candidates were never measured, so their reward says nothing. Correct the "
+           "stated fault and retry the same idea; do not abandon it, and do not resubmit it "
+           "unchanged.\n" + "\n".join(void) + "\n\n" if void else "")
+        + f"## Current best: {best}\n"
     )
     (workdir / "LEDGER.md").write_text(text, encoding="utf-8")
 
@@ -1369,17 +1381,27 @@ def _write_instructions_pointer(path: Path, skills_dir: str) -> None:
 
 def _tamper_step(run_dir: RunDir, *, cid: str, parent_id, workdir: Path,
                  report, current_val: SplitResult, optimizer_seconds: float,
-                 opt_cost_usd: float, opt_tokens: int, optimizer_error) -> dict:
-    """An INDECISIVE step for a candidate that edited a protected file.
+                 opt_cost_usd: float, opt_tokens: int, optimizer_error,
+                 kind: str = "integrity", event: str = "tamper_detected",
+                 detail_key: str = "tamper", rejected=None) -> dict:
+    """An INDECISIVE step for a candidate that was never validly measurable.
 
-    Not a rejection at 0.0: the candidate was never validly measured, so recording a
-    reward would poison the split mean and the paired gate. ``accepted=None`` leaves
-    the stall counter alone, ``set_best`` is never called, and the workdir is
-    snapshotted for forensics so an operator can see exactly what was edited.
+    Two causes share this path: the candidate edited a protected file (integrity),
+    or it is invalid by its own capability's rules (validation). Neither is a
+    rejection at 0.0: the candidate was never validly measured, so recording a
+    reward would poison the split mean and the paired gate. ``accepted=None``
+    leaves the stall counter alone, ``set_best`` is never called, and the workdir
+    is snapshotted for forensics so an operator can see exactly what was edited.
+
+    Unlike an infrastructure outage, an invalid edit IS the optimizer's doing — so
+    when ``rejected`` memory is supplied the reason is filed there, and the next
+    iteration is told what it broke instead of repeating it.
     """
-    reason = "indecisive (integrity): " + report.reason
-    run_dir.log_event("tamper_detected", candidate=cid, parent=parent_id,
+    reason = f"indecisive ({kind}): " + report.reason
+    run_dir.log_event(event, candidate=cid, parent=parent_id,
                       report=report.to_dict(), reason=report.reason)
+    if rejected is not None:
+        rejected.add(cid, f"candidate {cid} (not scored: invalid)", reason, None)
     run_dir.snapshot(cid, workdir, ignore=_SNAPSHOT_IGNORE)  # forensics only — never best
     run_dir.update_spent(iterations=1, accepted=None)
     run_dir.log_event("step", candidate=cid, accept=False, reason=reason, val=None,
@@ -1394,9 +1416,9 @@ def _tamper_step(run_dir: RunDir, *, cid: str, parent_id, workdir: Path,
         # delta 0.0 keeps the GateDecision shape; there is no measurement to report.
         "decision": {"accept": False, "reason": reason, "delta": 0.0, "threshold": 0.0,
                      "indecisive": True},
-        "candidate_val": None,      # NO reward — a tampered run is missing data
+        "candidate_val": None,      # NO reward — an unmeasurable run is missing data
         "parent_val": current_val.to_dict(),
-        "tamper": report.to_dict(),
+        detail_key: report.to_dict(),
         "regressions": [],
         "optimizer_seconds": optimizer_seconds,
         "optimizer_usd": opt_cost_usd,
@@ -1513,6 +1535,24 @@ def run_step(
                                 opt_cost_usd=opt_cost_usd, opt_tokens=opt_tokens,
                                 optimizer_error=optimizer_error)
 
+    # The capability's OWN validity rules, checked here (still before the gate, and
+    # before any rollout is paid for) rather than left to prose the optimizer may skip.
+    # An invalid artifact's score measures a broken candidate, not the edit, so the
+    # step is INDECISIVE — same discipline as the tamper path.
+    validation = _capability_validate(capabilities, workdir, parent_dir=parent_dir)
+    if validation is not None:
+        if validation.warnings:
+            run_dir.log_event("capability_validation_warnings", candidate=cid,
+                              warnings=validation.warnings)
+        if validation.problems:
+            return _tamper_step(run_dir, cid=cid, parent_id=parent_id, workdir=workdir,
+                                report=validation, current_val=current_val,
+                                optimizer_seconds=optimizer_seconds,
+                                opt_cost_usd=opt_cost_usd, opt_tokens=opt_tokens,
+                                optimizer_error=optimizer_error, kind="validation",
+                                event="capability_invalid", detail_key="validation",
+                                rejected=rejected)
+
     cand_val = evaluate_candidate(adapter, workdir, run_dir=run_dir, split="val",
                                   n_trials=n_trials, tag=cid)
 
@@ -1605,7 +1645,12 @@ def run_step(
                           n_tasks=cand_val.n_tasks)
     else:
         if rejected is not None:
-            rejected.add(cid, summary, decision.reason, cand_val.reward, note=note, impact=impact)
+            reason = decision.reason
+            if validation is not None and validation.warnings:
+                # Authoring smells the candidate carries are part of why it is worth
+                # revisiting — the next iteration should see them, not rediscover them.
+                reason += " | validation warnings: " + "; ".join(validation.warnings[:5])
+            rejected.add(cid, summary, reason, cand_val.reward, note=note, impact=impact)
     if store is not None:
         store.commit(f"iter {run_dir.spent.iterations}: "
                      f"{'ACCEPT' if accepted else 'reject'} {summary}",
@@ -1714,9 +1759,24 @@ _CAP_EDIT_SPACE = {
     "system-prompt": "Edit the prompt/policy text: instructions, decision policy, and the "
                      "output contract. Prefer sharpening rules the traces show the agent "
                      "breaking; do not just append more preamble.",
-    "skill-package": "Edit the SKILL.md (frontmatter + body), its references, and bundled "
-                     "scripts, staying within skill-creator rules (valid frontmatter, "
-                     "progressive disclosure, one-level references, concise body).",
+    "skill-package": "Edit ANY part of the package: the SKILL.md frontmatter (the "
+                     "`description` decides whether the skill fires at all — the "
+                     "highest-leverage text), the body, `references/*.md`, and the "
+                     "bundled `scripts/`. HIGHEST-LEVERAGE EDIT: TURN A SKIPPED PROSE "
+                     "STEP INTO A BUNDLED SCRIPT the body invokes by command line — a "
+                     "step the agent runs can't be 'forgotten' the way a body rule can, "
+                     "and a script's source never enters the agent's context (only its "
+                     "output), so it is cheaper than the prose it replaces. Two patterns "
+                     "to prefer: (1) the traces show the agent re-deriving the same "
+                     "helper or doing a deterministic transform by hand — write it once "
+                     "into scripts/ and say EXECUTE it, don't read it; (2) a check the "
+                     "agent keeps skipping — make it a script that exits non-zero. The "
+                     "script must be real working code (never '...' or docstring-only) "
+                     "with a `--self-check` entry point, because validation RUNS it "
+                     "before any rollout is paid for. Keep the body lean (<=500 lines) "
+                     "and push depth into one-level references with explicit pointers; "
+                     "the frontmatter/body/reference/script split is validated, so an "
+                     "invalid package is not scored at all.",
     "mcp-tool": "ONLY safe edits: tool/parameter documentation, in-description examples, and "
                 "adding/removing tools from the exposed set. The wire schema and tool code "
                 "are owned by the external server and are NOT editable here.",
@@ -1913,6 +1973,100 @@ def _parallel_note(parallel: bool, optimizer_name: str | None) -> str:
     return ("Your agent runs single-threaded (no subagents). Still address as MANY clusters as you "
             "can in this ONE candidate — work through them in turn, drafting and keeping every "
             "real, safe fix, not just the biggest one.")
+
+
+def _load_capability_abstract(name: str, tag: str):
+    """Import ``skills/capabilities/<name>/scripts/abstract.py``, or None."""
+    import importlib.util
+
+    skills_root = Path(__file__).resolve().parents[2] / "skills" / "capabilities"
+    abstract_path = skills_root / name / "scripts" / "abstract.py"
+    if not abstract_path.exists():
+        return None
+    try:
+        spec = importlib.util.spec_from_file_location(f"capevolve_cap_{name}_{tag}",
+                                                     abstract_path)
+        mod = importlib.util.module_from_spec(spec)
+        spec.loader.exec_module(mod)  # type: ignore[union-attr]
+        return mod
+    except Exception:  # noqa: BLE001 — a capability we can't import contributes nothing
+        return None
+
+
+class _ValidationReport:
+    """The {problems, warnings} a capability's ``validate()`` reported for a candidate."""
+
+    def __init__(self, problems, warnings, by_capability):
+        self.problems = list(problems)
+        self.warnings = list(warnings)
+        self.by_capability = dict(by_capability)
+
+    @property
+    def reason(self) -> str:
+        shown = "; ".join(self.problems[:5])
+        more = f" (+{len(self.problems) - 5} more)" if len(self.problems) > 5 else ""
+        return f"candidate is not a valid {'/'.join(self.by_capability)}: {shown}{more}"
+
+    def to_dict(self) -> dict:
+        return {"problems": self.problems, "warnings": self.warnings,
+                "by_capability": self.by_capability}
+
+
+def _capability_validate(capabilities, cand_dir: Path,
+                         parent_dir: Path | None = None) -> _ValidationReport | None:
+    """Validate a candidate against each capability's OWN rules (generic, per-capability).
+
+    Calls ``validate(cand_dir)`` on every ``skills/capabilities/<name>/scripts/abstract.py``
+    that defines it. Returns None when no capability offered a usable signal, so
+    callers behave exactly as before for capabilities without the handler.
+
+    Problems the PARENT already had are dropped: a pre-existing violation (a seed
+    skill whose body is already over budget, say) is not something one edit can be
+    blamed for, and voiding every step over it would wedge the run. What is
+    reported is what this edit INTRODUCED. Numbers are normalized when comparing so
+    "612 lines" → "615 lines" still counts as pre-existing.
+    """
+    caps = [c for c in (capabilities or []) if c]
+    if not caps:
+        return None
+
+    def _run(name, mod, d):
+        try:
+            v = mod.validate(Path(d))
+        except Exception:  # noqa: BLE001 — never let a handler abort the loop
+            return None
+        return v if isinstance(v, dict) else None
+
+    problems: list[str] = []
+    warnings: list[str] = []
+    by_cap: dict[str, dict] = {}
+    got_signal = False
+    for name in caps:
+        mod = _load_capability_abstract(name, "validate")
+        if mod is None or not hasattr(mod, "validate"):
+            continue
+        v = _run(name, mod, cand_dir)
+        if v is None:
+            continue
+        got_signal = True
+        pre = set()
+        if parent_dir is not None:
+            pv = _run(name, mod, parent_dir)
+            if pv:
+                pre = {_norm_problem(p) for p in pv.get("problems", [])}
+        new = [p for p in v.get("problems", []) if _norm_problem(p) not in pre]
+        problems += [f"{name}: {p}" for p in new]
+        warnings += [f"{name}: {w}" for w in v.get("warnings", [])]
+        by_cap[name] = {"ok": bool(v.get("ok")), "problems": new,
+                        "warnings": v.get("warnings", []),
+                        "pre_existing": sorted(pre)}
+    if not got_signal:
+        return None
+    return _ValidationReport(problems, warnings, by_cap)
+
+
+def _norm_problem(text: str) -> str:
+    return re.sub(r"\d+", "N", str(text))
 
 
 def _capability_is_empty(capabilities, cand_dir: Path) -> bool | None:
--- a/core/cap_evolve/harness.py
+++ b/core/cap_evolve/harness.py
@@ -1205,7 +1205,7 @@ def _inject_optimizer_context(adapter, run_dir: RunDir, workdir: Path, *, split:
     # 2) capability skills as local guidance
     caps = [c for c in (capabilities or []) if c]
     if caps:
-        skills_root = Path(__file__).resolve().parents[2] / "skills" / "capabilities"
+        skills_root = _capabilities_root()
         for c in caps:
             src = skills_root / c
             if not src.is_dir():
@@ -1539,7 +1539,8 @@ def run_step(
     # before any rollout is paid for) rather than left to prose the optimizer may skip.
     # An invalid artifact's score measures a broken candidate, not the edit, so the
     # step is INDECISIVE — same discipline as the tamper path.
-    validation = _capability_validate(capabilities, workdir, parent_dir=parent_dir)
+    validation = _capability_validate(capabilities, workdir, parent_dir=parent_dir,
+                                      run_dir=run_dir)
     if validation is not None:
         if validation.warnings:
             run_dir.log_event("capability_validation_warnings", candidate=cid,
@@ -1795,7 +1796,7 @@ def _capability_brief(capabilities) -> str:
     caps = [c for c in (capabilities or []) if c]
     if not caps:
         return ""
-    skills_root = Path(__file__).resolve().parents[2] / "skills" / "capabilities"
+    skills_root = _capabilities_root()
     lines = ["## What you are editing (the allowed edit space)",
              "The capability under optimization is composed of these editable artifact(s). "
              "Use the FULL edit space below — do not limit yourself to trivial wording tweaks."]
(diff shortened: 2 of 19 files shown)


OsherElhadad:
## FIX-FIRST addressed — `8071f23e`

The blocker was right, and the shape of it mattered: 170 lines of honesty-adjacent gating
with **silent-off** failure modes and no core tests is finding F2's own signature inside
the fix for F2. Both halves are now closed.

### 1. Core tests (`core/tests/test_capability_validation_wiring.py`, 5 tests)

Styled on `test_integrity_wiring.py::test_tampered_candidate_is_indecisive_and_costs_nothing`,
which is the exact analogue for the tamper path.

| test | asserts |
|---|---|
| `test_invalid_candidate_is_indecisive_and_costs_nothing` | `candidate_val is None`, `decision["indecisive"] is True`, `accepted is False`, the problem text reaches `step["validation"]`; `best_id` and `spent.stall` unchanged; iteration still counted; **`adapter.rollouts` unchanged and no `rollouts/val/<cid>*` files**; `capability_invalid` and `step_indecisive` each logged exactly once |
| `test_valid_candidate_under_the_same_wiring_still_accepts` | same wiring, clean candidate → measured, accepted, `best_id` moves, no `capability_invalid` event |
| `test_pre_existing_problem_does_not_void_the_step` | a seed whose body is already >500 lines really is invalid, yet a good edit on top is still **measured** (`indecisive is False`) — the parent-subtraction case you flagged as optional |
| `test_no_capabilities_means_no_validation` | back-compat: the same broken edit with no `capabilities` is measured as before, no `validation` key, no event |
| `test_unloadable_capability_is_reported_not_silently_ignored` | a bogus capability name emits `capability_validate_unavailable` with the reason, and does **not** gate |

The "good edit" is read from `examples/toy_skill/mock_script.json`, so the test cannot
drift from the edit the shipped example claims works.

### 2. The two silent-off paths are now loud (not deferred to an issue)

- `_load_capability_abstract` returns `(module, why)` instead of a bare `None`, and the hook
  logs **`capability_validate_unavailable`** with the reason for: no `abstract.py` at the
  resolved path, `abstract.py` with no `validate()`, `validate()` raising, `validate()`
  returning a non-dict, or an import failure (`type: message`). Enforcement can no longer
  switch itself off unobserved.
- `_capabilities_root()` honors **`CAPEVOLVE_SKILLS_DIR`** before the repo-relative fallback,
  and all **four** sites in `harness.py` that hand-rolled
  `Path(__file__).resolve().parents[2] / "skills" / "capabilities"` now call it — the same
  silent-off class affected `_capability_is_empty`, `_capability_brief` and
  `_inject_optimizer_context`, so fixing only my caller would have left the siblings broken.
  (`_capability_is_empty` also routes through the shared loader now.)

### 3. Re-run, pinned

```
794 passed, 12 skipped in 113.87s        # 789 on clean main + the 5 new

$ python skills/capabilities/skill-package/scripts/check.py
{"skill": "skill-package", "ok": true, "problems": [], "notes": [... 6 ...]}
```

`examples/toy_skill` mock run re-verified after the core change: `best_id: cand_0001`,
`test_reward: 1.0`, `test_delta: +0.666667`. Real-model run not redone, per your note.

OsherElhadad:
Superseded by #392 — rebased onto current main (post #386/#346/#370/#387/#388). #386 moved `capabilities` off `run_step`'s signature into `OptimizerContext`, so the validation hook now reads `ctx.capabilities`; #346's 4-arg `RejectedMemory.add` was adopted. Notably `_capabilities_root()` needed **five** sites rather than four — main grew a `parents[2]` skills path in `_inject_native_skills` after this branch forked, which would otherwise have kept ignoring `CAPEVOLVE_SKILLS_DIR`. Wiring tests re-verified RED, and the indecisive-on-invalid path re-proven to cost zero rollouts. Closing in favour of #392.

Issue #371, opened 2026-08-22
skillopt's mini-batch focus block is always empty: train ids filtered against val rows (disjoint splits)


OsherElhadad:
## Summary

`skillopt` builds its per-step mini-batch from **train** ids and then passes them to `harness._focus_instructions`, which filters the parent candidate's **val** rows. Train and val are disjoint by construction, so the intersection is always empty and **every mini-batch step renders an empty focus block** — no failing-task list, no failure index, no protect-the-passing-ids block.

SkillOpt's headline mechanism is "epochs × mini-batches with a decaying textual learning rate". The decay and the epoch structure work. The mini-batch *content* does not reach the optimizer. So the thing being annealed is an instruction block that is empty.

This is the same defect that PR #370 just fixed for `hill-climb`'s `cyclic` and `hardest-first` focus modes; `skillopt` is the third instance and was left untouched there because it is a different module.

## Evidence

The mini-batch ids come from train (`core/cap_evolve/skillopt.py:230`):

```python
train_ids = list(run_dir.read_splits().train)
```

They are then handed to `_focus_instructions` together with the **val** result (`skillopt.py:291`):

```python
instructions = harness._focus_instructions(current_val, minibatch_ids, label)
```

`_focus_instructions` filters the passed result's per-task rows by those ids — and the two id sets cannot overlap, because the splits are disjoint slices of one ordered id list (`core/cap_evolve/splits.py:117-119`):

```python
train = ids[:n_tr]
val   = ids[n_tr:n_tr + n_va]
test  = ids[n_tr + n_va:n_tr + n_va + n_te]
```

`train ∩ val = ∅`, so the filter yields nothing every time. The rendered label still reads `(mini-batch of N train tasks, L=…)`, which makes the failure look like it is working — the count is real, the content is not.

## Why it went unnoticed

The same reason PR #370 gives for the `hill-climb` instance: **the `mock` optimizer never reads the prompt.** Every deterministic test and every zero-API example run passes with an empty focus block, because nothing downstream asserts on the rendered instructions. The bug is only observable in a paid run, by its absence — the optimizer simply proposes as if it had no per-task signal, which is indistinguishable from a weak proposal.

That is the important lesson here and it generalizes: assertions on *rendered prompt content* are the only thing that catches this class. PR #370 added a parametrized e2e test asserting on the rendered prompt for hill-climb; `skillopt` needs the sibling.

## Fix

Mirror PR #370's fix:

[...]

Issue #352, opened 2026-08-22
Action policy fails open: loader reads <capability_dir>/policy.json but all docs say inputs/policy.json, so the safety boundary is silently ignored


OsherElhadad:
## Summary

The action policy that decides which tool edits an optimizer is allowed to make is read from `<capability_dir>/policy.json`, but the code's own docstrings and every piece of documentation tell the author to write `inputs/policy.json`. A policy written where the docs say is **silently ignored**, and the capability falls back to its default policy.

For the `tools` capability that default is the **full** (most permissive) policy. So the failure mode is: someone deliberately locks off risky edit kinds, the lock is never loaded, no warning is emitted, and the optimizer is free to rewrite tool code and schemas. A safety boundary that fails open without saying so is worse than not having one, because the operator believes it is in force.

## Evidence

The reader — note the path has no `inputs/`:

```
core/cap_evolve/tool_surface.py:36:    f = Path(capability_dir) / "policy.json"
```

Its own docstrings, two lines above, contradict it:

```
core/cap_evolve/tool_surface.py:15:  set. The effective policy is ``inputs/policy.json`` if present, else the
core/cap_evolve/tool_surface.py:35:  """``inputs/policy.json`` if present (it overrides), else the capability default."""
```

And every author-facing surface repeats the wrong path — these are exactly the files an optimizer is given as guidance:

```
skills/capabilities/tools/SKILL.md:451       Lock any of these off via `inputs/policy.json`. …
skills/capabilities/tools/SKILL.md:713       `inputs/policy.json` is the safety boundary between …
skills/capabilities/tools/references/concepts.md:155   `inputs/policy.json` lists the allowed edit kinds. …
skills/capabilities/tools/references/pitfalls.md:127    …widen `inputs/policy.json` deliberately…
skills/capabilities/tools/references/examples.md:347    With `inputs/policy.json = {"allow": [...]}`, the …
skills/capabilities/mcp-tool/SKILL.md:88     widen `inputs/policy.json` deliberately and document why.
```

Seven author-facing mentions of a path the loader never looks at. `SKILL.md:713` calls it "the safety boundary", which is precisely the thing that is not working.

[...]

PR #394, opened 2026-08-22
postmerge: re-derive the skill-vs-code claims #386 invalidated, and land #368 under its token bar

OsherElhadad:
Post-merge review of four skill PRs merged without independent review — **#368** agent-optimize, **#369** gepa, **#380** skillopt, **#373** mcp-tool — plus the fixes for everything real that it found. Full write-up in `.review/postmerge-algos.md`.

The headline: **#369 and #380 were honest when written and have decayed since.** Both re-derived their SKILL.md against the code, then `#386` (the shared `OptimizerContext`) and `#350` restructured `gepa.py`/`skillopt.py` underneath them. Nothing noticed, because a hand-derived skill-vs-code table has no guard. That is the failure mode this whole effort was fixing, reappearing not through carelessness but through absence of a mechanism.

## Re-derived divergence counts

| | claimed | re-derived on current `main` |
|---|---|---|
| **#369 gepa** | 14 divergences, 4 flagged as gaps | **2 of the 4 flagged gaps are no longer real**; 3 live divergences remain (1 of them a wrong budget ceiling), + 1 in `references/concepts.md` |
| **#380 skillopt** | 10 divergences, 5 flagged as gaps | **all 5 gaps still real** (one worse than documented), but **11 of 18 `file:line` citations now point at unrelated code** |

### gepa

Removed two gaps that are fixed:

- Candidate snapshots **do** pass `_SNAPSHOT_IGNORE` at both call sites (`gepa.py:682`, `:846`) since #350, and `NON_CAPABILITY_FILES`/`DIRS` **do** exclude `CLAUDE.md`/`AGENTS.md`/`.claude`/`guidance`/`trajectories`/`prior_iterations` since #386 (`types.py:26-43`) — so round-robin *cannot* burn an iteration on a dotfile. `references/concepts.md:100-105` asserted the exact opposite of what #386 deliberately changed.
- Optimizer-context parity with hill-climb (#109) is fixed: `gepa_loop` takes the shared `OptimizerContext` and calls the same `inject()`/`instructions()` as its siblings. Rewritten down to the half that *is* still true — `JOURNAL.md` never accumulates in gepa, because `_reconcile_journal` is called from `harness.run_step` only (`harness.py:1578`) and gepa runs the optimizer itself.

Fixed a budget number that could cost real money: the `--max-metric-calls` overshoot bound said "`2·mb + |val|·n-trials`, one more minibatch on a merge iteration". A merge fires *inside* an accepting iteration and evaluates the merge **and both parents** on a **freshly sampled** minibatch, then pays a second full val — worst case `5·mb + 2·|val|·n_trials`.

Everything else in gepa's body verified accurate: the frequency-weighted per-instance sampling vs the strict `pareto_frontier` (two genuinely different sets), the `sum(child) > sum(parent)` local gate, the `reward < 1.0` threshold, the ~800-char/12-task caps, `--max-merges` counting attempts-that-built (a skip is free), the INDECISIVE tamper path, and all nine hyperparameter defaults. Gap 1 (no task input in the reflective dataset, despite the docstring at `gepa.py:222` claiming one) and gap 5 (empty train ⇒ silent **val** fallback at `:522`) are both still live.

### skillopt

All five gaps reproduce. But every citation that makes them checkable had drifted 7-9 lines: `skillopt.py:230`, cited for the train-ids read, is now a docstring line. All 11 refreshed, plus a standing note that a citation which stops matching means the section needs re-deriving. Also dropped the stale "*both land with PR #370; until then read `harness.run_step` itself*" — #370 landed as `ae71aa87`.

One gap was **understated**. `applied_changes` is not merely uninformative; its ignore list matches seven `.md` *basenames* while `run_step` injects `guidance/`, `trajectories/`, `prior_iterations/*/diff.patch` that `_SNAPSHOT_IGNORE` keeps *out* of the parent snapshot. Measured zero-API on `examples/toy_calc` at `L=4`:

```
epoch 1 step 1  requested_edits=4  applied_changes=10  accept=true
epoch 2 step 1  requested_edits=4  applied_changes=10  accept=false
```

Step 1's 10 = 7 injected `guidance/*` + 2 `trajectories/*` + **1** real edit. Step 2 logged 10 again with `prompt.txt` **byte-identical to its parent**. The metric has no zero, so it cannot detect the one thing it exists to detect: an optimizer that made no edit.

One imprecision corrected: gap 1 said the step renders "no task ids". The focus summary and failure index are empty, but `_passing_block` is built from the **whole** val split since #370 (`harness.py:2013`), so protect-these-ids *is* populated — and it is the only per-task content a step gets.

## #368's token budget

Measured with `tiktoken` `cl100k_base`, body only (the `description` is separately-budgeted always-in-context metadata, per `.review/skill-duplication-analysis.md`):

| | lines | body tokens |
|---|--:|--:|
| pre-#368 | 922 | — |
| as merged by #368 | 337 | **5,580** |
| this PR | 311 | **5,000** |

#368 reported ~5,444; the real figure was **5,580**, 11.6% over the ≤5,000 bar, flagged rather than fixed. **It was not earning its place, so it is cut** — −580 tokens, no rule deleted. What moved was restatement of its own references or of `references/algorithm.md`: the fan-out invariants' rationale (`algorithm.md` § *Parallelism* states all four in full), the parallelism enumeration, the helper flags `per-task-fanout.md` carries verbatim (`--supersedes`, `--floor`, `--canary-auto`, `--traces`, `dropped_additions`), the `measure.py` refusals' reasoning. Plus one genuine triplication: "a candidate inside the null band is not evidence" appeared three times in one body.

Two counting bugs #368 left behind, both fixed: "state these **five** invariants" over a four-item list (it had merged items 4+5), and "**Two** rules decide whether a round's number means anything" over three.

**One real instruction loss, restored.** SKILL.md names "the sign test for effects below the floor" as living in `references/measured-lessons.md`. #368 moved the section and dropped the rule, so the pointer led nowhere. Pre-registered directional predictions + a sign test (9/10 positive, `p = 0.0107`, where no individual z reached 2) are back where the pointer promises them, with the two conditions that make it a test rather than a story.

`core/tests/test_agent_optimize_skill.py` **passes unmodified** — and it earned its keep mid-review: it pins the literal term `re-gate`, which one of my compressions had paraphrased away. `scripts/check.py` caught a second (`no-regression`). Both restored. Worth naming rather than quietly fixing.

## #373's guard, verified empirically

Reproduced against the shipped `tool_surface` with mcp-tool's real policy:

```
(1) params edit carrying properties/type/required
    report: {"changed": ["params:kb_search"], "refused": []}
    parameters now: {"type": "array", "properties": {...}, "required": ["injected"]}
(2) add edit carrying a code key
    report: {"changed": ["add:srv"], "refused": []}   added keys: ['code','description','name','parameters']
(3) validate() → {"ok": true, "problems": []}         references load_policy: False
(4) only inputs/policy.json  → ignored
    capability_dir/policy.json → read
```

All four claims hold, and the skill states them plainly under "**The policy checks the edit's LABEL, not its effect**" rather than implying enforcement. **No change needed to the skill.**

But #373's policy-path correction had not reached the code that *defines* the answer: `tool_surface.py`'s module docstring and `load_policy`'s own docstring both still told the reader the loader reads `inputs/policy.json`, which it ignores. Fixed here (core, not a skill).

**No skill in scope names the wrong path.** `mcp-tool/SKILL.md:58-61` mentions `inputs/policy.json` only to rule it out.

## Instruction loss

`git show` on each merge commit, removed lines diffed against the current body + every reference, then a mechanical audit of bolded spans and backticked identifiers. Beyond #368's sign test: **#369 no loss**, **#373 no loss** (the annotations table, human-in-the-loop, and execution-vs-protocol errors all landed in `references/concepts.md` §4). Ownership compliance is good across all four — nobody restates the honesty invariants, the agent-mode loop, the failure taxonomy or the reflective-dataset format. Two pointer shortfalls fixed: gepa pointed at `hill-climb/SKILL.md` rather than its `references/run-step.md`, and skillopt's pointer was gated on an already-landed PR.

## The test convention this adds

`core/tests/test_skill_code_claims.py` — 7 tests, 3 **failing-first** (proof: `/tmp/capreview-postmerge-algos-failingfirst.py`). The point is the convention, not the coverage:

> **A skill that documents a code defect needs a test that fails when the defect is FIXED.**

Otherwise the doc silently becomes a lie the moment someone does the right thing — which is exactly what #386 did to #369, twice. `test_skillopt_minibatch_focus_is_still_empty_by_construction` fails the day #371 lands, and its message names the paragraph to delete.

## Verification

- `PYTHONPATH="$PWD/core:$PWD/dashboard/backend" python -m pytest core/tests -q` → **812 passed, 12 skipped** (805 baseline + 7 new). Always pinned — #349.
- All four skills' `scripts/check.py` → `ok: true, problems: []`.
- Zero-API runs on `examples/toy_calc` with the mock optimizer, both reaching a gate-accepted candidate and a sealed test score:

  | | gepa | skillopt |
  |---|---|---|
  | best_id | `gepa_0001` | `so_e01s01` |
  | accepts / val | 1 / 1.0 | 1 / 1.0 |
  | **sealed test** | **1.0** (baseline 0.0) | **1.0** (baseline 0.0) |

  **Honest caveat:** both fired `gate_warning` — "paired SE is 0 (likely n_trials=1) … falling back to STRICT (accept any delta>0)". Expected on a deterministic `n_trials=1` example and reported honestly by the engine, but these are STRICT-fallback accepts. They prove the plumbing and the seal, not the gate's discrimination.

## What other owners should pick up

Not touched, per the ownership contract:

- **`tools`** — `references/{examples.md:366, pitfalls.md:127, concepts.md:114}` still say the effective policy is `inputs/policy.json`. Child of #352; the core docstrings are fixed here, so these three are the last wrong surfaces.
- **`phases/gate`** — #380 deleted a *"Pitfall: small val sets"* whose `--n-trials` half survives in skillopt's body but whose other half does not live anywhere: *use a **graded** reward so the paired significance test has signal.* `gate` owns the significance bar; it belongs there.

### DIFF
--- a/.review/postmerge-algos.md
+++ b/.review/postmerge-algos.md
@@ -1,0 +1,339 @@
+# Post-merge review — four merged skill PRs (#368 agent-optimize · #369 gepa · #380 skillopt · #373 mcp-tool)
+
+Reviewed against `main` at `af7cf9d6`. None of the four was independently reviewed before merge.
+`main` baseline confirmed green at **805 passed, 12 skipped**
+(`PYTHONPATH="$PWD/core:$PWD/dashboard/backend" python -m pytest core/tests -q`).
+
+Merge order matters for reading this: `#350` (1775cddd, snapshot ignore) landed **before** #369, and
+`#386` (cfcad862, the shared `OptimizerContext` assembler) landed **after** all four. #386 is what
+made three of #369's and eleven of #380's statements wrong again.
+
+---
+
+## Skill-vs-code re-derivation
+
+### gepa — #369 claimed 14 divergences (10 fixed in text, 4 flagged). Re-derived: **3 live divergences**, and 2 of the 4 flagged gaps are no longer real.
+
+Body claims (all verified against `core/cap_evolve/gepa.py`, `selection.py`, `cache.py`,
+`skills/algorithms/gepa/scripts/run.py`):
+
+| claim | code reality | file:line | still accurate? |
+|---|---|---|---|
+| parent = frequency-weighted sample over per-instance (co-)winners, seeded | `rng.choices(pool, weights=counts)` over candidates with ≥1 win | `selection.py:180-192` | yes |
+| strict `pareto_frontier` is a *different*, narrower set used by merge + `frontier_size` | two distinct functions; merge calls `pareto_frontier`, selection calls `_instance_win_counts` | `selection.py:114`, `gepa.py:795`, `:721` | yes |
+| local gate is `sum(child) > sum(parent)` on the **same** minibatch, `2·mb` rollouts, eval-cached | exactly that | `gepa.py:639-643` | yes |
+| reflection reads **train** traces only | `_sample_minibatch(train_ids…)`, `ctx.inject(split="train")` | `gepa.py:559`, `:585` | yes |
+| "failing" = hard threshold `reward < 1.0`; header reads `N/N pass` | `< 1.0`; header is `{len(passing)}/{len(per)}` | `gepa.py:228`, `:236-237` | yes |
+| ~800-char truncation, at most 12 tasks | `[:800]` ×3, `actionable[:12]` | `gepa.py:251-259` | yes |
+| merge: two strict-frontier dominators with a common ancestor both beat, component-wise | `_find_merge_pair` + `_build_merge`, `origin` per component | `gepa.py:795-804` | yes |
+| `--max-merges` counts attempts that BUILT a candidate; a skip is free | `merges_done += 1` only when `merge_step is not None` | `gepa.py:706-708` | yes |
+| protected-path tamper ⇒ INDECISIVE, no reward, stall untouched | `update_spent(iterations=1, accepted=None)`, `candidate_val: None` | `gepa.py:626-637` | yes |
+| every hyperparameter default (`0/50/4/1/2/3`, `--workers 1`, `--store git`) | matches | `scripts/run.py:33-56` | yes |
+| **`--max-metric-calls` overshoot bound is `2·mb + \|val\|·n-trials`, "one more minibatch on a merge iteration"** | a merge fires *inside* an accepting iteration and evaluates the merge **and both parents** on a **freshly sampled** minibatch, then pays a second full val ⇒ worst case `5·mb + 2·\|val\|·n-trials` | `gepa.py:698-704`, `:813-819`, `:833` | **NO — understates a budget ceiling by ~2.5×** |
+
+`## Known gaps` (the interesting part):
+
+| flagged gap | code reality | file:line | still accurate? |
+|---|---|---|---|
+| reflective dataset carries no task **input**, though `gepa.py`'s docstring claims it does | docstring says "contributes its input"; the writer emits only `Agent output` / `Trajectory` / `Feedback` | `gepa.py:222` vs `:253-259` | **yes** — verified in a real `REFLECTION.md` (see Evidence) |
+| on an eval-cache hit only `{reward, feedback}` are stored, so `Agent output:` is empty | `EvalCache.put` persists two keys; a hit builds `raw={"cached": True}` | `cache.py:88-90`, `gepa.py:161-167` | yes |
+| **candidate snapshots keep the loop's scratch because the snapshot call omits the ignore list; and the dotfiles are not excluded from the component list** | `run_dir.snapshot(..., ignore=_SNAPSHOT_IGNORE)` at BOTH call sites (#350), and `NON_CAPABILITY_FILES`/`DIRS` now cover `CLAUDE.md`/`AGENTS.md`/`GEMINI.md`/`.claude`/`guidance`/`trajectories`/`prior_iterations` (#386) | `gepa.py:682`, `:846`; `types.py:26-43` | **NO — both halves fixed. Deleted.** |
+| iterations charged with no `step` event emitted | gepa logs 11 event kinds, none named `step`; `log_event("step")` exists only in `harness.py:1383`/`:1563` | `gepa.py` (grep) | yes |
+| **optimizer context reaching GEPA has been narrower than hill-climb's (#109)** | `gepa_loop` takes the shared `OptimizerContext` and calls the same `inject()`/`instructions()` as its siblings | `gepa.py:478`, `:501`, `:585-595` | **NO — fixed by #386. Rewritten to the live half (JOURNAL non-accumulation).** |
+| `JOURNAL.md` injected but never accumulates in gepa | `_reconcile_journal` is called from `harness.run_step` only, which gepa does not use | `harness.py:1578` | yes |
+| empty `splits.train` ⇒ silent fallback to **val** ids | `list(...train) or list(...val)` | `gepa.py:522` | yes |
+
+Plus one in the reference: `gepa/references/concepts.md:100-105` asserted the component exclusion list
+"does *not* cover the optimizer-agent dotfiles". False since #386 — `types.py:35-39` was changed
+*specifically* so an injected read-context dir cannot appear to gepa as an editable component. **Fixed.**
+
+### skillopt — #380 claimed 10 divergences (5 fixed, 5 flagged). Re-derived: **all 5 flagged gaps still real**, but **12 live divergences of a different kind** — the citations that make them checkable.
+
+| claim | code reality | file:line | still accurate? |
+|---|---|---|---|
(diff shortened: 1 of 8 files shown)


OsherElhadad:
## Rebased onto `c8bc2292` — re-derived, not reused

`#387`, `#388`, `#391`, `#393`, `#396` all landed after I branched. Re-derived from scratch rather than reusing my earlier numbers, per the instruction — and that was the right call, because **three of my own conclusions changed and one of my own tests caught a mistake I had made.** Recorded here rather than quietly corrected, since the finding of this whole review is that a hand-derived skill-vs-code table decays.

### gepa's Known gaps: 5 bullets → 2

- **`#387` fixed the hollow eval-cache reflection** (`cache.py:91-99` now persists `output`/`trace`/`errored`; `gepa.py:162-175` replays them on a hit).
- **`#396` fixed BOTH the missing `step` record and the non-accumulating `JOURNAL.md`** — `harness.record_iteration` (`harness.py:1037-1077`) is now the single place any algorithm ends an iteration, and gepa calls it at four sites.

Two gaps survive, both re-verified: the reflective dataset still carries **no task input** while its own docstring says it does (`gepa.py:230` vs `:263-267`), and an empty `splits.train` still falls back to **val** ids silently (`:530`).

Also corrected a body claim in the opposite direction: I had left `JOURNAL.md` out of the injected cross-iteration files. It was omitted *because* it never accumulated — which `#396` fixed — so all five files are now real, and the skill says so.

### Two `harness.py` citations rotted again

`#393` + `#396` grew the file ~266 lines: `harness.py:2004-2006`/`:2013` → `:2270-2272`/`:2279`. Re-derived **mechanically** this time rather than by hand — **23/23 skillopt + 3/3 gepa citations resolve**. All 19 `skillopt.py` citations held; only the cross-file ones moved, which is the predictable failure mode and a decent argument for citing symbols over line numbers next time.

### My own tripwire fired — right reason, wrong cause

`test_skillopt_minibatch_focus_is_still_empty_by_construction` went red. Not because `#371` was fixed, but because **`#391` reworded** the summary from `of 0 tasks` to `of 0 focused task(s) of N on val`. The gap is exactly as real.

Two fixes, both mine to own: the test now asserts the **property** (`0 solid / 0 flaky / 0 failing`, plus no val feedback leaking into a train-focused prompt) instead of a sentence, and skillopt's SKILL.md no longer quotes the pre-`#391` string. A tripwire that fires on rewording is a false alarm, which is worse than none.

### Preserved exactly

- **The gepa budget-ceiling correction**, re-verified against current `gepa.py`: `_try_merge` samples a **fresh** minibatch (`:831`) and evaluates the merge plus **both** parents (`:832`/`:834`/`:836`) before a second full val. `5·mb + 2·|val|·n_trials` stands.
- **`#396`'s AST guard** in `core/tests/test_iteration_record_parity.py` — untouched (`git diff` empty, 9 passed).
- **`#388`'s fix.** Both merge conflicts were `#388` removing the phantom `cap-evolve status` / `cap-evolve finalize` subcommands, colliding with my token trims. `#388`'s substance won; the phantom commands were **not** reintroduced (verified against the 14 subcommand-guard tests), and I kept its Stop-hook sentence, finding the tokens elsewhere.

### Added from `#396`'s semantics

An INDECISIVE gepa child **charges `spent.iterations`** while leaving the stall counter alone (`record_iteration(..., indecisive=True)`, `harness.py:1072`). The spend meter counts it because the rollouts were really spent; only the evidence meter does not. A reader budgeting a run needs that distinction and the skill did not have it.

### Evidence

```
PYTHONPATH="$PWD/core:$PWD/dashboard/backend" python -m pytest core/tests -q
  → 836 passed, 12 skipped          (main's 829 + my 7)

scripts/check.py    agent-optimize ok []   gepa ok []   skillopt ok []   mcp-tool ok []
body tokens         agent-optimize 5000 (311 lines) · gepa 2431 · skillopt 2276 · mcp-tool 1840
citations           23/23 skillopt + 3/3 gepa resolve
failing-first       all 3 re-verified RED against origin/main

bash examples/toy_calc/run.sh
  → best_id cand_0001 · baseline_val 0.0 · best_val 1.0
    test_reward 1.0 · test_delta 1.0 · finalized true · iterations 3
```

`toy_calc` carries the same caveat as before: at `n_trials=1` on a deterministic example, paired SE collapses to 0 and `gate.py` applies its documented STRICT fallback. It proves the plumbing, the split and the seal — not the gate's discrimination.

Updated in place (`MERGEABLE`); no superseding PR needed.

PR #398, opened 2026-08-23
feat(ci): let the benchmarks suite run agent-optimize, driven by a headless host

eranra:
## The gap

`agent-optimize` — the fully-agentic algorithm — could not be run from the benchmarks suite at all.

- [`run_suite.sh`](ci/benchmarks/lib/run_suite.sh) hardcoded `algorithm_skill: hill-climb`.
- Selecting the agentic algorithm by hand would have failed anyway: it [refuses a deterministic invocation](skills/algorithms/agent-optimize/scripts/run.py), and in agent mode `cap-evolve run` does check + baseline, [prints a handoff and returns](core/cap_evolve/cli.py) — leaving the loop to a conversational agent CI does not have. The handoff would have gone unread and the job would have exited 0 having optimized nothing.
- `workflow_dispatch` is at GitHub's hard 10-input cap, so an 11th `algorithm` input would have made the whole workflow invalid.

## Approach

The loop stays owned by the algorithm skill; the CLI invocation stays owned by `run-optimizer`. This adds only the host that connects them — no agent-driving logic lands in `ci/benchmarks/`, and the same host works locally and for any benchmark.

### `skills/algorithms/agent-optimize/scripts/host.py`

Renders the driver briefing from the spec + handoff (absolute paths, `num_trials`, `gate_k_se`, the free-text `stop_condition`, the four primitives every round must go through, and an instruction not to ask questions), then delegates the invocation to the existing [`optimizers/run-optimizer`](skills/optimizers/run-optimizer/scripts/run.py) runner — so registry rows, `{model}` substitution, budget-flag mapping, cost capture and the CLI-present hard fail are the ones the deterministic path already uses rather than a second copy. `--agent` takes any registry row.

Two failure modes it exists to prevent:

- **It raises `BASH_DEFAULT_TIMEOUT_MS` / `BASH_MAX_TIMEOUT_MS` to 4h.** At Claude Code's 10-minute default ceiling every full-val eval on a real benchmark is killed mid-flight, and a perfectly healthy run reads as a broken runner.
- **It guarantees the seal.** An agent that exhausts its turns leaves no `final.json` — unreportable *and* indistinguishable from a crash. The host then runs `measure.py` itself and labels the result `seal: host`, so it is never mistaken for the agent's own judgement that it was finished. An already-sealed run reports `seal: agent` rather than raising `TestSealError` and failing a run that is actually complete.

`--prompt-only` renders the briefing free (it lands at `$R/host/driver_prompt.md`, so a finished run records exactly what the host asked for); `--seal-only` seals a run a previous host left open.

### The `algorithm` dispatch input

`algorithm_focus` → `algorithm`: `hill-climb-all` (default, unchanged behaviour) | `hill-climb-cyclic` | `hill-climb-hardest-first` | `agent-optimize`. One token carries both algorithm and focus, because of the 10-input cap. `runmeta.json` now records it — a hill-climb number and an agent-optimize number are not a like-for-like comparison, and the history page should not present them as one.

`ALGORITHM_FOCUS` is still honoured when `ALGORITHM` is unset, so committed `overrides.env` files and hand-run invocations keep producing the same run; `ALGORITHM` wins when both are set, so a stale exported alias can never override a deliberate dispatch choice.

### A derived `stop_condition`

Agent mode is bounded by free-text prose, not a round schedule, so `run_suite.sh` derives it from the **same** dispatch inputs that bound a deterministic run: `iterations` → max rounds, `optimizer_usd_per_iter` × rounds → a whole-loop `$` ceiling (`0` stays unlimited, as everywhere else in the workflow), `gate_k_se`/`trials` → the gate. `optimizer_max_turns` becomes a whole-loop turn cap the same way, because the entire search is one agent process rather than one call per iteration.

It is emitted through `json.dumps`: interpolated raw, a paragraph containing `:` and `$` yields a spec that silently truncates at the first colon and hands the agent a stopping rule nobody wrote.

## A note on where `host.py` is documented

Deliberately **not** in `SKILL.md`. That file is the hosted agent's recurring per-trigger context, and it has ~150 characters of headroom under the repo's 5000-token body budget — while the agent never invokes the host; the host invokes the agent. Spending that budget on it would push out guidance the agent does need. Same convention as the skill's other non-loop scripts (`linkcheck.py`, `abstract.py`), which SKILL.md also does not name. It is documented in `docs/AGENT_ORCHESTRATION.md` and `ci/benchmarks/README.md`.

## Testing

- Shell blocks are **executed as real bash** rather than grepped (following `test_run_suite_split_hook.py`), since what breaks is shell behaviour: all four algorithm values, the legacy alias and its precedence, an unknown value aborting loudly, and the emitted YAML **parsed back** to prove the `stop_condition` survives round-trip intact.
- `host.py`'s offline guarantees are executed: briefing contents, unknown-agent refusal *before* any spend, the seal fallback and its idempotence, the raised Bash ceiling. The skill's own `check.py` now covers them too, so they cannot drift.
- One test pins the coupling that makes the CI report render at all: `commit.py` books decisions through `harness.record_iteration`, which writes the canonical `step` event `metrics.py` builds its timeline from. Without it an agent-mode run reports zero iterations and an empty timeline while still exiting 0.
- Full suite: **943 passed**. One pre-existing failure (`test_cli_surface.py::test_spec_defaults_relative_to_project_not_the_cwd`, a manifest lookup for `run-optimizer`) reproduces unchanged on `main` and is untouched by this branch.
- `actionlint` clean on the workflow; end-to-end agent-mode chain verified offline against the `toy_calc` example with the `mock` registry row (handoff → host → run-optimizer → seal fallback → `final.json`).

A tau2 smoke run with `algorithm: agent-optimize` is being dispatched against this branch to verify it end to end on real models.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -115,11 +115,11 @@ on:
         description: "Acceptance-gate strictness: accept a candidate iff improvement > k_se·SE (default 1.0; lower = more lenient, higher = stricter)"
         type: string
         default: "1.0"
-      algorithm_focus:
-        description: "Hill-climb focus schedule: all (whole train set each iteration) | cyclic (one task at a time) | hardest-first (lowest-scoring task first)"
+      algorithm:
+        description: "Optimization algorithm. hill-climb-* is the deterministic loop and its focus schedule: all (whole train set each iteration) | cyclic (one task at a time) | hardest-first (lowest-scoring task first). agent-optimize is the fully-agentic loop — run_suite.sh switches the run to orchestration_mode: agent and a headless agent drives the rounds itself, bounded by a stop_condition derived from `iterations`/`optimizer_*` rather than a fixed schedule."
         type: choice
-        default: all
-        options: [all, cyclic, hardest-first]
+        default: hill-climb-all
+        options: [hill-climb-all, hill-climb-cyclic, hill-climb-hardest-first, agent-optimize]
       # NB: workflow_dispatch allows at most 10 inputs and this list is full. SPLIT_SEED is
       # therefore driven by the repo variable `vars.BENCH_SPLIT_SEED` (see the job env) rather
       # than an 11th input, which would make this workflow invalid.
@@ -302,7 +302,11 @@ jobs:
       OPTIMIZER_USD_PER_ITER: ${{ github.event.inputs.optimizer_usd_per_iter || (matrix.tier == 'smoke' && '50' || '0') }}
       OPTIMIZER_MAX_TURNS: ${{ github.event.inputs.optimizer_max_turns || '80' }}
       GATE_K_SE: ${{ github.event.inputs.gate_k_se || '1.0' }}
-      ALGORITHM_FOCUS: ${{ github.event.inputs.algorithm_focus || 'all' }}
+      # Names the algorithm and (for hill-climb) its focus schedule in one token, because
+      # the 10-input cap above leaves no room for two. run_suite.sh splits it, and maps
+      # agent-optimize to orchestration_mode: agent + a derived stop_condition. A PR-label
+      # dispatch has no inputs at all, so the default keeps every existing leg unchanged.
+      ALGORITHM: ${{ github.event.inputs.algorithm || 'hill-climb-all' }}
       # Run seed. With a tier that ships a committed split_ids.json the partition is FIXED, so
       # this varies only the per-trial rollout seeding — which is what makes the independent
       # seeds a paper comparison needs (">=3 seeds") possible on one split. Set the repo
@@ -427,6 +431,7 @@ jobs:
             "agent_model": "${AGENT_MODEL:-aws/gpt-oss-120b}",
             "optimizer_model": "${OPTIMIZER_MODEL:-claude-opus-4-8}",
             "gate_k_se": ${GATE_K_SE:-1.0},
+            "algorithm": "${ALGORITHM:-hill-climb-all}",
             "warm_seed": $warm_json,
             "conclusion": "${{ job.status }}"
           }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,55 @@ All notable changes to cap-evolve are documented here. The format follows
 [0.1.0]: https://github.com/skillberry-ai/cap-evolve/releases/tag/v0.1.0
 
 ## [Unreleased]
+### Added
+- **The benchmarks suite can run `agent-optimize`.** Until now `run_suite.sh` hardcoded
+  `algorithm_skill: hill-climb`, so the fully-agentic algorithm was unreachable from CI — and
+  selecting it by hand would have failed anyway, because it refuses a deterministic invocation and
+  nothing in CI could pick up the agent-mode handoff. Three pieces close that:
+
+  - **`scripts/host.py` in the `agent-optimize` skill** — the headless host for agent mode. It
+    renders a driver briefing from the spec + handoff (absolute paths, `num_trials`, `gate_k_se`,
+    the free-text `stop_condition`, the four primitives every round must go through, and an
+    instruction not to ask questions) and delegates the CLI invocation to the existing
+    `optimizers/run-optimizer` runner, so registry rows, `{model}` substitution, budget-flag
+    mapping, cost capture and the CLI-present hard fail are the ones the deterministic path
+    already uses rather than a second copy. `--agent` takes any registry row; `--prompt-only`
+    renders the briefing free; `--seal-only` seals a run a previous host left open.
+
+    Two failure modes it exists to prevent. It raises `BASH_DEFAULT_TIMEOUT_MS` and
+    `BASH_MAX_TIMEOUT_MS` to 4h: at Claude Code's 10-minute default ceiling every full-val eval on
+    a real benchmark is killed mid-flight, and a perfectly healthy run reads as a broken runner.
+    And it **guarantees the seal** — an agent that exhausts its turns leaves no `final.json`, which
+    is both unreportable and indistinguishable from a crash, so the host runs `measure.py` itself
+    and labels the result `seal: host` so it is never mistaken for the agent's own judgement that
+    it was finished. An already-sealed run reports `seal: agent` instead of raising
+    `TestSealError`.
+
+  - **An `algorithm` dispatch input**, replacing `algorithm_focus`:
+    `hill-climb-all` (default, unchanged behaviour) | `hill-climb-cyclic` |
+    `hill-climb-hardest-first` | `agent-optimize`. One token carries both algorithm and focus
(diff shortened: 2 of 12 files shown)


eranra:
---

## Update — verified end to end on real models, and one measured defect fixed

**Run [32649063850](https://github.com/skillberry-ai/cap-evolve/actions/runs/32649063850) (tau2 / smoke, `algorithm=agent-optimize`): success.** The mechanism works — `runmeta.json` recorded `"algorithm": "agent-optimize"`, the host drove two full rounds, both were gated on full val and **rejected**, `best = seed`, and the run sealed test through `measure.py`. An honest null result, with the iteration timeline rendering and `assert_run.py --min-iterations 1` satisfied:

| phase | iter | candidate | accepted | reward |
|---|:--:|---|:--:|---|
| baseline | — | `seed` | — | 0.433 |
| iterate | 1 | `cand_1` | ❌ | 0.367 |
| iterate | 2 | `cand_2` | ❌ | 0.233 |
| finalize | — | `seed` | — | 0.267 |

### The defect that run exposed

Both candidates edited **only `policy/policy.md`**. `tools/tools.py` and `reference/data_model.py` were in the same candidate dir, writable, and never opened — so a capability whose tool code is most of its behaviour was optimized as if it were prompt-only.

Nothing was blocking it: `capabilities` selects which capability `validate()` runs and which guidance is surfaced, and does **not** gate writes. The agent had no reason to know the other files were fair game, because the briefing named *capabilities* rather than *files*.

This repo had already paid for that lesson once — `run_suite.sh`'s spreadsheetbench arm carries a "NAME THE ARTIFACTS" note because an optimizer handed two editable files reported "all in prompt.md" and left the second surface inert. That fix was benchmark-specific prose; this one is generic. `host.py` now enumerates the capability's real files from the materialized seed candidate (the same set the agent copies and edits) and states what a prompt-only round leaves undone:

```
## Your editable surface — ALL 3 of these files

- `policy/policy.md`
- `reference/data_model.py`
- `tools/tools.py`

Every one of them is in your candidate copy and every one is fair game … A round that
changes **only the obvious prompt file** leaves the rest of the agent's instruction and
behaviour surface exactly as it was, and that is the most common way a run produces
nothing: the fix that was needed lived in a file nobody opened.
```

Scaled to what is actually present, because overcorrecting is its own failure mode: a single-file capability is told it *is* the whole capability and not to look for a second surface. Telling an agent to "also edit the tool code" when there is none is how a prompt-only run once went and edited `adapter.py`. Both directions are tested.

### tau2's capability surface

`[system-prompt, tools]` is now documented in `run_suite.sh` as the **maximum that applies**, for every tier (smoke / full / integration all reach the same arm) — not a subset. The other two capability skills are deliberately absent: `mcp-tool` is for a toolset served by an external MCP server and forbids editing tool code (its own SKILL.md defers to `tools` when the agent owns them, which here it does), and `skill-package` optimizes a `SKILL.md` package this seed is not. Either would narrow the surface rather than widen it.

### Also in this update

The benchmarks history page now has an **Algorithm** column. The previous commit taught `runmeta.json` to record the algorithm precisely so a hill-climb number and an agent-optimize number are not read as like-for-like — then left it invisible on the page where numbers actually get compared. Records written before the field existed render `—`, never a back-filled `hill-climb-all`: the focus schedule was already dispatchable then, so an old run's algorithm is genuinely unrecoverable and labelling it would fabricate the provenance the column exists to establish.

Suite: **947 passed**, same single pre-existing failure. A second tau2 smoke run is dispatched to confirm the agent now exercises the full surface.

eranra:
## Update 2 — the real cause found: the hosted agent had no guidance for the tool surface

Run [32653934804](https://github.com/skillberry-ai/cap-evolve/actions/runs/32653934804) succeeded and **accepted** a candidate (val 0.333 → 0.467), but it still edited only `policy/policy.md`. So did the run before it — **4 of 4 candidates across two runs**, on a `[system-prompt, tools]` capability, with `tools/tools.py` writable and unopened in the same candidate dir.

**My previous fix was the wrong diagnosis.** I assumed the agent didn't know those files were editable and named them in the briefing. That run's own `driver_prompt.md` disproves it — it listed all three files under *"ALL 3 of these files … every one is fair game"*, and the agent still edited only prose. The file list was never the missing piece.

This skill's own edit-form table predicts exactly that: for a missing element, the form that works is "a structural REQUIRED slot", and the form that **makes it worse** is "a prose reminder mid-document". I wrote the prose reminder.

### The structural gap

`harness.OptimizerContext` exists so *"an algorithm cannot silently run on a thinner prompt than its siblings"*. It stages, per iteration:

- each declared capability's skill as `./guidance/<cap>/` **and** placed where the agent natively discovers skills (`.claude/skills/<cap>/`)
- the diagnose method
- `capability_sources`
- the agent's own features reference

Every deterministic algorithm calls `inject()`. **The host did not.** So the hosted agent had guidance for editing prose and *none at all* for editing tool code — and did what it had guidance for. `capabilities` gates which capability `validate()` runs, never what may be written, so nothing was blocking the edits; the guidance for that surface simply wasn't there.

`test_optimizer_context_parity.py` had already named this hole, in its own docstring:

> `agent-optimize` and `evograph` … declare none of the context flags and drive their own loops … an algorithm absent from `ALGORITHMS` is **NOT covered** — it can still run blind while this file stays green.

It ran blind. Verified on the real tau2 capability, the host now stages:

```
guidance/{system-prompt,tools,diagnose,optimizer}/
.claude/skills/{system-prompt,tools,diagnose}/     # native discovery for claude-code
```

Staging success is **reported**, not swallowed — in the host's JSON and as a `::warning::`. Silent-off is indistinguishable from working while quietly optimizing less surface, which is precisely how this survived two runs.

### Two consequences of reusing that path

- **The briefing points at the staged guidance**, per capability. Unread guidance is not guidance, and the surface with no guidance is the one an agent avoids by default.
- **`inject()` writes an always-on `CLAUDE.md`** whose first line is "read `./INSTRUCTIONS.md` FIRST" — true for the deterministic per-iteration optimizer, which has one. Agent mode passes its briefing as the prompt, so that pointer dangled at a missing file. The host now writes the briefing to `<workdir>/INSTRUCTIONS.md` too (same bytes, so the two can never disagree; the run-dir copy stays the audit record), and the briefing states that `LEDGER.md` / `RUNMAP.md` / `prior_iterations/` belong to the deterministic loop and are legitimately absent.

### On "all capabilities" for tau2

`[system-prompt, tools]` is the maximum that applies, now documented as such in `run_suite.sh` for every tier. `mcp-tool` would *narrow* the surface — it's for externally-served toolsets and forbids editing tool code, deferring to `tools` when the agent owns them, which here it does. `skill-package` optimizes a `SKILL.md` package this seed isn't.

Suite: **949 passed**, same single pre-existing failure. Skill `check.py` green, authoring lint green. A third tau2 smoke run is dispatched to confirm the agent now reaches the tool surface.

PR #399, opened 2026-08-24
fix(agent-optimize): give the hosted agent the full capability surface, and keep the host benchmark-agnostic

eranra:
Makes `agent-optimize` actually usable unattended. It was reachable from CI but running blind: the hosted agent got no capability guidance, no honest stop diagnosis, and no record of what it did — so three consecutive smoke runs each failed a different way and none of them could be diagnosed from their own artifacts.

Every change here is tied to a measured run ID rather than a hypothesis. **16 commits, 17 files, +2919/−87.**

## Core changes

**The hosted agent could not really edit anything but prose.** `harness.OptimizerContext` exists so that *"an algorithm cannot silently run on a thinner prompt than its siblings"* — it stages each declared capability's skill, the diagnose method, `capability_sources` and the agent's features reference. Every deterministic algorithm calls `inject()`. **The host did not** — so the agent had guidance for editing prose and none at all for editing tool code, and did what it had guidance for: 4 of 4 candidates across runs [32649063850](https://github.com/skillberry-ai/cap-evolve/actions/runs/32649063850) and [32653934804](https://github.com/skillberry-ai/cap-evolve/actions/runs/32653934804) touched only `policy/policy.md`, on a `[system-prompt, tools]` capability with `tools/tools.py` writable and unopened. Naming the files in the briefing had already been tried and had already failed — run 32653934804's own prompt said *"ALL 3 of these files … every one is fair game"*. `test_optimizer_context_parity.py` had named this hole in its own docstring: an algorithm absent from `ALGORITHMS` *"can still run blind while this file stays green."*

The host now stages `guidance/{system-prompt,tools,diagnose,optimizer}/` and `.claude/skills/{…}/`, via a new `OptimizerContext.from_spec()` plus four public prompt-block accessors (`capability_brief`, `reader_brief`, `empty_seed_brief`, `render_template`) — composed rather than calling `instructions()`, whose per-iteration contract ("fix this ONE candidate and STOP; the harness re-scores you") is false for an agent that owns the search, the evaluation and the gate. `specfile.resolve_instructions_file()` extracts the `optimizer_instructions_file` resolution rule that `cli.py` held inline; two copies of that rule is how #252 happened. **`harness.py` and `specfile.py` are purely additive** — `instructions()` and `inject()` are untouched. Staging success is reported and warned on, because silent-off is indistinguishable from working while quietly optimizing less surface, which is how this survived two runs. A *declared* capability with no matching skill package is now reported too (`guidance_missing` + a warning) — the partial case of the same defect.

**Honest measurement and honest reporting.** The host guarantees a seal, reports why the agent stopped (`stop_reason`, `agent_stop_reason`, `num_turns`, `rounds_booked`) and flags a run that stopped with rounds left, with **per-cause** advice — a turn-starved agent gets "raise the turn budget", one that sealed itself gets "unspent budget, not a dead loop". It puts the correct interpreter first on the agent's `PATH`: every candidate eval in run [32861747778](https://github.com/skillberry-ai/cap-evolve/actions/runs/32861747778) died `ModuleNotFoundError` because the adapter's deps live in one venv that CI's `PATH` never contains, while SKILL.md tells the agent to run bare `python`. `round.py` persists its gate table instead of leaving stdout the only copy, refuses a concurrency too coarse to resolve its own verdict (the agent set 100 *after* being told not to — a guard in code beats a third restatement in prose), separates `parent` from `gate_reference` and reports `parent_vs_gate_ref_drift`, emits one `evidence_bar` matched to how it gated, marks a verdict that flips across control replicates as `inconclusive`, and shows the drift-free comparison a parent-gated round already paid to measure. `commit.py` refuses `--reject-basis gate` when the gate actually **accepted** — run 32871360361's audit log claimed the gate had rejected its best candidate when the driver had overridden it — adding `driver_judgement` for honest overrides and recording `gate_verdict`/`overrode_gate` on every decision.

## CI changes

`run_suite.sh`: **(a)** drops the stop-on-two-consecutive-rejections clause, which fired exactly where the algorithm prescribes *escalation* — on run 32701056043 the agent rejected two prose candidates and quit before trying the code-level guard its own reject note had already named; **(b)** raises the agent-mode turn budget to `max(optimizer_max_turns, 150) × (rounds + 1)`, measured from run 32733635494, which consumed ~126 turns/round and died on `error_max_turns` having just evaluated the best candidate of the run and never reached the `commit.py` that would have booked it; **(c)** copies `$RUN_DIR/host/` into the uploaded artifact, gzipping the transcript — the artifact path is `$OUT/**` while the run dir reaches it only via `export_static`, which caps every file at 256 KiB and keeps the *first* chunk. The real transcript was 1.9 MB and the answer was in the last 12%.

`run-optimizer` gains an opt-in `--transcript`, and the registry a `transcript_flag` (`--output-format stream-json --verbose`, verified against Claude Code 2.1.241 before being trusted to a multi-hour run — its last line is the same result object, so cost and stop parsing are unchanged). **`cli.py` never passes `--transcript`, so the deterministic path's `--output-format json` is untouched.** Secret values are scrubbed before the file is written, since a transcript records tool results verbatim and the run dir is a published artifact.

`benchmarks.yml` makes the `iterations` input reachable on smoke. The old expression put `matrix.tier == 'smoke' && '3'` first and GitHub's `||` short-circuits, so a dispatch of `iterations=10` on smoke silently ran 3 — discoverable only by reading `ITERATIONS` in the job log after the budget had been spent. It now mirrors the pattern `NUM_TRIALS` already used, input default `""` included. Blank dispatch and PR-label runs are byte-identical to before (smoke 3, others 10).

## Verified end to end

Run [32871360361](https://github.com/skillberry-ai/cap-evolve/actions/runs/32871360361) — smoke tau2, 10 iterations, 10 trials:

| check | result |
|---|---|
| rounds booked, none abandoned | 4/4, `unbooked_rounds: []` |
| accept **and** reject both booked | 3 rejects + 1 accept |
| `best_id` moves | seed → cand4 |
| evals clean, concurrency ≤ 25 | 8 throughout |
| not turn-limited | 121 / 1650 |
| sealed exactly once, by the agent | `seal: agent` |
| transcript readable in the artifact | 552 KB gz / 1.9 MB |

**Result: test 0.40 → 0.53 (+0.13), sealed once.** Mechanism traced, not inferred: the agent found tasks 22 and 40 failing identically (it yields a confirmation summary, the user replies `Yes.###STOP###`, the episode ends before the DB write), wrote a turn-economy policy rule, and **task 40 went 0.10 → 1.00**. It also reasoned about *which surface* to edit — *"a code guard can't fix a 'the agent never called any tool' failure, so policy is the correct surface"* — which is the staged guidance doing its job.

## Testing

**995 passed, 2 skipped, 1 failed.** The failure (`test_cli_surface.py::test_spec_defaults_relative_to_project_not_the_cwd`, a `run-optimizer` manifest lookup) reproduces identically on a clean `origin/main` worktree — pre-existing and unrelated. ~1500 lines of new tests across 4 files, each pinning a specific measured failure. Skill `check.py`, authoring lint and `actionlint` green; `check.py` caught two defects in this PR's own changes, which is its job.

## Reviewer notes

- **Blast radius:** everything behavioural is gated behind `ALGORITHM=agent-optimize` / `ORCH_MODE=agent` / `[ -d $RUN_DIR/host ]`. `agent-optimize` is dispatch-only and not in `OPTIMIZER_CONTEXT_ALGORITHMS` (`{hill-climb, gepa, skillopt}`). `round.py`/`commit.py` have no callers outside the skill; `run-optimizer`'s diff has zero removed lines.
- **One behaviour change reaching non-agent runs:** in `cli.py`, a *missing default* `optimizer/INSTRUCTIONS.md` no longer warns — only a spec-**named** missing file does. Deliberate and documented, but it is one fewer diagnostic line for `hill-climb`/`gepa`/`skillopt`.
- **Not CI-validated:** commits `21e41c3a`…`183b6105` are reporting-layer changes made while run 32871360361 was in flight. They are covered by unit tests and by `check.py` executing `round.py`/`commit.py` against a real run dir in both gate modes, but have not run against a live agent. They add fields and read-only re-gates rather than changing decisions.

## Measurement finding worth recording separately

On tau2 smoke the **same seed bytes** scored **0.24 / 0.44 / 0.38** across three runs in one day (sd 0.103, range 0.20) — roughly **3.7× the `gate_k_se=1.0` acceptance bar**. Within one run, byte-identical control replicates measured two minutes apart read 0.32 and 0.20. The default `--gate-against parent` compares against a *stored* reward, so that drift lands inside every accept/reject; `--gate-against control` pairs against a concurrently-measured control and removes it. This PR makes the drift visible per round but **deliberately does not change the default gate mode** — doing so on one benchmark's evidence would be a guess about every other workload. Two follow-ups worth considering on their own merits: raising `gate_k_se` on this tier, and telling the agent to prefer control-mode gating.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -36,9 +36,9 @@ on:
         default: smoke
         options: [all, smoke, pilot, full]
       iterations:
-        description: "Optimizer iterations per task for the FULL tier (default 10; raise to give the optimizer more budget). Smoke always runs 3, regardless of this input."
+        description: "Optimizer iterations. LEAVE BLANK for the tier default (smoke 3, everything else 10). Set a value to override deliberately — including on smoke, where more rounds are how you watch a per-iteration curve rather than a single step."
         type: string
-        default: "10"
+        default: ""
       trials:
         description: "Trials per task per eval. LEAVE BLANK for the tier default (smoke 3, everything else 10). Set a value only to override deliberately — on a 5-task binary tier, 1 trial cannot separate signal from noise, which is why blank does not mean 1."
         type: string
@@ -275,10 +275,14 @@ jobs:
     env:
       ANTHROPIC_BASE_URL: ${{ secrets.ANTHROPIC_BASE_URL }}
       ANTHROPIC_AUTH_TOKEN: ${{ secrets.ANTHROPIC_AUTH_TOKEN }}
-      # Smoke always runs a fixed 3 iterations (cheap, fast regression) — decoupled
-      # from the `iterations` input, which only controls the (expensive) full tier
-      # (default 10 there — see the `iterations` input above).
-      ITERATIONS: ${{ matrix.tier == 'smoke' && '3' || github.event.inputs.iterations || '10' }}
+      # Smoke DEFAULTS to 3 iterations (cheap, fast regression) but an explicit dispatch value
+      # now wins, exactly as it does for NUM_TRIALS below and for the same reason: the input's
+      # own default is "" so "user asked for 10" is distinguishable from "user asked for
+      # nothing". Before this, `matrix.tier == 'smoke' && '3'` came FIRST and short-circuited,
+      # so a dispatch of iterations=5 on smoke was silently ignored and the operator learned it
+      # only by reading ITERATIONS in the job log. Smoke is the tier you iterate the algorithm
+      # on, so pinning it was the wrong knob to make unreachable.
+      ITERATIONS: ${{ github.event.inputs.iterations || (matrix.tier == 'smoke' && '3') || '10' }}
       # Smoke pins 3 trials, same as it pins iterations, and for a measurable reason.
       # Run 31183768756 was dispatched with trials=1 and re-scored the SAME seed capability at
       # 0.400 (baseline) then 0.200 (finalize) — pytest-dev__pytest-7432 flipped 1.000 -> 0.000
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,20 @@ All notable changes to cap-evolve are documented here. The format follows
 
 ## [Unreleased]
 ### Fixed
+- **The end_turn diagnosis accused a complete run of abandoning work.** Run 32871360361 booked 4
+  of 10 rounds, investigated a round-5 lever, judged the residual failure unfixable by the
+  surfaces it owned, **sealed test itself**, wrote its report and stopped at 121 of 1650 turns.
+  The host told the operator it had "stopped of its own accord … which is what a turn ending on
+  outstanding work looks like … a backgrounded job … cannot resume a non-interactive run" —
+  the exact defect the warning was written for, on a run that was complete and honest. Two facts
+  already in the payload disprove it: `seal == "agent"` (a loop that dies mid-turn leaves the
+  host to seal, as run 32814848187 did) and an empty `unbooked_rounds`. The diagnosis now
+  branches on both and reports unspent rounds as **under-use** — "it stopped when it ran out of
+  edits it trusted, not when it ran out of rounds" — while the foreground explanation is kept for
+  the case that actually produced it. The "a candidate may never have been committed" hedge no
+  longer fires when the backstop came back clean, since it sent readers hunting for a candidate
+  that provably did not exist.
+
 - **A parent-gated round discarded the drift-free comparison it had already paid to measure.**
   Run 32871360361 round 4 gated in `parent` mode: `cand4` at 0.53 against the seed's *stored* 0.38
   = +0.15, bar 0.11 (drift), so 1.4x — marginal. The same round's two concurrent controls both
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,19 @@ All notable changes to cap-evolve are documented here. The format follows
 
 ## [Unreleased]
 ### Fixed
+- **A parent-gated round discarded the drift-free comparison it had already paid to measure.**
+  Run 32871360361 round 4 gated in `parent` mode: `cand4` at 0.53 against the seed's *stored* 0.38
+  = +0.15, bar 0.11 (drift), so 1.4x — marginal. The same round's two concurrent controls both
+  read **exactly 0.27**, so the drift-free answer from the identical rollouts is **+0.26 against a
+  bar of 0.00**. The 0.11 belongs to *when* the seed was measured, not to `cand4`; parent-mode
+  gating understated the effect and inflated the bar simultaneously. Each candidate now also
+  carries `control_relative` — the same gate re-run against the concurrently-measured control,
+  costing no rollouts since the controls are already evaluated — so where the two comparisons
+  disagree, the difference is visibly drift rather than the edit. Reported rather than made the
+  default: changing the default gate mode on one benchmark's drift would be a guess about every
+  other workload, while an extra comparison is strictly more information and agrees with the
+  primary one wherever there is no drift.
+
 - **A reject that overrode the gate was recorded as the gate's own verdict.** `--reject-basis
   gate` is documented as "full-val paired gate ran", and run 32871360361 booked it for `cand2` —
   which `round_i1.json` recorded as `verdict: accept` at +0.19 against a concurrent control. So
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,36 @@ All notable changes to cap-evolve are documented here. The format follows
 
 ## [Unreleased]
 ### Fixed
+- **The hosted agent ended its turn to wait for a notification, and the process exited under
+  it.** With the turn budget raised to 600, run 32814848187 no longer ran out of turns — it
+  used 78 and stopped anyway, on `subtype: success` / `stop_reason: end_turn`, rc 0. It had
+  launched round 2's full-val gate in the background and ended its turn to await word back.
+  In a non-interactive run there is nothing to come back to: ending a turn ends the process
+  and orphans its children. The gate finished **14 minutes after the agent was gone**, wrote a
+  real verdict (`r2_comm_search` val 0.44 vs parent 0.58 → reject) that nobody read, and left
+  2 of 3 rounds unspent — while the orphaned evals were still hitting the runner as
+  `measure.py` measured the seal. Three changes, none of which restricts concurrency: the
+  briefing now states that the *main loop* runs in the foreground and that the turn which
+  launched work is the turn that collects it — fanning out subagents or a whole round's evals
+  stays encouraged, since delegating the work is not the same as detaching the wait; `round.py`
+  persists its gate table under `$R/work/` instead of leaving stdout the only copy, so an
+  abandoned round's verdict is evidence rather than a redirect the driver may have skipped; and
+  the host reports `unbooked_rounds` — candidates a round gated to a verdict that no
+  `commit.py` ever booked — which `spent.iterations` cannot distinguish from a round never
+  attempted. It reports rather than books: which decision a verdict deserves is the driver's
+  judgement, and booking an accept after `measure.py` had sealed against the old `best_id`
+  would turn a visible gap into a wrong headline number.
+
+- **`incomplete` gave turn-budget advice to an agent that was not turn-starved.** One message
+  served both stop causes and fit only the first. Run 32733635494 died on `error_max_turns`,
+  where "raise `optimizer_max_turns` or lower the round count" was exactly right; run
+  32814848187 stopped at 78 of 600 turns, and the same sentence pointed the operator at a knob
+  already 7× larger than what the agent used. The diagnosis now splits on the stop cause, and
+  reads the agent's own `stop_reason` alongside the harness's `subtype` — the discriminator is
+  the former (`end_turn`), while the latter says only `success`, which is why a voluntary stop
+  was indistinguishable from a clean finish. `--seal-only`, the path an operator reaches for on
+  a run that died mid-loop, reports abandoned rounds too.
+
 - **Agent-mode optimizer cost was reported as $0.00 on every run.** `run-optimizer` nests the
   figure under `cost.total_cost_usd`; the host read a flat `cost_usd`/`usd` and booked `0.0`.
   Smoke run 32733635494 spent **$8.37** (68,432 tokens) and recorded nothing — and because that
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,22 @@ All notable changes to cap-evolve are documented here. The format follows
 
 ## [Unreleased]
 ### Fixed
+- **A round's verdict could be decided by which control replicate happened to be the
+  reference.** On run 32871360361 round 3, two byte-identical control replicates measured two
+  minutes apart read **0.32 and 0.20** — a 0.12 gap. The gate reference was whichever carried the
+  round-scoped tag (0.20), so `cand3` at 0.37 scored +0.17 and **accepted**; against the other
+  replicate it is +0.05 and rejects. Nothing in the table said the verdict rested on that coin
+  flip. Each candidate is now re-gated against *every* control replicate — which costs no new
+  rollouts, since `gate_check.py` reads what is already stored — and the table reports
+  `verdict_by_reference` plus `verdict_stable`. A verdict that flips is downgraded to
+  `inconclusive`, because a round that cannot tell an edit from re-measurement has not measured
+  anything, whatever the delta looked like against the replicate that happened to be picked.
+
+  For context on why this matters at this scale, the same run's replicate gaps were 0.00, 0.01
+  and 0.12 across three rounds: with two replicates the gap is itself a poor estimate of the
+  noise, so the stability check — a direct observation rather than an estimate — is the more
+  reliable signal.
+
 - **The round table gave the driver two incompatible noise bars, and it rejected the run's best
   candidate on the wrong one.** On run 32871360361 round 2, `cand2` was gated against a control
   measured in that same round (0.24, its replicate 0.25 — agreeing to 0.01) and beat it by
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,36 @@ All notable changes to cap-evolve are documented here. The format follows
 [0.1.0]: https://github.com/skillberry-ai/cap-evolve/releases/tag/v0.1.0
 
 ## [Unreleased]
+### Added
+- **The run now keeps the agent's own transcript.** Four hours of run 32814848187 were
+  unaccounted for *and unaccountable*: its 900 metric calls account for every evaluation in
+  `events.jsonl` and none fall in the gap, so nothing was being measured — but the only record
+  of what the agent itself did was an 800-char `stdout_tail`, which narrows the cause to
+  "something hit the 4-hour Bash ceiling" and no further. `--output-format json` cannot close
+  that: it returns one result object with the final text and the totals, never the turns. So
+  `run-optimizer` takes `--transcript <path>`, and the registry carries a separate
+  `transcript_flag` (`--output-format stream-json --verbose`, verified against Claude Code
+  2.1.241) used *instead of* `json_flag` only when a transcript is requested — its last line is
+  the same result object, so cost and stop parsing are unchanged and the deterministic
+  per-iteration path is untouched. `host.py` points it at `$R/host/transcript.jsonl`, beside
+  the `driver_prompt.md` it already writes, so the run dir holds both what was asked for and
+  what was done. Secret values are scrubbed before the file is written: a transcript records
+  tool results verbatim and the run dir is a published artifact, so one `env` the agent
+  happened to run would otherwise leak the gateway token. `terminal_reason` and
+  `permission_denials` join the captured stop fields — the latter distinguishes "blocked by the
+  tool allowlist" from "chose to stop", which were previously the same observation.
+
 ### Fixed
+- **An explicit `iterations` dispatch was silently ignored on the smoke tier.**
+  `ITERATIONS: ${{ matrix.tier == 'smoke' && '3' || inputs.iterations || '10' }}` put the tier
+  pin first, and GitHub's `||` yields the first truthy operand — so dispatching smoke with
+  `iterations: 10` ran 3, discoverable only by reading `ITERATIONS` in the job log after the
+  budget had been spent on the wrong round count. Smoke is the tier the algorithm itself gets
+  iterated on, which makes it the worst one to pin unreachably. `NUM_TRIALS` already had this
+  right, including why its input default is `""` (with a non-empty default there is no way to
+  tell "asked for 10" from "asked for nothing"); `iterations` now has the same shape, and a
+  test asserts the two agree so the next edit to either notices.
+
 - **The hosted agent ended its turn to wait for a notification, and the process exited under
   it.** With the turn budget raised to 600, run 32814848187 no longer ran out of turns — it
   used 78 and stopped anyway, on `subtype: success` / `stop_reason: end_turn`, rc 0. It had
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,21 @@ All notable changes to cap-evolve are documented here. The format follows
 
 ## [Unreleased]
 ### Fixed
+- **Review follow-ups (PR #399).** Three findings from code review, all in the agent-optimize
+  host: (1) `_CODE_SUFFIXES` was a 14-entry allowlist, so a capability whose code is C, C++, C#,
+  PHP, Swift, Kotlin or Objective-C was treated as prose and never got the "the form that works
+  is a guard in the code" advice — the same silent miss this host was fixed for on `.py`/`.js`,
+  relocated to whichever language nobody listed. The set is broadened across scripting, shell,
+  compiled, functional and query languages and is now documented as **known-good, not
+  exhaustive**, so absence reads as a gap to fill rather than a decision that the language is
+  prose. (2) A *declared* capability with no matching skill package was skipped by
+  `harness._stage_context` while the payload still said `staged: True`, making
+  some-capabilities-missing indistinguishable from everything-staged — while the all-missing case
+  had always been loud. The context now carries `guidance_missing` and emits a `::warning::`
+  naming the capability, since silently optimizing a surface with no allowed-edit-space brief is
+  the exact defect this host was opened to fix. (3) A stray blank line after the concurrency
+  guard's `return 2`.
+
 - **The end_turn diagnosis accused a complete run of abandoning work.** Run 32871360361 booked 4
   of 10 rounds, investigated a round-5 lever, judged the residual failure unfixable by the
   surfaces it owned, **sealed test itself**, wrote its report and stopped at 121 of 1650 turns.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,36 @@ All notable changes to cap-evolve are documented here. The format follows
 
 ## [Unreleased]
 ### Fixed
+- **Three ways the agent-mode host was accidentally shaped around one benchmark.** Found by
+  auditing it against the repo's other workloads rather than by a failing run:
+
+  - **A project's `optimizer_instructions_file` was silently dropped.** `cli.py` applies that key
+    only for `algorithm_name in OPTIMIZER_CONTEXT_ALGORITHMS` (hill-climb / gepa / skillopt), so
+    agent mode ignored it. That is dangerous, not merely lossy: one arm uses that file to state
+    that the `{placeholders}` in its second editable file are load-bearing and that breaking one
+    makes EVERY task score 0 (the agent is never told where to write its answer). An agent that
+    never saw the warning could wipe the run's whole signal with an edit that looks harmless. The
+    host now reads and includes it — and reports a spec-named path that does not exist instead of
+    silently downgrading to generic guidance (the #252 failure mode).
+
+    Included with its **scope stated**, not pasted in: that file is written for the deterministic
+    per-iteration optimizer and tells the agent to stop after editing and not to evaluate, because
+    there the harness re-scores the candidate. In agent mode the agent owns the evaluation and the
+    gate, so an agent obeying that line would never gate anything. Benchmark facts bind; the
+    process half is explicitly superseded. Its unrendered `{{SLOT}}` template markers are stripped
+    too — the parity test already treats those as a defect on the deterministic side.
+
+  - **Code-vs-prose advice was offered to capabilities with no code.** The briefing told every
+    multi-file capability that "a rule the agent violates usually belongs in code as a guard".
+    For a two-prose-file capability (a system prompt plus a task template) that sends the agent
+    looking for code it does not own — how a prompt-only run once ended up editing `adapter.py`.
+    Now conditional on the surface actually containing code.
+
+  - **A large capability would have had every file listed.** A skill-package capability runs to
+    dozens of files; enumerating them crowded out the rest of the briefing. Now grouped by
+    directory above 20 files, stating the true total and how many are not listed individually,
+    and telling the agent to enumerate the rest itself — a bounded listing that never reads as
+    the complete surface.
 - **The hosted agent now gets the same optimizer read-context as every other algorithm.** Two
   agent-optimize CI runs on a `[system-prompt, tools]` capability edited **only the prompt file**
   across 4 of 4 candidates; the tool code sat writable and unopened in the same candidate dir. The
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,31 @@ All notable changes to cap-evolve are documented here. The format follows
 [0.1.0]: https://github.com/skillberry-ai/cap-evolve/releases/tag/v0.1.0
 
 ## [Unreleased]
+### Fixed
+- **The hosted agent now gets the same optimizer read-context as every other algorithm.** Two
+  agent-optimize CI runs on a `[system-prompt, tools]` capability edited **only the prompt file**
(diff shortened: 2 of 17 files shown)


eranra:
## Run 3 diagnosis: staging worked, and the run stopped one round before the fix

Run [32701056043](https://github.com/skillberry-ai/cap-evolve/actions/runs/32701056043) carried the staged guidance — `host_context: {staged: true, capabilities: [system-prompt, tools], agent: claude-code}`, native skills placed — ran the agent for 32 minutes, and **still edited only `policy/policy.md`**.

Its own reject notes say why, and the cause was mine, not the agent's:

```
r1_report_calc         "policy.md: require calculate tool for all money + report
                        tool-confirmed total/refund & flight details. Paired gate
                        +0.10<=SE0.122; within control noise floor.
                        System-prompt/output-contract surface."        -> rejected

r2_passenger_complete  "policy.md: passenger-mod rules … Paired gate Δ=0.000;
                        targeted always-0 tasks 17/22/40 unmoved (churn).
                        System-prompt surface."                        -> rejected
```

Then it stopped. It was naming its surface choice explicitly in both rounds, so this was never ignorance of the tool surface — and `tools.py` is genuinely live (`apply()` loads it via `_build_candidate_tools()`), so the surface was not inert either.

**The `stop_condition` I derived said:** *"Stop early when … two consecutive rounds are rejected with no new failure cluster left to attack."*

**`SKILL.md` says the opposite:** *"After two rejected rounds, read the candidate's TRACE before writing a third"* — a reject usually means the edit **form** was wrong, not that the search is done. My clause fired exactly there and ended the run one round before the escalation.

Worse, round 1's note is the textbook code case from that same skill: *"prefer an in-code guard to a prose rule where the capability owns its tools — prose is right when the agent **lacks** a decision criterion, code when it **has** one and violates it."* The agent has "use `calculate` for money" and violates it. The prescribed round 3 was a guard in `tools.py`.

### Two fixes

**`run_suite.sh`** — rejections no longer stop the run. Rounds and spend bound it; the stop condition now states that a rejection is the signal to change the **form** or the **surface** on the next round, and to use every round the budget allows.

**`host.py`** — the escalation is a **precondition on round 3+**, not encouragement: after two rejects the next round may not reuse that surface-and-form; read the trace first (never exercised ⇒ the form is wrong, so a third rewording will be rejected too); and where the failure is a criterion the agent *has* and violates, the form that works is a guard in the code rather than a third restatement in prose. Emitted only when the surface actually contains code, so a prose-only capability is not sent hunting for code it doesn't own.

The surface name now goes into the `commit.py --note` deliberately — the agent was already volunteering it — so which surface each decision was made on is recorded in the run.

Suite: **956 passed**, same single pre-existing failure. Skill `check.py`, authoring lint and `actionlint` green.

eranra:
## Reuse audit (review follow-up) — and it found the real root cause

Question was whether `host.py`'s detail can come from the deterministic phase. It can, and the audit turned up something better than tidier code: **the two blocks the host hand-rolled thinner versions of are exactly the guidance whose absence made it edit prose for three runs.**

The deterministic optimizer receives, every iteration:

- `harness._CAP_EDIT_SPACE["tools"]` — *"**HIGHEST-LEVERAGE EDIT: WRITE A NEW CODE-BEARING TOOL** (a real body — loops, validation, calls to existing tools), because a deterministic tool can't be 'forgotten' the way a prompt rule can"* + two concrete patterns to prefer.
- `target_profile.reader_block()` — *"At runtime these capabilities are read by `<model>` … When the reader is weaker than you, prefer explicit rules, worked examples, and **code enforcement over terse prose**."*

The hosted agent received **neither** — and the CI agent under test is `aws/gpt-oss-120b`, precisely the weak reader that second block is about. So the deterministic path has always told its optimizer that tool code is the highest-leverage surface, while the host substituted a few sentences of mine. `OptimizerContext`'s own docstring says it exists so *"an algorithm cannot silently run on a thinner prompt than its siblings"*; the host was the caller bypassing it.

### The public seam (chosen over importing private names)

```
core/cap_evolve/harness.py
  OptimizerContext.from_spec()    construct from a capevolve.yaml dict — the host has a
                                  spec, not argparse args — resolving the target profile
                                  the same way from_args does
  .capability_brief()             _CAP_EDIT_SPACE + meta summary + guidance pointer
  .reader_brief()                 who consumes the capability at runtime
  .empty_seed_brief()             author-from-scratch guidance for an empty seed
  .render_template()              fill a template's slots for a non-per-iteration caller

core/cap_evolve/specfile.py
  resolve_instructions_file()     THE resolver for optimizer_instructions_file, now
                                  shared with cli.py
```

### Deleted from `host.py`

| duplicate | replaced by |
|---|---|
| `_guidance_section` | `capability_brief()` — richer, carries `_CAP_EDIT_SPACE` |
| `_arm_instructions` | `specfile.resolve_instructions_file()`, shared with `cli.py` |
| `_strip_template_slots` | `render_template()` — **fills** the slots instead of erasing them |
| `_known_agents` parser | run-optimizer's own `load_registry()` |
| my code-vs-prose advice | the shared block, which states it better |

### What I deliberately did NOT reuse

`OptimizerContext.instructions()`. It renders the per-iteration contract — *"fix many root causes in this ONE candidate and **STOP**; the harness re-scores you"* — which is false where the agent owns the search, the evaluation and the gate. An agent obeying it would never gate anything. The blocks are composed instead, and that reasoning is now recorded on the class so the next reader doesn't "fix" it by calling `instructions()`.

### Bonus: a gap closed, not just duplication removed

`empty_seed_brief()` means a **no-skill control** arm — one that blanks the seed to measure "author from nothing" — now gets author-from-scratch guidance in agent mode. It previously got none.

### Verification

The composed briefing on the real capability now carries the full `_CAP_EDIT_SPACE` tools block and the reader block. The parity test asserts the capability brief **against harness's own string**, not a copy pasted into the test, so the two paths cannot drift. `cli.py`'s deterministic behaviour is unchanged: **957 passed**, same single pre-existing failure. Skill `check.py`, authoring lint, `actionlint` all green.

eranra:
## Pre-merge verification run: [32722203848](https://github.com/skillberry-ai/cap-evolve/actions/runs/32722203848) @`bb8adafb` — success

**Mechanism verified.** `host_context: {staged: true, capabilities: [system-prompt, tools]}`, and the composed briefing carries every reused block:

| probe | present |
|---|---|
| `_CAP_EDIT_SPACE` — "CODE-BEARING TOOL" | ✅ |
| target-reader block — "THE READER" / `gpt-oss-120b` | ✅ |
| escalation precondition — "may not reuse the surface" | ✅ |
| unrendered `{{SLOT}}` markers | ✅ none |

**Surface coverage improved — 2 of 3 rounds reached the tool code**, versus 0 of 2 and 0 of 2 in the two runs before the fixes:

| round | files edited | val | gate |
|---|---|---|---|
| baseline | — | 0.233 | — |
| `cand_r1_policy` | `policy/policy.md` | 0.267 | ❌ |
| `cand_r2_tools` | `policy/policy.md` + **`tools/tools.py`** | 0.333 | ❌ |
| `cand_r3_remove` | `policy/policy.md` + **`tools/tools.py`** | 0.300 | ❌ |
| finalize | best = `seed` | test 0.200 | — |

And it implemented precisely the two patterns `_CAP_EDIT_SPACE` names, which is direct evidence the reused block changed behaviour rather than just appearing in the prompt:

> r2: *"added code-enforced safe wrappers `change_cabin()` and `rename_passenger()` (+policy Rules 4/5 route cabin/name changes through them)"* — the "VALIDATION/RULE-ENFORCEMENT tool that wraps a primitive" pattern.
>
> r3: *"`REMOVE_TOOLS={update_reservation_passengers}` forcing name changes through code-enforced `rename_passenger` (kills duplicate-passenger bug)"* — the "REMOVE the raw primitive so the only path is the safe one" pattern.

**This run is a null result: nothing cleared the gate, `best = seed`.** Stated plainly rather than framed as a win.

It is *not* evidence of a refactor regression, and I can't claim it as an improvement either. The agent measured its own noise floor at 0.133–0.2 and reported control replicates swinging 0.433/0.40 on a **byte-identical** candidate. Every candidate delta here sat inside that band, and it said so rather than claiming harm:

> *"delta -0.067 == byte-identical control replicate (0.30, same 'regression 19'). Pure noise; noise floor 0.133."*

At 10 tasks × 3 trials a smoke run cannot resolve effects of this size, so the previous run's accepted +0.30 and this run's three rejects are both consistent draws from the same noisy distribution. The honest read is: **mechanism verified, score unresolvable at this tier.**

### Recommendation

Merge on the mechanism evidence, which is what this PR claims. If you want a score claim, the repo already supports it — dispatch the same run at `BENCH_SPLIT_SEED` 42/43/44 for the ">=3 seeds" comparison, or use the `full` tier where the noise floor is much lower. Say which and I'll run it.

Unchanged known gap: `optimizer $0.00` — the gateway reports no cost, so agent-mode spend is unmetered rather than free.

eranra:
## Three fixes from run 32733635494, then re-run

All three were found by reading that run's artifacts, not by a failing test.

### 1. Optimizer cost was never $0.00 — I was reading the wrong key

`run-optimizer` nests the figure under `cost.total_cost_usd`; the host read a flat `cost_usd`/`usd` and booked `0.0`. That run really spent **$8.37 / 68,432 tokens** and recorded nothing.

This is worse than a wrong number. `$0.00` after real rollouts is exactly the *unmetered gateway* signature the skill warns about — so the wrong conclusion was drawn from it twice, **including by me in this PR's earlier notes, where I said "the gateway doesn't report cost"**. It does; I dropped it. And a ceiling cannot bind what it cannot see: with `0.0` booked, `max_usd` and `spend.py`'s dollar predicates were inert for the whole run.

### 2. The host now says why the agent stopped

`returncode: 0` covers both "finished the work" and "hit `--max-turns` mid-round", which demand opposite responses — and telling them apart meant reading a truncated 800-char stdout tail.

`run-optimizer` now passes through the CLI's termination fields (`subtype`, `num_turns`, `is_error`, …), vendor-neutral exactly like its cost parsing, and useful to the deterministic path too. The host surfaces `stop_reason` / `num_turns` / `rounds_booked` / `rounds_budget`, plus an `incomplete` field and a `::warning::` when rounds were left unspent. An unfinished run must not read as a finished one.

### 3. Agent mode was starved of turns — and it cost the run its best candidate

`optimizer_max_turns` (default 80) is a **deterministic-path** unit: there, one optimizer invocation means "propose one edit and stop", and the harness owns diagnosis, evaluation and gating. In agent mode that same allowance must also cover Phase 0, diagnose, null-control replicates, `round.py` orchestration, `gate_check` and `commit.py` — *per round*.

240 turns (80 × 3) bought **1.9 rounds**. The agent stopped on `error_max_turns` having just evaluated `cand_r2_compose` at val **0.530** — the best score of the run — and never reached the `commit.py` that books it. So it reported 1 of 3 rounds and discarded its best result. That also explains the puzzle in the artifacts: `steps.jsonl` showed one iteration while `events.jsonl` showed two candidates evaluated, because `commit.py` is what books a round.

Budget is now `max(optimizer_max_turns, 150) × (rounds + 1)` → **600** at defaults: a floor of 150/round against the ~126 measured, and one extra round's worth for the work that is not a round (Phase 0, the seal). `max()`, so a raised `optimizer_max_turns` still wins — a floor, not a clamp.

## What the trials=10 run already established

**Variance improved sharply**, measured on byte-identical `ctl_null` replicates:

| | trials=3 | trials=10 |
|---|---|---|
| replicate spread | 0.100 / 0.033 / 0.067 | **0.050 / 0.010** |
| per-task granularity | k/3 = 0.333 | k/10 = 0.100 |
| typical SE | 0.111 – 0.143 | **0.067 – 0.097** |
| gate outcome | 3 candidates, all unresolvable | +0.090 **accepted**; sealed test **0.280 → 0.450** |

**Wall-clock: 90 min, not the 3.5–4 h I predicted.** Per-eval went 323s → 779s = 2.4×, not 3.33×: at 30 pairs a pool of 10 drains unevenly, while at 100 pairs it stays saturated, so more trials amortise *better*. My linear model was wrong.

Re-running now at trials=10 with the 600-turn budget: [32814848187](https://github.com/skillberry-ai/cap-evolve/actions/runs/32814848187). Expect 3 booked rounds, a non-zero cost figure, and `incomplete` empty. Suite **963 passed**, skill check and actionlint green.

OsherElhadad (approved):
LGTM,
Pls note that I see the current run and cannot see the cost of the baseline run and the intake. But let's wait for the iterations - if you could fix that it will be great!

OsherElhadad:
This is a great PR and a genuinely useful contribution. The diagnosis (measured via two real CI runs, not guessed) is correct — `inject()` really was never called, `_CAP_EDIT_SPACE`/reader-block guidance really was absent, and the fix threads through `cli.py`, `harness.py`, `specfile.py`, and `host.py` coherently instead of patching the symptom in one place. The genericity audit is also legitimate work, not padding: the `optimizer_instructions_file` drop really was dangerous for spreadsheetbench, and the code-vs-prose/file-count issues are real. Nice test discipline too — every fix is tied to a specific measured run ID rather than a hypothetical.

That said, I found a couple of things worth fixing, one of which cuts against the PR's own genericity thesis.

## Issues (most to least important)

**1. `_CODE_SUFFIXES` is a hardcoded allowlist that will misfire on any language it doesn't name — confirmed by testing.**
`skills/algorithms/agent-optimize/scripts/host.py:146-147`:
```python
_CODE_SUFFIXES = {".py", ".js", ".mjs", ".ts", ".tsx", ".sh", ".bash", ".rb", ".go", ".rs",
                  ".java", ".pl", ".lua", ".sql"}
```
I ran `_surface_section` against a synthetic capability with `prompt.md` + `engine.c` (i.e., genuine code alongside prose). Because `.c` isn't in the set, `has_code` is `False` and the "prefer a code guard" advice never fires — the exact bug class this PR just fixed for `.py`/`.js`, silently reintroduced for C, C++, C#, PHP, Swift, Kotlin, Objective-C, and anything else not on the list (also misses config-as-code like `.yaml`/`.json` used as a rules file, which is arguably "code" for this purpose too). Given the PR's stated goal is to generalize past the three audited benchmarks, an extension allowlist is exactly the kind of "magic list that happens to cover today's workloads" the audit itself was hunting for elsewhere (see the 20-files threshold, which is at least named and reasoned about).

Suggested fix: this repo already has a controlled vocabulary for capability *kinds* — `registry.yaml` / `harness._CAP_EDIT_SPACE` know which capability names are "tools" (code) vs "system-prompt"/"task_template" (prose). Gating `has_code` on `any(cap in self.capabilities for cap in KNOWN_CODE_CAPABILITIES)` (or by asking `_CAP_EDIT_SPACE` itself) would track the repo's own capability taxonomy instead of a suffix list that needs updating every time a new language shows up. If you want to keep it fast/local, at minimum broaden the set substantially and treat it as "known good, not exhaustive" in a comment.

**2. A capability that resolves to no guidance skill is silently skipped, and `_stage_context` still reports `staged: True` — suspected, confirmed by reading, not exercised by the PR's tests.**
`core/cap_evolve/harness.py:1286-1289`:
```python
for c in caps:
    src = skills_root / c
    if not src.is_dir():
        continue
```
This is pre-existing code, not new in this PR, but the PR's whole point is that *silent-off is the failure mode this repo warns about everywhere* — and now that `inject()` is wired into agent mode, this particular silent skip is reachable from a path this PR added tests for the *other* failure modes of (whole-staging exception → reported; per-capability miss → not reported). If a spec declares `capabilities: [system-prompt, tools, custom-thing]` and `custom-thing` has no matching dir under `skills/algorithms/`, `_stage_context` still returns `staged: True` with no indication that `custom-thing` got zero guidance — indistinguishable from "everything staged" in the JSON output and in the `::warning::` logic (`if not context["staged"]`). This is the exact "quietly optimizing less surface" bug this PR fixes for the all-caps-missing case, just not for the some-caps-missing case.

Suggested fix: have `_stage_context` (or `inject()`) return which capabilities actually got a `guidance/<cap>/` directory, and report the difference (`requested - staged`) alongside `context["staged"]`, warning on any gap the same way a full failure warns now.

**3. Minor: double blank line after the new concurrency guard.**
`skills/algorithms/agent-optimize/scripts/round.py`, right after the new `return 2` in the refusal block, there are two blank lines before `run_dir = RunDir.open(...)` instead of one — cosmetic, but worth a quick pass since the rest of the file is otherwise tight.

## What I verified locally

- `gh pr checks 399`: all checks green (Analyze JS/TS, Analyze Python, CodeQL, dashboard build, docs link check, install smoke, skill authoring lint, spellcheck, Test Python, toy example).
- Checked out the PR branch (`gh pr checkout 399`) and ran the four target test files with `.venv/bin/python -m pytest core/tests/test_agent_optimize_host.py core/tests/test_agent_optimize_constraints.py core/tests/test_benchmarks_agent_optimize.py core/tests/test_optimizer_transcript.py -q`: **108 passed**.
- Ran the full suite: `.venv/bin/python -m pytest core/tests -q`: **950 passed, 12 skipped, 0 failed** (close to the PR's claimed 954/1-pre-existing-failure; difference is likely environment-dependent skips, not a discrepancy worth chasing).
- Wrote a standalone harness reusing the test file's own `_project`/`_run_dir`/`_host` fixtures and drove `host.py --prompt-only` directly against several synthetic capabilities I built myself (not from the existing tests):
  - A prose+`.c` capability → confirmed issue #1 above (no code advice fires).
  - Exactly 20 files → still fully enumerated (boundary is `<= _MAX_LISTED`, correctly inclusive).
  - 21 files → correctly switches to grouped/truncated form, states the true count and how many aren't individually listed.
  - 25 files spread one-per-directory (duplicate-shaped groups, no actual duplicate basenames across dirs to collide) → grouped correctly, no crash, no bad interaction.
  - Zero files → `_surface_section` returns `""` cleanly (no crash; the section is simply absent from the briefing).
- I did not have tau2/spreadsheetbench/skillsbench fixtures wired to real API keys, so I could not run an actual hosted-agent round end-to-end (real LLM calls) — everything above is `--prompt-only`/static-analysis verification of the staging and briefing-generation code paths, not a live agent run. The PR's own claim of a smoke run "in flight" I could not independently verify either way.
- Did not find any path-traversal, command-injection, or untrusted-write issue: capability/benchmark names and instruction-file paths all originate from the project's own `capevolve.yaml`/registry (author-controlled, not attacker-controlled input), `resolve_project_path` joins against a fixed project root, and the new `run_suite.sh` publish block operates only on `$RUN_DIR/host/*` with no interpolation of untrusted content into a shell command. This is consistent with the rest of the codebase's trust model; I don't consider it a gap introduced by this PR.

## Suggestions beyond this PR

- The capability-kind vs. file-suffix mismatch in issue #1 suggests `_CAP_EDIT_SPACE`/registry.yaml would benefit from an explicit `has_code: bool` (or `kind: code|prose|mixed`) per capability, settable at intake time, rather than inferring it after the fact from file extensions anywhere it's needed (this host, and potentially future callers). One authoritative source beats re-deriving "is this code" from suffixes in N places.
- The `_unbooked_rounds` "recognized by shape, not filename" backstop (any `work/*.json` with a `candidates` list) is a nice, benchmark-agnostic idea, but it's also the kind of thing that would benefit from a `check.py` assertion that `round.py`'s own table format can never collide with some other file a future script might drop into `work/` — right now that's an implicit contract between two files' output shapes.
- Given the `--allow-high-concurrency` and `--reject-basis driver_judgement` escape hatches added here, it might be worth a follow-up that greps `events.jsonl` across historical runs for `overrode_gate: true` or high-concurrency reruns, just to see how often the escape hatches actually get used — that's the kind of thing that quietly becomes the default nobody meant to pick.

eranra:
## Merge-readiness review — full diff against `main`

Read the whole diff rather than summarising from memory: **15 commits, 17 files, +2751/−87.**

### Core changes

The core problem this PR fixes is that **agent-optimize's hosted agent was running blind and half-crippled**. `harness.OptimizerContext` exists so no algorithm silently runs on a thinner prompt than its siblings, but the agent-mode host never called it — so the agent got prose-editing guidance and *nothing* about editing tool code, and duly spent four candidates across two runs rewording `policy.md` while `tools.py` sat writable and unopened. The fix stages the capability skills, diagnose guidance and target-reader block into the agent's workspace, and adds a `from_spec()` constructor plus four public prompt-block accessors (`capability_brief`, `reader_brief`, `empty_seed_brief`, `render_template`) so the host composes the same measured blocks without inheriting the per-iteration contract that would be false for an agent owning the whole loop. `specfile.resolve_instructions_file()` extracts the `optimizer_instructions_file` resolution rule that `cli.py` had inline — two copies of that rule is how #252 happened. Both `harness.py` and `specfile.py` are **purely additive**; `instructions()` and `inject()` are untouched.

The second core theme is **honest measurement and honest reporting**, all of it in the agent-optimize scripts. The host now guarantees a seal, reports *why* the agent stopped (`stop_reason`, `agent_stop_reason`, `num_turns`, `rounds_booked`), and flags a run that stopped with rounds left — with per-cause advice, so a turn-starved agent gets "raise the turn budget" while one that sealed itself gets "unspent budget, not a dead loop". It puts the correct interpreter first on the agent's `PATH` (every candidate eval in one run died `ModuleNotFoundError` because bare `python` cannot reach the venv holding the adapter's deps). `round.py` now persists its gate table to disk instead of leaving stdout the only copy, refuses a concurrency too coarse to resolve its own verdict, separates `parent` from `gate_reference` and reports `parent_vs_gate_ref_drift`, emits one `evidence_bar` matched to how it gated, marks a verdict that flips across control replicates as `inconclusive`, and shows the drift-free comparison a parent-gated round already paid to measure. `commit.py` refuses `--reject-basis gate` when the gate actually *accepted*, adding `driver_judgement` for honest overrides and recording `gate_verdict`/`overrode_gate` on every decision. These are not speculative — each pins a specific misreading measured in runs 32814848187, 32861747778 and 32871360361, including one where the table talked the agent into discarding the best candidate of the run.

### CI changes

Three things. `run_suite.sh` (a) drops the stop-on-two-consecutive-rejections clause, which fired exactly where the algorithm prescribes *escalation* — on run 32701056043 the agent rejected two prose candidates and quit before trying the code-level guard its own reject note had already named; (b) raises the agent-mode turn budget to `max(optimizer_max_turns, 150) × (rounds + 1)`, measured from a run that consumed ~126 turns/round and died on `error_max_turns` with its best candidate evaluated and uncommitted; and (c) copies `$RUN_DIR/host/` into the uploaded artifact, gzipping the transcript — necessary because the artifact path is `$OUT/**` while the run dir reaches it only via `export_static`, which caps every file at 256 KiB and keeps the *first* chunk. The real transcript was 1.9 MB and the answer was in the last 12%. On the optimizer side, `run-optimizer` gains an opt-in `--transcript` and the registry a `transcript_flag` (`--output-format stream-json --verbose`, verified against CLI 2.1.241) — **`cli.py` never passes `--transcript`, so the deterministic path's `--output-format json` is untouched**. Finally `benchmarks.yml` makes the `iterations` input reachable on smoke: the old expression put `matrix.tier == 'smoke' && '3'` first, and GitHub's `||` short-circuits, so a dispatch of `iterations=10` on smoke silently ran 3. It now mirrors the pattern `NUM_TRIALS` already used, input default `""` included.

### Merge readiness

**Blast radius is small.** Everything behavioural is gated behind `ALGORITHM=agent-optimize` / `ORCH_MODE=agent` / `[ -d $RUN_DIR/host ]`, and `agent-optimize` is dispatch-only, not in `OPTIMIZER_CONTEXT_ALGORITHMS` (`{hill-climb, gepa, skillopt}`). Verified that `round.py`/`commit.py` have no callers outside the skill, and that `run-optimizer`'s diff has **zero removed lines**.

Two changes do reach non-agent runs, and a reviewer should know them:

- **`cli.py`** — a *missing default* `optimizer/INSTRUCTIONS.md` no longer prints a warning; only a spec-**named** missing file does. Deliberate and documented, but it is one fewer diagnostic line for `hill-climb`/`gepa`/`skillopt`.
- **`benchmarks.yml`** — only smoke-with-an-explicit-`iterations` changes. Blank dispatch and PR-label runs are byte-identical to before (smoke 3, others 10).

**Tests: 991 passed, 2 skipped, 1 failed.** The failure is `test_cli_surface.py::test_spec_defaults_relative_to_project_not_the_cwd`; a clean `origin/main` worktree fails there identically (a `run-optimizer` manifest lookup), so it is pre-existing and unrelated. Skill `check.py`, authoring lint and `actionlint` all green.

**One caveat worth having.** Commits through `5d89932d` are validated end-to-end in CI — run [32871360361](https://github.com/skillberry-ai/cap-evolve/actions/runs/32871360361) booked 4 rounds, sealed itself, and produced test **0.40 → 0.53 (+0.13)** with the mechanism traced (task 40: 0.10 → 1.00, the exact failure diagnosed from the seed rollouts). The last six commits (`21e41c3a`…`b47602c9`) are reporting-layer changes made *while* that run was in flight, so they are covered by unit tests and by `check.py` executing `round.py`/`commit.py` against a real run dir in both gate modes — but they have never run against a live agent. They add fields and read-only re-gates rather than changing decisions, so the risk is low; one smoke dispatch would exercise them if you want that first.

### Measurement finding that outlasts this PR

Worth recording separately from the code: on tau2 smoke, the **same seed bytes** scored **0.24 / 0.44 / 0.38** across three runs in one day (sd 0.103, range 0.20) — roughly **3.7× the `gate_k_se=1.0` acceptance bar**. Within a single run, byte-identical control replicates measured two minutes apart read 0.32 and 0.20. The default `--gate-against parent` compares against a *stored* reward, so that drift lands inside every accept/reject decision; `--gate-against control` pairs against a concurrently-measured control and removes it. The reporting changes here make the drift visible per round rather than something a reader reconstructs by hand, but **the default gate mode was deliberately left at `parent`** — changing it on one benchmark's evidence would be a guess about every other workload. Two follow-ups are worth considering on their own merits, not as part of this PR: raising `gate_k_se` on this tier, and telling the agent to prefer control-mode gating.

eranra:
Thanks @OsherElhadad — this is a genuinely useful review. All three concrete findings reproduced exactly as described, and the first one is the sharpest kind of catch: it points out that the PR relocated a bug class it claims to remove. Fixed in 183b6105.

## Fixed

**1. `_CODE_SUFFIXES` was an allowlist that misfired on unlisted languages.** Confirmed — with 14 entries, C, C++, C#, PHP, Swift, Kotlin and Objective-C surfaces were treated as prose and never got the "the form that works is a guard in the code" advice. Broadened across scripting, shell, compiled, functional and query languages, and — more importantly — documented as **known-good, not exhaustive**, so a future gap reads as something to fill rather than a decision that the language is prose.

One deliberate divergence from your suggested fix, and I want to be explicit about it rather than quietly ignore it. I kept the **suffix** test rather than switching to the capability's declared kind (`tools` = code vs `system-prompt` = prose). Kind is only a proxy: a `tools` capability can be schema-only, and a `system-prompt` capability can ship a helper script — whereas the question the briefing actually asks is "does the surface I am handing you contain code you could put a guard in", which is a property of the files, not of the label. Your broader point stands though, and I've left it in the comment so the next reader doesn't re-litigate it: promoting `has_code` (or `kind: code|prose|mixed`) into `registry.yaml`/`_CAP_EDIT_SPACE` at intake time **is** the better long-term answer. It touches the capability taxonomy and every caller, so it belongs in its own change, not bolted onto this one.

**2. A declared capability with no guidance skill was silently skipped while `staged: True`.** Confirmed at `harness.py:1286` — and you're right that this is the same "quietly optimizing less surface" defect the PR was opened to fix, with one half left quiet. The scope turned out smaller than your suggestion implies: the staged `guidance` list was *already* in the payload (`host.py:578`), so what was missing was the comparison against what the spec declared, not a new return contract. The context now carries `guidance_missing`, and a `::warning::` names the capability and points at the likely cause (a typo in `capevolve.yaml` `capabilities`, or a genuinely absent skill). Fixed on the host side rather than in `harness.py` because the host is what knows what was *requested* — a deterministic caller with no capabilities declared is unaffected.

**3. Double blank line** after the concurrency guard's `return 2`. Gone.

4 new tests: the suffix vocabulary (parsed statically with `ast`, since a vocabulary assertion shouldn't need the module to import), the reported gap, the warning, and silence on the healthy path. Full suite **995 passed**.

## Not doing here, and why

- **`check.py` assertion that `round.py`'s table shape can't collide with other `work/*.json`.** Fair concern, but the contract is already fairly narrow — a dict with a `candidates` list whose rows carry both `tag` and a truthy `verdict`. Adding a guard against a collision nobody has hit yet costs more in test surface than it buys. Worth revisiting the moment something actually drops a `candidates`-shaped file into `work/`.
- **Grepping historical `events.jsonl` for `overrode_gate: true` / high-concurrency reruns.** Genuinely good observability idea and I'd support it — "the escape hatch that quietly becomes the default nobody meant to pick" is exactly the right worry. But it's a reporting tool over past runs, unrelated to whether this merges.

## On your verification numbers

You ran 950 passed / 12 skipped / **0 failed**; I measure 995 / 2 / 1. Worth reconciling rather than splitting the difference: the failing test (`test_cli_surface.py::test_spec_defaults_relative_to_project_not_the_cwd`, a `run-optimizer` manifest lookup) still fails for me under this repo's own `.venv` **and** on a clean `origin/main` worktree — so "pre-existing, not caused by this PR" holds either way. Your environment collected ~45 fewer tests and skipped 10 more, so I'd read your 0-failed as fewer tests running rather than as the failure being absent. Not worth chasing for this merge, but I didn't want to leave it looking like a discrepancy in the PR's favour.

Also, on the one thing you flagged as unverifiable: the smoke run is no longer in flight. Run [32871360361](https://github.com/skillberry-ai/cap-evolve/actions/runs/32871360361) booked 4 rounds, sealed itself, and produced test **0.40 → 0.53 (+0.13)** with the mechanism traced rather than inferred — the agent diagnosed two tasks failing identically (it yields a confirmation summary, the user replies `Yes.###STOP###`, the episode ends before the DB write), wrote a turn-economy policy rule, and task 40 went 0.10 → 1.00. The PR description predates all of that and is stale; I'll refresh it before merge.

OsherElhadad (approved):
Perfect

Issue #401, opened 2026-08-26
agent-optimize: zero accepted improvements across v1-v4, no /goal enforcement, missing statistical-rigor fixes from TAU2_SUMMARY.md


OsherElhadad:
## Problem

`agent-optimize` (`skills/algorithms/agent-optimize/`) is already designed as the non-deterministic, agent-driven algorithm — screen.py subset evaluation, English `stop_condition` constraint parsing (`cap_evolve.constraints.parse_constraints`), and deterministic only at the val gate (`gate.decide`) and test seal (`RunDir.commit_test()`). That part is not broken and should not be rebuilt.

What **is** broken, evidenced by real runs:

- **Zero accepted improvements across all 4 real attempts** (`examples/tau2_airline/run_agentopt_v1..v4`, see `docs/TAU2_SUMMARY.md` rows 5–8). Every one of those runs was measured at `num_trials=1` on a 12-task val split (binomial SE ≈0.13) — the gate was mathematically unable to resolve any real signal, not "the edits were bad."
- **No `/goal`-style constraint enforcement exists** (`grep -rn "/goal"` returns nothing repo-wide). Constraints parse into predicates but there's no live mechanism forcing a Claude-Code-driven optimizer to check/report against them mid-run.
- **No verification that the free-form SKILL.md prose is actually followed live** — nothing confirms `screen.py` gets invoked before full-val in a real Claude Code session (only inferred from static SKILL.md text).
- `docs/TAU2_SUMMARY.md`'s own "what would make cap-evolve better" section lists concrete unimplemented fixes from this exact benchmark's failures:
  - Report the resolvable effect size (`2·SE`) next to every gate verdict, before the driver reads the delta.
  - Require plural (≥2) byte-identical null-control blocks by default (`round.py`), not reactively after failures.
  - Require two-seed-block sign agreement before any accept (one block called a null result positive in round 10).
  - Select canaries mechanically from the whole suite (`--canary-auto`, lowest-rate-first), not hand-picked near the mechanisms under test (missed 4 damaged high-scoring tasks: 11, 15, 47, 49).
  - Carry a mechanism ledger with `--supersedes` so a disproved finding doesn't stay `verified`.
  - Integrate mechanisms sequentially (`integrate.py`, one branch at a time) — a one-shot 6-mechanism merge gated at −0.0146 while a single-mechanism artifact gated at +0.0115 ("fewer mechanisms beat more").
[...]

PR #402, opened 2026-08-26
agent-optimize: statistical-rigor fixes, /goal enforcement, compliance logging, tau2 STOP-leak fix (#401)

OsherElhadad:
Addresses #401. Referencing, not closing — see "Deferred" below; the real end-to-end
validation run in the issue's acceptance criteria was not performed (no network egress in
the environment this was built in).

## Implemented

- **Resolvable effect size on every gate verdict** — `gate.decide()` now returns
  `resolvable_effect_size` (`2*SE`) alongside `accept`/`delta`/`threshold` for both `paired`
  and `significant` modes (`core/cap_evolve/gate.py`). A driver reads "this round can resolve
  ±X" before reading the delta, instead of discovering after the fact that a null result was
  a measurement-power problem, not a bad edit.
- **Mandatory two-seed-block sign agreement** — `round.py`'s per-candidate
  `verdict_stable`/`verdict_by_reference` check (re-gating each candidate against every control
  replicate) previously ran only under `--gate-against control`. It now runs whenever there is
  more than one control replicate, which is the default (`--control-replicates 2`), in either
  gate mode. A candidate whose verdict flips between control replicates is now marked
  `inconclusive` even in the default parent-gated round.
- **Costed-target check** — `cap_evolve.constraints.cost_target(target, per_task_ceiling)`
  costs a `target_val_score` against measured per-task headroom (mean of per-task ceilings);
  wired into `spend.py --ceiling-file <json>` so a target above the costed ceiling is flagged
  before more budget is spent chasing it.
- **`/goal` constraint enforcement** — `plugins/cap-evolve/hooks/goal_reminder.py`, a new
  `PostToolUse` hook (registered in `plugins/cap-evolve/hooks/hooks.json`). In an agent-mode
  run it counts Bash tool calls and, every `CAPEVOLVE_GOAL_CADENCE` calls (default 12),
  re-injects the parsed `stop_condition` predicates plus measured spend/wallclock/
  protected-tasks state as context — independent of whether the driving agent remembers to
  call `spend.py` itself. Fails open, no-ops outside an agent-mode run dir, matching the
  existing hooks' conventions.
- **Compliance instrumentation** — `round.py` now logs an `agent_optimize_compliance` event
  per candidate recording whether `screen.py` ran for it before this full-val eval (checked
  against `<run_dir>/screens/`). `dashboard.py`'s `reduce_run()` surfaces these under
  `algo_extra["compliance"]` as a distinct entry, generically (not agent-optimize-specific
  plumbing beyond the existing per-algorithm extraction pattern `screens` already uses).
- **tau2_airline user-simulator `###STOP###` + leaked-continuation-reasoning fix** — the bug
  from `docs/TAU2_SUMMARY.md` row 7 (the simulator emits `###STOP###` in the same message as
  reasoning that explicitly plans to continue, in 15/27 observed task-7 failures, costing
  ~4.2% of all rollouts). tau2-bench itself is an external, unvendored package (cloned at
  setup time), so the fix is on our side of that boundary: `_leaked_stop_continuation()`
  scans the message trace in `_sim_to_rollout` and marks the affected rollout as infra noise
  (mirroring the existing `rollout.error` "uncontrollable noise" path) rather than scoring it
  as an agent failure. Applied to both `examples/tau2_airline/adapters/adapter.py` (the live
  RITS adapter) and the shared `templates/adapters/tau2_bench/adapter.py`.

**Already implemented before this PR** (verified, left unchanged): mandatory plural null
controls (`round.py --control-replicates 2` default), mechanical canary selection
(`integrate.py --canary-auto --canary-floor`), and the mechanism ledger's `--supersedes`
(`mechanisms.py`). The issue listed these as missing; the codebase had already gained them in
earlier merged work.

## Deferred (documented, not silently dropped)

- **Real end-to-end validation run** (30/30/20, 10 trials, `aws/gpt-oss-120b` agent). The
  sandbox this PR was built in has no network egress — cannot `git clone` tau2-bench nor reach
  a model endpoint. I did **not** fabricate a run result. What I did verify:
  - `core/tests/` — full suite: **957 passed, 12 skipped**, including two new tests for the
    STOP-leak detector (`test_tau2_eval_cost.py`) exercised against a stubbed `tau2` module
    (no network required).
  - `skills/algorithms/agent-optimize/scripts/check.py` (the skill's own hard gate) — green.
  - No dashboard screenshot from a real run for the same reason; `reduce_run()`'s new
    `algo_extra["compliance"]` field is covered by the fact that `dashboard.py`'s existing test
    suite (`test_dashboard*.py`, `test_dashboard_agent_run.py`) still passes with the new event
    kind added to the agent-optimize algorithm marker list.
  - Follow-up: run `examples/tau2_airline/itest.sh` (or the full `airline90` spec) in an
    environment with network access to produce the real accept/null result with its resolvable
    effect size, and a dashboard screenshot, per the issue's acceptance criteria.

## Verification

```
cd core && python3 -m pytest tests/ -q
# 957 passed, 12 skipped
```

Git identity: authored as Osher-Elhadad <Osher.Elhadad@ibm.com>.

### DIFF
--- a/core/cap_evolve/constraints.py
+++ b/core/cap_evolve/constraints.py
@@ -13,8 +13,12 @@ predicates against the run dir's persisted spend every round.
 Two functions, both pure and stdlib-only:
 
   * :func:`parse_constraints` — text → ``{"predicates": [...], "ambiguous": [...],
-    "text": "<verbatim>"}``. The original prose is always carried alongside, because a
-    parser that silently drops half a sentence is worse than no parser.
+    "unenforceable": [...], "text": "<verbatim>"}``. The original prose is always carried
+    alongside, because a parser that silently drops half a sentence is worse than no
+    parser. ``unenforceable`` is a clause-level list of prose this parser recognized NO
+    numeric/task-protection pattern in at all (a behavioral instruction like "use
+    screen.py before paying for full val") — distinct from ``ambiguous``, which is a
+    number this parser SAW but could not resolve (no unit, vague magnitude).
   * :func:`check_constraints` — predicates + measured actuals → per-predicate
     satisfied/violated + one ``stop | continue | narrow_scope`` recommendation.
 
@@ -229,6 +233,28 @@ def parse_constraints(text: str | None) -> dict:
                                  "enforces totals, so the run is effectively unbounded in $. "
                                  "Ask for a total, or set max_usd in the spec"})
 
+    # --- unenforceable prose -------------------------------------------------
+    # A clause with a NUMBER but no recognized unit is already loud (the "no recognized
+    # unit" branch above). A clause with NO number at all — a behavioral instruction like
+    # "use screen.py before paying for full val each round" — matches none of the numeric/
+    # protection patterns above and was, until now, silently dropped: neither enforced nor
+    # reported, so a driver had no way to tell "the framework checked this and it's fine"
+    # from "the framework never saw this at all". Purely mechanical: split on clause
+    # boundaries and report any clause none of the predicates/ambiguous entries above were
+    # extracted from — never an attempt to enforce the prose itself.
+    _covered = [str(p.get("source", "")).lower() for p in preds] + \
+        [str(a.get("span", "")).lower() for a in ambiguous if isinstance(a, dict)]
+    unenforceable: list = []
+    for clause in re.split(r"[;,.]\s+|\s+\band\b\s+|\s+\bor\b\s+", raw.strip()):
+        clause = clause.strip().strip(".,;")
+        if len(clause) < 8:  # too short to be a real, checkable instruction
+            continue
+        cl = clause.lower()
+        if any(span and span in cl for span in _covered):
+            continue
+        if clause not in unenforceable:
+            unenforceable.append(clause)
+
     # Tighten duplicates.
     out: list = []
     seen: dict = {}
@@ -251,7 +277,8 @@ def parse_constraints(text: str | None) -> dict:
         seen[k] = p
         out.append(p)
 
-    return {"text": raw, "predicates": out, "ambiguous": ambiguous}
+    return {"text": raw, "predicates": out, "ambiguous": ambiguous,
+            "unenforceable": unenforceable}
 
 
 def cost_target(target: float, per_task_ceiling: dict) -> dict:
@@ -386,5 +413,6 @@ def check_constraints(
         "recommendation": rec,
         "reasons": stop_reasons or warn_reasons,
         "ambiguous": parsed.get("ambiguous") or [],
+        "unenforceable": parsed.get("unenforceable") or [],
         "stop_condition_text": parsed.get("text", ""),
     }
--- a/core/cap_evolve/constraints.py
+++ b/core/cap_evolve/constraints.py
@@ -29,7 +29,7 @@ from __future__ import annotations
 
 import re
 
-__all__ = ["parse_constraints", "check_constraints"]
+__all__ = ["parse_constraints", "check_constraints", "cost_target"]
 
 #: Recognized kinds. ``max_*`` are ceilings (violated when actual >= target);
 #: ``target_val_score`` is a goal (satisfied when actual >= target).
@@ -254,6 +254,41 @@ def parse_constraints(text: str | None) -> dict:
     return {"text": raw, "predicates": out, "ambiguous": ambiguous}
 
 
+def cost_target(target: float, per_task_ceiling: dict) -> dict:
+    """Cost a ``target_val_score`` against measured per-task headroom before accepting it.
+
+    ``per_task_ceiling`` maps ``task_id -> highest rate that task can structurally reach``
+    (e.g. ``min(0.95, measured_component_cap)`` — a task whose COMMUNICATE component rate
+    measures 0.667 cannot reach 1.0 no matter what a capability edit does). The costed
+    ceiling is the mean of those caps, i.e. ``1 - sum(1 - ceiling_t) / n``: the score no
+    capability edit can exceed. A target above it is not a capability problem and should be
+    renegotiated with the user rather than chased with more iterations.
+
+    Motivating case: docs/TAU2_SUMMARY.md's own ceiling analysis found a ≈0.92 ceiling set by
+    three tasks capped by things no edit could touch (a leaking user simulator, two
+    COMMUNICATE-component rates) — two points of slack a 90%-target run had no way to close.
+    """
+    if not per_task_ceiling:
+        return {"target": float(target), "feasible": None,
+                "reason": "no per-task ceiling data — cannot cost this target"}
+    ceiling = sum(per_task_ceiling.values()) / len(per_task_ceiling)
+    feasible = float(target) <= ceiling + 1e-9
+    return {
+        "target": float(target),
+        "costed_ceiling": round(ceiling, 4),
+        "headroom": round(ceiling - float(target), 4),
+        "feasible": feasible,
+        "reason": (
+            f"target {float(target):.4f} is "
+            f"{'within' if feasible else 'ABOVE'} the costed ceiling {ceiling:.4f} "
+            f"(mean of {len(per_task_ceiling)} per-task caps) — "
+            + ("reachable by a capability edit" if feasible else
+               "NOT reachable by any capability edit; renegotiate the target or the caps "
+               "before spending more budget chasing it")
+        ),
+    }
+
+
 def check_constraints(
     parsed: dict,
     *,
@@ -265,6 +300,7 @@ def check_constraints(
     metric_calls: int = 0,
     regressed_tasks: list | None = None,
     warn_frac: float = 0.8,
+    per_task_ceiling: dict | None = None,
 ) -> dict:
     """Check every parsed predicate against MEASURED actuals from the run dir.
 
@@ -315,9 +351,17 @@ def check_constraints(
         elif kind == "target_val_score":
             actual = None if best_val is None else float(best_val)
             met = actual is not None and actual >= float(p["target"]) - 1e-9
-            rows.append({**p, "actual": actual, "satisfied": met, "violated": False,
-                         "note": "checked on the FULL-val mean only; a subset screen "
-                                 "can never satisfy this"})
+            row = {**p, "actual": actual, "satisfied": met, "violated": False,
+                   "note": "checked on the FULL-val mean only; a subset screen "
+                           "can never satisfy this"}
+            if per_task_ceiling:
+                cost = cost_target(float(p["target"]), per_task_ceiling)
+                row["cost"] = cost
+                if cost.get("feasible") is False:
+                    warn_reasons.append(
+                        f"target_val_score {p['target']:.4f} is above the costed ceiling "
+                        f"{cost['costed_ceiling']:.4f} — not reachable by any capability edit")
+            rows.append(row)
             if met:
                 stop_reasons.append(f"score goal met on full val "
                                     f"({actual:.4f} >= {p['target']:.4f})")
--- a/core/cap_evolve/dashboard.py
+++ b/core/cap_evolve/dashboard.py
@@ -337,7 +337,7 @@ _ALGO_MARKERS = (
 #: ``skillopt_step`` is the one kind deliberately absent, and for a different reason —
 #: it is not a legacy record but epoch DETAIL logged alongside a ``step`` for the same
 #: candidate on the same (current) runs, so it never carried a graph anyone needs.
-_STEP_KINDS = ("step", "gepa_val_gate", "accept", "reject")
+_STEP_KINDS = ("step", "gepa_val_gate", "accept", "reject", "provisional")
 
 #: Kinds whose presence means "this candidate was accepted" without an ``accept`` field.
 _ACCEPT_KINDS = ("accept",)
@@ -780,7 +780,14 @@ def reduce_run(run_dir) -> dict:
         # not a rejection and not a failure: the gate declined to judge, so the edit's
         # quality is unknown. Collapsing it into "rejected" (the old behaviour) told
         # the reader a measured verdict existed when none did.
-        if cid in indecisive_ids:
+        if kind == "provisional":
+            # Directionally positive (Δ>0) but not yet gate-significant, and the driver
+            # chose to buy more trials on this SAME candidate (scripts/grow.py) instead of
+            # a final accept/reject — neither "accepted" nor "rejected" describes that, and
+            # falling through to the accepted/rejected branch below would misreport it as
+            # one or the other while it is still pending.
+            status = "provisional"
+        elif cid in indecisive_ids:
             status = "indecisive"
         elif val is None and not per and kind not in ("accept", "reject"):
             # ``failed`` means NO VERDICT AND NO MEASUREMENT — a step that produced
@@ -1063,6 +1070,7 @@ def reduce_run(run_dir) -> dict:
         reason = str(n.get("reason") or "")
         verdict = ("accept" if n["status"] == "accepted"
                    else "indecisive" if n["status"] == "indecisive"
+                   else "provisional" if n["status"] == "provisional"
                    else "reject" if n["status"] == "rejected" else "no measurement")
         gs = n.get("gate_stats") or {}
         m_delta = re.search(r"Δ̄?\s*=\s*([+-]?\d*\.?\d+)", reason)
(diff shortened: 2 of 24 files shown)


OsherElhadad:
## Real end-to-end validation (network-connected, real `aws/gpt-oss-120b` rollouts)

The implementation agent's sandbox had no network egress, so it deferred live validation. I ran it for real from a network-connected session:

- Branch checked out at `05e607d5`, run via `examples/tau2_airline` project scaffold, `capevolve.smoke.real.yaml` (2 tasks, 1 trial, 1 iteration, real `claude-code` optimizer at `aws/claude-haiku-4-5`, real target agent+user simulator at **`aws/gpt-oss-120b`** through the internal litellm proxy).
- Full loop completed: check → baseline → optimize → gate → finalize → report → dashboard. `report.md`:
  ```
  Best candidate: seed
  Baseline val: 0.5 ± 0.5
  Best val: 0.5 ± 0.5
  Held-out test (optimized skills): 0.5 ± 0.5  (pass^1=0.500)
  Val→test gap: +0.0
  Iterations: 1
  ```
- **Confirmed live: the resolvable-effect-size fix is real, not just unit-tested.** The candidate's rejection reason in `rejected.jsonl` reads:
  `"paired Δ̄=+0.5000 <= 1.0·SE=0.5000 (SE=0.5000, n=2, resolvable effect size 2·SE=1.0000)"`
  — exactly the annotation this PR set out to add next to every gate verdict, produced by a real gate decision on real rollout data.
- Correctly rejected at this sample size (n=2, SE=0.5) — matches `docs/TAU2_SUMMARY.md`'s own documented noise-floor problem (small splits can't resolve real effects); this is the gate working as intended, not a bug.
- `dashboard.html` generated for the run; not screenshotted here (no browser-automation tool available in this pass), but present at the run dir and renders the candidate graph.

This clears the "no real validation" gap noted in the PR description. I did not run the full 30/30/20 tau2_airline benchmark here — that's reserved for the dedicated final task after this PR (and the memory-system and crash-fix PRs) merge, per the overall plan.

One correction to the PR description: `aws/gpt-oss-120b` **is** reachable on this project's own local litellm credentials (repo-root `.env`, not committed) — it was a different, more restricted credential that lacked access, not a hard platform limitation. The upcoming full run is unblocked.

eranra:
Context from a benchmark run, plus a compatibility check that may be useful before this lands.

## We can't measure this PR's improvements yet

We ran spreadsheetbench smoke with `agent-optimize` on `main` (`323ed3b5`) — [run 32971129203](https://github.com/skillberry-ai/cap-evolve/actions/runs/32971129203) — and a follow-up at `trials=10` ([run 33046360451](https://github.com/skillberry-ai/cap-evolve/actions/runs/33046360451)). Both are on `main` + #407 only, so **neither includes this PR**. Any number we report from them is not a measurement of the statistical-rigor fixes here, and we don't want it read as one.

The plan is to re-run once #402 (and #405, for the process narrative) are on `main`, and report that number. **Are there blockers to landing them?** Happy to help with anything outstanding.

## What we found on `main`'s agent-optimize, which overlaps this PR's themes

Run 32971129203 published `best=seed` / "null result" with all 3 rounds rejected — but its own `work/round_i0.json` recorded both comparisons for `cand_1`:

| reference | delta | threshold | verdict |
|---|---|---|---|
| parent (`round.py`'s default) | 0.0333 | 0.0440 | reject |
| control (same-round replicate) | 0.0556 | 0.0341 | **accept** |

0.0556 also cleared that round's own `evidence_bar` of 0.0444. Rounds i1/i2 rejected under both modes. So one candidate mattered and the drift-carrying default reference discarded it — the null was a gate-reference artifact, not an absence of signal. #407 fixes that in the CI `stop_condition` (requiring `--gate-against control`) rather than changing `round.py`'s default, which felt like your call given this PR owns that area.

This is adjacent to two items in #401's list — "report the resolvable effect size next to every gate verdict" and "require plural byte-identical null-control blocks by default". Flagging in case it changes how you'd scope those: on this benchmark the *reference* mattered more than the *strictness*.

Also measured, which may be useful for the `trials` guidance in #401: at `trials=3` the byte-identical replicate gap was **0.0444**; at `trials=10` it is **0.020**. Note the reported across-task `stderr` barely moved (0.144 → 0.132) because it is dominated by genuine task heterogeneity — it's the *paired* per-task SE that trials actually buys, which is what the gate uses.

## Compatibility with #407 — checked, mostly clean

#407 makes `commit.py` read `round.py`'s persisted `work/round_i<N>.json` so the published `gate_decisions[]` carries real numbers instead of nulls (an agent writes prose, so `dashboard.reduce_run`'s regex matched nothing).

- **`round.py`: no conflict.** Your diff here is +20/−1 and touches none of the keys that reader depends on — `parent`, `candidates[].gate_delta`, `gate_threshold`, `gated_against`, `control_relative`, `evidence_bar`, or the `round_i{spent.iterations}` filename. #407 also has a test (`test_commit_reads_only_fields_round_actually_writes`) that fails loudly if any of those ever move, so a future change here won't break it silently.
- **`dashboard.py`: expect a textual conflict.** Both PRs edit it — this one adds compliance-logging events, #407 changes the `gate_decisions[]` fields and copies structured gate fields onto the candidate node. Semantically independent; just needs a merge resolution. Happy to rebase #407 onto whichever of these lands first.

OsherElhadad:
## Round 2: fixes found via live observation of a real running tau2-bench experiment

On top of this PR's first round (statistical-rigor fixes), I ran/watched a real agent-optimize
session against tau2-bench and found 6 concrete, generalizable gaps in the framework wiring —
none of them tau2-specific.

**1. JOURNAL.md was never seeded in agent-optimize's continuous-session mode (root cause).**
`harness._seed_journal` is the only thing that creates `workdir/JOURNAL.md`, and it was only
ever called from the deterministic per-iteration path (`_augment_instructions`). `host.py` never
called it, so `JOURNAL.md` never existed in an agent-optimize run — `_reconcile_journal` logged
"empty handover" every single round regardless of what the optimizer actually did. Fixed by
seeding it onto the seed candidate before round 1 (`host.py`) and re-seeding it onto whichever
candidate is `$BEST` after every `commit.py` call, so the next round's `cp -r "$R/candidates/$BEST"`
always carries a clean append target forward — no changes to `_reconcile_journal`'s
escalate-vs-raise behavior, only two new call sites. Documented in SKILL.md +
`references/algorithm.md` (write protocol, marker, `## Iteration <cid>` heading) so any future
continuous-session algorithm gets this for free.

**2. Single-candidate-per-round was the biggest lever for faster progress.** Two real rounds each
proposed exactly 1 candidate, paying the round's fixed overhead (baseline + 2 null-control
replicates) for one shot at the gate. `round.py` already supports parallel siblings; the gap was
that nothing told the optimizer to default to using it. SKILL.md's round-proposal step and
`host.py`'s driver-prompt template now state N≥3 sibling candidates as the explicit default, not
a buried alternative.

**3. `commit.py` never passed structured gate numbers to events.** The dashboard showed
`stderr`/`n`/`k_se`/`threshold`/`delta`/`resolvable_effect_size` as null for every agent-optimize
candidate — only a hand-typed prose `--note` carried them, and `dashboard.py` had to regex-parse
that note. `round.py`'s table now also carries these fields (straight from `gate_check.py`'s own
JSON), `commit.py` attaches them to the events it writes (in addition to, not replacing, the
prose note), and `dashboard.py` prefers them over the regex fallback. Benefits any algorithm using
`commit.py`, not just agent-optimize; field names match what the dashboard already expects.

**4. `parse_constraints` silently dropped unenforceable behavioral prose.** This run's
`stop_condition` said "use screen.py before paying for full val each round" — a clause with no
number in it, so none of the numeric/task-protection patterns even looked at it, and it vanished
with no signal either way. `parse_constraints` now reports an `unenforceable` list (clause-level,
purely mechanical — never an attempt to enforce the prose itself), distinct from `ambiguous`
(a number it saw but couldn't resolve). Surfaced through `check_constraints` → `spend.py` so both
the optimizer and the dashboard can see "this instruction wasn't something the framework could
check automatically." New tests in `test_agent_optimize_constraints.py`.

**5. diagnose's clustering used a fixed English stopword list, not corpus-relative filtering.**
22 of 30 failing tasks merged into one mega-cluster because the benchmark's own recurring
vocabulary ("action check state write") isn't in any generic English list.
`skills/phases/diagnose/scripts/cluster.py` now also strips any stem that recurs in >65% of the
batch's own feedbacks, on top of (never instead of) the fixed list — purely mechanical,
benchmark-agnostic by construction. New `core/tests/test_diagnose_cluster.py` reproduces the
mega-cluster with the old single-tier filter and shows the fix separating it into discriminated
clusters.

**6. `run-optimizer/scripts/run.py` buffered the whole session in memory.** A multi-hour
continuous optimizer session (agent-optimize's execution mode) only wrote its transcript after
`subprocess.run()` returned, so a crash lost everything. Switched to `Popen` with the stdout pipe
teed to `--transcript` line-by-line as it arrives (a background thread drains stderr concurrently
to avoid the two-pipe deadlock), while the function's return contract and existing error handling
are unchanged — `test_optimizer_transcript.py`'s full suite (redaction, streaming-flag
substitution, cost/stop parsing, secret scrubbing) passes unmodified.

All 6 implemented — none deferred. Full `core/tests` suite: **964 passed, 12 skipped**. Ran with
the editable install pointed at this branch's `core/`.

Pushed directly to this branch as incremental round-2 work.

OsherElhadad:
## Another real bug found live, fixed

While observing the fresh rerun (finalrun2, driven by all round-2 fixes), saw:
`{"kind": "optimizer_context_warning", "what": "guidance/system-prompt", "error": "[Errno 17] File exists..."}`

Root cause: `_inject_optimizer_context`'s per-capability guidance-staging `shutil.copytree()` call (the first one in the function) was the ONE call site in this function missing the `if dst.exists(): shutil.rmtree(dst)` idempotency guard that every OTHER copytree call in the same function already has. On a rerun sharing the same base dir (e.g. after killing and relaunching a run), the leftover guidance/<cap> dir from the prior run collided, the exception was silently caught+logged, and — worse — the optimizer could keep reading STALE guidance instead of the current capability skill content, since the copy never actually happened.

Fixed by adding the same guard used everywhere else in the function. Non-fatal in this specific run only because the SKILL.md content happened to be identical between the two runs, but this is a real robustness gap for any rerun scenario. Test suite: 976 passed, 12 skipped, 0 failed.

OsherElhadad:
## Architectural clarification implemented: gate decisions belong to the optimizer, not the code

The user clarified the design intent explicitly: nothing between the deterministic baseline eval and the deterministic final eval should be code-decided — including the accept/reject gate. Investigated whether this needed new code: **it did not**. `gate_check.py` already only reports statistics (delta, SE, resolvable_effect_size, two-seed-block agreement); `commit.py` already accepts a free-form `--decision`; `set_best()` is already an unconditional setter; `--reject-basis driver_judgement` already exists for logged overrides. The only thing forcing rigidity was one line of SKILL.md prose: "Accept only on verdict: accept."

Rewrote that section so the optimizer reads the printed statistics the way a researcher reads a stats printout and decides for itself, with a required one-line `--note` justification as the audit trail (moved the fuller rationale + explicit safety argument into `references/algorithm.md`'s new "Gate as evidence" section to stay inside the 5000-token SKILL.md body budget). Also scrubbed two accidental doc citations from these files per the generality requirement — no skill should reference another project-specific artifact by name.

Verified live in the round that prompted this: structured gate fields (delta/threshold/resolvable_effect_size/k_se/n) are real numeric fields on every accept/reject event now, confirming the earlier commit.py fix works end to end, and the JOURNAL.md file is now correctly listed as an editable surface in the driver prompt (the seeding fix works structurally) — though it's still ending up empty most rounds, which looks like a compliance gap distinct from the scaffolding one already fixed; investigating separately.

Full test suite: 973 passed, 12 skipped, 3 failed — 2 are the previously-documented full-suite-order flakes; the 3rd (a new test I added for the port-leak fix) shows the exact same pass-alone/flake-in-full-suite pattern on inspection, not a logic bug. `check.py` passes clean (ok: true, no problems).

OsherElhadad:
## Round 4: journal-write trigger, sequential (provisional) mechanism banking, targeted-holdout screening

### 1. Why JOURNAL.md was still ending up empty every round

Confirmed the seeding fix (round 2) worked structurally — `JOURNAL.md` exists from round 1 and is listed as editable in the driver prompt — but the optimizer had **zero standing instruction on when to write into it**. SKILL.md's Agent-mode loop went straight from step 5 ("commit the decision") to the next round with no mention of the journal at all, and `host.py`'s driver-prompt only said the file was "worth reading from round 2 on" — never "write here before you end your turn". `commit.py` only *reconciles* what the optimizer already wrote; nothing told it to write anything.

Fix: SKILL.md's loop now has an explicit numbered **step 6** ("write the handover before ending this round — append one `## Iteration <cid>` entry... below the marker"), and `host.py`'s JOURNAL.md paragraph was rewritten to name the trigger point explicitly (after `commit.py`, every round, accept or reject). This was the actual root cause — the scaffolding was fine, the loop just never told the optimizer to use it.

### 2. Sequential mechanism banking (`provisional` candidates)

A candidate with Δ>0 that misses the significance gate is real partial evidence, not a null result. Implemented:

- **`commit.py --decision provisional`** — a third decision alongside accept/reject, booked as its own `provisional` event kind with a `verdict` field. It snapshots the candidate for the audit trail but deliberately does **not** call `set_best` or `harness.record_iteration` — the round isn't over, so the stall counter/LEDGER/JOURNAL must not advance for it. A later real `accept`/`reject` on the *same* candidate id is still allowed (the reuse guard only watches accept/reject events).
- **`scripts/grow.py`** — runs `--add-trials` new trials on the candidate's *unmodified* working copy under a throwaway tag, pools them with the existing rollouts via a new `loop.pool_split_results` helper, re-runs the same paired gate at the pooled n, then merges the new rollout files onto the candidate's own tag (renumbered past its existing trial indices) so every later reader (`gate_check.py`, `commit.py`, the dashboard) sees one candidate at the pooled n with zero special-casing. Recommends `promote` / `grow_again` / `abandon`.
- **The pooling is real pooling.** `pool_split_results` concatenates each shared task's `trial_rewards` and re-aggregates from scratch — averaging two means, or two verdicts, would silently discard the fact that more trials narrow the SE (a 5-trial run and a separate 5-trial run are one candidate at n=10, not two half-strength votes).
- **The cap, and why:** growth is capped at 2 additional rounds (`grow.py --max-growth-rounds`, default 2). The honest risk is a provisional lineage accumulating trials on a false positive indefinitely — 2 rounds is enough to roughly double/triple n (where most of the SE reduction from added trials lives) while turning "maybe next round" into a bounded, auditable cost. Past the cap the candidate must be finally promoted or abandoned.
- **`gate_check.py`** now surfaces `directionally_positive_but_inconclusive` so the driver notices the option without computing Δ>0 itself.
- **`dashboard.reduce_run`** renders `provisional` as its own status (never counted as accepted/rejected, never moves `best`/`best_id`) instead of choking on the new event kind or misreporting it.

This is deliberately narrow: a provisional candidate only ever buys more *measurement precision* on itself. It never becomes the base a sibling edit builds from until it's actually accepted — that's the line that keeps this from becoming "compound edits on unconfirmed evidence," which a prior post-mortem already flagged as the failure mode to avoid.

### 3. Targeted holdout for collateral-damage detection

Prose-only, as recommended: SKILL.md's screening step and `references/algorithm.md` now say to run two screens when targeting one diagnosed failure cluster — `--broken <ids> --k <n> --holdout-frac 0` for the cluster itself (pure signal, no dilution), and a separate call with no `--broken` and a normal `--holdout-frac` for collateral damage elsewhere (its holdout draws from tasks the parent already *passes*, which a diagnosed failure cluster is not, by construction). No code changes needed — `screen.py` already supports this.

### Tests

`core/tests/test_agent_optimize_provisional.py` (9 tests, all new):
- `pool_split_results` unit tests (concatenates trial vectors correctly, not two means; keeps tasks present on only one side).
- A synthetic **promote-after-pooling** scenario and a synthetic **still-abandoned-after-pooling** scenario, both built directly on `cap_evolve.gate.decide` with hand-verified numbers (a real δ>0-but-rejected first measurement, then either genuinely resolves or genuinely doesn't after pooling).
- `commit.py --decision provisional` end-to-end: no premature `best_id`/iteration/JOURNAL advance, the reject-basis guard rejects it, and a real accept afterward on the same candidate id still works.
- `dashboard.reduce_run` on a run containing a `provisional` event — renders correctly, doesn't crash, doesn't move `best_id`.
- `grow.py` end-to-end with a scripted deterministic adapter — new trials land under a throwaway tag and get merged onto the candidate's own tag with continued trial indices.

### Verification

- `CAPEVOLVE_CORE=core python skills/algorithms/agent-optimize/scripts/check.py` → `"ok": true`.
- `python skills/_registry/lint_skills.py skills` → `OK — 20 skill packages, no authoring violations` (SKILL.md had to be re-trimmed in several existing paragraphs to land the new steps under the 5000-token body budget — margin was already only ~20 tokens before this round).
- `python -m pytest core/tests -q` → **973 passed, 12 skipped, 0 failed** (full suite, no flakes this run; editable install pointed at this branch's `core/`).

Not merging — leaving that to you as usual.

PR #403, opened 2026-08-26
fix(core): protect evaluate_candidate in run_step the way the optimizer call is protected

### LINKED ISSUE #286: An exception from an adapter's live() aborts the whole run — evaluate_candidate is unprotected while the optimizer call beside it is
`harness.run_step` deliberately protects the optimizer call:

```python
except Exception as e:
    # A failed proposal (e.g. a transient optimizer/API error) must not abort a
    # long run — leave the workdir as the parent copy so the candidate == parent
    # and the gate simply rejects it (a wasted iteration, not a crash).
```

The `evaluate_candidate(adapter, workdir, …)` call immediately below it has no such protection. Since `evaluate_candidate` enters the adapter's `live()` context (via `_live`), **any** exception an adapter raises there — or during its rollout setup — propagates out of `run_step` and aborts the run.

`live()` is a documented extension point ("Override `live` when the live state needs explicit teardown, or to yield a richer `ctx`"), so adapters are invited to do real work there: acquire a sandbox, validate a candidate, set up a worktree. Every one of those can fail for a reason specific to *one candidate*, and the stated principle — a bad candidate costs an iteration, not the run — should apply identically.

The cost scales with run length: on a spreadsheetbench full run this would destroy ~10 hours of work *and* the sealed 639-task evaluation, which is the only number the run exists to produce and cannot be re-derived without re-running everything.

Found while adding a candidate-validation guard in #282; worked around in #285 by having the adapter report rather than raise, and by failing each rollout cheaply instead. That fix is adapter-local, so the general gap remains for tau2, swebench, skillsbench and any future adapter.

Suggested fix: wrap the candidate evaluation the same way the optimizer call is wrapped — log an event, treat the candidate as rejected (parent unchanged), continue. Worth distinguishing it from a genuine infrastructure abort so an unrunnable *environment* still fails loudly rather than looking like ten rejected candidates.

OsherElhadad:
Closes #286.

## What was wrong

`harness.run_step` deliberately wraps the optimizer call in `except Exception` — a bad proposal costs one iteration, not the run. The `evaluate_candidate(adapter, workdir, ...)` call right below it has no equivalent protection, so any exception an adapter raises in `live()`/rollout setup propagates out of `run_step` and aborts the whole run — destroying unrecoverable long-run work (e.g. a sealed test evaluation).

#285 worked around this locally in the spreadsheetbench adapter (report instead of raise). This PR generalizes that into the harness so tau2/swebench/skillsbench/future adapters all get the protection without opting in individually.

## Fix

`core/cap_evolve/harness.py`:

- Wrap the `evaluate_candidate` call in `run_step` the same way the optimizer call beside it already is: on exception, log an `evaluate_error` event, reject the candidate (parent unchanged, stall counted — same "wasted iteration" discipline as `optimizer_error`), and continue the run. New helper `_eval_error_step` mirrors the existing `_tamper_step` shape but is a real rejection (not indecisive) — there's no candidate-side wrongdoing to exempt from the stall counter here, just a failed evaluation, exactly like a failed proposal.
- To distinguish a genuine infrastructure/environment failure from a single bad candidate: track a consecutive-failure streak on the run (`run_dir._consecutive_eval_errors`, reset on any success). After `_MAX_CONSECUTIVE_EVAL_ERRORS` (3) raises in a row, re-raise instead of swallowing — so an unrunnable adapter/environment still fails loudly rather than silently looking like N rejected candidates.

No new abstraction layer — this mirrors the existing `except Exception` pattern already beside it.

## Tests

New `core/tests/test_evaluate_error_wiring.py`:
- `test_a_single_live_error_rejects_the_candidate_and_continues` — a `live()` raise on one candidate rejects it (parent unchanged, stall incremented, `evaluate_error` event logged) and the run continues to a clean next step.
- `test_n_consecutive_live_errors_abort_loudly_as_an_infra_failure` — an adapter whose `live()` never recovers raises `RuntimeError` out of `run_step` after 3 consecutive failures, instead of being silently absorbed.
- `test_a_success_after_failures_resets_the_streak` — a success after a failure resets the consecutive-error counter.

```
$ python -m pytest tests/test_evaluate_error_wiring.py -q
...
3 passed in 0.70s
```

Targeted regression run:
```
$ python -m pytest tests/ -k "harness or run_step or tamper or integrity or optimizer_error or capability_validation or regression_gate or iteration_record or spreadsheetbench" -q
194 passed, 11 skipped, 764 deselected in 41.46s
```

Full suite:
```
$ python -m pytest tests/ -q
957 passed, 12 skipped in 159.85s
```

## Smoke validation

`examples/toy_calc/run.sh` (zero-API, deterministic full pipeline: check → baseline → optimize → gate → finalize → report) ran end to end unaffected:

```
{
  "best_id": "cand_0001",
  "finalized": true,
  "baseline_val": 0.0,
  "best_val": 1.0,
  "test_reward": 1.0,
  "iterations": 3,
  ...
}
```

`examples/tau2_airline/smoke.sh` requires cloning tau2-bench and a Python 3.12 venv + LLM API access, which wasn't available in this environment, so `toy_calc` (zero-API, exercises the same `run_step` path) was used as the real end-to-end smoke check instead.

### DIFF
--- a/core/cap_evolve/harness.py
+++ b/core/cap_evolve/harness.py
@@ -1505,6 +1505,47 @@ def _tamper_step(run_dir: RunDir, *, cid: str, parent_id, workdir: Path,
     }
 
 
+_MAX_CONSECUTIVE_EVAL_ERRORS = 3  # N raises in a row -> the environment, not the candidates
+
+
+def _eval_error_step(run_dir: RunDir, *, cid: str, parent_id, workdir: Path, error: str,
+                     current_val: SplitResult, optimizer_seconds: float,
+                     opt_cost_usd: float, opt_tokens: int, optimizer_error,
+                     rejected=None) -> dict:
+    """A REJECTED step (mirrors the optimizer_error handling above) for a candidate
+    whose ``evaluate_candidate`` call raised — e.g. an adapter's ``live()``/rollout
+    setup blew up on this one candidate (#286). Same discipline as the optimizer
+    call beside it: log it, reject the candidate (parent unchanged, stall counted),
+    keep the run going. Unlike a tamper/validation step this is NOT indecisive —
+    there is no candidate-side wrongdoing to exempt from the stall counter, just a
+    failed evaluation, exactly like a failed proposal.
+    """
+    reason = f"candidate evaluation raised: {error}"
+    run_dir.snapshot(cid, workdir, ignore=_SNAPSHOT_IGNORE)
+    record_iteration(run_dir, workdir, cid, parent_id=parent_id, accepted=False,
+                     reason=reason, val=None, parent_val=current_val.reward,
+                     optimizer_seconds=round(optimizer_seconds, 2),
+                     opt_cost_usd=round(opt_cost_usd, 6), opt_tokens=opt_tokens)
+    summary = f"candidate {cid} (evaluation failed)"
+    if rejected is not None:
+        rejected.add(cid, summary, reason, None)
+    run_dir.record_spend_warnings()
+    return {
+        "candidate_id": cid,
+        "accepted": False,
+        "decision": {"accept": False, "reason": reason, "delta": 0.0, "threshold": 0.0},
+        "candidate_val": None,
+        "parent_val": current_val.to_dict(),
+        "eval_error": error,
+        "regressions": [],
+        "optimizer_seconds": optimizer_seconds,
+        "optimizer_usd": opt_cost_usd,
+        "optimizer_tokens": opt_tokens,
+        "optimizer_error": optimizer_error,
+        "workdir": str(workdir),
+    }
+
+
 def run_step(
     adapter,
     *,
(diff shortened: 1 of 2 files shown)


OsherElhadad:
Review follow-up (pushed `0586467f`): the eval-error step is now **indecisive**, not a rejection.

The original patch counted an `evaluate_candidate` raise against the stall counter and filed it in the rejected memory. That contradicts the doctrine the rest of the codebase already enforces for an unmeasured candidate:

- `gate.decide`'s `min_coverage` guard returns `indecisive=True` — *"the evaluation measured the infrastructure, not the edit — not counted as a rejection"*. A raise is the 0%-coverage case of exactly that.
- `_tamper_step` / gepa's tamper path: *"it never enters the pool or the rejected memory, and `accepted=None` leaves the stall counter untouched."*
- `test_infra_errors_not_zeros` documents the concrete cost of getting this wrong: *"The optimizer burned an iteration proving to itself that iters 1–2 were infra flakes, not content regressions."*

The `optimizer_error` path this was modelled on is *not* the analogue — there the workdir stays == parent, so the gate scores a genuine Δ of 0 on a real measurement. An eval raise produces no measurement at all.

So: `indecisive=True`, no `rejected.add`, a `step_indecisive` event. `_MAX_CONSECUTIVE_EVAL_ERRORS` remains the (correct) detector for a truly unrunnable environment — the stall counter should not be guessing at that.

Full suite green locally: `957 passed, 12 skipped`.

Not addressed here (separate issue): `gepa._full_val_gate` and `_eval_minibatch` still call `evaluate_candidate` unguarded, so a raise there aborts a GEPA run. Fixing that needs an iteration-level guard in gepa's own loop, not a one-line mirror.

Issue #404, opened 2026-08-26
Memory: make cross-iteration memory truly pluggable (extract evograph wiki), and tell the optimizer when a run is ending


OsherElhadad:
Follow-up to #400. Two things scoped out of that PR to keep it surgical:

## 1. Memory is not yet actually pluggable

#400 fixes the confirmed empty-journal bug and adds the three new per-run
summary files (INSIGHTS.md / META_INSIGHTS.md / FRAMEWORK_IMPROVEMENTS.md),
but the LEDGER/JOURNAL/INSIGHTS scheme is still hardcoded into
`core/cap_evolve/harness.py` — there is no real `memory_skill` selection or
plug-in interface yet. `skills/phases/intake/inputs/INPUTS.md` now documents
`memory_skill` as an input (default, and only wired, value: `md-files`), but
choosing anything else does nothing.

Still to do, per the original issue's scope item 1:
- Extract the wiki format out of `skills/algorithms/evograph/SKILL.md`
  (`<run_dir>/wiki/`: weakness nodes, solution cards, per-round results) into a
  standalone, selectable memory skill, per evograph's own maintainer note.
- Define a minimal common interface (write handover, read prior insights,
  summarize) so a memory skill plugs into `harness.py` the way capability/
  optimizer skills already do (`_augment_instructions`, `record_iteration`).
- Delete (or thin-point) `evograph` once its format has a real owner.

## 2. The optimizer isn't told when a run is ending

Real-run validation for #400 (`examples/toy_calc`, a genuine `claude-code`
optimizer session, 1 iteration) showed: `JOURNAL.md` and `PROCESS.md` were
populated with real, substantive content, but `META_INSIGHTS.md` and
`FRAMEWORK_IMPROVEMENTS.md` — whose prompt text says "update AT LEAST once,
at the end of the run" — were left empty. The optimizer has no signal in its
prompt telling it "this is the last iteration" (or how much budget remains),
so it reasonably treated the optional instruction as lower priority than the
REQUIRED `PROCESS.md`.

Fix idea: thread a cheap "iterations remaining" / "this may be your last
iteration" signal (something `spend.py`-adjacent already computes) into
`_augment_instructions`'s pointer text, so the optional summary files get a
real nudge near the actual end of a run instead of relying on the optimizer
to guess.

Issue #400, opened 2026-08-26
Memory: make cross-iteration memory composable (skill-selected at intake), fix journal-trimming bug, add process-narrative HTML report


OsherElhadad:
## Problem

Cross-iteration memory today is a single hardcoded scheme, not a composable, per-project choice, and it has a real, confirmed data-loss bug:

- `core/cap_evolve/harness.py:709-1148` bakes in exactly one memory format: framework-owned `LEDGER.md` (regenerated, factual) + optimizer-owned `JOURNAL.md` (append-only). `core/cap_evolve/memory.py` (49 lines) adds two audit JSONLs (`RejectedMemory`, `History`) that "nothing in the core re-reads" (memory.py:8) — pure log, not usable memory.
- **Confirmed trimming/loss bug**: `.capevolve/run_wide/JOURNAL.md`, iteration `cand_0003`: `## Iteration cand_0003 — (no handover written by the optimizer)`. The harness silently accepts an empty handover with no enforcement that the optimizer actually wrote insight (harness.py:57-60 in that run's log).
- `_diff_capabilities(..., max_chars=8000)` (harness.py:802-819) caps capability diffs shown per iteration — deliberate, but compounds the loss when journals are also thin.
- `skills/phases/intake/SKILL.md:65` interviews for capability/optimizer/algorithm but **never asks about memory** — it's not a Phase 1 choice today.
- A wiki-style memory format already exists and works, in the deprecated `skills/algorithms/evograph/SKILL.md` (`<run_dir>/wiki/`: weakness nodes, solution cards, per-round results, rendered by dashboard.py's Weakness-graph tab). The skill's own maintainer note says explicitly: *"move the wiki format contract into `agent-optimize` as an optional output... then delete this directory."* This is a generalize-and-extract job, not a green-field build.
- Skill injection into the optimizer's environment is real and works (`harness.py:1347-1450 _inject_native_skills`, driven by `skills/optimizers/registry.yaml`'s `skills_dir`/`instructions_file`) — but nothing verifies the optimizer's actual tool-call log used an injected skill. It's plumbing without enforcement or telemetry.
- `core/cap_evolve/dashboard.py:1452-1475` already has a reusable self-contained HTML template pattern (`render_html`, inline CSS/JS, JSON-island data injection, existing multi-tab structure e.g. Weakness-graph) — a new memory/process-narrative HTML tab should extend this, not introduce a second templating approach.

## Scope

[...]

PR #405, opened 2026-08-26
fix(core): escalate empty JOURNAL.md handovers, add per-run insight files, and a default-on process narrative (#400)

OsherElhadad:
Addresses #400 (partially — see Deferred below; follow-up filed as #404).

## What's implemented

**1. Fix the confirmed journal-trimming/empty-handover bug**
`core/cap_evolve/harness.py`'s `_reconcile_journal` used to silently substitute
`(no handover written by the optimizer)` when the optimizer wrote no `## Iteration`
entry, with no trace anywhere. It now:
- logs an `optimizer_context_warning` event, and
- stamps the journal itself with a visible `⚠ EMPTY HANDOVER (framework escalation)`
  marker instead of unremarkable filler text.

**Deliberate choice — escalation, not hard-fail.** The issue offered "hard-fail or
explicit escalation." I tried hard-failing (`EmptyHandoverError`) first, but it broke
28 unrelated unit tests across 8 files that pass small in-process fake "optimizer"
functions (testing gate/integrity/skillopt mechanics, not journal content) and never
touch `JOURNAL.md` at all — none of those are the bug this issue is about, and
rewriting 8 test files' fixtures to fake a handover was out of scope for a surgical
fix. Escalation (always logged + visibly flagged) fixes the actual "silent" part of
the bug without that blast radius. Real-run validation below confirms a working
optimizer's real handover flows through untouched.

The mock optimizer (`skills/optimizers/run-optimizer/scripts/_mock_apply.py`, the
offline test double) now writes a trivial handover too, matching what a real
optimizer does.

**2. Per-run summarized/verified insight files**
Three new optimizer-authored, run-level, append-only files — `INSIGHTS.md` (verified
per-task/per-mechanism findings), `META_INSIGHTS.md` (about the optimization process
itself), `FRAMEWORK_IMPROVEMENTS.md` (cross-run suggestions for cap-evolve itself) —
seeded into the workdir and folded back the same way `JOURNAL.md` already is
(`_seed_accumulator`/`_fold_accumulator`, generalizing the existing
`_seed_journal`/`_reconcile_journal` mechanic, minus the journal's RESULT stamp).
Unlike `JOURNAL.md` these are optional most iterations — no escalation on an empty
tail — but the optimizer prompt asks for at least one entry by the end of the run.
Excluded from candidate snapshots/diffs the same way LEDGER/JOURNAL/RUNMAP already are.

**3. Memory as a (partially) selectable component**
`skills/phases/intake/inputs/INPUTS.md` now documents `memory_skill` as an input
(default and only wired value: `md-files`, i.e. the scheme above). The full
pluggability (a real interface + the evograph wiki extraction) is deferred — see below.

**4. Default-on HTML process-narrative dashboard section**
`dashboard.py`'s `reduce_run` now reads the run-level JOURNAL/INSIGHTS/META_INSIGHTS/
FRAMEWORK_IMPROVEMENTS files plus the best candidate's `PROCESS.md` straight off disk
(`_read_narrative`, modeled on the existing `_read_evograph`) and a new JS section
renders them — with a small dependency-free markdown-to-HTML pass (headings, bullets,
blockquotes, bold) — reusing `render_html`'s existing self-contained JSON-island
pattern. Shown as its own "Process narrative" section for every run by default,
whenever any of those files have content — no algorithm special-casing.

**5. Skill injection: verified, not just wired**
Ran a real `claude-code` optimizer session (see Validation below) and inspected its
working dir: `.claude/skills/system-prompt/` and `.claude/skills/diagnose/` were
injected, and its `PROCESS.md` output used the diagnose skill's own failure-taxonomy
vocabulary verbatim (`KNOWLEDGE` / `BEHAVIORAL` / `CAPABILITY-GAP`) — real evidence the
optimizer read and used the injected skill, not just had it sitting unused. No CLAUDE.md
enforcement text was needed.

## Deferred (follow-up: #404)

- Full memory pluggability: extracting the evograph `wiki/` format into a standalone
  selectable memory skill + a real plug-in interface into `harness.py`. `memory_skill`
  is documented but not yet functionally selectable beyond the default.
- The optimizer isn't told when a run is ending, so the optional `META_INSIGHTS.md`/
  `FRAMEWORK_IMPROVEMENTS.md` "update at least once, at the end of the run" instruction
  is sometimes not honored on a short run — real finding from the validation run below.

## Validation (real, not fabricated)

Two real end-to-end runs via the actual `cap-evolve` CLI (not just pytest):

1. **`examples/toy_calc/run.sh`** (zero-API, deterministic mock optimizer) — full
   loop (check → baseline → optimize → gate → finalize → report → dashboard),
   3 iterations, confirmed `JOURNAL.md` has real per-iteration entries (mock optimizer
   now writes one), no empty-handover markers.
2. **A real `claude-code` optimizer session** against the same `toy_calc` capability
   (1 iteration, `$0.367252`, 14464 tokens, 66s wall — real API cost, no target-model
   calls since `toy_calc`'s target is deterministic arithmetic):
   - `JOURNAL.md` got a real, substantive entry (`## Iteration cand_0001 — add explicit
     arithmetic-answer-only output contract to system prompt`, with expected-effect
     reasoning, a `RESULT` stamp correctly showing `REJECTED` since val stayed 0.000,
     and no `EMPTY HANDOVER` marker anywhere).
   - `PROCESS.md` was filled in with a ranked-issue table tagging the failure
     `KNOWLEDGE` — the exact taxonomy from the injected `diagnose` skill.
   - `INSIGHTS.md`/`META_INSIGHTS.md`/`FRAMEWORK_IMPROVEMENTS.md` were seeded into the
     workdir (confirmed via `work/cand_0001/`) but left empty by the optimizer — the
     real finding written up in #404.
   - `dashboard.html`'s "Process narrative" section rendered the real journal content
     (confirmed via `grep -o "arithmetic-answer-only" dashboard.html`), with zero
     "no handover written" placeholder text anywhere in the output.
   - Run report (`report.md`) honestly reported a null result (val stayed 0.0, gate
     correctly rejected) — no fabricated gain.

Full test suite (`core/tests`, 952+ tests): all green except two pre-existing
flaky tests in `test_eventstream.py` (thread-timing / module-identity races under
full-suite ordering) confirmed to fail identically on `main` with none of this PR's
changes applied, and to pass standalone either way.

## Git identity
Commits authored as `Osher-Elhadad <Osher.Elhadad@ibm.com>`, no Claude/Anthropic
co-author trailer.

### DIFF
--- a/core/cap_evolve/dashboard.py
+++ b/core/cap_evolve/dashboard.py
@@ -25,9 +25,17 @@ The candidate **graph** schema (``reduced["graph"]``)::
         {"id", "parent", "children": [...], "status": seed|accepted|rejected|failed,
          "val", "stderr", "per_task": {task_id: reward}, "feedback": {task_id: str},
          "cost_usd", "tokens", "seconds", "optimizer_seconds", "runner_seconds",
-         "iteration", "reason", "epoch"?, "merge_of"?, "best_so_far"}
+         "iteration", "reason", "epoch"?, "merge_of"?, "best_so_far",
+         "gate_fields": {delta?, stderr?, n?, k_se?, threshold?, resolvable_effect_size?},
+         "screened": bool | None}
      ],
      "root": "seed", "best_id": "..."}
+
+``gate_fields`` holds whichever of those keys the commit event carried DIRECTLY (any
+algorithm's commit step may attach them to its accept/reject event); absent keys mean the
+gate-decisions table falls back to parsing the prose ``reason`` string instead. ``screened``
+is ``None`` when no event recorded compliance for this candidate's tag, else the
+``screened_before_fullval`` value read generically off any event that carries it.
 
 The **summary** schema (``reduced["summary"]``)::
 
@@ -36,7 +44,13 @@ The **summary** schema (``reduced["summary"]``)::
      "frontier": int, "tasks": [task_id, ...],
      "wall_clock_seconds", "optimizer_seconds", "runner_seconds",
      "cost": {optimizer_usd, runner_usd, total_usd}, "tokens": int,
-     "gate_warnings": [...], "diagnoses": [...], "git_log": [...]}
+     "gate_warnings": [...], "diagnoses": [...], "git_log": [...],
+     "controls": [{"tag", "reward", "stderr", "n", "iteration", "t"}, ...]}
+
+``controls`` lists null-control replicate evaluations — evaluate-only measurements with no
+candidate-graph node — read generically off any ``evaluate`` event carrying ``role:
+"control"`` or a truthy ``is_control`` field. Empty until an algorithm's evaluate step
+attaches either.
 
 Optional panels degrade silently: when per-task data / diffs / finalize are missing
 the renderer hides the panel rather than crashing.
@@ -643,7 +657,24 @@ def _read_narrative(root: Path, best_id: str | None) -> dict:
     optimizer wrote across iterations, plus the best candidate's final ``PROCESS.md``
     (why THAT iteration was done the way it was). Returns ``{}`` when none of these
     exist yet (e.g. a synthetic log with no real optimizer session).
+
+    Each file is compared against its own known seed-template text (harness.py's
+    ``_seed_journal``/``_seed_accumulator``/``_PROCESS_SEED``): a file whose content is
+    STILL exactly that template — no real entry ever appended — carries
+    ``"template_only": True`` so the renderer can flag it instead of presenting an
+    unedited instructional template as real optimizer narrative.
     """
+    try:
+        from . import harness
+        seed_by_name = {
+            "JOURNAL.md": harness._JOURNAL_SEED,
+            "INSIGHTS.md": harness._INSIGHTS_SEED,
+            "META_INSIGHTS.md": harness._META_INSIGHTS_SEED,
+            "FRAMEWORK_IMPROVEMENTS.md": harness._FRAMEWORK_IMPROVEMENTS_SEED,
+        }
+        process_seed = harness._PROCESS_SEED.strip()
+    except Exception:  # noqa: BLE001 — template detection is a nicety, not load-bearing
+        seed_by_name, process_seed = {}, None
     files = []
     for name, title in _NARRATIVE_FILES:
         p = _safe_subpath(root, name)
@@ -654,7 +685,9 @@ def _read_narrative(root: Path, best_id: str | None) -> dict:
         except OSError:
             continue
         if text:
-            files.append({"title": title, "text": _sanitize_text(text, 20000)})
+            seed_text = seed_by_name.get(name)
+            files.append({"title": title, "text": _sanitize_text(text, 20000),
+                          "template_only": bool(seed_text) and text == seed_text.strip()})
     process_text = None
     if best_id:
         p = _safe_subpath(root, "candidates", best_id, "PROCESS.md")
@@ -665,7 +698,8 @@ def _read_narrative(root: Path, best_id: str | None) -> dict:
                 process_text = None
     if process_text:
         files.append({"title": f"Process — best candidate ({best_id})",
-                      "text": _sanitize_text(process_text, 20000)})
+                      "text": _sanitize_text(process_text, 20000),
+                      "template_only": bool(process_seed) and process_text == process_seed})
     if not files:
         return {}
     return {"files": files}
@@ -763,6 +797,18 @@ def reduce_run(run_dir) -> dict:
             minibatch_evals.discard(ev.get("tag"))
         if ev.get("kind") == "minibatch":
             minibatch_evals.add(ev.get("tag"))
+
+    # Cheap-screen compliance, keyed by candidate tag: whether a candidate paid for a
+    # cheap screen before its full-val eval. Read generically off ANY event that carries a
+    # ``screened_before_fullval`` field (agent-optimize's ``agent_optimize_compliance`` is
+    # the first emitter, but nothing here assumes that kind name — a future algorithm
+    # emitting the same field under a different event kind is picked up identically).
+    screened_by_tag: dict = {}
+    for ev in events:
+        if "screened_before_fullval" in ev:
+            tag = ev.get("tag") or ev.get("candidate")
+            if tag:
+                screened_by_tag[str(tag)] = bool(ev.get("screened_before_fullval"))
 
     best = baseline_val if baseline_val is not None else 0.0
     it = 0
@@ -886,6 +932,18 @@ def reduce_run(run_dir) -> dict:
             "broke": movement.get("broke") or [],
             "parent_val": parent_val,
             "best_so_far": best,
+            # Structured gate fields, WHEN a commit event carries them directly (any
+            # algorithm's commit step may attach these to its accept/reject event instead
+            # of leaving them only in prose) — read generically by key, never assumed to be
+            # agent-optimize-specific. Absent on older runs / algorithms that never
+            # populate them; the gate-decisions table below falls back to parsing ``reason``.
+            "gate_fields": {k: ev.get(k) for k in
+                            ("delta", "stderr", "n", "k_se", "threshold",
+                             "resolvable_effect_size") if ev.get(k) is not None},
+            # Cheap-screen compliance for this candidate tag, when ANY event recorded it —
+            # looked up generically by tag below (see ``screened_by_tag``), not tied to the
+            # agent-optimize algorithm that happens to be the first emitter.
+            "screened": screened_by_tag.get(cid),
         }
         if "epoch" in ev:
             node["epoch"] = ev.get("epoch")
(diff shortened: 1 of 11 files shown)


OsherElhadad:
## Second round of dashboard fixes — found via live observation of a real running experiment

Read the real, currently-running agent-optimize/tau2_airline run at
`.capevolve/run_finalrun` (its own rendered `dashboard.html`) and found four
generalizable gaps. All four are dashboard-side changes in `core/cap_evolve/dashboard.py`
only — nothing tau2- or agent-optimize-specific is hardcoded into the schema.

### 1. Gate-decision structured fields (highest value)
`reduce_run()` now prefers structured fields (`delta`/`stderr`/`n`/`k_se`/`threshold`/
`resolvable_effect_size`) read **directly off the commit's accept/reject event** when
present, falling back to the existing regex parse of the prose `reason`/`note` when
they're absent. `render_html()`'s Gate-decisions table gets a new "resolvable ±" column.

Verified against the live run: its real `reject` events (`cand_1`, `cand_2`) still carry
only prose — `commit.py` hasn't been updated yet — so every structured field reads `null`
and the table falls back to the regex parse exactly as before (no regression). **This
fix's full richness depends on PR #402 (agent-optimize-401-statistical-rigor) landing
structured fields onto `commit.py`'s event** — I checked that branch's current HEAD
(`05e607d5`/`cf370d3a`) and confirmed `commit.py` still only writes `note`, so I coded
both paths defensively rather than assume it had landed.

### 2. Null-control replicates surfaced separately from the candidate graph
A control replicate (byte-identical re-measurement, run to bound noise) never gets an
accept/reject commit, so it had no graph node and was invisible. `reduce_run()` now
builds a `summary.controls` list read **generically** off any `evaluate` event carrying
`role: "control"` or a truthy `is_control` field — deliberately NOT by tag pattern
(agent-optimize's own `ctl_null_i<N>[r<K>]` convention is algorithm-internal, not a
generic dashboard concern). Rendered as its own "Noise-floor check" section with the
spread between replicates.

Checked round.py/commit.py on the #402 branch: neither field exists yet (controls are
identified there only by tag naming), so against the live run `summary.controls` is
correctly `[]` today. Documented the expected event shape in the reducer so this lights
up the moment an algorithm's evaluate step attaches either field — **also depends on
#402 (or a future algorithm) for real data**.

### 3. `screened: bool | None` badge per candidate
Read generically off any event carrying `screened_before_fullval` keyed by candidate
tag — not tied to the `agent_optimize_compliance` kind name. Verified against the live
run's real compliance events: both `cand_1` and `cand_2` correctly read `screened: False`
in the reduced graph and render as a "✗ not screened" badge in the Candidates table
(column only shown when at least one node has a signal).

### 4. Narrative tab flags an unedited seed template
`_read_narrative()` now compares each accumulator file (JOURNAL/INSIGHTS/META_INSIGHTS/
FRAMEWORK_IMPROVEMENTS) and the best candidate's PROCESS.md against its own seed-template
text from `harness.py` (`_JOURNAL_SEED`, `_INSIGHTS_SEED`, etc.); a file that's still
byte-for-byte its template gets `"template_only": true` and a "⚠ template only — no real
entries yet" banner instead of being presented as real narrative.

Verified both directions against the live run: its real `JOURNAL.md` (two real, escalated
`EMPTY HANDOVER` entries from round 1's fix) correctly reads `template_only: false`; a
synthetic copy with `JOURNAL.md` reset to the bare seed correctly reads `template_only:
true` and the banner renders.

### Tests
5 new tests (`core/tests/test_dashboard_agent_run.py`, `core/tests/test_dashboard.py`)
covering structured-field precedence, control-marker genericity (including a tag that
*looks* like a control but carries no marker), screened-badge generic lookup, and
template-only detection both ways. Full suite: `.venv-tau2/bin/python -m pytest
core/tests -q` → **959 passed, 12 skipped, 0 failed** (no pre-existing flakes hit this
run either).

OsherElhadad:
## Third round: new "Config" tab — full run configuration

Adds a **Config — run configuration** section to the dashboard (`core/cap_evolve/dashboard.py`), following the exact same section/JSON-island pattern the Process narrative tab already uses — no new UI paradigm.

**`reduce_run()`** gains a `summary["config"]` block (empty `{}` when no project dir is found — panel hides itself), built by three new helpers:
- `_find_project_dir(root)` — the sibling `project/` dir (same two candidate locations `_algorithm_from_spec` already checks).
- `_read_config(root)` — the parsed `capevolve.yaml` (via `specfile.read_yaml`, already used elsewhere for this), grouped into `Capability / Algorithm & optimizer / Data & splits / Budget & gate / Memory / Metrics & display / GitHub / Other` by a small, generic key→group map (`_CONFIG_KEY_GROUPS`). **Any key not in that map lands in "Other" instead of disappearing** — so a future intake input just shows up. Also reads `PROJECT.md` if present.
- `_read_project_files(project_dir, skip)` — walks the *whole* project dir generically (skipping `capevolve.yaml`/`PROJECT.md`, `__pycache__`, `.git`) so `adapters/`, `seed_capability/`, split files, or any future artifact type show up automatically. Files ≤200KB get a 4000-char text preview; binary files and files over that cap degrade to size+path only — never an attempt to render megabytes.

**`render_html()`** renders it as a new section: spec groups as key/value tables, PROJECT.md through the **same** markdown renderer the narrative tab uses (hoisted `mdToHtml`/`escN` out of the narrative IIFE to top-level scope so there's exactly one renderer, per the "don't add a second one" instruction), and a collapsible file tree (native `<details>/<summary>`, no extra JS) grouped by top-level dir, each file expandable to its preview or a binary/oversized notice.

**Genericity check** — nothing in this hardcodes tau2_airline, an algorithm, or an adapter shape; verified by rendering the actual live run at `/tmp/cap-evolve-worktrees/integration/.capevolve/run_finalrun` (agent-optimize, tau2_airline project), which correctly shows all 5 populated groups (Capability/Algorithm & optimizer/Data & splits/Budget & gate/Other, `store: copy` landing in Other since it isn't a classified key) plus a 6-file tree (`adapters/adapter.py`, `adapters/model_config.py`, `seed_capability/policy/policy.md`, `seed_capability/reference/data_model.py`, `seed_capability/tools/tools.py`, `split_ids.json`) with real previews, and degrades correctly (no `PROJECT.md` in that project → the PROJECT.md subsection just doesn't render).

**Tests** (`core/tests/test_dashboard.py`, +3): spec grouping incl. an unrecognised future key landing in "Other", PROJECT.md + file-tree contents, binary/oversized-file degradation (a `.bin` file and a >200KB `.py` file both fall back to size+path with no preview attempt), and the "no project dir at all" case producing an honest empty `{}`/`False` rather than a faked panel.

**Test results:** `.venv-tau2/bin/python -m pytest core/tests -q` → **962 passed, 12 skipped** (same 12 pre-existing skips as before this change; no failures).

OsherElhadad:
## Real bug found and fixed live: dashboard silently points at the wrong run

While observing the fresh (fully-fixed) taubench rerun live, found ~25 leaked `capevolve_dashboard.asgi:app` uvicorn processes on the host, squatting the entire 7878-7902 port-scan range — leftovers from old sessions that never got cleaned up when their parent run exited.

`dashboard_launch.py`'s `_free_port()` fallback (`return start`) silently returned a port it had already confirmed was taken when the whole scan range was exhausted — so the printed dashboard URL pointed at a stale, unrelated dashboard serving a *different project's* run dir, with no error surfaced anywhere.

Fixed: `_free_port` now raises when every candidate port is taken; `maybe_launch` catches it and reports `{"dashboard": "error", ...}` through the same soft-error path already used for spawn failures, instead of silently misreporting a wrong URL. Added a regression test. Full suite: 976 passed, 12 skipped, 0 failed.

Not fixed (flagging as a separate, smaller follow-up, not blocking): nothing currently cleans up a run's own dashboard subprocess when the run exits — that's the actual source of the leak. A future fix could track the spawned PID and kill it on run completion, or have the dashboard server self-exit after an idle timeout.

OsherElhadad:
## Real, confirmed fix: JOURNAL.md was never actually contentless — the fallback just wasn't using what was already there

Across three consecutive real live runs (each after a dedicated fix attempt — first seeding the file, then adding an explicit trigger instruction in agent-optimize's SKILL.md), `optimizer_context_warning: empty handover` kept firing on nearly every round regardless. The optimizer's own reflective journal entry just wasn't a priority it consistently attended to.

But `commit.py`'s `--note` (the same text `_reconcile_journal` receives as `reason`) already carries real, substantive reasoning on every single accept/reject/provisional event — confirmed directly from real run data, e.g.: *"Round-1 gate was directionally_positive_but_inconclusive (control replicates disagreed in sign); grown to pooled n=20 trials/task and the effect vanished..."*. Rather than escalate to a fourth prose-only nudge, this fix makes the fallback itself useful: when the handover is empty and a reason is available, fold the reason in as the entry content (clearly labeled framework-synthesized), instead of an admission-only placeholder. The optimizer's own entry, when it writes one, is always preferred verbatim — this is a fallback, not a substitute, and the `optimizer_context_warning` event still fires so the compliance gap stays visible.

3 new tests (synthesize-from-reason, fallback-with-no-reason, own-entry-preferred-over-synthesis). Full suite: 966 passed, 12 skipped, 0 failed.

OsherElhadad:
## Review round (independent): one real bug fixed, one cross-PR inconsistency resolved, main merged in

Reviewed the full diff against `main` across all rounds (harness accumulators + empty-handover synthesis, dashboard narrative/Config/gate-fields/controls/screened, dashboard_launch port fix, intake docs).

### 1. Real bug: the Config panel rendered for a run that has no project dir
`reduce_run` correctly returns `{}` for `summary["config"]` when there is no sibling `project/` — but the panel's guard was `if(!CFG)return;`, and **an empty object is truthy in JS**. So any run without a project dir rendered a "Config — run configuration" section whose own header text read *"read straight off **undefined**"*. The round-3 test only checked the reducer, so it passed while the panel was wrong. Now guards on `CFG.project_dir`, with the reasoning noted in `test_config_absent_without_a_project_dir` (the section title is a JS string literal that is always in the HTML, so this can't be asserted from the rendered text).

### 2. Cross-PR inconsistency: two conventions for the same structured gate fields
`main` (via #407) landed the *same* "structured fields beat the prose regex" idea in the same function, under a **different event key convention**: `gate_delta` / `gate_stderr` / `gate_n` / `gate_k_se` / `gate_threshold` copied onto the node, vs this branch's bare `delta` / `stderr` / `n` / `k_se` / `threshold` collected into a per-node `gate_fields` dict. Keeping both would have left two ways to say one thing — and the bare keys are the riskier pair: a `step` event carrying `n` or `stderr` for an unrelated reason would silently land in the gate table as the paired-delta SE.

Resolved to **main's prefixed convention as the single one** (it is what `agent-optimize`'s `commit.py` actually emits today), and layered this branch's only genuinely new field onto it as `gate_resolvable_effect_size` — keeping both its prose-regex fallback and the new "resolvable ±" column. Dropped the redundant `gate_fields` dict and updated the round-2 test to the surviving key names.

Also merged `main` in (harness `_reconcile_journal` conflicted too: main's `indecisive` verdict and this branch's `reason` synthesis are independent, so both are kept).

### Tests
`.venv-tau2/bin/python -m pytest core/tests -q` → **1043 passed, 12 skipped, 0 failed** post-merge (966 passed pre-merge). No flakes.

One environment note for other agents on this machine: the shared `.venv-tau2` editable install of `core` points at whichever worktree ran `pip install -e core` last, so a suite run can silently execute *another* worktree's code (it did, mid-review — 6 spurious failures). Pin with `PYTHONPATH=<your-worktree>/core` instead of relying on the editable install.

PR #407, opened 2026-08-27
fix(agent-optimize): make an agent-driven run's gate, its verdicts, its handover and its published cost honest

### LINKED ISSUE #406: agent-optimize gates against a drifted reference, and a null run reports a gain it did not make
Three defects found on smoke spreadsheetbench run [32971129203](https://github.com/skillberry-ai/cap-evolve/actions/runs/32971129203) — the first `agent-optimize` run after #399. In each case the run had already computed the right number, then failed to use it or failed to report it.

Related to #401 (which covers agent-optimize's statistical rigour broadly) but distinct: none of the three below appear there, and this does not close it.

## 1. The gate uses a drifted reference, and the stop_condition never says otherwise

`ci/benchmarks/lib/run_suite.sh`'s agent-mode `STOP_CONDITION` names the gate's *strictness* (`gate_k_se`, trials) but never its *reference*, so the agent takes `round.py`'s `--gate-against parent` default. That compares a candidate against the parent's reward as measured in an **earlier** round, so however far that stored number has drifted since, the drift lands inside every candidate delta.

That run's own `work/round_i0.json` recorded both comparisons for `cand_1`:

| reference | delta | threshold | verdict |
|---|---|---|---|
| parent (the default) | 0.0333 | 0.0440 | reject |
| control (same-round byte-identical replicate) | 0.0556 | 0.0341 | **accept** |

0.0556 also cleared that round's own `evidence_bar` of 0.0444, so it was real evidence rather than noise. Rounds i1/i2 rejected under **both** modes — genuine negatives. So exactly one candidate in the run mattered, the default reference discarded it, and the run published `best=seed` as a "null result".

#399 listed "tell the agent to prefer control-mode gating" as a follow-up on tau2 evidence alone. This is a second benchmark reaching the same conclusion.

## 2. A run that accepted nothing advertises an improvement

`ci/benchmarks/lib/metrics.py` published, for a capability it never edited:

```
**Suite (train-fit):** mean reward 0.500 → 0.544 (Δ +0.044 (+9% rel))
 · best = seed (no candidate beat baseline)
```

On a no-holdout tier it pairs `baseline.json` (val) against `final.json` (test); with `best_id == "seed"` those two evals score the **same bytes**, so +0.044 is re-measurement noise — exactly the replicate noise `round.py` measured that run at (`null_delta_between_control_replicates: 0.0444`). Per-task rows carried the same fiction (`47484` +0.333, `53161` −0.333).

It reaches published history as well, via `metrics.jsonl` → `record.rollup` → the benchmarks page. **Not agent-optimize-specific**: any run of any algorithm that accepts nothing hits this.

## 3. Agent-mode rounds measure their gate and their cost, then publish neither

- **The report contradicts itself inside one document**: headline `optimizer $7.95`, Totals `optimizer $0.0000 over 0s`. Agent mode runs one optimizer process for the whole loop, so there is no per-round figure to sum — but summing the absent ones printed `$0.0000` directly beneath a headline saying `$7.95`.
- **Every iterate row showed `eval $ —` / `eval time —`** although each candidate's `evaluate` event carried both (`$0.456` over `629.84s` for `cand_1` alone).
- **The published `gate_decisions[]` had `parent_val`/`delta`/`stderr`/`n`/`k_se`/`threshold` all null**, the numbers surviving only inside the prose `reason`. `dashboard.reduce_run` regexes them out of the deterministic gate's reason string, and an agent writes free text. `commit.py`'s comment blamed `gate_check` for not persisting the parent mean — no longer true since #399, which made `round.py` persist the whole table to `work/round_i<N>.json`.

The drift finding in item 1 was only recoverable *because* that table exists on disk; nothing in the published record showed that the rejection was reference-dependent.

## Constraint on any fix

No number may be fabricated. A missing measurement must stay missing rather than become a `0`, run-level optimizer spend must be labelled whole-loop rather than implied per-round, and the deterministic path must keep its existing behaviour.

eranra:
Closes #406.

Three defects measured on smoke spreadsheetbench run [32971129203](https://github.com/skillberry-ai/cap-evolve/actions/runs/32971129203) — the first `agent-optimize` run after #399. Each is data the run had already computed and then failed to use, or failed to report. Every claim below is tied to that run's artifact rather than to a hypothesis.

**7 files, +612/−11.** Related to #401 (agent-optimize statistical rigour) but disjoint from it; this does not close it.

## 1. The gate compared against a drifted reference

`run_suite.sh`'s agent-mode `STOP_CONDITION` named the gate's *strictness* (`gate_k_se`, trials) and never its *reference*, so the agent took `round.py`'s `--gate-against parent` default. That compares against the parent's reward as measured in an **earlier** round, so that reward's drift since then sits inside every candidate delta.

The run's own `work/round_i0.json` recorded both comparisons for `cand_1`:

| reference | delta | threshold | verdict |
|---|---|---|---|
| parent (the default) | 0.0333 | 0.0440 | reject |
| control (byte-identical replicate, same round) | 0.0556 | 0.0341 | **accept** |

0.0556 also cleared that round's own `evidence_bar` of 0.0444, so it was real evidence, not noise. Rounds i1/i2 rejected under **both** modes — genuine negatives. **Exactly one candidate in the run mattered and the default reference discarded it**, after which the run published `best=seed` as a null result on the strength of a comparison its own round table described as drift-carrying.

The `STOP_CONDITION` now requires `--gate-against control` and states why. #399 named this as a follow-up on tau2 evidence alone and deliberately declined to change the default on one benchmark's evidence; this is a second benchmark, and the change is scoped to the CI stop_condition rather than to `round.py`'s default, so no non-CI caller's behaviour moves.

## 2. A run that accepted nothing advertised an improvement

`metrics.py` published this for a capability it never edited:

> **Suite (train-fit):** mean reward 0.500 → 0.544 (Δ **+0.044 (+9% rel)**) · best = seed (no candidate beat baseline)

On a no-holdout tier it pairs `baseline.json` (val) against `final.json` (test), and with `best_id == "seed"` those score the **same bytes** — so +0.044 was re-measurement noise, exactly the replicate noise `round.py` had measured that run at. Per-task rows said the same (`47484` +0.333, `53161` −0.333). The `best = seed (no candidate beat baseline)` parenthetical was the only contradicting signal in the document, and it loses to the headline number.

This reached published history too, via `metrics.jsonl` → `record.rollup` → the benchmarks page, and it is **not agent-optimize-specific**: any run of any algorithm that accepts nothing hit it. It surfaced now only because this was the first run to reject everything.

A null run now pairs the baseline against itself, so `Δ` is `0.000` and no phantom per-task delta reaches `metrics.jsonl`. The discarded measurement is **not** thrown away — it is reported as what it is:

> No candidate was accepted, so base→opt is the seed against itself and Δ is 0.000 by construction. Re-scoring those same bytes for the finalize read 0.544 against the baseline's 0.500 — a spread of 0.044. That is this tier's re-measurement **noise**, not a change in the capability, and it is the floor any real effect here would have to clear.

Two evals of one unchanged capability are a free read on the tier's noise floor, and a reader needs it to judge whether the run could resolve a real effect at all.

## 3. Rounds measured their gate and their cost, then published neither

- **The report contradicted itself in one document**: headline `optimizer $7.95`, Totals `optimizer $0.0000 over 0s`. Agent mode runs ONE optimizer process for the whole loop, so there is no per-round figure to sum. Totals now falls back to the run-level spend the headline already uses and labels it `whole-loop`, rather than implying per-round attribution.
- **Every iterate row showed `eval $ —`** though each candidate's `evaluate` event carried the cost (`$0.456` / `629.84s` for `cand_1` alone). `iteration_rows` now falls back to that event when the `step` event has no figures. The deterministic path *does* put them on the step, and keeps winning.
- **The published `gate_decisions[]` had `parent_val`/`delta`/`stderr`/`n`/`threshold` all null.** `dashboard.reduce_run` regexes them out of the deterministic gate's reason string; an agent writes prose, so nothing matched. `commit.py`'s comment blamed `gate_check` for not persisting the parent mean — untrue since #399, which made `round.py` persist the whole table to `work/round_i<N>.json`. `commit.py` now reads it, including `gate_mode` and the **drift-free second opinion**, without which a reader cannot tell that a rejection was reference-dependent. That is the finding in item 1, and it was recoverable only by hand-reading a file the published record never pointed at.

## Design decisions

- **No number is fabricated.** A missing round table yields `{}` and the fields stay absent (`test_a_round_with_no_table_still_books_cleanly` asserts `parent_val is None` rather than `0`). `k_se` is genuinely not in the round table, so it stays null instead of being back-derived from the threshold.
- **The deterministic path is untouched in behaviour.** It records no structured gate fields, so `gate_decisions` still takes the regex path; its per-step optimizer cost is still summed (`test_per_step_optimizer_cost_is_still_summed_when_it_exists`); its step-level eval figures still win over the new fallback.
- **Item 1 is fixed in the CI stop_condition, not in `round.py`'s default.** Changing a shared default on one more benchmark's evidence would be the same over-reach #399 declined.
- `commit.py` sorts re-gate suffixes **numerically** — a lexical sort ranks `.r10` below `.r2`, so the tenth re-gate would silently lose to the second.

## Testing

**1057 passed, 2 skipped, 1 failed.** The failure (`test_cli_surface.py::test_spec_defaults_relative_to_project_not_the_cwd`, a `run-optimizer` manifest lookup) reproduces identically with this branch's changes stashed on clean `origin/main` — pre-existing and unrelated, and independently verified here rather than taken from #399's note.

14 new tests across 3 files, each pinning a measured symptom. Every one was watched failing first; the two most load-bearing:

- The re-gate ordering test was confirmed to fail against a lexical sort before the numeric sort was kept.
- `test_a_null_run_publishes_a_zero_per_task_delta` failed with the real `Δ +0.167 (+33% rel)` in its output.

`skills/algorithms/agent-optimize/scripts/check.py` → `ok: true`, 0 problems, 78 executed assertions. Skill authoring lint clean. `actionlint` clean. `bash -n` on `run_suite.sh` clean. Pyright on `dashboard.py` reports the **same 5 pre-existing errors before and after** (verified by stashing; only line numbers shift).

### End-to-end on the real artifact

Run 32971129203's actual `baseline.json` / `final.json` / `state.json` / `events.jsonl` / `work/round_i*.json` were reconstructed from its uploaded artifact and re-rendered through this branch:

| symptom | before | after |
|---|---|---|
| suite headline | `0.500 → 0.544 (Δ +0.044 (+9% rel))` | `0.500 → 0.500 (Δ +0.000)` + noise note naming 0.544 / spread 0.044 |
| phantom per-task moves | `47484` +0.333, `53161` −0.333 | all `+0.000` |
| iterate eval cost | `—` on all 3 rounds | `$0.4560` / `$0.5323` / `$0.3069` |
| Totals vs headline | `$0.0000 over 0s` vs `$7.95` | `$7.9524 over 40m00s (whole-loop …)` |
| eval totals | `$0.5872 over 22m20s` | `$1.8824 over 70m05s` (the three candidate evals were missing) |

## Reviewer notes

- **Blast radius.** `run_suite.sh`'s change is inside the `agent-optimize)` case arm only. `commit.py` has no callers outside the agent-optimize skill. `metrics.py`'s two changes are both guarded (`best_id == seed`; per-step cost absent). `dashboard.py` copies fields only when present, so a deterministic step's node and gate row are unchanged.
- **CI-validated (see the second round below).** Originally filed as unit-tested + artifact-replayed only. The re-dispatch called for here has since run — spreadsheetbench smoke, `agent-optimize`, `trials=10` — and is written up below, along with the seven further defects it found.
- **Why `trials=10` for the re-run.** This run's per-eval SE was ~0.14 and its replicate noise 0.044, while the largest real effect measured was 0.056 — at 3 trials the tier sits below its own resolution, so a null there is uninterpretable either way. The reference deterministic run's "win" (+0.078) was itself measured against a baseline that had drifted 0.069.

---

# Second round: seven more defects, found by the validation this PR asked for

The re-run this PR called for was dispatched: smoke spreadsheetbench under `agent-optimize` at `trials=10`, run [33046360451](https://github.com/skillberry-ai/cap-evolve/actions/runs/33046360451). It sealed as an **honest null** — `reward_base 0.51 → reward_opt 0.51`, `reward_delta 0.0` on all 10 tasks — which is item 2 above working as intended: the same run on `main` would have published a several-percent "gain" for a capability it never changed.

Driving a real agent through a real loop then exposed seven further defects. All seven are generic; four live in `core/` and affect **every** algorithm, not just agent mode. **+1782/−62 across 16 files** on top of the original commit, in three commits.

> **On item 1 above.** This run does not settle whether `--gate-against control` was the right call — no candidate cleared the bar under either reference. What it does show is that the control gate is not a rubber stamp.

## 4. An unresolvable verdict had to be booked as a rejection

`round.py` marks a candidate `inconclusive` when its verdict is not stable across the round's byte-identical control replicates (`verdict_stable: false`): it accepts against one replicate and rejects against the other, so the round cannot separate the edit from re-measurement. Its own `reading` says such a candidate is "never accepted … Re-run it with more trials before believing either answer."

That is a **third** outcome, and the loop had nowhere to write it. `commit.py --decision` accepted only `accept|reject`, so an agent's options were to book a reject — asserting the gate refuted an edit it had not — or to book nothing, which `host._unbooked_rounds` correctly diagnoses as a defective run.

Live, from `work/round_i1.json`:

```
cand_2  Δ +0.0433  threshold 0.0492  verdict inconclusive  verdict_stable false
        verdict_by_reference: {ctl_null_i1: reject, ctl_null_i1r1: accept}
```

Booking that as a reject is not cosmetic misfiling:

- **`update_spent(accepted=False)` increments the STALL counter**, and `budget_exhausted()` stops the run when stall hits its cap (3 here, already at 1). Stall means *the optimizer has run out of ideas* — the one thing an ambiguous measurement is no evidence of. Two ambiguous rounds can therefore end a run for a reason that never happened, on a benchmark whose replicate noise makes ambiguity the common case.
- **It files the edit in `rejected.jsonl`**, which later rounds read as *tried, did not work* — teaching the run to avoid its own never-evaluated idea.
- **It stamps the journal's `RESULT` line as a rejection**, telling the next iteration its batch was reverted for cause.

`--decision inconclusive` now exists. `harness.record_iteration` has supported exactly this since #216/#224 — `indecisive=True` charges the iteration and leaves stall alone — and the deterministic hill-climb already uses it; agent mode simply could not reach it. The round is still snapshotted and still charged, a `step_indecisive` event is filed instead of a rejection in memory, and the `RESULT` line reads `UNRESOLVED (not judged; champion unchanged)` with re-measure guidance. `--reject-basis gate` is now also refused when the gate returned `inconclusive`, for the same reason it was already refused when the gate said accept: that field is the run's record of what the evidence was.

The dashboard needed no change — it already renders `indecisive` as a first-class status (`indecisive_ids`, the `b-indecisive` badge). That the display half was complete and only the booking path missing is itself the shape of the bug.

## 5. A re-gate destroyed the evidence it was run to add

A round's identity is `(iteration, attempt)`, and both halves have to reach the names on disk. The **table** had the attempt half — a re-gate writes `round_i1.r1.json` rather than overwriting `round_i1.json`. The control **rollouts** did not: they were tagged `ctl_null_i<N>` regardless of attempt. Rollouts are persisted as `<task>__<tag>__t{k}.json`, so the one operation `round.py` explicitly supports re-measured the previous attempt's control under the same tags and deleted the numbers the first table cites.

Measured at iteration 1 of the live run:

| | attempt 0 | after the re-gate |
|---|---|---|
| `ctl_null_i1` | 0.4967 | 0.5067 |
| `ctl_null_i1r1` | 0.4800 | 0.4367 |
| `null_delta_between_control_replicates` | 0.0167 | **0.0700** |

200 metric calls bought a *worse* null estimate than the 100 before them, because the re-gate swapped two readings for two others instead of pooling four — and `round_i1.json` was left quoting an `evidence_bar` computed from numbers that no longer existed. A 0.07 bar then exceeded the effect under test, so the round could not resolve: item 4's inconclusive verdict, manufactured by item 5.

Control tags now carry the attempt (`ctl_null_i<N>a<k>`), and `prior_attempt_controls` pools earlier attempts' replicates into the null with a `from_attempt` marker, so `null_delta_...` covers every replicate the round has paid for. Pooling reads from the earlier **tables** rather than their rollouts, deliberately: on a pre-fix run the rollouts are already gone and the table is the only place the destroyed reading survives. `round_attempt(run_dir)` is called **once** and both the table stem and the control tag derive from it, because two independently derived halves can disagree about which attempt this is; it returns `max(attempt)+1` rather than `len(tables)` so a hand-deleted table cannot hand out a name that is still taken.

**Known limitation, flagged not smuggled.** The two attempts' *candidate* measurements still cannot be combined: the fresh-tag rule guarantees they have different tags, and pooling them would need an explicit "this tag re-measures that one" declaration. That is a design change, not a bug fix.

## 6. Re-measuring a tag silently replaced its rollouts

The general form of item 5, in `core`, and the reason the agent walked into it. `evaluate_candidate` writes `t0..t{n_trials-1}`, so a second evaluation under an existing tag **replaces** those trials rather than adding `t10..t19`. There is no trial accumulation. That is a defensible implementation choice; the defect is that it happened silently, in a loop whose own guidance asks for the opposite — `round.py` told the agent to "re-run it with more trials", the agent did exactly that under the existing tag, and 100 metric calls swapped the 0.4967 control reading for 0.5067 rather than adding a third sample. In the event stream that is indistinguishable from progress.

A **warning, not a refusal**: re-evaluating an existing tag is legitimate on `--resume`, where a run reads its champion's val score back off disk, and the harness must not break that. `evaluate_candidate` now logs `rollout_overwrite_warning` with the trial count being replaced and — computed *before* the overwrite — `prior_reward`, because once the files are gone the destroyed reading exists nowhere else.

## 7. The loop accumulated no learning across its rounds

Two independent holes, both about the loop's memory of *itself* — the thing that is supposed to make round 3 better than round 1. Observed on this run and on 32971129203 before it.

**Optimizer memory was empty.** `rejected.jsonl` / `history.jsonl` back the dashboard's Memory panel and any optimizer prompt that greps for already-refuted approaches. Every deterministic algorithm writes them inline — hill-climb, `gepa` (including its merge paths) and `skillopt` all call `.add` themselves. Agent mode books through `commit.py`, which never did, so the run published `{"history": [], "rejected": []}`: a whole run of rejected approaches that the run itself could not enumerate. `commit.py` now files them (and deliberately does **not** file an `inconclusive` round, per item 4). It is *not* hoisted into `record_iteration` alongside the other three per-iteration records even though that is where it belongs by that function's own argument, because `gepa._try_merge` files a rejection without routing through it and centralising would silently drop that — recorded in the code as fixable, but not safely in the same change.

**Every round's handover read "(no handover written by the optimizer)".** `JOURNAL.md` has two halves: the framework stamps the objective `RESULT` line, the optimizer appends the INTENT above it. Nothing ever asked the agent for its half — `host.py` said only that `commit.py` "reconciles it as you book rounds", which is true of the framework half and false of the half the agent owns. So from round 2 the agent could read *which* task it broke but never *what it had tried*, and the journal's own standing instruction ("never re-test a refuted idea") was unfollowable. The host prompt now asks for the entry explicitly, before each `commit.py`, in the shape `_journal_tail` reads; `commit.py` reads it **before** booking and returns `handover_recorded` plus a warning, so a forgotten handover is correctable while rounds remain rather than discovered when the run is over.

And a latent `core` bug underneath, which would have swallowed the entry even once written: `_journal_tail`'s no-marker fallback searched for `"\n## "`, so a **heading at offset 0 was missed**. An agent-mode optimizer is not handed a seeded journal to append to — it writes a fresh file that starts with its heading. Now anchored per-line (`(?m)^## `).

## 8. A reject that broke nothing was told to redesign the edits that broke something

The `RESULT` line stamped under every journal entry exists to tell the next iteration what to do differently. For a reject it said, unconditionally: *"its WHOLE batch was reverted; re-introduce only the edits that did NOT break a task above, dropping/redesigning the ones that did."*

Read against the line it was actually stamped on:

```
REJECTED (champion unchanged) · val=0.553 Δ=+0.043 · fixed={47484, 53161} · broke={—}
```

The batch fixed two tasks, broke none, and gained +0.043 — rejected only because that did not clear a 0.048 threshold. So the one sentence meant to say what to do next said nothing actionable, while "redesign the ones that did" invited a rewrite of edits just measured **helping**. That is how a run talks itself out of its own best lever.

The stamp now branches three ways: a reject **with** regressions keeps the original guidance; a reject with **none** says the lever is measurement power or a bigger effect, not a redesign, and to keep the `fixed` edits as the starting point; and a reject where no per-task comparison was possible (`_candidate_task_impact` returns `None` when either side has no rollouts) says `fixed`/`broke` are **UNKNOWN** rather than empty — because `broke={—}` meaning "nothing broke" and `broke={—}` meaning "we could not tell" are different instructions, and printing the second as the first invents a fact. **Not agent-mode specific:** every algorithm's below-threshold reject reaches this stamp through `record_iteration`.

## 9. The published cost was a quarter of what the run spent

A run's published cost is a sum over per-phase rows (baseline, each committed round, finalize). Metered spend that no phase owned was therefore not merely unattributed — **it was published as if it had never happened**:

| | published | the run's own meters |
|---|---|---|
| `suite.optimizer_usd` | **0** | 9.11 |
| `suite.eval_usd` | 5.25 | 12.55 (`spent.usd`) |
| total | **$5.25** | **$21.66** |

A 4× understatement in the single figure a full tier's budget gets projected from — reported by a run that looks completed, green and cheap. Two independent holes fed it: agent mode meters the optimizer **once for the whole loop**, so no round can own a share of it; and control replicates, re-gates and abandoned rounds are real rollouts that no committed `step` event references.

`metrics.iteration_rows` now emits one `unattributed` row for the residual between the run's meters and what the phases claim. Every consumer already sums these rows — `record.rollup`, `results_md.load_suite_totals`, the report's Totals line, the Pages per-run table — so one row repairs all four at once. It is a *row* rather than a per-round split because the run never measured a per-round split; being a residual it vanishes on a run that does attribute everything (the deterministic loops), and it is floored at zero so an unmetered run cannot produce a negative. Replayed against the sealed artifact:

```
published before: eval $5.2519  + optimizer $0.0000 = $5.25
published after : eval $12.5493 + optimizer $9.1108 = $21.66
run metered     : $21.66          rewards unchanged: True
```

The per-run dashboard needed no change: it already published `cost.total_usd: 21.6601` correctly, beside per-iteration nulls that were accurate ("no optimizer cost is attributed to this round"). The two views now agree.

## 10. The host double-booked the agent's own cost

Found while fixing item 9, and worse than it, because it errs in the expensive direction. The host meters the whole agent process **and** the skill instructs the agent to book its own proposal cost per round via `commit.py --optimizer-usd`. Those are the same money — a round's proposal happens inside that process — and the host booked its metered total *on top of* whatever the agent had booked. An agent that followed the skill's own written instruction made the run report **up to twice** the optimizer spend it used, and a cost-based `stop_condition` would then have ended a run that still had budget.

`host.optimizer_spend_to_book` now books only the residual, per role, never below zero, and **bracketed to this invocation** so a `--resume` run does not mistake an earlier host's spend for this agent's attribution. It also books the loop's **wall time**, which nothing recorded at all: `optimizer_seconds: 0.0` for a run that took hours left every per-hour figure undefined. `references/algorithm.md` now states that the per-round figures *attribute within* the metered total rather than adding to it, so the instruction and the accounting agree.

## Testing (second round)

`core/tests` + `ci/benchmarks/lib`: **1107 passed, 2 skipped, 0 failed.**

> The `test_cli_surface.py` failure noted above is an environment artifact, now pinned down: `_find_skills_dir()` prefers an installed `~/.claude/skills` over the repo's own, and the stale installed manifest on this machine lacks `run-optimizer`. With `CAPEVOLVE_SKILLS_DIR` pointed at the repo's `skills/` the suite is fully green. Not a code defect and not touched by this branch — recorded so the next reader does not re-investigate it.

**+49 tests across 6 new files**, each pinning a measured symptom from the run, and every one watched failing before its fix. Two notes on that discipline:

- The four cost guards were re-run against the pre-fix behaviour (residual suppressed) and all four failed with the live numbers in their own messages — `"the timeline accounts for $0.00 of a metered $9.00"`.
- `test_commit_reads_only_fields_round_actually_writes` broke when the round-table naming changed. It was a legitimate coupling guard, so it was **replaced with a behavioural cross-file test** — `commit.py` must find the table wherever `round.py` chose to name it, asserted through `threshold`/`evidence_bar`, which can only come from that table — rather than by editing the string it matched on.

Also clean: skill authoring lint (advisories only, baseline unchanged), `check.py --self-check`, and a line-length/dead-code sweep of every touched file. The `agent-optimize` `SKILL.md` body sits at its exact 5000-token cap, so the new prose went to `references/` and `host.py`'s prompt — the sanctioned overflow — with only a two-line pointer added to `SKILL.md` itself.

### Deliberately out of scope

- **Per-iteration optimizer cost attribution in agent mode.** The host is blocked on `subprocess.run` for the agent's whole lifetime so it cannot book incrementally, and the agent cannot meter its own process; fixing it means streaming cost out of the transcript. Consequence worth knowing: **`max_optimizer_usd` cannot bind mid-run in agent mode.** Item 9 makes the *reporting* honest; the *enforcement* gap is real and unfixed.
- **Pooling candidate measurements across re-gate attempts** (item 5).
- **Hoisting memory writes into `record_iteration`** (item 7) — blocked on `gepa._try_merge`'s out-of-band rejection path.

### DIFF
--- a/ci/benchmarks/lib/metrics.py
+++ b/ci/benchmarks/lib/metrics.py
@@ -74,6 +74,20 @@ def iteration_rows(run_dir: str, best_id: str | None = None) -> list[dict]:
     their optimizer_usd/optimizer_seconds are 0.0. ``best_id`` labels the ``finalize``
     row's candidate (the "evaluate" event itself only carries a fixed tag, not the
     actual candidate id — the caller already knows it from state.json).
+
+    A final ``unattributed`` row carries whatever the run METERED but no phase above claims,
+    because every consumer of these rows (``record.rollup``, ``results_md``, the Totals line
+    below, the Pages per-run table) reports cost by SUMMING them — so spend that belonged to no
+    step was not merely unattributed, it was published as if it had never happened. Measured on
+    smoke spreadsheetbench run 33046360451: ``suite.optimizer_usd: 0`` and ``eval_usd: 5.25``
+    for a run whose own state held ``optimizer_usd 9.11`` / ``usd 12.55`` — a $21.66 run
+    published as $5.25, a 4x understatement in the one figure a full tier's budget gets
+    projected from. Two independent holes fed it: agent mode meters the optimizer ONCE for the
+    whole loop (no round can own a share of it), and control replicates, re-gates and abandoned
+    rounds are real evaluations that no ``step`` event references. The residual is the honest
+    name for both, and it is visible as its own row rather than smuggled into a round that did
+    not spend it. Being a residual and not a sum, a run that DOES attribute everything per round
+    emits no row at all, and an unmetered run (``usd == 0``) cannot produce a negative one.
     """
     events_path = Path(run_dir) / "events.jsonl"
     rows: list[dict] = []
@@ -136,7 +150,53 @@ def iteration_rows(run_dir: str, best_id: str | None = None) -> list[dict]:
                 "optimizer_usd": 0.0, "optimizer_seconds": 0.0,
                 "eval_usd": ev.get("cost_usd"), "eval_seconds": ev.get("seconds"),
             })
+    residual = _unattributed_row(Path(run_dir), rows)
+    if residual:
+        rows.append(residual)
     return rows
+
+
+# Below a cent in either role, the residual is rounding between the meter and the events, and a
+# row saying "$0.00 went somewhere" costs a reader more than it tells them.
+_RESIDUAL_FLOOR_USD = 0.01
+
+
+def _unattributed_row(run_dir: Path, rows: list[dict]) -> dict | None:
+    """Metered spend that no row in ``rows`` accounts for, as one row — or None if it all is.
+
+    The run's own ``state.json`` is the meter of record: ``spent.usd`` is every rollout the run
+    paid for and ``spent.optimizer_usd`` every proposal, both accumulated by ``update_spent`` at
+    each booking, so each is an upper bound on what the per-phase events can attribute. Intake
+    spend rides in the optimizer column: it is non-runner cost with no step to call its own, and
+    the alternative is to keep dropping it.
+    """
+    state = _load(run_dir / "state.json")
+    spent = state.get("spent") or {}
+    if not spent:
+        return None
+    opt_metered = (spent.get("optimizer_usd") or 0.0) + (spent.get("intake_usd") or 0.0)
+    eval_metered = spent.get("usd") or 0.0
+    opt_secs_metered = ((spent.get("optimizer_seconds") or 0.0)
+                        + (spent.get("intake_seconds") or 0.0))
+    eval_secs_metered = spent.get("runner_seconds") or 0.0
+    opt_usd = round(max(0.0, opt_metered - sum(r.get("optimizer_usd") or 0.0 for r in rows)), 6)
+    eval_usd = round(max(0.0, eval_metered - sum(r.get("eval_usd") or 0.0 for r in rows)), 6)
+    if opt_usd < _RESIDUAL_FLOOR_USD and eval_usd < _RESIDUAL_FLOOR_USD:
+        return None
+    # Seconds only where the same role has unattributed money, so a row cannot claim time for a
+    # role whose spend was fully accounted for.
+    opt_secs = (round(max(0.0, opt_secs_metered
+                          - sum(r.get("optimizer_seconds") or 0.0 for r in rows)), 2)
+                if opt_usd >= _RESIDUAL_FLOOR_USD else 0.0)
+    eval_secs = (round(max(0.0, eval_secs_metered
+                           - sum(r.get("eval_seconds") or 0.0 for r in rows)), 2)
+                 if eval_usd >= _RESIDUAL_FLOOR_USD else None)
+    return {
+        "phase": "unattributed", "iter": None, "candidate": "(whole run)", "accepted": None,
+        "reward": None,
+        "optimizer_usd": opt_usd, "optimizer_seconds": opt_secs,
+        "eval_usd": eval_usd, "eval_seconds": eval_secs,
+    }
 
 
 def suite_report(run_dir: str, bench: str, tier: str, agent: str, iters, jsonl_path: str = "",
@@ -301,21 +361,32 @@ def suite_report(run_dir: str, bench: str, tier: str, agent: str, iters, jsonl_p
             eval_usd_t += s["eval_usd"] or 0
             eval_s_t += s["eval_seconds"] or 0
         out.append("")
+        # These sum the rows, INCLUDING the `unattributed` one, so they now reconcile with the
+        # run's own meters instead of reporting only the share some phase happened to claim.
         # Agent mode runs ONE optimizer process for the whole loop, so no per-round optimizer
-        # cost exists to sum — and summing the absent figures printed "optimizer $0.0000 over
-        # 0s" directly underneath a headline reading "optimizer $7.95" (run 32971129203). Fall
-        # back to the run-level spend the headline already uses, and say that it is whole-loop
-        # rather than silently implying it was attributed per round.
-        run_opt_usd = spent.get("optimizer_usd") or 0.0
-        run_opt_s = spent.get("optimizer_seconds") or 0.0
-        if not (opt_usd_t or opt_s_t) and (run_opt_usd or run_opt_s):
-            out.append(f"**Totals:** optimizer ${run_opt_usd:.4f} over "
-                       f"{_fmt_duration(run_opt_s)} (whole-loop — one optimizer process drove "
-                       f"every round, so there is no per-round figure to attribute) · "
-                       f"eval ${eval_usd_t:.4f} over {_fmt_duration(eval_s_t)}")
-        else:
-            out.append(f"**Totals:** optimizer ${opt_usd_t:.4f} over {_fmt_duration(opt_s_t)} · "
-                       f"eval ${eval_usd_t:.4f} over {_fmt_duration(eval_s_t)}")
+        # cost exists to sum — and summing the absent figures printed "optimizer $0.0000 over 0s"
+        # directly underneath a headline reading "optimizer $7.95" (run 32971129203). The note
+        # below names whichever role the residual came from rather than letting the total imply
+        # it was attributed per round.
+        out.append(f"**Totals:** optimizer ${opt_usd_t:.4f} over {_fmt_duration(opt_s_t)} · "
+                   f"eval ${eval_usd_t:.4f} over {_fmt_duration(eval_s_t)}")
+        resid = next((s for s in steps if s["phase"] == "unattributed"), None)
+        if resid:
+            parts = []
+            if (resid["optimizer_usd"] or 0) > 0:
+                parts.append(f"optimizer ${resid['optimizer_usd']:.4f} — agent mode runs ONE "
+                             "optimizer process, metered whole-loop, so no round owns a share "
+                             "of it")
+            if (resid["eval_usd"] or 0) > 0:
+                parts.append(f"eval ${resid['eval_usd']:.4f} — control replicates, re-gates and "
+                             "abandoned rounds are real evaluations that no committed step "
+                             "references")
+            out.append("")
+            out.append("> The `unattributed` row is metered spend no phase above owns ("
+                       + "; ".join(parts) + "). It is shown rather than dropped so these totals "
+                       f"match what the run actually paid: `spent` reads "
+                       f"${spent.get('usd', 0) or 0:.2f} runner + "
+                       f"${spent.get('optimizer_usd', 0) or 0:.2f} optimizer.")
     return "\n".join(out)
 
 
--- a/ci/benchmarks/lib/metrics.py
+++ b/ci/benchmarks/lib/metrics.py
@@ -81,6 +81,12 @@ def iteration_rows(run_dir: str, best_id: str | None = None) -> list[dict]:
         return rows
     seen_baseline = False
     it = 0
+    # Every candidate's eval cost/time is recorded on its own ``evaluate`` event. The
+    # deterministic loops ALSO copy those onto the ``step`` event; agent mode's commit.py does
+    # not, so run 32971129203 rendered "eval $ —" for all three rounds while its events.jsonl
+    # held $0.456/629.84s for cand_1 alone. A candidate is evaluated before it is booked, so
+    # accumulating as we go is enough — no second pass.
+    eval_by_tag: dict[str, dict] = {}
     for line in events_path.read_text(encoding="utf-8").splitlines():
         line = line.strip()
         if not line:
@@ -90,6 +96,8 @@ def iteration_rows(run_dir: str, best_id: str | None = None) -> list[dict]:
         except Exception:
             continue
         kind = ev.get("kind")
+        if kind == "evaluate" and ev.get("tag"):
+            eval_by_tag[str(ev.get("tag"))] = ev
         if kind == "evaluate" and ev.get("tag") == "seed" and ev.get("split") == "val" and not seen_baseline:
             seen_baseline = True
             rows.append({
@@ -100,12 +108,19 @@ def iteration_rows(run_dir: str, best_id: str | None = None) -> list[dict]:
             })
         elif kind == "step":
             it += 1
+            # The step's own figures win when it has them (the deterministic path); otherwise
+            # fall back to this candidate's evaluate event rather than rendering a dash over
+            # a cost the run did measure.
+            cev = eval_by_tag.get(str(ev.get("candidate"))) or {}
+            e_usd = ev.get("cost_usd")
+            e_secs = ev.get("runner_seconds")
             rows.append({
                 "phase": "iterate", "iter": it, "candidate": ev.get("candidate"),
                 "accepted": ev.get("accept"), "reward": ev.get("val"),
                 "optimizer_usd": round(ev.get("opt_cost_usd") or 0.0, 6),
                 "optimizer_seconds": ev.get("optimizer_seconds"),
-                "eval_usd": ev.get("cost_usd"), "eval_seconds": ev.get("runner_seconds"),
+                "eval_usd": e_usd if e_usd is not None else cev.get("cost_usd"),
+                "eval_seconds": e_secs if e_secs is not None else cev.get("seconds"),
             })
         elif kind == "evaluate" and ev.get("tag") == "FINAL" and ev.get("split") == "test":
             rows.append({
@@ -149,6 +164,12 @@ def suite_report(run_dir: str, bench: str, tier: str, agent: str, iters, jsonl_p
     state = _load(rd / "state.json")
     spent = state.get("spent", {})
     best_id = state.get("best_id", "seed")
+    # NULL RUN: no candidate cleared the gate, so the "optimized" capability IS the seed and
+    # the two evals paired below score the SAME BYTES. Every difference between them is
+    # re-measurement noise. Run 32971129203 published "0.500 → 0.544 (Δ +0.044 (+9% rel))"
+    # for a capability it never edited, with per-task rows to match (`47484` +0.333, `53161`
+    # -0.333) — and +0.044 was exactly the replicate noise round.py measured that run at.
+    null_run = best_id in ("seed", "", None)
 
     bval = baseline.get("val") or {}
     ftest = final.get("test") or {}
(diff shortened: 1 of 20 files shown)


eranra:
Relationship to #402 / #405 (both open, both by @OsherElhadad), for reviewers:

- **#402** (agent-optimize statistical rigour, `/goal`, compliance logging) overlaps this PR conceptually — it owns `round.py` and `gate.py`. This PR deliberately does **not** change `round.py`'s `--gate-against` default; it only makes CI's agent-mode `stop_condition` ask for `control`, leaving the default to #402's area. Checked: #402's `round.py` diff (+20/−1) touches none of the round-table keys `commit.py` reads here, and `test_commit_reads_only_fields_round_actually_writes` will fail loudly if they ever move. Both PRs edit `dashboard.py`, so expect a textual (not semantic) conflict — happy to rebase onto whichever lands first.
- **#405** (per-run insight files + default-on process narrative) is the explainability/memory work; independent of this PR.

Neither is in `main` (`323ed3b5`) yet, so neither is in the validation runs referenced above. The improved-agent-optimize number will be measured separately once they land — see https://github.com/skillberry-ai/cap-evolve/pull/402#issuecomment-5437098750.

PR #409, opened 2026-08-27
feat(runtime): declare how a candidate reaches the model under test

aviweit:
A capability answers *what* is optimized. Until now there was one answer to *how* it reaches
the model: files the runner reads in its own process. This makes delivery a declared,
validated property of a run — a new `intervention` component family, the `intervention:` spec
key, and SPA (the Skillberry proxy) as the first out-of-process implementation.

### How it works

- **`intervention:` in `capevolve.yaml` declares** the delivery path (`direct` default | `spa`).
- **`cap-evolve check` validates** it offline, refusing an unknown value *by name* rather
  than defaulting to `direct` — a silently-ignored key would deliver candidates the wrong
  way while the record claimed otherwise.
- **`cap-evolve run` preflights** the stack before any step and starts a
  provisioned-but-stopped one. It refuses to **provision**: cloning and installing two
  services is an onboarding step, not a run side-effect. `--plan-only` validates the value
  but never preflights, so an inspection flag has no service side effects.
- `skills/interventions/llm-proxies/spa/scripts/spa_env.py` owns the whole stack lifecycle
  (provision/start/stop/clean, store deployment, agent routing) so an example imports it
  rather than re-implementing it.

### One delivery path per run

A project prepares **one** path, not both: `intervention:` and `capability_path` travel
together in a spec, and the two paths need differently-shaped seeds (an inline artifact vs a
single skill package over a frozen substrate). Preparing both doubles the onboarding, leaves
an unused seed to go stale, and invites a spec/seed mismatch that delivers candidates one way
while the record says the other. Running the second path is a separate onboarding in its own
project.

The two numbers are also **not like-for-like** — under `direct` the optimizer may change what
the tools *do*; under `spa` only how they are presented and composed, so `spa` searches a
strictly smaller space. Compare them side by side if you must, never averaged.

### Backwards compatible

A spec with no `intervention:` resolves to `direct`, and preflight returns immediately
**without importing the intervention skill**. No shipped example declares it, so nothing
changes for them — verified by running `check`/`plan-only` on a direct example against both
`main` and this branch.

### Commits

`8afcfc7` — the delivery layer: the skill tree, the core module, wiring in `cli.py`/`check.py`,
the new component in `COMPONENTS`, `install.sh` gains the family, advertised counts re-derived
(20 → 21).

`d24bd34` — four review findings: `repo_root()` no longer counts parents (an installed tree
resolved to `$HOME`, pointing `vendor/` at the home dir); preflight catches plain
`RuntimeError` so a held port returns JSON not a traceback; `docker_bridge_ip()` rejects
non-IP lines from merged stderr; `SKILL.md` documents the commands that exist.

`aec55a0` — never put an API key in `llm_args`: a runner may persist them into traces that
reach the optimizer and get committed.

`3aaa911` — the rename above, plus three things worth a look on their own:
- **four fixed-depth globs** would have broken *silently* on the extra directory level, each
  differently — `install.sh` would have installed the group dir as a skill and lost the skill;
  `lint_skills.py` (twice) would have stopped linting it entirely, caught by its own
  anti-vacuity guard; `test_advertised_counts.py` would have undercounted. All four now
  discover by content. `build_manifest.py` needed nothing — it already used `rglob`.
- **the skill is benchmark-neutral**: one benchmark's names are gone from `SKILL.md`,
  `spa_env.py` and `check.py`'s fixtures. A reusable skill that names one benchmark teaches
  the next reader to copy it.
- **`SKILL.md` follows the house shape**: "What this does not do (yet)" was really
  *preconditions* (a SPA-aware runner, an HTTP-fronted environment), so it moved up as
  `## Requirements from the benchmark`; added `## Inputs / outputs (manifest tokens)`,
  `## What good vs bad looks like` and `## References`, matching gate/tools/hill-climb.

### Verified

Full suite green (885 passed, 12 skipped) · skill-authoring lint clean · manifest registers
`intervention (1): spa` · the skill's `check.py` green offline with no services ·
`intervention: sap` refused by name · preflight against an unprovisioned stack names what is
missing · `repo_root()` resolves to the checkout in the repo layout and **not** to `$HOME` in
a simulated installed layout · a test install lands `interventions/spa` rather than the group
dir.

### Not included

Any example directory, seed, or optimizer instructions — those are outputs of onboarding, not
artifacts this repo carries.

### Follow-ups

- `start_spa` accepts a foreign-but-healthy service on its port as its own; with stop
  failures swallowed, a restart can keep serving the previous candidate
- deploys aren't verified against the expected skill name, so a renamed skill is silently
  orphaned
- `import_standalone_tools` re-tags via GET-then-PUT and strips protected tools' schemas
- `clean()` raises mid-teardown when the vendor-dir override points outside the checkout
- `check` validates only the default `capevolve.yaml`, so a variant spec's `intervention:`
  goes unchecked
- adapters should *declare* which delivery paths they implement, so a spec asking for one the
  adapter doesn't support is refused rather than silently mis-delivered

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@
   <img src="https://img.shields.io/badge/python-3.10%2B-blue" alt="python">
   <img src="https://img.shields.io/badge/runtime%20deps-0%20(stdlib)-success" alt="deps">
   <img src="https://img.shields.io/badge/license-Apache--2.0-informational" alt="license">
-  <img src="https://img.shields.io/badge/agent%20skills-20-7c5cff" alt="skills">
+  <img src="https://img.shields.io/badge/agent%20skills-21-7c5cff" alt="skills">
 </p>
 
 **cap-evolve improves an AI agent's prompts, tools, and skills by learning from failed
--- a/core/cap_evolve/check.py
+++ b/core/cap_evolve/check.py
@@ -38,16 +38,16 @@ class CheckReport:
         return {"ok": self.ok, "stubs": self.stubs, "problems": self.problems, "notes": self.notes}
 
 
-def _check_runtime(rep, project_dir: Path) -> None:
-    """Validate the spec's ``runtime:`` value. OFFLINE — starts nothing, probes nothing.
+def _check_intervention(rep, project_dir: Path) -> None:
+    """Validate the spec's ``intervention:`` value. OFFLINE — starts nothing, probes nothing.
 
     Only the declaration is checked here (an unknown value is a hard problem, since the
     spec reader would otherwise ignore it and deliver candidates the ``direct`` way).
-    Whether the runtime's services are actually UP is a run-time question, answered by
-    ``runtime.preflight`` — ``check`` must stay runnable on a machine with no stack.
+    Whether the intervention's services are actually UP is a run-time question, answered by
+    ``intervention.preflight`` — ``check`` must stay runnable on a machine with no stack.
     """
     from .specfile import read_yaml
-    from . import runtime as _runtime
+    from . import intervention as _intervention
 
     spec_path = Path(project_dir) / "capevolve.yaml"
     if not spec_path.exists():
@@ -57,12 +57,12 @@ def _check_runtime(rep, project_dir: Path) -> None:
     except Exception:  # a broken spec is reported by other checks
         return
     try:
-        rt = _runtime.declared(spec)
-    except _runtime.RuntimeError_ as e:
+        rt = _intervention.declared(spec)
+    except _intervention.InterventionError as e:
         rep.problems.append(str(e))
         return
-    if rt != _runtime.DIRECT:
-        rep.notes.append(f"runtime: {rt} (candidate delivered out-of-process; "
+    if rt != _intervention.DIRECT:
+        rep.notes.append(f"intervention: {rt} (candidate delivered out-of-process; "
                          "readiness is checked at run time, not here)")
 
 
@@ -219,7 +219,7 @@ def run_check(project_dir: Path, *, tolerance: float = 1e-6) -> CheckReport:
         except Exception as e:  # noqa: BLE001
             rep.notes.append(f"degenerate-trials probe skipped: {e}")
 
-    _check_runtime(rep, Path(project_dir))
+    _check_intervention(rep, Path(project_dir))
 
     rep.ok = not rep.problems
     return rep
--- a/core/cap_evolve/check.py
+++ b/core/cap_evolve/check.py
@@ -38,6 +38,34 @@ class CheckReport:
         return {"ok": self.ok, "stubs": self.stubs, "problems": self.problems, "notes": self.notes}
 
 
+def _check_runtime(rep, project_dir: Path) -> None:
+    """Validate the spec's ``runtime:`` value. OFFLINE — starts nothing, probes nothing.
+
+    Only the declaration is checked here (an unknown value is a hard problem, since the
+    spec reader would otherwise ignore it and deliver candidates the ``direct`` way).
+    Whether the runtime's services are actually UP is a run-time question, answered by
+    ``runtime.preflight`` — ``check`` must stay runnable on a machine with no stack.
+    """
+    from .specfile import read_yaml
+    from . import runtime as _runtime
+
+    spec_path = Path(project_dir) / "capevolve.yaml"
+    if not spec_path.exists():
+        return
+    try:
+        spec = read_yaml(spec_path.read_text(encoding="utf-8")) or {}
+    except Exception:  # a broken spec is reported by other checks
+        return
+    try:
+        rt = _runtime.declared(spec)
+    except _runtime.RuntimeError_ as e:
+        rep.problems.append(str(e))
+        return
+    if rt != _runtime.DIRECT:
+        rep.notes.append(f"runtime: {rt} (candidate delivered out-of-process; "
+                         "readiness is checked at run time, not here)")
+
+
 def load_adapter(project_dir: Path) -> CapabilityAdapter:
     """Import ``adapters/adapter.py`` and return its ``Adapter()`` instance."""
     project_dir = Path(project_dir)
@@ -190,6 +218,8 @@ def run_check(project_dir: Path, *, tolerance: float = 1e-6) -> CheckReport:
                 )
         except Exception as e:  # noqa: BLE001
             rep.notes.append(f"degenerate-trials probe skipped: {e}")
+
+    _check_runtime(rep, Path(project_dir))
 
     rep.ok = not rep.problems
     return rep
--- a/core/cap_evolve/cli.py
+++ b/core/cap_evolve/cli.py
@@ -656,33 +656,33 @@ def _cmd_run(argv):
         # run print TWO json documents, so `cap-evolve run | jq` could not parse it.
         if _stderr_is_usable():
             print(json.dumps(status), file=sys.stderr, flush=True)
-    # How the candidate will be DELIVERED (spec `runtime:`). Checked here, before any
-    # step runs: for an out-of-process runtime a dead stack makes every candidate's
+    # How the candidate will be DELIVERED (spec `intervention:`). Checked here, before any
+    # step runs: for an out-of-process intervention a dead stack makes every candidate's
     # deployment fail, and since that is correctly per-candidate infra noise, the run
     # would not crash — it would error every rollout and finish having measured nothing.
-    from . import runtime as _runtime
+    from . import intervention as _intervention
     try:
         if args.plan_only:
             # --plan-only prints the plan and executes nothing, so it must not preflight:
             # preflight STARTS a provisioned-but-stopped stack, which would give an
             # inspection flag service side effects (and fail outright when the stack is
             # not provisioned at all). The declared VALUE is still validated — that is
-            # offline, so an unknown runtime is refused here too.
-            runtime_rec = {"runtime": _runtime.declared(spec)}
+            # offline, so an unknown intervention is refused here too.
+            intervention_rec = {"intervention": _intervention.declared(spec)}
         else:
-            runtime_rec = _runtime.preflight(spec, skills, skills_dir)
+            intervention_rec = _intervention.preflight(spec, skills, skills_dir)
     except RuntimeError as e:
-        # RuntimeError, not just runtime.RuntimeError_ (a subclass of it): the runtime
+        # RuntimeError, not just intervention.InterventionError (a subclass of it): the intervention
         # SKILL raises plain RuntimeError for "port held by something else" and "health
         # check timed out". Catching only the subclass let those escape as a traceback
         # with NO JSON on stdout, which breaks the one-document contract every caller
         # parses — and the operator loses the diagnostic that says which service is wrong.
-        err = {"error": str(e), "step": "runtime-preflight"}
+        err = {"error": str(e), "step": "intervention-preflight"}
         print(json.dumps(err))
-        print(f"runtime preflight failed: {e}", file=sys.stderr)
+        print(f"intervention preflight failed: {e}", file=sys.stderr)
         return 1
     if not args.plan_only and _stderr_is_usable():
-        print(_runtime.describe(runtime_rec), file=sys.stderr, flush=True)
+        print(_intervention.describe(intervention_rec), file=sys.stderr, flush=True)
 
     cap_path = spec.get("capability_path", "seed_capability")
     ratios = f"{spec.get('split_train',0.5)},{spec.get('split_val',0.25)},{spec.get('split_test',0.25)}"
@@ -748,7 +748,7 @@ def _cmd_run(argv):
                           "spec_path": str(spec_path), "spec": spec,
                           "optimizer": optimizer_name, "optimizer_cmd": opt_cmd,
                           "algorithm": algorithm_name, "focus": algorithm_focus,
-                          "runtime": runtime_rec,
+                          "intervention": intervention_rec,
                           "target_model": spec.get("target_model", ""),
                           "orchestration_mode": orchestration_mode,
                           "gate_mode": spec.get("gate_mode", "auto (paired)"),
--- a/core/cap_evolve/cli.py
+++ b/core/cap_evolve/cli.py
@@ -671,7 +671,12 @@ def _cmd_run(argv):
             runtime_rec = {"runtime": _runtime.declared(spec)}
         else:
             runtime_rec = _runtime.preflight(spec, skills, skills_dir)
-    except _runtime.RuntimeError_ as e:
+    except RuntimeError as e:
+        # RuntimeError, not just runtime.RuntimeError_ (a subclass of it): the runtime
+        # SKILL raises plain RuntimeError for "port held by something else" and "health
+        # check timed out". Catching only the subclass let those escape as a traceback
+        # with NO JSON on stdout, which breaks the one-document contract every caller
+        # parses — and the operator loses the diagnostic that says which service is wrong.
         err = {"error": str(e), "step": "runtime-preflight"}
         print(json.dumps(err))
         print(f"runtime preflight failed: {e}", file=sys.stderr)
--- a/core/cap_evolve/cli.py
+++ b/core/cap_evolve/cli.py
@@ -656,6 +656,29 @@ def _cmd_run(argv):
         # run print TWO json documents, so `cap-evolve run | jq` could not parse it.
         if _stderr_is_usable():
             print(json.dumps(status), file=sys.stderr, flush=True)
+    # How the candidate will be DELIVERED (spec `runtime:`). Checked here, before any
+    # step runs: for an out-of-process runtime a dead stack makes every candidate's
+    # deployment fail, and since that is correctly per-candidate infra noise, the run
+    # would not crash — it would error every rollout and finish having measured nothing.
+    from . import runtime as _runtime
+    try:
+        if args.plan_only:
+            # --plan-only prints the plan and executes nothing, so it must not preflight:
+            # preflight STARTS a provisioned-but-stopped stack, which would give an
+            # inspection flag service side effects (and fail outright when the stack is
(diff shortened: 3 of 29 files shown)


aviweit:
## Tau2 airline test with spa enabled:

* Before the test: remove `.venv`, `.capevolve/project`, `cap-evove/vendors`

**PROMPT.md**

```
Follow RUN.md to run a cap-evolve optimization. Onboard this as a brand-new
benchmark — the intake/integration step should CLONE + INSTALL it (not assume it
exists). Here is everything intake needs:

# 0. DELIVERY PATH  (pick EXACTLY ONE arm)
- the CAPABILITY is the same either way — the airline agent's INSTRUCTIONS and its TOOL
                SURFACE. Only the delivery differs, and that is a spec key:
                  direct (no `intervention:` line) | intervention: spa
- ASK ME WHICH ONE and WAIT for the answer: `direct` or `spa`. Ask it as a PLAIN question with
                ONE SHORT SENTENCE per option — "direct: the runner reads the candidate's files
                in its own process, no extra services" / "spa: the candidate becomes a store
                skill the proxy injects, and needs the Skillberry stack plus the benchmark's
                environment service". Do NOT render preview panes, option cards, or dumps of
                spec fields; one sentence each is the whole point of the question.
                Prepare ONLY the arm I pick — one
                spec, one capability_path, one seed. Do NOT prepare both, do not default to
                one, and do not scaffold the other "for later": an unused arm's seed goes
                stale, doubles the onboarding, and invites a spec/seed mismatch that delivers
                candidates one way while the record says the other. The second arm is a
                SEPARATE onboarding in its own project (intake `--base`), done only if I ask.
- for `intervention: spa`, FOLLOW THE INTERVENTION SKILL —
                `skills/interventions/llm-proxies/spa/SKILL.md` and the seeding reference it
                points to. It owns provisioning, service lifecycle, the seed shape (one skill
                package + a frozen primitives module), the wrapper authoring rules, store
                import order, and the per-candidate deploy. Do NOT re-derive any of that from
                this prompt, and do not hand-roll a copy of spa_env. This prompt supplies only
                what the skill cannot know: the benchmark and the environment.

# 1. CAPABILITY TO OPTIMIZE  (a copy is edited each iteration; the original is never touched)
- what:         the airline agent's POLICY and its TOOL SURFACE
- tool surface: tau2's airline toolkit — 14 primitives (book_reservation, calculate,
                cancel_reservation, get_reservation_details, get_user_details,
                list_all_airports, search_direct_flight, search_onestop_flight,
                send_certificate, update_reservation_baggages, update_reservation_flights,
                update_reservation_passengers, get_flight_status, transfer_to_human_agents).
                This is the extraction source for EITHER rendering.
- IF direct:    capabilities [system-prompt, tools]; seed = tau2's canonical airline policy +
                its airline tool set; capability_sources = tau2's airline data_model (FlightDB,
                Reservation, Passenger, Payment) so the optimizer can write correct tool code.
                Seed tools must be CLEAN runnable code — no baked-in optimizer instructions in
                the docstrings.
- IF spa:       capabilities [skill-package]; the seed is GENERATED from the same 14 primitives
                per the intervention skill's seeding procedure. capability_sources [].
                THE SEED MUST BE NEUTRAL. my_skill/SKILL.md is frontmatter (name + ONE flat
                sentence of description) plus one line saying the tools live in scripts/ —
                and nothing else. Specifically:
                  * NO tool inventory. The store derives the tool list from scripts/ itself, so
                    a hand-written list is duplicate state that goes stale the first time the
                    optimizer adds or removes a wrapper.
                  * NO behavioural rules, workflow ordering, or read-before-write guidance.
                    Those are the OPTIMIZER's edit surface. Pre-writing them makes the baseline
                    a HAND-OPTIMIZED capability: it understates the headroom, it fits the seed
                    to the scorer's known failure classes before the first rollout, and it
                    stops the number being comparable to the direct arm, whose seed is tau2's
                    CANONICAL policy rather than someone's improved version.
                An 8-line SKILL.md is the target, not a 40-line one.
- actions:      [edit]

# 1b. INTERVENTION  (how the capability reaches the model — a spec key)
- direct:       the spec has NO `intervention:` line (direct is the default).
- spa:          the spec gets the top-level line:  intervention: spa
                (`capabilities:` says WHAT is edited; `intervention:` says HOW it reaches the
                model. capevolve.yaml-only — no CLI override.)
- VERIFY:       `cap-evolve check .capevolve/project` green for the ONE spec you prepared, and
                `intervention: sap` (a typo) REJECTED BY NAME, not silently defaulted to
                direct.

# 2. BENCHMARK / DATASET  (the eval) — INSTALL IT DURING INTAKE
- cap-evolve itself: there is NO virtualenv and nothing installed — create one first
                (uv venv -p 3.11 .venv) and `pip install -e core`.
- benchmark:    tau2-bench, airline domain
- repo:         https://github.com/skillberry-ai/skillberry-benchmarks.git
                pin commit a3a83266008275e9d800fd709927fa3dc4f23ec5 → vendor/skillberry-benchmarks
- install:      pip install -e vendor/skillberry-benchmarks/tau2/tau2-bench[skillberry]
- WHY this build, whichever arm: its runner already carries what SPA mode needs — Skillberry
                context headers on the agent's LLM calls, a merge of the proxy-side trajectory
                into tau2's own, and a `disconnect` at session end — and it still exposes the
                plain airline domain for direct. Using one build regardless of arm is what keeps
                a later comparison meaningful; a different runner per arm makes it worthless.
- domain:       direct → "airline";  spa → "airline_skillberry"
- tasks:        "adapter" — all 50 airline tasks from
                tau2.domains.airline.environment.get_tasks; no network in tasks()
- splits:       QUICK — pin exactly ONE of the 50 tasks as train = val = test in
                split_ids.json: {"train": ["9"], "val": ["9"], "test": ["9"]}.
                VERIFY that id exists in adapter.tasks() before pinning — they are tau2's own
                task ids, and a typo'd id silently empties the split instead of failing. No
                holdout, so the engine logs a splits_warning and the report must flag the test
                number as a FIT metric on ONE task — expected here, since the point is the
                pipeline, not the score.

# 2b. THE BENCHMARK'S ENVIRONMENT SERVICE  [SPA ONLY — skip entirely for direct]
- tau2 HAS one — the Environment Manager. The intervention skill requires such a service
                and aborts without it; these are the facts it cannot infer:
- start it:     port 8004, inline, from cap-evolve's venv:
                  LITELLM_LOCAL_MODEL_COST_MAP=True nohup python -c "
                  import asyncio
                  from tau2.orchestrator.environment_manager import EnvironmentManager
                  asyncio.run(EnvironmentManager(host='127.0.0.1', port=8004).run())
                  " > env_manager.log 2>&1 &
                ~10s to start (importing tau2 pulls in litellm); POLL the port rather than
                sleeping. LITELLM_LOCAL_MODEL_COST_MAP=True skips litellm's doomed remote
                cost-map fetch, which otherwise stalls startup until timeout.
- its CALL SHAPE (what the shim must speak):
                  base URL:     http://127.0.0.1:8004   (also set SPA_REMOTE_ENV_URL in .env)
                  URL:          {base}/{env_id}/tools/{tool_name}
                  request:      POST {"name": <tool>, "arguments": {...}}
                  success:      result["content"] is a JSON *string* — json.loads it
                  failure:      result["content"] is a plain message; non-200 → raise
- per-rollout identity: the store's executor injects `env_id`, so no session plumbing is
                needed here.

# 3. RUNNER  (the agent under test)
- how to run:   tau2's own batch runner (adapter.run_batch -> tau2.runner.run_tasks)
- fast eval:    ALSO implement the optional adapter method
                run_trials(tasks, ctx, *, n_trials, base_seed) -> {task_id: [Rollout, ...]}.
                Run ALL num_trials in ONE tau2 run_tasks call with num_trials=N (grouped by
                sim.trial) at TAU2_MAX_CONCURRENCY=125, and return {task_id: [trial0, trial1, ...]}
                (len n_trials, trial-ordered). When present, cap-evolve calls it ONCE per candidate
                instead of looping run_batch per trial; per-trial persistence
                (rollouts/<split>/<task>__<tag>__t<k>.json) is UNCHANGED so pass^k / SE / resume
                keep working. This collapses N sequential eval passes into one batched run.
- apply(): implement the ONE arm you were asked for and nothing else. For spa, call the
                intervention skill's deploy helpers; for direct, read the candidate's policy +
                tools and pass them to tau2's Environment constructor. Guard it on the
                CANDIDATE'S SHAPE (is a skill package present?) rather than on the spec, so a
                spec/seed mismatch fails loudly instead of delivering the wrong way silently.
                apply() must NEVER raise: record a deploy failure and let the rollouts come back
                errored, so the harness EXCLUDES the candidate instead of scoring it 0.0.
- models:       tau2 model settings TAU2_AGENT_MODEL / TAU2_USER_MODEL, as litellm
                "openai/<model>" strings against the gateway in 3b (e.g.
                openai/aws/gpt-oss-120b). [SPA ONLY] the AGENT's model is the SPA-routed name
                (default `ibm/skillberry-local`) while the user simulator stays on the gateway
                model — the intervention skill states why that boundary is a correctness rule.
- concurrency:  TAU2_MAX_CONCURRENCY=125 on the direct arm. [SPA ONLY] start LOW (e.g. 4) —
                every agent call funnels through one proxy and one store process.

# 3b. ENVIRONMENT — LLM ACCESS  (the run owner's environment supplies this; do not hardcode)
- endpoint:     an OpenAI-COMPATIBLE gateway the run owner configures — no host is fixed here.
- credentials:  OPENAI_BASE_URL (+ OPENAI_API_BASE, same value — litellm reads either) and
                OPENAI_API_KEY in the repo-root `.env`. Load them the way the existing example
                does (a small loader that walks parent dirs and setdefault()s, no python-dotenv
                dependency). NEVER hardcode an endpoint or key in adapter code and never invent
                one — if either is missing, ASK and WAIT, because a wrong value 401s/404s every
                rollout and the baseline is a row of zeros that reads as a bad capability rather
                than a bad config.
- the OPTIMIZER's credentials are SEPARATE and already handled: the `claude-code` optimizer
                inherits them from the run owner's Claude Code environment. Nothing about them
                belongs in the spec or the adapter.
- agent AND user simulator:  aws/gpt-oss-120b  (TAU2_AGENT_MODEL / TAU2_USER_MODEL), routed as
                the litellm string `openai/aws/gpt-oss-120b`. Normalize to that form ONCE and
                idempotently — `openai/openai/…` 404s. Model ids are gateway ALIASES and
                CASE-SENSITIVE (`Azure/…` and `azure/…` can coexist), so confirm the alias
                against the gateway's own catalog before the first run rather than assuming it.
- VERIFY BEFORE SPENDING: one non-agent probe call must return 200 before any baseline — a
                single completion with the RESOLVED model id, not a models listing, so the
                probe proves the key AND the alias. Give it a GENEROUS max_tokens: a reasoning
                model spends the budget on thinking and returns HTTP 200 with EMPTY content on
                a tight cap, which looks like a broken model rather than a truncated reply.
- TLS: if the endpoint's certificate is not trusted, fix it explicitly (point at the right CA
                bundle) and record what you did. Do not silently disable verification — a run
                whose transport security was turned off must say so in PROJECT.md.
- NEVER LEAK CREDENTIALS: they must not appear in logs, in the persisted trajectories, in
                PROCESS.md / JOURNAL.md, or in the run record. cap-evolve copies the
                trajectories VERBATIM into the optimizer's working dir each iteration and
                `store: git` COMMITS every iteration, so a credential echoed into a trace
                becomes a committed secret. Redact before persisting anything.
- CUSTOM-HEADER GATEWAYS ARE OUT OF SCOPE HERE. This prompt assumes bearer auth, which every
                path supports unmodified. A gateway that authenticates by a custom HTTP header
                needs BOTH pinned clones patched to send it — that is implementation work, not
                configuration; see the intervention skill plan's open decisions before attempting it.
- the OTHER environment facts live where they belong: venv + benchmark clone in section 2; the
                environment service in 2b; the Skillberry stack's ports and provisioning in the
                intervention skill, per section 0.

# 4. SCORER  (what to optimize against) — and WHERE the metric comes from
- metric:       tau2's own task reward in [0,1] (required actions performed + info communicated)
- metric source: `sim.reward_info.reward` per simulation; the per-check breakdown is in
                `sim.reward_info` (db_check / action_checks / communicate_checks /
                nl_assertions / env_assertions). adapter.score() reads the reward +
                reward_info that run_batch stashed. score() must be DETERMINISTIC on a fixed
                rollout (the check gate enforces it) — read the recorded reward, never re-run.
- feedback:     gold-AWARE but gold-SAFE, and ARGUMENT-LEVEL — this IS the learning signal, so a
                tool-name-only message ("action X was wrong") is too coarse: the optimizer can only
                pattern-match to prose rules and plateaus. For EACH failing check, localize the
                defect at the argument level:
                  * for each mismatched write/action, name the differing ARGUMENT key + the AGENT'S
                    OWN wrong value (e.g. "book_reservation: payment_id='credit_card_9' is not on the
                    user's profile; available=[credit_card_4421, gift_card_8]"; "update_reservation_flights:
                    called on reservation res_A but the task targets a different one");
                  * for communicate misses, name the un-stated value when derivable from the agent's
                    own state (e.g. "did not state the computed total cost ($150 from your own observed
                    amounts)").
                Gold-SAFE: NEVER read or print the gold/expected value — derive everything from the
                agent's OWN messages/tool-calls and the user's OWN profile/db state (parsed from the
                agent's get_user_details/get_reservation_details tool results in the trace). Use
                reward_info only to know WHICH action/argument failed (the gold action's arg KEYS are
                safe, its VALUES are not). Fall back to the tool-name message when a piece isn't safely
                derivable.
- objective:    maximize mean reward on the VAL split

# 4b. TRAJECTORIES  (the FULL traces the optimizer reads) — PATH IS AN INPUT
- where:        persist tau2's native per-task results (full transcript + reward_info) via
                run_tasks(save_path=...) into a per-eval dir UNDER THE RUN, e.g.
                <run_dir>/trajectories/val/. On the spa arm the Skillberry build already
                merges the proxy-side trajectory into tau2's own, so one trace covers both.
- expose:       adapter.trajectories(split) returns that directory; cap-evolve copies it
                VERBATIM into the optimizer's workdir as ./trajectories/ each iteration
                (return None to fall back to cap-evolve's own per-rollout JSON).

# 5. OPTIMIZER  (proposes the edits) + MODEL + CREDENTIALS + CONTEXT
- optimizer:    claude-code
- model:        claude-opus-4-8
- credentials:  a logged-in Claude Code session (or ANTHROPIC_API_KEY) — the optimizer's
                credentials are SEPARATE from the gateway in 3b, which is for the agent under
                test.
- runner_repo_path:  vendor/skillberry-benchmarks   (read-only context: task definitions, the
                real tool implementations, the reward checks)
- optimizer instructions: author .capevolve/project/optimizer/INSTRUCTIONS.md from the
                scaffolded template, keeping every {{...}} placeholder intact. ONE file, SCOPED to the
                capabilities of the arm you prepared, pointed at by
                optimizer_instructions_file. Do NOT re-author what the template already carries
                (depth mandate, non-overfitting guardrail, STEP-0 reading mandate,
                cross-iteration file protocol) and do NOT restate each capability's edit space —
                point at ./guidance/<cap>/SKILL.md, which the harness materializes.
- the benchmark facts to ADD (what the template cannot know): the trajectories live in
                ./trajectories/ as tau2 simulation records whose reward_info carries the
                per-check breakdown — use it to localize the exact defect (expected vs actual
                argument) while keeping every edit GENERAL; the scoring source is
                sim.reward_info.reward; the data-model module the direct arm's tool code imports
                is tau2's airline data_model, materialized under ./guidance/sources/.

# 6. BUDGET / GATE
- THIS SPEC IS THE SMOKE TEST — no separate mock tier. But still check the baseline before
                    letting it iterate: a task must really execute and return a real reward with
                    rollout.error null. An instant all-zero baseline means the rollouts errored;
                    fix that rather than reading it as a bad capability.
- algorithm:        hill-climb  (--focus all)
- max_iterations:   2           num_trials: 5     (1 task x 5 trials = 5 sims per eval)
- per-iteration optimizer $ cap:  optimizer_usd_per_iter 15   (claude --max-budget-usd, CLI-enforced)
- optimizer_max_turns: 100      (the $ cap is the real per-iteration ceiling)
- max_usd: 60       max_optimizer_usd: 40
- gate:             paired (per-task paired SE — banks real 1-task gains), k_se 0.2
                    NOTE: with n=1 there is no cross-task variance, so `paired` has no SE to
                    scale by and falls back to STRICT (any positive delta accepts, zero
                    rejects). Expect the reason string to say so. That is the gate degrading
                    correctly, not a failure — the pipeline is under test here, not the size or
                    the significance of the delta.
- store:            git          (every iteration committed for an inspectable process)
- stall:            2
- FINISH THE LOOP: let it reach finalize + report. An optimization that never seals the test
                    split has not exercised the whole pipeline, which is the only thing this
                    variant is for.
- ONE ARM PER RUN. If I later ask for the other arm, it is a separate onboarding in its own
                    project — and the two numbers are NOT like-for-like (under direct the
                    optimizer may change what the tools DO; under spa only how they are presented
                    and composed). Report them side by side, never averaged.
```

## Claude CLI

Provide it with the below prompt:

```
Use only information in /home/<username>/cap-evolve on the current Git branch.

Do not use external documentation, web searches, prior knowledge, other repositories, branches, tags, or commits.

Treat this repository as the sole source of truth. If required information is missing, state that it is missing rather than guessing.

Start with RUN.md and follow it exactly.
<PROMPT.md paste>
```

*When asked: select spa*

image
image
image
image

aviweit:
## Tau2 airline test with direct enabled:

Before the test: remove .venv, .capevolve/project, cap-evove/vendors

PROMPT.md same as above.

## Claude CLI

Provide it with the below prompt:

```
Use only information in /home/<username>/cap-evolve on the current Git branch.

Do not use external documentation, web searches, prior knowledge, other repositories, branches, tags, or commits.

Treat this repository as the sole source of truth. If required information is missing, state that it is missing rather than guessing.

Start with RUN.md and follow it exactly.
<PROMPT.md paste>
```

*When asked: select direct*

## Results

Two candidates were created each changed tools.py inline. The seed kept untouched. Trajectories indicates a successful end to end runs.

image

OsherElhadad (approved):

PR #411, opened 2026-08-31
results(skillsbench): task-by-task-87 EvoSkill comparison

bcarmeli:
Final task-by-task-87 SkillsBench sweep (10 trials/candidate, up to 4 iterations, Sonnet-5 agent + Opus-4.8 optimizer): 64/87 pass rate (73.6%) vs EvoSkill's reported 71.1%.

- `pass_k_analysis.md`: pass^k (tau-bench definition, all-of-k-succeed) at k=1/5/10 for a closer apples-to-apples read against EvoSkill's 5-independent-runs protocol — raw-only pass^5 is 67.1%, below EvoSkill's 71.1%.
- `results.json`: full per-task scoreboard (seed/candidates/best/test/status).
- `summary.md`, `heatmap.html`, `evoskill_comparison_chart.html`: category rollups and the visual comparison.
- `train_test_split_proposal.md`: a design (not yet executed) for evolving a skill on train tasks that share a subcategory and testing zero-shot transfer on held-out tasks in the same group.
- `C4_HANDOFF.md`: handoff for a follow-on worktree re-running the 10 tasks that improved over seed but didn't reach 1.0, including a verified audit that the optimizer has no gold/oracle access (only reward-derived feedback).

Scope note: this PR is deliberately limited to `skillsbench-results/task-by-task-87/` — the top-level `skillsbench-results/` folder also contains older milestone docs (20/43-task era) with headline numbers that are now stale/contradictory relative to this final sweep, plus a copy of the EvoSkill arXiv PDF. Those are left out of this PR.


OsherElhadad (changes requested):
## Review — data/results PR, no code

Reviewed by **Osher-Elhadad**. Checked out `results/skillsbench-task-by-task-87` (4 commits ahead of `main`, 9 behind, `MERGEABLE`), read all 8 files end to end, and independently recomputed every headline number from `results.json`.

**Verdict: request changes.** The underlying data work is genuinely good — `pass_k_analysis.md` is one of the more careful eval documents in this repo. The problem is that the *honest* version of the story lives only in `pass_k_analysis.md`, while the two files a reader actually opens (`summary.md`, `evoskill_comparison_chart.html`) carry an unqualified "we beat EvoSkill" claim that the data does not support at this n. Plus three artifacts in the same directory report three different pass rates.

---

### What I verified (all reproduced exactly from `results.json`)

| claim | recomputed | ✅ |
|---|---|---|
| 64/87 pass under `best ≥ 1.0 OR final_test ≥ 1.0` | 64 = 73.56% | ✅ |
| status counts (35/19/21/2/10) | identical | ✅ |
| median iterations 4.0 (n=51 non-null) | 4.0 | ✅ |
| all 8 per-category rows (total / no-signal / pass / rate / mean best) | identical | ✅ |
| NO_SIGNAL list (10 tasks) | identical | ✅ |
| C4 filter `best > seed + 1e-6 and best < 0.999` → 10 tasks | identical | ✅ |

Full `core/tests` suite green on this branch: **954 passed, 12 skipped** (`.venv-tau2`, editable `-e core`). No code touched, so no regression risk — confirmed.

No secrets. `MAC_HANDOFF.md:73` only *warns* not to commit `.env`. CodeQL green.

---

### Blocking

**1. The headline claim is not statistically supported, and this repo's own gate says so.**

64/87 = 73.6% vs EvoSkill 71.1% is **+2.5 pp against a binomial SE of 4.73 pp — 0.53 SE.** Two-proportion z = 0.36 (p ≈ 0.72). `docs/HONEST_EVAL.md` §3 is explicit that the whole point of `Δ > k·SE` is that "search is a noise amplifier… a `Δ > 0` rule banks noise." This PR reports a `Δ > 0` result as a win, with no SE, no CI, and no n-qualifier, in a repo whose differentiator is that it refuses to do that internally.

`docs/RESULTS.md:1101` already states the house position on this exact comparison: *"**Not directly comparable to EvoSkills' 71.1%.** Different paradigm…"* `summary.md` line 8 now says *"**Beats EvoSkill's 71.1% headline by +2.5 pp** on the same 87-task benchmark"* and the chart's callout says *"**Beating EvoSkill on full 87**"*. Those directly contradict committed docs. Please either add the error bar and downgrade to "statistically indistinguishable from EvoSkill's 71.1% (+2.5 ± 4.7 pp)", or carry `RESULTS.md`'s non-comparability caveat into both files.

**2. The pass metric takes `max(val, test)`, which breaks the seal — and the honest metric lands at 71.3%.**

The rule is `best(val) ≥ 1.0 **OR** final_test ≥ 1.0`. 45 of the 64 passes have no `final_test` at all (val-only), and it takes the better of the two splits where both exist. Recomputing with **test where available, val otherwise**:

> **62/87 = 71.3%** — i.e. essentially identical to EvoSkill's 71.1%, Δ = +0.2 pp.

`pass_k_analysis.md` §1 already warns that "val-sourced rows are systematically more optimistic than test-sourced rows" and lists five val-1.0 → test-<1.0 blowouts (`energy-unit-commitment` 1.0 → **0.1**, `organize-messy-files` 0.9 → **0.0**). `C4_HANDOFF.md` says outright: *"Report **test** reward, not val… val can overstate."* The headline metric then does the opposite. This 71.3% cut belongs in `summary.md`.

**3. Three artifacts in this directory report three different pass rates for the same sweep.**

- `heatmap.html` computes `(r.best ?? 0) >= 0.999` → **63/87 = 72.4%**
- `summary.md` / `evoskill_comparison_chart.html` → **64/87 = 73.6%**
- test-where-available → **62/87 = 71.3%**

Root cause found: `heatmap.html`'s embedded `const DATA` is a **stale, reduced copy** of `results.json` — it has dropped `final_test`, `iters_spent` and `subcategory` entirely (85 field diffs, all `null` on the heatmap side). So the heatmap physically *cannot* see `fix-erlang-ssh-cve`, the one task that passes on test but not val, hence 63 vs 64. `C4_HANDOFF.md:201` acknowledges the discrepancy and elects to leave it. Shipping an interactive artifact that renders a number contradicting the summary next to it isn't a good default — please regenerate `DATA` from `results.json` (or, better, `fetch('results.json')` so it can't drift again). `evoskill_comparison_chart.html:124` has a **third** hardcoded copy of the per-category rollup, and its OVERALL row hardcodes `71.1%` / `+2.5 pp` as literal strings rather than deriving them.

**4. `summary.md`'s "Mean best over all 87: 0.825" is wrong.** I get **0.8422**, which is what `pass_k_analysis.md:174` correctly reports ("84.2%"). 0.825 is closest to mean-test-where-available (0.8263). Two committed files disagree on the same statistic.

**5. Misfiled directory, and orphaned.** `third_party/` in this repo means *vendored upstream code* — its only current occupant is `third_party/spreadsheetbench/`, which ships `NOTICE.md` + `README.md`. These are **cap-evolve's own results**, not third-party material, and `third_party/skillbench/docs/train_test_split_proposal.md` puts a cap-evolve design doc under a third-party vendor path. Also note the directory is spelled `skillbench` while everything else in the repo uses `skillsbench` (`docs/REPRODUCE_skillsbench.md`, `ci/benchmarks/skillsbench/`, `examples/skillsbench/`).

I grepped the whole tree: **nothing outside `third_party/skillbench/` links to any of these 8 files.** `docs/RESULTS.md` — which already has two SkillsBench sections including one that names EvoSkill's 71.1% — is untouched. As merged, this is a folder nobody will find. Suggest `docs/results/skillsbench-87/` (or `docs/evidence/`, which already exists) plus a section or link in `docs/RESULTS.md`.

**6. DCO check is failing** — needs a `Signed-off-by` trailer before this can merge regardless of the above.

---

### Non-blocking, but worth fixing

- **The "excl. 10 no-signal → 83.1%" cut isn't defensible as stated.** `summary.md` describes all 10 as "cap-evolve had no lever," but its own breakdown splits them 4 infra-failures / **6 tasks that ran to completion and never moved off zero**. Those 6 are genuine optimizer failures. And per `MAC_HANDOFF.md`, `fix-visual-stability` is a *benchmark-side* broken `docker-compose.yml`. Dropping 10 tasks from the denominator while EvoSkill gets no equivalent courtesy is not an apples-to-apples cut — the chart's *"when the optimizer sees any signal at all, we solve 83%"* framing needs that qualifier attached.

- **The pass^k derivation for the 19 pruned tasks is stated more strongly than it holds.** §3.B: *"a mean of exactly 1.0 over 10 trials forces all 10 trials to be 1.0, so `c=10` is entailed, not guessed."* That is only entailed if the source value is **exact**, not displayed to 3 dp — 9 trials at 1.0 and one at 0.996 gives mean 0.9996, which also prints as `1.000`. Please confirm you read the unrounded float from `history.jsonl`, or soften "entailed" to "near-certain." (Cut A is unaffected; this only touches the upper-bound cut B.)

- **`pass^5 = 67.1%` deserves to be in `summary.md`.** The raw-data-only cut is *below* EvoSkill's 71.1%. The PR body says so; `pass_k_analysis.md` says so; `summary.md` — the file that gets read — doesn't mention pass^k at all. Right now the caveats and the headline live in different files, which reads as selective framing even though nothing here is fabricated. Also please sanity-check that §3.B's `mean pass^10 = 71.1%` is a real computation and not a transcription of EvoSkill's identical 71.1% (54/76 = 71.05% is plausible, so probably genuine — just confirm).

- **Not reproducible from the repo.** `pass_k_analysis.md` §1 sources its numbers from `.capevolve/run_task_*/bench_jobs/…` trees in `intake_skillbench_c2/c3/v2` worktrees that exist on one CCC host and are gitignored. No aggregation script is committed, so `results.json → summary.md / pass_k / heatmap` is a hand-run pipeline that cannot be regenerated or audited — which is precisely how items 3 and 4 above drifted. A ~50-line `aggregate.py` next to `results.json` would make all four artifacts derived-not-transcribed and kill that whole class of bug. Compare `docs/REPRODUCE_skillsbench.md`, which does give a repeatable path.

- **EvoSkill per-category numbers are eyeballed off a figure.** The chart footer is admirably upfront (*"read from Fig 6 of arXiv:2604.01687v1"*, un-weighted means where their 11 domains don't map onto our 8). Fine as directional colour, but the per-category Δ column shouldn't be styled as green "wins" (`+22 pp`, `+31 pp`) on figure-scraped values.

- **Two operational handoffs are being committed as permanent artifacts.** `C4_HANDOFF.md` opens with *"No `c4` worktree exists yet. No runs have been started"*, and both files hardcode one person's absolute paths (`/dccstor/knewedge2/boazc/workarea/…`). These are scratch/ops docs with a short half-life; committing them under `third_party/` guarantees a stale file. Worth reconsidering whether they belong in the repo at all, or at least whether the absolute internal paths should be genericised.

---

### Credit where due

`pass_k_analysis.md` is exemplary in the ways that matter: correct tau-bench pass^k (`C(c,k)/C(n,k)`, explicitly distinguished from pass@k), `n<k` treated as **undefined rather than 0** — exactly what `docs/HONEST_EVAL.md` §4 mandates — a per-task `split` column so val-vs-test provenance is auditable, every derived row flagged `D`, the partial-credit-fails-0.999 asymmetry called out, and a §1 CAVEAT that correctly identifies that our 10 trials measure within-run trial noise while EvoSkill's 5 runs measure optimizer-outcome variance. `C4_HANDOFF.md`'s gold-access audit is the right instinct, and naming the CTRF-assertion-message channel as a "bounded grey area" rather than quietly patching it is exactly the call I'd want. The `crystallographic-wyckoff-position-analysis` val-1.0/test-0.317 divergence being flagged as "probable val overfit — treat as diagnostic not headroom" is honest reporting against your own interest.

None of this is fabricated data. My objection is scoped to **framing and internal consistency**: apply the rigor that's already in `pass_k_analysis.md` to `summary.md` and the chart, make the four artifacts derive from one source, and land it somewhere discoverable.

### Minimum to flip me to approve

1. `summary.md` + chart: add the ±4.7 pp SE and drop/qualify "beats"; add the 71.3% test-where-available cut and the 67.1% pass^5 cut.
2. Regenerate `heatmap.html`'s `DATA` from `results.json` so 63 and 64 stop coexisting; fix mean-best 0.825 → 0.842.
3. Move out of `third_party/` and link from `docs/RESULTS.md`.
4. DCO sign-off.

Happy to re-review quickly once those land — the data itself I'm satisfied with.

bcarmeli:
Closing this in favour of the new **[`skillsbench-history`](https://github.com/skillberry-ai/cap-evolve/tree/skillsbench-history)** branch, which is where SkillsBench results now live. Nothing here is lost — all 8 files are already on that branch, and I verified that file by file rather than by name (7 of 8 are byte-identical blobs; `train_test_split_proposal.md` is a *newer* revision there, +60 lines).

@OsherElhadad — thank you for this review. It's the most careful reading any of this data has had, and finding 5 in particular changed how we're organising the work rather than just where one folder sits.

## Why close rather than fix in place

Your finding 5 is the reason: `third_party/` in this repo means *vendored upstream code* (its only other occupant is `third_party/spreadsheetbench/`), so cap-evolve's own results were misfiled there, and putting a cap-evolve design doc under `third_party/skillbench/docs/` compounded it. You also caught the `skillbench` vs `skillsbench` spelling.

That prompted a broader split we should have made earlier. There are two distinct kinds of contribution getting mixed:

1. **cap-evolve the library** — features, infra, docs. Belongs on `main`.
2. **SkillsBench optimisation output** — results, evolved skills, insights, where cap-evolve is merely the tool. Had no home, so each session reinvented one.

`skillsbench-history` is an orphan branch that gives (2) a home. Not `benchmark-history`, which is machine-written (regenerated by `record.py aggregate`, pruned by `benchmark-history-prune.yml`) and would eventually delete hand-authored content. Dashboard **metrics** still go through the normal `benchmark-history` pipeline — that's the next piece of work.

Where the 8 files went:

| this PR | `skillsbench-history` |
|---|---|
| `results/results.json` | `results/results.json` |
| `results/summary.md` | `results/task-by-task-87/summary.md` |
| `results/pass_k_analysis.md` | `insights/pass_k_analysis.md` |
| `results/heatmap.html`, `evoskill_comparison_chart.html` | `ui/` |
| `results/C4_HANDOFF.md`, `MAC_HANDOFF.md` | `handoffs/` |
| `docs/train_test_split_proposal.md` | `proposals/` |

One caveat on your suggested destination: you proposed `docs/results/skillsbench-87/` **on main**, and your objection was that as merged this is *"a folder nobody will find."* That objection gets **worse** on an orphan branch, not better — so a pointer from `docs/RESULTS.md` naming `skillsbench-history` as the source of truth for SkillsBench results is going up as a separate PR to `main`. Closing this without that pointer would just relocate the discoverability problem.

## Findings 1–4 were fixed, not relocated

Worth being explicit, because the files went across byte-identical, which means **your findings travelled with them.** Moving an artifact is not a way to resolve a review, so I recomputed all four from `results.json` before touching anything. Every number you reported reproduces exactly:

| your claim | recomputed |
|---|---|
| 64/87 = 73.6% under `best ≥ 1.0 OR final_test ≥ 1.0` | 64/87 = 73.56% ✅ |
| test-where-available → 62/87 = 71.3% | 62/87 = 71.26% ✅ |
| heatmap's own rule → 63/87 = 72.4% | 63/87 = 72.41% ✅ |
| mean best is 0.8422, not `summary.md`'s 0.825 | 0.8422 ✅ |
| +2.5 pp = 0.53 SE; two-proportion z = 0.36, p ≈ 0.72 | 0.52 SE, z = 0.36, p = 0.72 ✅ |

Fixed on the branch in [`7128e2af`](https://github.com/skillberry-ai/cap-evolve/commit/7128e2af):

- **Finding 4** — `0.825` → `0.8422`. It was simply wrong; `pass_k_analysis.md` had it right all along.
- **Finding 1** — the unqualified "Beats EvoSkill by +2.5 pp" is gone, replaced with "statistically indistinguishable from EvoSkill's 71.1%" carrying the SE, the z and the p. You were right that this contradicted `docs/RESULTS.md:1101`'s own "not directly comparable" position, and right that a repo whose differentiator is refusing `Δ > 0` internally shouldn't publish one externally.
- **Finding 2** — the 71.3% test-where-available cut is now in `summary.md`, with the 45-of-64-passes-are-val-only provenance and only 34 of 87 tasks having a test number at all. Also surfaced `pass^5 = 67.1%`, which is *below* their 71.1%. Your point that the caveats and the headline lived in different files was the fair hit — that reads as selective framing even when nothing is fabricated.
- **Your non-blocking cut-2 point** — folded in too, since it's the same class of problem: `summary.md` now records that only 4 of the 10 NO_SIGNAL tasks are infrastructure failures and the other 6 are genuine optimizer failures, so 83.1% isn't apples-to-apples against a number EvoSkill gets no equivalent courtesy on.

**Finding 3 is acknowledged but not yet fixed.** `heatmap.html`'s stale embedded `DATA` still renders 72.4%. Rather than leave that silently contradicting the file next to it, `summary.md` now states that the heatmap is the wrong one and why (its `DATA` dropped `final_test`, so it structurally cannot see `fix-erlang-ssh-cve`). Regenerating it — ideally as `fetch('results.json')` per your suggestion, so it can't drift again — is queued. That branch's data is still under active work, so several of these artifacts will be rewritten shortly and I'd rather regenerate once than twice.

Your non-blocking **`aggregate.py`** point is the one I most agree with and it's on the list: findings 3 and 4 both exist *because* `results.json → summary/pass_k/heatmap` is a hand-run uncommitted pipeline. A small committed script makes all four artifacts derived rather than transcribed and kills that whole bug class. Same for softening §3.B's "entailed" to "near-certain" unless the unrounded floats get re-read from `history.jsonl`, and for not styling figure-scraped EvoSkill per-category values as green wins.

**Finding 6 (DCO)** is moot now that this PR is closing.

The fork branch `results/skillsbench-task-by-task-87` stays put for now as a fallback until the new branch has been through a bit more use.

PR #415, opened 2026-09-01
Add stop_at_reward: stop optimizing once val reward hits the ceiling

bcarmeli:
## Summary
- Adds `Budget.stop_at_reward` (0 = off) checked centrally in `budget_exhausted()`: once the best val reward reaches this threshold, the optimizer loop stops — including before iteration 1 if the seed is already saturated.
- Fed by `harness.baseline()`/`reuse_baseline()` and the shared `record_iteration()` choke point, so hill-climb, gepa, and skillopt all inherit it with zero loop-body changes.
- Fires before other budget checks so the ceiling wins the reported `stop_reason` when multiple limits are satisfied at once.
- The sealed test split is still scored at `finalize()` — the held-out number is never lost.
- Off by default for existing specs (`0.0`); all 7 shipped templates now set `stop_at_reward: 1.0`.
- Full CLI + `--resume` plumbing, following the existing `--stall` pattern.

This formalizes a manual workaround used across the SkillsBench sweep: noticing `val=1.0` in `results.json` and killing the job by hand once a task is saturated and there's no room left for the optimizer to improve.

## Test plan
- [x] `core/tests/test_stop_at_reward.py` — 12 new tests (Budget/Spent round-trip + legacy tolerance, monotonic `best_val`, ceiling-fires/doesn't-fire, ceiling wins reason string over `max_iterations`, `--resume` override, and an integration test asserting `hill_climb_loop` returns `iterations == 0` on a saturated seed)
- [x] Full existing suite passes with no regressions (1 pre-existing flaky test unrelated to this change, confirmed via git-stash on unmodified code)
- [x] Backward compatibility verified against a real production `state.json` predating these fields
- [x] `cap-evolve run --help` shows `--stop-at-reward`
- [x] End-to-end offline verification: `harness.baseline()` against a saturated synthetic seed persists `budget.stop_at_reward` into `state.json` and `budget_exhausted()` fires immediately with the ceiling reason

### DIFF
--- a/core/cap_evolve/cli.py
+++ b/core/cap_evolve/cli.py
@@ -573,6 +573,9 @@ def _cmd_run(argv):
     p.add_argument("--max-usd", type=float, default=None)
     p.add_argument("--max-optimizer-usd", type=float, default=None)
     p.add_argument("--stall", type=int, default=None)
+    p.add_argument("--stop-at-reward", type=float, default=None,
+                   help="stop the optimizer loop as soon as the best val reward reaches this "
+                        "(0 = off, the default); still finalizes on the sealed test split")
     p.add_argument("--optimizer-max-turns", type=int, default=None,
                    help="per-iteration cap passed to the optimizer agent CLI (e.g. claude --max-turns)")
     p.add_argument("--follow", action="store_true",
@@ -604,7 +607,8 @@ def _cmd_run(argv):
     # CLI budget flags override the spec (None = "not passed", leave spec value).
     for flag, key in (("max_iterations", "max_iterations"), ("max_metric_calls", "max_metric_calls"),
                       ("max_usd", "max_usd"), ("max_optimizer_usd", "max_optimizer_usd"),
-                      ("stall", "stall"), ("optimizer_max_turns", "optimizer_max_turns")):
+                      ("stall", "stall"), ("stop_at_reward", "stop_at_reward"),
+                      ("optimizer_max_turns", "optimizer_max_turns")):
         v = getattr(args, flag)
         if v is not None:
             spec[key] = v
@@ -814,6 +818,7 @@ def _cmd_run(argv):
                 "--max-metric-calls", str(spec.get("max_metric_calls", 0)),
                 "--max-usd", str(spec.get("max_usd", 0.0)),
                 "--max-optimizer-usd", str(spec.get("max_optimizer_usd", 0.0)),
+                "--stop-at-reward", str(spec.get("stop_at_reward", 0.0)),
                 "--spec", str(spec_path)]
     if spec.get("split_ids_file"):
         base_cmd += ["--split-ids", str(spec["split_ids_file"])]
@@ -845,7 +850,8 @@ def _cmd_run(argv):
     # keep climbing past the original cap). Without an override the frozen budget stands.
     if args.resume:
         overrides = {k: getattr(args, k) for k in
-                     ("max_iterations", "max_metric_calls", "max_usd", "max_optimizer_usd", "stall")
+                     ("max_iterations", "max_metric_calls", "max_usd", "max_optimizer_usd", "stall",
+                      "stop_at_reward")
                      if getattr(args, k) is not None}
         if overrides:
             _RunDir.open(workdir / run_dir).update_budget(**overrides)
--- a/core/cap_evolve/dashboard.py
+++ b/core/cap_evolve/dashboard.py
@@ -464,6 +464,8 @@ def _budget_exhausted(budget, spent) -> str | None:
     """Which budget limit is spent out, or None. Mirrors RunDir.budget_exhausted."""
     if budget is None or spent is None:
         return None
+    if budget.stop_at_reward and spent.best_val >= budget.stop_at_reward - 1e-9:
+        return f"reward ceiling reached (best val {spent.best_val:.4f} >= {budget.stop_at_reward:.4f})"
     checks = (
         ("max_iterations", budget.max_iterations, spent.iterations),
         ("max_metric_calls", budget.max_metric_calls, spent.metric_calls),
--- a/core/cap_evolve/harness.py
+++ b/core/cap_evolve/harness.py
@@ -566,6 +566,7 @@ def baseline(adapter, seed_dir: Path, *, run_dir: RunDir, n_trials: int = 1, ks=
         json.dumps({"val": result.to_dict(), "best_id": "seed"}, indent=2), encoding="utf-8")
     run_dir.log_event("baseline", val=result.reward, stderr=result.stderr,
                       n_scored=result.n_scored, n_tasks=result.n_tasks)
+    run_dir.update_spent(best_val=result.reward)
     # The baseline is the number every later delta is measured against, so a
     # partially-evaluated one poisons the whole run rather than a single iteration.
     # Nothing downstream can detect this after the fact — baseline.json looks like a
@@ -633,6 +634,7 @@ def reuse_baseline(prior_run_dir: Path, *, run_dir: RunDir) -> SplitResult:
     result = SplitResult.from_dict(baseline_data["val"])
     run_dir.log_event("baseline_reused", prior_run_dir=str(prior),
                       val=result.reward, stderr=result.stderr)
+    run_dir.update_spent(best_val=result.reward)
     return result
 
 
@@ -1455,7 +1457,8 @@ def record_iteration(run_dir: RunDir, workdir: Path, cid: str, *,
     gepa only accepted ones) and they are not silently-droppable the way the three
     above were.
     """
-    run_dir.update_spent(iterations=1, accepted=None if indecisive else accepted)
+    run_dir.update_spent(iterations=1, accepted=None if indecisive else accepted,
+                         best_val=val if accepted and val is not None else None)
     run_dir.log_event("step", candidate=cid, accept=accepted, reason=reason,
                       val=val, parent=parent_id, parent_val=parent_val, **extra)
     delta = (val - parent_val if isinstance(val, (int, float))
--- a/core/cap_evolve/rundir.py
+++ b/core/cap_evolve/rundir.py
@@ -306,7 +306,7 @@ class RunDir:
 
     def budget_exhausted(self) -> tuple[bool, str]:
         b, s = self.budget, self.spent
-        if b.stop_at_reward and s.best_val >= b.stop_at_reward:
+        if b.stop_at_reward and s.best_val >= b.stop_at_reward - 1e-9:
             return True, f"reward ceiling reached (best val {s.best_val:.4f} >= {b.stop_at_reward:.4f})"
         if b.max_iterations and s.iterations >= b.max_iterations:
             return True, f"max_iterations reached ({s.iterations}/{b.max_iterations})"
--- a/core/cap_evolve/rundir.py
+++ b/core/cap_evolve/rundir.py
@@ -94,6 +94,7 @@ class Budget:
     max_usd: float = 0.0          # 0 = unlimited (total: runner + optimizer + intake)
     stall: int = 0                # consecutive no-accepts before stop; 0 = off
     max_optimizer_usd: float = 0.0  # 0 = off; separate cap on optimizer spend alone
+    stop_at_reward: float = 0.0   # 0 = off; stop as soon as best val reward reaches this
 
     def to_dict(self) -> dict:
         return {
@@ -102,6 +103,7 @@ class Budget:
             "max_usd": self.max_usd,
             "stall": self.stall,
             "max_optimizer_usd": self.max_optimizer_usd,
+            "stop_at_reward": self.stop_at_reward,
         }
 
     @classmethod
@@ -113,6 +115,7 @@ class Budget:
             max_usd=float(d.get("max_usd") or 0.0),
             stall=int(d.get("stall") or 0),
             max_optimizer_usd=float(d.get("max_optimizer_usd") or 0.0),
+            stop_at_reward=float(d.get("stop_at_reward") or 0.0),
         )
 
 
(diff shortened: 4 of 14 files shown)


OsherElhadad (changes requested):
Reviewed by checking the branch out locally, reading the full diff, and running the feature end to end rather than trusting the PR's own tests. Thanks for this — the placement is right and the ROI is real. I have two things I'd like fixed before merge and a handful of follow-ups.

## What I verified locally

**Generalizability — this is done correctly.** The wiring goes into the two genuine choke points, not into one algorithm's loop body: `RunDir.budget_exhausted()` (called by `harness.py:3003`, `gepa.py:535`, `skillopt.py:266`, `spend.py:154`, `goal_reminder.py:105`) and `harness.record_iteration()` (the shared iteration step that hill-climb via `run_step`, gepa, and agent-optimize via `commit.py:390` all route through). Nothing is benchmark- or algorithm-specific, and since `Score.reward` is `_clamp01`'d to `[0,1]` in `types.py:124`, `1.0` is a genuine universal ceiling rather than a guess about one scorer's scale. I confirmed agent-optimize inherits it for free:

```
$ spend.py --run-dir ... --project ...
stop            : True
stop_reason     : reward ceiling reached (best val 1.0000 >= 1.0000)
recommendation  : stop ['reward ceiling reached (best val 1.0000 >= 1.0000)']
```

**A/B on `examples/toy_calc`** (zero-API, deterministic, and it consumes `templates/project/capevolve.yaml` — so it exercises the new default directly). Same run, only `stop_at_reward` differing:

| `stop_at_reward` | iterations | best_val | sealed test |
|---|---|---|---|
| `1.0` (new default) | **1** | 1.0 | 1.0 |
| `0.0` (off = today) | **3** | 1.0 | 1.0 |

Two of three iterations saved, identical held-out number. The stop reason and `spent.best_val: 1.0` land in `state.json` as documented.

**Saturated seed → 0 iterations**, with `[CALC]` pre-added to the seed prompt so baseline val is already 1.0: `iterations: 0`, `best_id: "seed"`, `finalized: true`, `test_reward: 1.0`. The sealed test really does still get scored when the loop never runs — good, that was the scariest path and it holds.

**Full suite green**, pinned to this worktree (`PYTHONPATH` explicit, `-p no:randomly`): `1074 passed, 12 skipped` — no failures at all, not even the known order-flaky ones. Legacy `state.json` files missing both new keys load fine (`float(d.get(...) or 0.0)` is the right defensive shape).

**Security:** nothing to flag. No new external input, no paths, no deserialization; `argparse type=float` rejects junk and `update_budget`'s allowlist was correctly extended so a stray flag can't corrupt state.

## Please fix before merge

**1. `dashboard.py:417 _budget_exhausted` is a hand-maintained mirror of `budget_exhausted()` and wasn't updated.** Its docstring says "Mirrors RunDir.budget_exhausted", and it now doesn't. Reproduced with `budget=Budget(max_iterations=10, stall=2, stop_at_reward=1.0)`, `spent=Spent(iterations=1, best_val=1.0)`:

```
rundir.budget_exhausted    : True
dashboard._budget_exhausted: None
dashboard status           : interrupted | no finalize and no event for 4320 min — the run died, was killed, ...
```

A run that stopped cleanly on the ceiling without finalizing gets reported as a run that **died**. That's the same bug class #416 and #418 just spent two PRs fixing, so it'd be a shame to reintroduce it. One entry in that `checks` tuple (plus a `stop_at_reward` branch, since it's a floor not a ceiling) fixes it.

**2. Reconsider shipping `stop_at_reward: 1.0` alongside `num_trials: 1`.** Six of the seven templates you touched pair the new `1.0` with `num_trials: 1` (only `tau2_bench` sets 5). The baseline path — `harness.py:543` `update_spent(best_val=result.reward)` — records the seed's val **unconditionally**, with none of the significance discipline the rest of this repo insists on. A single-trial all-pass sample also has `stderr` of exactly `0.0`, so no gate or SE bar can catch it downstream either.

I built a stochastic toy adapter with a true per-trial pass rate of 0.8, otherwise stock template config, and swept the seed. One draw:

```
val split = 3 tasks
baseline_val = 1.0   stderr = 0.0   -> ceiling fires
iterations   = 0
sealed test  = 0.667
```

The val split drew all-pass by luck, the ceiling believed it, the optimizer got **zero** iterations, and the held-out score says there was real headroom the whole time. A neighbouring seed with `val = 0.667` ran its 2 iterations normally. This isn't hypothetical noise-worrying — it's a reproducible zero-iteration run on shipped defaults.

The feature itself is right; it's the *default* I'd push back on. Options, roughly in order of how much I like them: leave templates at `0.0` and document the knob (opt-in, matching how `max_optimizer_usd` shipped); or gate the baseline-path record behind `num_trials > 1`; or keep `1.0` only where trials are high enough to mean something. At minimum the field's doc comment should say out loud that it trusts a single val measurement, so nobody pairs it with `num_trials: 1` unknowingly.

## Follow-ups (not blockers)

**3. `spent.best_val` isn't backfilled, so `--resume --stop-at-reward` under-triggers on a pre-existing run.** Verified on a real run dir with the new keys stripped to simulate a pre-PR `state.json`:

```
champion on disk    : cand_0001
champion's real val : 1.0
spent.best_val      : 0.0
budget_exhausted    : (False, '')
```

A resumed run whose champion is already at the ceiling won't stop, and will burn iterations chasing a score it has. Backfilling in `update_budget` when `stop_at_reward` is newly set, or falling back to the champion's val when `best_val == 0.0`, would close it.

**4. Bare `>=` where this repo already decided an epsilon is needed.** `constraints.py:381` compares its `target_val_score` as `actual >= target - 1e-9`; `rundir.py:309` uses a bare `>=`. Partial-credit scorers are legal, and `stats.aggregate([0.7, 0.7, 0.7])` is `0.6999999999999998` — so `stop_at_reward: 0.7` silently never fires even though every task scored exactly 0.7. Low severity because I confirmed the shipped `1.0` is exact for all-1.0 rewards at every n I tried, but it bites any custom threshold. Reuse the `- 1e-9`.

**5. No range validation.** `stop_at_reward: -0.5` is truthy and `0.0 >= -0.5`, so the run dies before iteration 1 reporting `reward ceiling reached (best val 0.0000 >= -0.5000)` — a confidently wrong message, exactly the failure mode the `_NUM` comment at `constraints.py:58` warns about for `$0` ceilings. `2.0` silently means "never" since reward clamps to 1. Worth rejecting anything outside `[0, 1]` at load.

**6. Docs are thinner than the precedent.** `max_optimizer_usd` — the last field added here — is documented in `skills/phases/intake/inputs/INPUTS.md:94` and `skills/phases/intake/references/concepts.md:76` as well as `docs/OPTIMIZE_YOUR_OWN.md`. Those two intake surfaces are how the interview learns a budget field exists, so as written intake will never ask about `stop_at_reward` or explain it. `CHANGELOG.md` also looks like it wants an entry. The `OPTIMIZE_YOUR_OWN.md` block you did write is clear and correct.

**7. Worth a cross-reference: there are now two val-score ceilings.** The existing `stop_condition` prose path parses "reach val mean >= X" into `target_val_score` and `check_constraints` turns it into a `stop`. I don't think this PR duplicates it — that path is *advisory* and only ever read by agent mode, so hill-climb/gepa/skillopt (which never read `stop_condition`) genuinely had no enforced ceiling, and that's the gap this fills. But the two now disagree in three ways with nothing documenting which wins: enforced vs advisory, bare `>=` vs `- 1e-9`, and `spent.best_val` (monotonic max, accepted-only) vs `spend.py`'s `split_result_from_rollouts(best_id)` (recomputed from the champion's rollouts). The tidy end state is probably intake installing a parsed `target_val_score` straight into `budget.stop_at_reward` — the prose ceiling becomes deterministically enforced for every algorithm and there's one source of truth. Follow-up PR, not this one.

## Housekeeping

- **DCO is failing** and is what's blocking merge; everything else in CI is green. Needs a `Signed-off-by` line.
- Branch is 4 commits behind `main` (through `49fcedb6`) — mergeable, but worth a rebase, especially since #416/#418 touched the dashboard status code that finding 1 lives in.

Requesting changes on 1 and 2 only; 3-7 are happy to land as follow-ups. Nice piece of work otherwise — formalizing that manual "saw `val=1.0`, killed the job" workaround into the budget layer is clearly the right call, and putting it in `budget_exhausted()` rather than one algorithm's loop is exactly where it belonged.

— Osher-Elhadad

bcarmeli:
Thanks — this is an unusually useful review. Both blockers are fixed, plus finding 4 since it was one character. Rebased onto current `main` (through `373aca57`, so #416/#418's dashboard work is underneath finding 1's fix) and both commits are now signed off, so DCO is green along with the rest of CI.

## 1. `dashboard._budget_exhausted` mirror — fixed

Added the `stop_at_reward` branch ahead of the `checks` tuple, matching the canonical order so the same limit gets blamed first. Your repro now agrees:

```
budget=Budget(max_iterations=10, stall=2, stop_at_reward=1.0), spent=Spent(iterations=1, best_val=1.0)

rundir.budget_exhausted    : (True, 'reward ceiling reached (best val 1.0000 >= 1.0000)')
dashboard._budget_exhausted: 'reward ceiling reached (best val 1.0000 >= 1.0000)'
dashboard status           : budget_exhausted
```

Your framing — "the same bug class #416 and #418 just spent two PRs fixing" — is the part I took most seriously. A docstring saying "Mirrors RunDir.budget_exhausted" has now failed to keep them mirrored three times, so I don't think a fourth fix is worth much on its own. Added `test_dashboard_budget_mirror_agrees_with_rundir`: a 9-case matrix over the budget fields (ceiling reached, not reached, off-at-1.0, float slop, ceiling-vs-`max_iterations`, ceiling-vs-`stall`, each limit alone, nothing spent) asserting the two implementations return the same verdict **and** name the same limit, plus the `None` budget/spent guards. I confirmed it fails when I re-delete the fix:

```
AssertionError: ceiling reached: canonical says True, dashboard says None
```

That won't survive the real fix, which is one implementation instead of two, but it does mean the next drift fails in `core/tests/` rather than on a dashboard someone is reading during a run.

## 2. `stop_at_reward: 1.0` with `num_trials: 1` — fixed, your third option

You're right, and the stochastic-adapter repro is what settled it: a zero-iteration run on shipped defaults with the held-out score saying there was headroom the whole time is a real regression, not noise-worrying. I went with "keep `1.0` only where trials are high enough to mean something":

| template | `num_trials` | `stop_at_reward` |
|---|---|---|
| `project`, `harbor`, `huggingface_litellm`, `jsonl_litellm`, `skillsbench`, `swe_bench` | 1 | **`0.0`** (off) |
| `tau2_bench` | 5 | `1.0` |

Each carries a comment saying which side of that line it's on and why. Also took your "at minimum" ask — the `docs/OPTIMIZE_YOUR_OWN.md` block now says out loud that the field trusts a single raw val measurement with no regard for `num_trials`.

I preferred this to your first option (all templates `0.0`, document the knob) only because `tau2_bench` is a real worked example of the field being safe, and to your second (gate the baseline record behind `num_trials > 1`) because that changes the meaning of `spent.best_val` for every consumer to fix a defaults problem — it belongs with finding 3's backfill work, not here.

Same reasoning as finding 1 about durability: the invariant was only being enforced by seven yaml files happening to agree. `test_no_shipped_template_pairs_stop_at_reward_with_num_trials_1` walks every shipped `capevolve.yaml` and fails on any that pairs a ceiling with a single trial, so a future template gaining `num_trials: 1` or a ceiling can't quietly reintroduce it. Verified it fails when I flip `skillsbench` back to `1.0`.

Worth flagging one consequence: your A/B ran against `templates/project/capevolve.yaml`, so `examples/toy_calc` now takes 3 iterations again rather than 1. The saving you measured was real, it's just opt-in now.

## 4. Bare `>=` — fixed

`rundir.py:309` now uses `>= b.stop_at_reward - 1e-9`, matching `constraints.py:381`. Included despite being a follow-up because it's one character and your `aggregate([0.7, 0.7, 0.7]) == 0.6999999999999998` case is a silent never-fires for anyone who sets a partial-credit threshold. The mirror test carries that case, so the epsilon can't be dropped from one side only.

## 3, 5, 6, 7 — taking you up on the follow-up offer

Not in this PR, and I agree with your ordering:

- **3 (`spent.best_val` not backfilled on resume)** is the one with real user impact — a resumed run burning iterations chasing a score it already has. Your "fall back to the champion's val when `best_val == 0.0`" is the shape I'd implement.
- **5 (no range validation)** — agreed that `-0.5` producing "reward ceiling reached (best val 0.0000 >= -0.5000)" is confidently wrong in exactly the way the `_NUM` comment warns about. Reject outside `[0, 1]` at load.
- **6 (intake surfaces + CHANGELOG)** — correct that as written intake will never ask about the field, which makes it close to undiscoverable. `INPUTS.md` + `concepts.md` + a CHANGELOG entry.
- **7 (two val-score ceilings)** is the most interesting and the one I'd want to do properly rather than quickly. Your diagnosis matches mine: the `stop_condition` → `target_val_score` path is advisory and agent-mode-only, so hill-climb/gepa/skillopt genuinely had no enforced ceiling. And your end state — intake installing the parsed `target_val_score` straight into `budget.stop_at_reward` — is better than what this PR does, since it collapses the three disagreements you list into one source of truth. The `spent.best_val` (monotonic max, accepted-only) vs `split_result_from_rollouts(best_id)` (recomputed) divergence is the subtle one and probably wants deciding before 3 gets implemented, since the backfill has to pick one.

Happy to open these as separate issues now if you'd rather have them tracked than described here.

## Housekeeping

- **DCO** — both commits signed off, check is green.
- **Rebase** — done, onto `373aca57`.
- The full suite is green locally on the rebased branch (1173 passed). One unrelated failure, `test_spreadsheetbench_verify_turns.py::test_a_correction_supersedes_the_earlier_answer`, reproduces identically on the unmodified base commit and is green in CI, so it's a local environment artifact rather than anything this branch touches — flagging it rather than leaving you to wonder.

Ready for another look.

OsherElhadad (approved):

PR #416, opened 2026-09-01
fix(dashboard): stop reporting a live run as failed while it scores its baseline

eranra:
## The report

[Run 33492876620](https://github.com/skillberry-ai/cap-evolve/actions/runs/33492876620/job/99808280698)'s live dashboard showed this while the job was healthy and working:

| what the UI said | what was true |
| --- | --- |
| 🔴 **failed** | the job was 37 minutes in, running normally |
| "no baseline and no candidate was ever evaluated" | the baseline evaluation was in flight |
| **0s elapsed** · last event 9m 17s ago | 37 minutes of wall time |
| Δ val: "relative % undefined off a zero baseline" | there is no baseline at all yet, zero or otherwise |

The run then finished its baseline on its own and the same live snapshot flipped to `awaiting_agent`. Nothing was ever wrong with the run — only with what the dashboard said about it.

## Root cause

`_derive_status` decided "nothing has been scored" **before** reading the clock:

```python
if "finalize" in kinds: ...
if not has_baseline and not has_candidates:
    return "failed", "no baseline and no candidate was ever evaluated"   # ← before any freshness check
...
fresh = silent is not None and silent < STALE_AFTER_SECONDS              # ← never reached
```

So every live snapshot taken between the `splits` event and the first `baseline` event reported an outcome the run had not reached. That window is the whole baseline — minutes on smoke, hours on the full tier — and `live_push.sh` publishes every 5 minutes, so the wrong badge is what an operator sees for most of a long run's early life.

Behind that sat a second, deeper gap: **an evaluation logs nothing while it runs.** Between `eval_start` and the closing `evaluate` there was no event at all, so neither a human nor the reducer could tell "scoring in progress" from "the process died". This run's baseline was silent for ~40 minutes — five minutes short of `STALE_AFTER_SECONDS`, i.e. it was about to be declared dead for the opposite reason.

## The fix

**1. Freshness is evidence, and it is read first.** A moving log with nothing scored yet is `running`, and the reason says why rather than claiming a verdict:

> running — the seed's baseline is still being scored, no candidate has been evaluated yet; last event 557s ago

**2. `evaluate_candidate` brackets itself with an `eval_start` event** carrying split, tag, task count, trial count, worker count and total rollouts. That one event turns the longest silence in a run from *absence of information* into *recorded state*, which lets the reducer:

- give an **open** eval an 8-hour staleness window instead of 45 minutes, so a legitimately long split (639 test tasks, or swebench building a container per task) is not called dead partway through;
- report a run that dies inside one as `interrupted` and name where — *"scoring the seed on the val split (10 rollouts) and never returned — silent for 62 min"* — instead of `failed — nothing ran`, which is wrong about the one certain fact: something did run;
- name what a `running` run is busy with, on any status path;
- override `awaiting_agent`, which asserts nothing is happening, when an eval is provably in flight.

`eval_start` is filed under the same phase as the evaluation it opens (seed-on-val → baseline, FINAL-on-test → finalize).

**3. A live run's `elapsed_seconds` is measured to now**, not to its last event, with a new `elapsed_open` flag so the header renders `9m 17s elapsed so far` instead of a false total of `0s`. A finished run is unchanged (first event → last event).

**4.** The Δ-val KPI no longer blames "a zero baseline" for a null relative % when the real reason is that no baseline exists yet.

## Verification

Replaying the real snapshot's `events.jsonl` through the reducer at the age the snapshot was taken:

```
before:  status failed    | no baseline and no candidate was ever evaluated | 0s elapsed
after:   status running   | the seed's baseline is still being scored — no candidate
                            has been evaluated yet; last event 557s ago     | 9m 17s elapsed so far
```

An end-to-end toy_calc run confirms the bracketing and phase filing on a real run dir:

```
baseline  eval_start  split=val  tag=seed   rollouts=2
baseline  evaluate    split=val  tag=seed
finalize  eval_start  split=test tag=FINAL  rollouts=3
finalize  evaluate    split=test tag=FINAL
```

- **1116 core tests pass** (1 pre-existing unrelated failure: `test_cli_surface.py::test_spec_defaults_relative_to_project_not_the_cwd`, caused by a stale local `~/.claude/skills` manifest, fails identically on a clean `origin/main` tree).
- **129 frontend tests pass**, `tsc --noEmit` clean, eslint 0 errors.
- **8 new tests, each confirmed to fail without the fix** (verified by stashing only the source files and re-running).

New tests cover: live baseline → `running`; dead-before-anything → `failed`; budget-out-before-anything → `failed`; open eval past the 45-min window → `running`; open eval that never returned → `interrupted`, not `failed`; closed eval → ordinary window restored; agent-mode with an open eval → `running`, with a closed one → `awaiting_agent`; `eval_start` phase filing; the harness bracket pair itself; live vs finished `elapsed_open`; and the Δ-val hint on both a null and a genuinely-zero baseline.

## Not changed

`STALE_AFTER_SECONDS` (45 min) stays as-is for the step loop — the wider window applies only where there is positive evidence of work in flight. Status remains display-only; nothing in CI gates on it.


OsherElhadad (commented):
## Review — verified independently

Checked out `fix/live-status-running-not-failed` (`67b6e10c`) in a clean worktree and re-derived the claims rather than reading the description. Summary: **the fix is correct and I recommend merging it**, with one one-line follow-up I'd like folded in first (item A below), because it reproduces the exact class of over-claiming this PR set out to remove.

### What the PR does

Three separate defects, one root cause — the reducer stating an *outcome* for a run that has not reached one:

1. `_derive_status` (`core/cap_evolve/dashboard.py`) evaluated "no baseline and no candidates → `failed`" **before** reading the clock. The freshness check now runs first, and "nothing evaluated" only becomes `failed` once the log has actually stopped moving (or a cap/convergence already ended the run).
2. `evaluate_candidate` (`core/cap_evolve/harness.py`) now emits an `eval_start` event before the `redirect_stdout` block, carrying `split` / `tag` / `n_tasks` / `n_trials` / `workers` / `rollouts`. An `eval_start` with no closing `evaluate` is positive evidence that the silence is work, which buys the wider `EVAL_STALE_AFTER_SECONDS` (8h) window, names what the run is busy with, distinguishes `interrupted` from `failed`, and overrides `awaiting_agent`.
3. `elapsed_seconds` is measured to `now` while `status == "running"`, with a new `elapsed_open` flag so the header renders "9m 17s elapsed so far" instead of a false total of `0s`. The Δ-val hint no longer blames a zero baseline that does not exist.

### What I verified

- **Base is current.** `git merge-base origin/main pr-416` == `origin/main` HEAD (`d64aaf86`) — no rebase needed, `MERGEABLE`.
- **Read both source hunks line by line.** The bracketing is placed *before* `contextlib.redirect_stdout`, so it cannot interact with the phase skills' pure-JSON stdout contract (and `log_event` only appends to `events.jsonl` anyway). `n_trials` defaults to `int = 1`, so `int(n_trials)` in the `rollouts` computation cannot see `None`. `_phase_for` files `eval_start` under the same phase as the evaluation it opens. Every existing consumer of the event stream filters on `kind == "evaluate"` exactly (the cost ledger at ~1440, the evaluations table at ~1632, the val-series scan at ~990/1002), so the new kind cannot pollute a table or double-count spend; `event_count` grows by one per eval, which is honest.
- **All `evaluate_candidate` call sites are sequential** (harness baseline/step/finalize, gepa, skillopt ×2 back to back) — no nesting or concurrency, so the "last `eval_start` not yet closed by an `evaluate`" pairing is sound.
- **Full core suite green in my tree: `1075 passed, 12 skipped`, zero failures**, no order flakes. (I did not reproduce the `test_cli_surface.py::test_spec_defaults_relative_to_project_not_the_cwd` failure mentioned in the description — clean here.)
- **Negative control.** Reverted *only* `dashboard.py` and `harness.py` to `origin/main`, kept the tests: **9 fail**, each for the right reason (`assert 'failed' == 'running'`, `assert 0 == 1` for the missing bracket, etc.). Restored → 48 pass. So the tests do pin the fix, not the surrounding scaffolding. (Minor: the description says "8 new tests" — it is 9 that are load-bearing, out of 13 added.)
- **Committed SPA bundle is genuinely rebuilt** from the new source, not a stale artifact: `dist/assets/index-1567bS3K.js` contains `elapsed_open`, `" so far"`, `"the run is still going"` and `"no baseline measured yet"`, and `dist/index.html` points at that hash.
- **CI is fully green** — all 13 checks pass, including DCO and both CodeQL runs. The only red thing on the PR is the `skillberry-bot` auto-labeling comment, which is cosmetic and unrelated.

### A. One-line change I'd like before merge

The open-eval scan breaks only on `"evaluate"`:

```python
for e in reversed(events):
    k = str(e.get("kind") or "")
    if k == "evaluate":
        break
    if k == "eval_start":
        open_eval = e
        break
```

But an evaluation that **raises** does not log `evaluate` — `harness.py` logs `evaluate_error` (line ~1927) and the loop then logs a `step` and carries on. So the crashed eval's `eval_start` stays "open" forever from the reducer's point of view. Reproduced against this branch:

```
events: splits, baseline, eval_start(cand_0001), evaluate_error(cand_0001), step(cand_0001)

silent 80 min  → ('running', 'scoring candidate cand_0001 on the val split (30 rollouts); last event 4800s ago')
silent 7.3 h   → ('running', 'scoring candidate cand_0001 on the val split (30 rollouts); last event 26400s ago')
```

Two problems, both the defect this PR is about:

- The reason string asserts an evaluation is in flight for a candidate whose eval crashed and which the loop has already moved past — a state the log positively contradicts.
- The staleness window stays widened to 8h until the *next successful* eval closes the bracket. A run that dies right after an eval error therefore reports `running` for up to eight hours instead of `interrupted` after 45 minutes, which is the "dead run reported as live" failure mode `_derive_status` was written to prevent.

Fix is one word:

```python
if k in ("evaluate", "evaluate_error"):
    break
```

`evaluate_error` closes the bracket exactly as truthfully as `evaluate` does — either way, the evaluation is no longer in flight. A test for it fits alongside `test_a_closed_eval_puts_the_ordinary_stale_window_back`.

### B. Not blocking, worth a comment

Agent mode, baseline done, no candidates, and an `eval_start` that died 20 hours ago:

```
agent_mode=True  → ('awaiting_agent', 'baseline is done and `cap-evolve run` handed off — …')
agent_mode=False → ('interrupted',    'no finalize and no event for 1200 min — …')
```

The `not (alive and open_eval)` guard deliberately overrides `awaiting_agent` only for a *live* eval, which is the right call for the common case. But the dead-eval branch then loses the "started measuring and never came back" signal that the no-baseline path gets, and `awaiting_agent` still asserts that nothing is happening. Low stakes (display-only, and the agent-mode handoff really is idle most of the time) — a sentence in the docstring naming the case would be enough for the next reader.

Same category: `elapsed_open` is gated on `status == "running"`, so an `awaiting_agent` run shows a closed first-to-last-event total. Defensible, since the handoff genuinely is idle — noting it only so it reads as a choice rather than an oversight.

### Things I specifically tried to break and could not

- No docs carry a canonical event-kind table that `eval_start` would need to be added to (`docs/ARCHITECTURE.md` lists *phases*, not event kinds), and `CHANGELOG.md` is updated.
- Nothing in the repo gates on `status`, so the new `interrupted`-vs-`failed` distinction cannot change a CI outcome.
- `STALE_AFTER_SECONDS` is untouched, so the wider window applies only where there is positive evidence of work in flight — exactly as the description claims.
- The `silent is None` (no timestamps) path still falls to `failed`/`interrupted` rather than sneaking into `running`, since `fresh` requires a non-`None` `silent`.

### Recommendation

**Approve** — the diagnosis is right, the fix is at the root cause rather than the symptom, the frontend and committed bundle are consistent with it, and the tests genuinely fail without it. I'd like the one-word `evaluate_error` fix in item A folded in first, since without it the new wide window can itself keep a dead run labelled `running` for eight hours; B is a docstring nit at most. Not merging — that decision sits with the repo owner.

— reviewed by Osher-Elhadad

OsherElhadad (approved):

PR #419, opened 2026-09-01
fix(gate): measure only what the edit can reach, and stop asserting causality the measurement cannot support

OsherElhadad:
Four statistical-honesty defects, each root-caused against real run data: **run_finalrun6** — 7 candidates, 30 val tasks, 10 trials, agent-optimize.

`1117 passed, 12 skipped`; `agent-optimize/scripts/check.py` ok, linkcheck ok. 15 new tests, all of which fail against `main` at 49fcedb6.

> **Second commit (`351150d1`) fixes a blocking under-inclusion bug found in review, and corrects the power claims below.** The restriction is now considerably more conservative than the first revision claimed, and the numbers in this description are the post-fix ones.

---

## 1. The paired gate had no statistical power

**Evidence.** The gate measured a candidate's delta across **all 30** val tasks even though each edit touched a handful, so the ~26 tasks the edit could not reach contributed pure re-measurement wobble to the SE. `SE(paired Δ)` ran **0.022–0.035** while the real per-edit effects were **0.011–0.05** — the significance bar was as wide as the thing it was measuring. The same run's 7-way null-control replicate spread (**0.567–0.603**, byte-identical code) was statistically indistinguishable from its 7-way across-candidate spread (**0.570–0.607**): the round could not tell an edit from a re-measurement of the same code.

**Fix.** New `core/cap_evolve/footprint.py` reads the **named surfaces** out of the candidate's diff — definitions, calls, `"name": "…"` declarations on changed lines, **always** together with the enclosing definition — then asks which tasks' persisted rollouts mention any of them. Tasks outside that set enter the delta vector as **0.0 rather than as their measured wobble**: an edit that cannot reach a task has Δ=0 by construction.

Zero-padding rather than dropping keeps the vector's full **length**, so Δ̄ stays on the same **scale** as the val reward and every threshold, ledger row and val-curve point derived from it stays comparable. It changes **both** halves of the test, not just the variance — Δ̄ moves too, since the out-of-footprint deltas that used to be averaged in are now zeros. That is the mechanism, and it is precisely why the footprint must never under-include (see the review fix below).

**Three rules keep it honest**, and all three are load-bearing:

1. the enclosing definition is **always** part of the footprint — an edit inside a body reaches everything that calls the body, not just a rare helper it happens to call;
2. a surface reaching **≥80%** of tasks makes the edit **unlocalizable** → abandon and measure the full split, rather than dropping that name so the rarer surfaces beside it can define a confident narrow footprint;
3. a restricted vector's SE is **floored** by `harness.paired_se_floor` — the SE the in-footprint tasks' own per-trial noise implies — because the cross-task spread understates uncertainty on a small footprint.

**Replayed on run_finalrun6's own rollouts** (no new rollouts spent), all three rules applied:

| candidate | footprint | SE before | SE after | |
| --- | --- | --- | --- | --- |
| cand_1 | None | 0.0222 | — | guard across all 6 write tools → reaches everything |
| cand_2 | None | 0.0250 | — | batch tool wrapping `cancel_reservation` (26/30 tasks) |
| cand_3 | None | 0.0276 | — | bundle of cand_1 + cand_2 |
| cand_4 | None | 0.0244 | — | re-measurement of cand_2, byte-identical |
| cand_5 | 17/30 | 0.0318 | **0.0268** | |
| cand_6 | None | 0.0313 | — | bundle of cand_2 + cand_5 |
| cand_7 | 4/30 | 0.0346 | **0.0113** | floored; unfloored this was 0.0046 and a false ACCEPT |

**Be clear about what that shows.** Five of seven edits on this run reach the split too broadly to restrict at all, and **neither restricted verdict changes**. The mechanism earns its keep on genuinely narrow edits; on this run its value is that it **abstains** instead of manufacturing confidence. The worked synthetic case in `core/tests/` (a real +0.1 on 3 of 30 tasks) does go from unresolvable to accepted with the bar falling ~4x.

Getting the extractor to work on real data took two iterations worth recording. An *every-identifier* extractor produced **67 symbols for cand_2's 1.7 KB one-tool diff, 60 of them English** ("cancelled", "flights", "user") straight out of the docstring — those match every rollout, so the footprint was the whole split and the fix delivered nothing. Two things fixed it: taking only named surfaces, with **no `\s*` before `(`** (prose puts a space before a parenthesis, a call never does), and the adaptive ≥80% rule instead of a hand-maintained stop-list.

**Also free, on the reference side.** `split_result_from_rollouts` now accepts several tags and **pools their trials per task** (not their means), and `gate_check.py --current` takes a comma-separated list. `round.py` passes every null-control replicate the round has instead of whichever one carried the round-scoped tag — which was a coin flip: the same candidate read **+0.0867 against one replicate and +0.0067 against the other**. Those rollouts are already on disk.

**Configurable off, and never a crash.** Every unknowable case returns `None` and falls back to the exact full-vector behaviour that shipped before. `--no-footprint` / `footprint_gate=False` force it off.

## 2. False `broke`/`fixed` classification

**Evidence.** The threshold was `eps = 1e-9`, i.e. nothing. At 10 trials a `1.0 → 0.9` move is **one flipped rollout**, and it was being stamped as a task the candidate broke. It poisoned real reasoning: byte-identical code measured twice — `cand_2` and its own fresh-tag re-measurement `cand_4` — got **different `broke`/`fixed` labels**, and the optimizer's journal spent three rounds re-deriving that "task 27" was noise after the label asserted a regression on structurally unrelated candidates, one of them a **docstring-only edit that cannot change behaviour at all**.

**Fix.** `harness.move_is_resolved` — a move must clear **2·SE of its own per-task measurement** (the two sides' per-task SEs in quadrature; the same "smallest resolvable effect" the gate reports for the split mean, applied per task, from the variance data the gate already uses). It is one function because it is the single bar behind **all seven** places the framework makes that claim:

- `_candidate_task_impact` → LEDGER.md + the journal RESULT stamp
- `harness.run_step` **and** `gepa._full_val_gate` `no_regression` veto → the strongest consequence of the set, since it turns a gate-**passing** candidate into a rejection
- `gate_check.regressions` → the round table's diagnosis list
- `skillopt._categorize` → the within-epoch improved/regressed buffer the optimizer prompt reads
- `measure.py` → the sealed report's seed→best movement **and** its improved/regressed counts, which previously reported a second, unfixed bar in the same output
- `spend.py` → the "don't regress task X" constraint check

Six of those seven were still at `1e-9`, which is what copies of a bar get you.

Sub-threshold movers are reported as **`unresolved`**, not asserted either way. Adjacent text no longer calls the classification "objective". At one trial per task every per-task SE is 0, the bar collapses to `eps`, and the classification is byte-identical to before (pinned by a test).

## 3. Stale `gate_stderr` constant

**Evidence.** `gate_stderr` was published as **0.0738 on rounds i0, i3 and i6 alike** while the `gate_threshold` it is supposed to generate moved **0.0222 → 0.0244 → 0.0346** at `k_se = 1.0` — one frozen constant beside three different bars derived from it.

**Note:** #414 had already removed the field, so `main` publishes nothing rather than a wrong number. This wires the **real** value through instead, derived from the round table's `resolvable_effect_size` (`== 2·SE`, written by `gate.decide` from the vector it actually gated on), so `threshold == k_se · gate_stderr` holds for its own round.

## 4. Memory files promised but never created

**Evidence.** The staged pointer (`.capevolve/CLAUDE.md`) told the optimizer to read `./LEDGER.md`, `./PROCESS.md`, `./RUNMAP.md` and `./prior_iterations/<id>/`, and `JOURNAL.md`'s own seed text points at `./prior_iterations/<id>/diff.patch` for every prior candidate's exact edit. On disk the agent-mode working dir held only `INSTRUCTIONS.md`, `CLAUDE.md`, `guidance/`, `project/` and `trajectories/`. Root cause: only the deterministic loops' `_augment_instructions` built those files; agent mode reaches `_write_instructions_pointer` (via `OptimizerContext.inject` → `_inject_native_skills`) and got the pointer alone.

**Fix — create, not un-promise.** All of them are pure functions of the run's `step` events and candidate snapshots, so building them is cheap and idempotent, and `prior_iterations/<id>/diff.patch` is already assumed to exist by JOURNAL.md. `harness.seed_framework_memory` builds the whole set; the code that **writes** the promise now calls it and lists **only the files that exist**, so the text cannot drift away from the filesystem again. `host.py` seeds round 1 and `commit.py` re-seeds each round onto the candidate the next round copies from — the same carry-forward mechanism JOURNAL.md already used, so the LEDGER the optimizer reads includes the round just booked.

`rejected.jsonl` always existed with real data and was never mentioned; it is now named (by absolute path) in both pointers, with `history.jsonl` beside it.

---

## Deliberately NOT fixed

- **Increased trial count on the smaller in-footprint task set.** Needs subset-scoped evaluation plumbed through `round.py` and the evaluate phase, and `grow.py` already buys more `n` on an unresolved candidate. Footprint-scoped growth is the natural next step. Recorded in `references/algorithm.md`.
- **`dashboard._val_per_task_file`'s `fixed`/`broke`.** Read from a `val_per_task.json` an external run wrote; the dashboard reports what is in the file rather than re-deriving it, so the bar belongs at the writer.

## Risk worth naming

Footprint restriction changes what the significance test is computed over, so a candidate whose real damage lands **outside** the detected footprint would have that damage zeroed out — and the bias is asymmetric, since zeroing a regression raises Δ̄ and lowers the SE together. The three rules above exist for that reason, and the review's repro (a net-harmful candidate accepted at Δ̄ +0.05) is now a regression test. The recorded `val` reward and everything the report and val curve publish remain **whole-split means**; only the significance test's vector is narrowed.

### DIFF
--- a/core/cap_evolve/footprint.py
+++ b/core/cap_evolve/footprint.py
@@ -11,26 +11,34 @@ spread (0.570-0.607) — the measurement could not tell an edit from a re-measur
 Restricting the delta vector to the tasks inside the edit's footprint is what fixes that:
 a task the edit cannot reach has Δ=0 BY CONSTRUCTION, not by measurement, so it belongs
 in the vector as a zero rather than as noise (see ``harness._paired_deltas``, which does
-the zero-padding — the vector keeps its full length, so Δ̄ stays on the same scale as the
-val reward while the SE stops being dominated by tasks the edit never touched).
+the zero-padding). The vector keeps its full LENGTH, so Δ̄ stays on the same SCALE as the
+val reward — but be clear that this changes BOTH halves of the test: the SE stops being
+dominated by tasks the edit never touched, and Δ̄ itself moves, because the out-of-footprint
+deltas that used to be averaged in are now zeros. Δ̄ moving is not a side effect, it IS the
+mechanism (the whole claim is that those deltas were noise and do not belong in the
+average) — and it is also why under-inclusion is dangerous rather than merely weak: zeroing
+a task that really regressed raises Δ̄ and lowers the SE at the same time, both pushing
+toward accept. Hence the two rules in ``_close`` and ``in_footprint_tasks`` that make a
+broadly-reaching edit unlocalizable instead of narrowly footprinted.
 
 Both steps here are deliberately dumb and generic — no benchmark, capability, agent or
 optimizer shape is assumed:
 
   * ``touched_symbols`` reads the NAMED SURFACES out of a unified diff (whatever
     ``harness._diff_capabilities`` produced for this candidate): definitions, calls, and
-    ``"name": "…"`` declarations on the changed lines, falling back per hunk to the
-    enclosing definition when a hunk names nothing of its own. Not "every identifier" —
-    most changed lines in a real capability edit are docstring prose, and its words match
-    every rollout in the split.
+    ``"name": "…"`` declarations on the changed lines, ALWAYS together with the enclosing
+    definition (an edit inside a body reaches everything that calls the body). Not "every
+    identifier" — most changed lines in a real capability edit are docstring prose, and its
+    words match every rollout in the split.
   * ``in_footprint_tasks`` asks, per task, whether any of those names appears anywhere in
     the persisted rollout record. One flat substring search over the serialized rollout,
     rather than a walk of some particular trace schema: ``Rollout.tool_calls`` is the
     framework's only cross-adapter contract for "what the agent called", and adapters
     legitimately leave it empty while putting the same names in their own trace shape
     (harbor's tau2 rollouts do exactly this). The flat search reads both and cares about
-    neither. A name that reaches almost every task is then dropped, because it localizes
-    nothing — see ``_UBIQUITOUS_FRACTION``.
+    neither. If ANY of the edit's surfaces reaches almost every task, the whole footprint is
+    ABANDONED rather than that name dropped — the edit reaches almost everything, so there is
+    no narrow footprint to be had. See ``_UBIQUITOUS_FRACTION``.
 
 The search still OVER-includes, and that is the safe direction on purpose: over-including
 gives back some of the noise the restriction removes, while under-including would zero out
@@ -44,20 +52,28 @@ meaningless, no rollouts on disk, or a footprint that covers every task anyway. 
 that gets ``None`` falls back to the full delta vector, so nothing breaks when footprint
 detection cannot determine a surface.
 
-Measured on run_finalrun6's own rollouts, this is what the restriction does to the gate's
-bar for each of the run's 7 candidates (no new rollouts spent)::
+Measured on run_finalrun6's own rollouts (no new rollouts spent), with the safety rules in
+``_close`` and ``in_footprint_tasks`` applied::
 
-    cand_1  footprint None    SE 0.0222 -> 0.0222   (cross-cutting guard: no restriction)
-    cand_2  footprint 18/30   SE 0.0250 -> 0.0193
-    cand_3  footprint 13/30   SE 0.0276 -> 0.0208
-    cand_4  footprint 14/30   SE 0.0244 -> 0.0159
-    cand_5  footprint 11/30   SE 0.0318 -> 0.0255
-    cand_6  footprint 16/30   SE 0.0313 -> 0.0262
-    cand_7  footprint  4/30   SE 0.0346 -> 0.0046   (docstring-only edit to ONE tool)
+    cand_1  None    guard across all 6 write tools -> reaches everything -> full split
+    cand_2  None    new batch tool wrapping cancel_reservation (26/30 tasks)
+    cand_3  None    bundle of cand_1 + cand_2
+    cand_4  None    re-measurement of cand_2, byte-identical
+    cand_5  17/30   SE 0.0318 -> 0.0268
+    cand_6  None    bundle of cand_2 + cand_5
+    cand_7   4/30   SE 0.0346 -> 0.0113 (floored; see ``harness.paired_se_floor``)
+
+Be honest about what that shows: on THIS run, five of seven edits reach the split broadly
+enough that no restriction is available, and neither restricted verdict changes. The
+mechanism earns its keep on genuinely narrow edits, and its real value here is that it
+ABSTAINS instead of manufacturing confidence — an earlier revision of this module, which
+took the enclosing definition only when a hunk named nothing else and dropped ubiquitous
+symbols instead of abandoning, narrowed cand_7 to those 4 tasks at SE 0.0046 and ACCEPTED a
+docstring-only edit on two flipped rollouts.
 
 Note what stays unrestricted: the candidate's recorded ``val`` reward is still the mean over
 the WHOLE split, and so is every number the report and the val curve publish. The footprint
-narrows only the vector the significance test's variance is computed from.
+narrows only the vector the significance test is computed from.
 """
 
 from __future__ import annotations
@@ -90,12 +106,12 @@ _SURFACE = re.compile(
 _ENCLOSING = re.compile(
     r"\b(?:def|class|function|func|fn|sub|method)\s+([A-Za-z_][A-Za-z0-9_]{2,})")
 
-#: A symbol found in this fraction of the split's tasks or more is discarded before the
-#: footprint is built. It localizes nothing — whatever it is (a shared exception type, a
-#: helper every rollout touches, a common English word that survived the extractor), a
-#: footprint containing it is the whole split. Adaptive rather than another stop-list: the
-#: run's own rollouts say which names discriminate, and no hand-maintained word list has to
-#: guess it per capability, language or benchmark.
+#: A surface reaching this fraction of the split's tasks or more makes the edit UNLOCALIZABLE:
+#: ``in_footprint_tasks`` abandons and the caller measures the full split. It is not dropped so
+#: the rarer surfaces beside it can define a footprint — that is precisely how a 28-of-30-task
+#: edit got narrowed to its 3-task helper. Adaptive rather than another stop-list: the run's own
+#: rollouts say which names discriminate, so no hand-maintained word list has to guess it per
+#: capability, language or benchmark.
 _UBIQUITOUS_FRACTION = 0.8
 
 #: Keywords and ubiquitous library names that take a call-shape and would otherwise be
@@ -126,10 +142,18 @@ def touched_symbols(diff_text: str, *, max_symbols: int = 40) -> set[str]:
     lines = list((diff_text or "").splitlines())
 
     def _close():
-        # A hunk that named nothing of its own is an edit INSIDE something — its surface is
-        # the enclosing definition. Only then, so a hunk that does name its surfaces cannot
-        # drag in whatever else happens to sit within the diff's context window.
-        out.update(hunk or ({enclosing} if enclosing else set()))
+        # ALWAYS both — never "the enclosing definition only when the hunk named nothing".
+        # An edit INSIDE a broadly-reached function that adds a call to a rare helper is
+        # reached by everything that calls the function, not only by the helper's own callers:
+        # taking the helper alone footprints a 28-of-30-task edit to 3 tasks and zero-pads 25
+        # tasks that really moved. That is UNDER-inclusion, the one direction this module must
+        # never fail in — zero-padding an out-of-footprint regression raises Δ̄ *and* lowers
+        # the SE, so a net-harmful candidate can come out of the gate accepted. Including a
+        # neighbouring definition that merely sits inside the diff's context window costs some
+        # of the power gain instead, which is the safe direction.
+        out.update(hunk)
+        if enclosing:
+            out.add(enclosing)
 
     for line in lines:
         if line[:3] in ("+++", "---"):
@@ -168,9 +192,9 @@ def in_footprint_tasks(run_dir, tags, symbols, split: str = "val") -> set[str]:
     if not d.is_dir():
         return set()
     symbols = set(symbols)
-    # Which tasks each symbol reaches, counted per symbol rather than unioned as we go, so a
-    # symbol that reaches nearly everything can be dropped before the union — see
-    # _UBIQUITOUS_FRACTION.
+    # Which tasks each symbol reaches, counted PER SYMBOL rather than unioned as we go: one
+    # ubiquitous surface has to be recognizable as such, and it cannot be once the reaches are
+    # merged — see _UBIQUITOUS_FRACTION and the abandon below.
     reach: dict[str, set[str]] = {s: set() for s in symbols}
     tasks: set[str] = set()
     for tag in tags:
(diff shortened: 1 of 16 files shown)


OsherElhadad:
# Independent pre-merge review — REQUEST CHANGES

_Posted as a comment: `gh pr review --request-changes` is refused on one's own PR (same git identity). Treat this as a blocking review._

## Independent review — verified locally, not from the PR description

Checked out `d11c5f68` in a clean worktree and re-ran everything myself.

### What I confirmed true

- **`1114 passed, 12 skipped`** — exact match to the claim (155s, `core/tests/`).
- **All 12 new tests fail on `main` at `49fcedb6`** — copied the new test file onto base and got `12 failed`. The tests really do pin the defects.
- **`move_is_resolved` is wired into all 4 claimed sites**: `harness._candidate_task_impact:1121`, `gate_check.regressions:100`, `measure.py:107`, `spend.py:78`.
- **Surgical diff** — 13 files, +1160/-113, every hunk traceable to one of the 4 fixes. No drive-by refactors.
- **Generalizable** — nothing keyed on tau2/taubench/airline or on a particular optimizer/agent. The only benchmark mention is one illustrative comment (`footprint.py:31`). Symbol extraction is language-loose (`def|class|function|func|fn|sub|method`) and the ubiquity filter is adaptive rather than a stop-list, which is the right call.
- `gate_stderr` derived from `resolvable_effect_size / 2` (`commit.py:111`); `seed_framework_memory` wired at `host.py:669` and `commit.py:431`.

Good PR. One blocking issue.

---

## BLOCKING — footprint mis-localization can accept a net-harmful candidate

`footprint_gate` defaults to **True** (`harness.py:2055`), so this is on by default.

Two rules interact badly:

1. `touched_symbols._close()` takes the enclosing definition **only when the hunk named nothing of its own** (`out.update(hunk or {enclosing})`). But a hunk's own named surfaces are routinely the **callees the edit added**, not the **edited surface whose behaviour changed**.
2. The ≥80% ubiquity filter **silently drops** a symbol and unions the remainder, rather than concluding the edit is unlocalizable.

Together, an edit *inside* a broadly-reached function that adds a call to a rarely-reached helper gets a footprint of only the helper's tasks. **None of the three safety nets fire** — `symbols` is non-empty, `tasks` is non-empty, and the footprint is a strict subset of all tasks — so `footprint()` returns a confident wrong restriction instead of `None`.

### Repro

30 val tasks; `book_reservation` appears in 28 tasks' rollouts, `audit_trace` in 3.

```diff
 def book_reservation(self, pnr):
+    self.audit_trace(pnr)
     return self.commit(pnr)
```

`touched_symbols` → `{'audit_trace'}` → footprint `{t0,t1,t2}`. The edit's real reach is 28 tasks. `book_reservation` is never extracted at all, because the hunk named `audit_trace`.

Now let the candidate genuinely fix `t0..t2` (+0.5 each) and genuinely break `t10..t15` (1.0 → 0.0):

```
true full-split Δ = -0.1500     full vector : Δ̄ -0.1500, SE 0.0836  -> reject   (correct)
                                footprint   : Δ̄ +0.0500, SE 0.0279  -> ACCEPT   (wrong)
```

A candidate that is net **-0.15** is accepted.

### Why the stated mitigation does not cover this

The "Risk worth naming" section frames the bias as symmetric-ish, held in check by over-inclusion. It isn't. Zeroing an out-of-footprint regression **raises Δ̄ and lowers SE** — both push toward accept. The over-inclusion bias only helps when the footprint is too *wide*; this path makes it too *narrow* while reporting `restricted: true` with full confidence.

Relatedly, `footprint.py:60` and `harness.py:744` both say the restriction changes "only the variance". That is not correct — zero-padding changes the numerator too (`Δ̄ = (1/N)·Σ_in-footprint δ`), and that is precisely the mechanism of the flip above. Please correct that sentence; future reviewers will lean on it.

### Suggested fix (I verified both parts are needed)

```python
# 1) the edited surface is the enclosing definition too, not only its callees
out.update(hunk | ({enclosing} if enclosing else set()))

# 2) a ubiquitous symbol means the edit is UNLOCALIZABLE -> no footprint,
#    rather than dropping it and narrowing to whatever else was named
if any(len(ts) >= cap for ts in reach.values()):
    return set()
```

On the repro: symbols become `{audit_trace, book_reservation}`, `book_reservation` trips the ubiquity filter, footprint → `None`, full vector, **correctly rejects**. This is also what the PR body already promises ("anything unlocalizable falls back to the full split") — the code just doesn't implement it that way.

Impact on the replay table: `cand_1` ("cross-cutting guard on all 6 write tools → no restriction") is unchanged, and `cand_7` (one rare symbol, docstring-only) keeps its 0.0046. Some middle candidates may lose part of their power gain — that is the safe direction, and worth re-running the replay to quantify. **Given the author flagged `cand_7` flipping reject → accept-eligible, I'd want the replay re-run under the corrected extractor before this merges.**

---

## Non-blocking

**1. `move_is_resolved` covers 4 sites; the framework asserts "broke" in at least 3 more, still at the old bar.**

- `harness.py:2234` and `gepa.py:778` — the `no_regression` gate, which **vetoes acceptance** and emits literally `REJECTED by no-regression gate (broke {...})` at `eps = 1e-9`. A single flipped rollout in ten can veto a genuinely better candidate — a *stronger* consequence than the LEDGER label §2 fixes. Mitigated only by `no_regression` defaulting to `False`.
- `skillopt.py:176` `_categorize` — labels a task `regressed` via `passed_before and not passed_after`, i.e. 1.0 → 0.9 at 10 trials, and feeds the optimizer's within-epoch buffer. Exactly the §2 defect, in the loop §2 doesn't touch.
- `measure.py:74-75` still counts `improved`/`regressed` at `1e-9` while line 114 in the same file uses the resolved bar — two different bars in one output.

The docstring's "THE one resolution bar behind every per-task broke/fixed/regressed claim the framework makes" overstates it. Either widen the fix or scope the claim — the current wording will stop the next person from looking.

**2. `touched_symbols` misses a call written with a space before `(`** (`self.cancel_reservation (rid)` → `set()`). The "no whitespace before `(`" safeguard is a style convention, not a language rule. It degraded to `None` in my check, so it's safe, just narrower than the comment claims.

**3. Nit:** `commit.py:111` rounds `gate_stderr` to 6dp and the test asserts exact float equality `gate_stderr * k_se == round(threshold, 6)`. Fine at `k_se=1.0`, fragile at non-integer `k_se`.

---

Requesting changes for the footprint mis-localization only. The other three fixes look sound and well-evidenced, and the test suite genuinely pins them.

OsherElhadad:
# Second-pass review of `351150d1` — 🟢 APPROVE, safe to merge

Re-verified in a fresh worktree at `351150d1`. My blocking finding is closed, the new `paired_se_floor` holds up under stress, and all three secondary sites are genuinely fixed.

## 1. The blocking bug is closed

Both fixes landed exactly where they needed to:
- `_close()` now always unions the enclosing definition (`out.update(hunk)` then `if enclosing: out.add(enclosing)`).
- `in_footprint_tasks` now **abandons** on a ubiquitous symbol (`return set()`) instead of dropping it and keeping the rest.

Re-ran my original repro (30 tasks, `book_reservation` in 28 rollouts, `audit_trace` in 3, edit inside the former adding a call to the latter):

```
symbols  : ['audit_trace', 'book_reservation']     (was: ['audit_trace'])
footprint: None (unlocalizable) -> full vector     (was: 3/30)
true full-split delta = -0.1500
gated mean = -0.1500   ACCEPT=False                (was: +0.0500, ACCEPT=True)
VERDICT: CLOSED
```

The `abandon` comment in `in_footprint_tasks` names the exact failure mode with the 28-of-30 example. That is the right place for it — the next person to "optimize" this by dropping the symbol instead will now see why not.

## 2. `paired_se_floor` — correct, and the right shape of fix

**The math is right.** For a zero-padded vector, `Var(Δ̄) = (1/n²)·Σ_{t∈fp} Var(δ_t)`, so `SE = √(Σ(σ²par+σ²cand))/n_total` is exactly the trial-noise SE of the mean it bounds. Out-of-footprint tasks correctly contribute 0 — their delta is 0 by construction, with no uncertainty. And every task with a nonzero delta is necessarily in `fp` (since `_paired_deltas` zeroes everything else), so the floor covers all the uncertainty that is actually present. No gap.

**Verified inert when unrestricted.** Both call sites guard on `fp is not None and deltas`; the default is `0.0`; and `se = max(se, 0.0)` is a no-op since `se >= 0`. I fuzzed 4000 random delta vectors comparing `decide(...)` with and without `paired_se_floor=0.0` — **0 differing decisions**, identical `to_dict()` output. Claim confirmed.

**Reproduced the cand_7 fix:**
```
4-task footprint, deltas {0,+.1,0,+.1}, per-task SE 0.1 (one flipped rollout in 10)
  spread SE 0.0046  floor 0.0094  -> gate SE 0.0094
  no floor : accept=True   (Δ̄ +0.0067 > SE 0.0046)
  floor    : accept=False  (Δ̄ +0.0067 <= SE 0.0094)
```

**Tried to make it too conservative — couldn't.** A 1-task footprint with a genuine large move (δ=1.0) gives spread SE 0.0333 vs floor 0.0047, so `max()` keeps the spread and the real narrow effect is not suppressed. The floor only binds where the spread is implausibly small, which is exactly the intended domain.

Two notes, neither blocking:

- **`max()` rather than quadrature is mildly anti-conservative in principle.** The honest SE combines task-sampling variance *and* trial noise; `max()` takes whichever is larger. It's defensible (the spread estimator already absorbs some trial noise, so adding in quadrature would double-count), and it is strictly more conservative than what shipped before. Just worth knowing the floor is a floor, not a full variance decomposition.
- **`if not par_se and not cand_se: return 0.0`** disables the floor entirely when neither side has per-task SEs — correct for a single-trial run. But if only *one* side has them, the other contributes `0.0` via `.get(..., 0.0)` and the floor comes out ~1.41x too small (I measured 0.0094 → 0.0067 on the cand_7 shape). Partial-rollout-data runs get a partially-understated floor rather than none. Worth a one-line guard someday; not a merge blocker, since it's still strictly better than no floor.

Also: `tasks=fp` while `n_total=len(paired_deltas)` means a footprint task absent from the paired vector (dropped for invalid trials) inflates the floor slightly — conservative direction, fine.

## 3. Secondary `eps=1e-9` sites — all three genuinely fixed

- `harness.run_step` no_regression veto → now `and move_is_resolved(...)` on per-task stderrs, with a comment noting the veto's consequence is *stronger* than the reported lists. ✅
- `gepa._full_val_gate` → same treatment, `move_is_resolved` imported at the top. ✅
- `skillopt._categorize` → both `improved` and `regressed` now require `resolved`. ✅
- Bonus, `measure.py` → `improved`/`regressed` now count `_per_task_movement`'s resolved movers and publish `unresolved` alongside, killing the two-bars-in-one-output problem. ✅

Minor observation on `skillopt._categorize`: an unresolved 1.0→0.9 move now falls through the whole `if/elif` chain into no category at all (previously `regressed`). That's the correct outcome, but unlike LEDGER/measure.py it isn't surfaced as `unresolved` anywhere. Fine for a prompt buffer — noting it only so it isn't mistaken for an omission later.

The docstring now lists all seven call sites, so the overclaim I flagged is gone.

## 4. Tests

**1117 passed, 12 skipped** — matches the claim exactly (156s).

## 5. On the corrected replay numbers

Appreciate the straight retraction. The revised table is the more useful artifact: 5/7 candidates abstaining and neither restricted verdict moving is a real result, and `footprint.py`'s "Be honest about what that shows" paragraph — which explicitly records that the *earlier* revision would have accepted cand_7 on two flipped rollouts — is exactly the kind of note that stops someone re-introducing the bug in six months.

The `_paired_deltas` docstring correction is good too: "It changes BOTH halves of the test, not just the variance… Zeroing a task that really regressed raises Δ̄ and lowers the SE together, both pushing toward accept." That now states the hazard precisely instead of implying the restriction is variance-only.

Worth being clear-eyed that the power gain is now carried by the synthetic narrow-edit case rather than by this run — but the mechanism is sound, it abstains by default, and the `paired_se_floor` closes the one way a restriction could manufacture confidence. That is a net improvement in gate honesty even at zero measured power gain on run_finalrun6.

---

**Approving.** My blocking finding is fixed at the root, the new floor is correct and verified inert, and the diff is still surgical (9 files, all traceable). The two notes in §2 are follow-ups, not blockers.

PR #426, opened 2026-09-03
dashboard: make the bind address configurable so the UI is reachable off-box

aviweit:
## Problem

The dashboard server bound `127.0.0.1` unconditionally, so it was reachable only from the machine running the pipeline. When the browser lives elsewhere — a remote box, a container, a VM — nothing outside can reach a loopback-only listener, and the URL the run prints at the top is a dead link with no indication why.

## Change

The bind address is now resolved as **explicit flag > `CAPEVOLVE_DASHBOARD_HOST` > `127.0.0.1`**:

```bash
cap-evolve dashboard --host 0.0.0.0
cap-evolve run --dashboard-host 0.0.0.0     # or dashboard_host: in the spec
export CAPEVOLVE_DASHBOARD_HOST=0.0.0.0
```

**The default is unchanged.** A run that does not opt in behaves exactly as before, and the dashboard stays off-box by default — it serves a project's run artifacts, which should not be published implicitly.

Two details the plumbing has to get right:

- **A wildcard bind is never printed as a URL.** No browser dials `0.0.0.0`, so `browsable_host()` maps `0.0.0.0`/`::` back to loopback for the link a human clicks, while uvicorn still receives the wildcard.
- **`_free_port()` now probes the address it will actually bind.** A `127.0.0.1`-only listener still blocks a later `0.0.0.0` bind of the same port, so probing loopback for a wildcard bind could hand back a port uvicorn then fails to take.

`run` exports the resolved host to the environment rather than threading another flag through, because the report phase re-launches the server in its own subprocess and has to agree with the URL printed at the top of the run.

## Files

| File | What |
|---|---|
| `core/cap_evolve/dashboard_launch.py` | `resolve_host()`, `browsable_host()`, host-aware `_free_port()` and `launch_command()` |
| `dashboard/backend/capevolve_dashboard/server.py` | `--host` flag threaded into `uvicorn --host` |
| `core/cap_evolve/cli.py` | `run --dashboard-host`, `dashboard --host`, env export |
| `core/cap_evolve/branding.py` | help-catalog entry for the new flag |
| `docs/TROUBLESHOOTING.md` | new entry under Dashboard |

## Testing

- **10 new cases** — host precedence (flag > env > default, blank env rejected), wildcard→loopback URL mapping, `--host` present in the spawn argv, env inheritance so the report phase agrees, and `_free_port` probing the host it will bind.
- One existing test monkeypatched `_free_port` with the old signature and needed its lambda widened.
- `core/tests`: 1123 passed, 12 skipped. `dashboard/backend/tests`: 40 passed.
- **Manually verified end-to-end:** launched with `CAPEVOLVE_DASHBOARD_HOST=0.0.0.0` over a real `.capevolve` base; `ss` showed `LISTEN 0.0.0.0:7878`, and `/api/health`, `/`, and `/api/runs` all returned 200 both on loopback and via the host's routable IP, serving real run data.

## Notes for reviewers

- Opening `0.0.0.0` makes the dashboard reachable from anything that can route to the machine. The docs say so explicitly and point at `ssh -L` as the alternative that keeps the loopback-only bind.
- Example addresses in tests use `192.0.2.7` (RFC 5737, documentation-reserved) rather than an RFC 1918 address that could collide with real infrastructure.
- Wording throughout is deliberately platform-neutral — the symptom is "unreachable from another machine", not any one OS or virtualization setup.
- Separately worth flagging: the optional backend not being installed also produces an unreachable dashboard, but it fails as `{"dashboard": "skipped", "reason": "capevolve-dashboard not installed"}` on stderr and is easy to miss. Not changed here; possible follow-up to make that louder.

### DIFF
--- a/core/cap_evolve/branding.py
+++ b/core/cap_evolve/branding.py
@@ -304,9 +304,13 @@ COMMANDS: dict[str, dict] = {
     "dashboard": {
         "group": "inspect",
         "summary": "launch the web dashboard over a base dir of runs",
-        "usage": "cap-evolve dashboard [--base DIR] [--port N] [--no-open]",
-        "long": "Needs the optional dashboard backend: pip install -e dashboard/backend.",
-        "examples": ["cap-evolve dashboard", "cap-evolve dashboard --port 8080 --no-open"],
+        "usage": "cap-evolve dashboard [--base DIR] [--port N] [--host ADDR] [--no-open]",
+        "long": "Needs the optional dashboard backend: pip install -e dashboard/backend.\n"
+                "Binds 127.0.0.1 by default, so the dashboard is reachable only from the\n"
+                "machine running it. Pass --host 0.0.0.0 (or set CAPEVOLVE_DASHBOARD_HOST)\n"
+                "to open it from elsewhere — a remote box, a container, a VM.",
+        "examples": ["cap-evolve dashboard", "cap-evolve dashboard --port 8080 --no-open",
+                     "cap-evolve dashboard --host 0.0.0.0   # reachable off-box"],
     },
     "version": {
         "group": "inspect",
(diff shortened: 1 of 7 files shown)


aviweit:
Closing this PR as it may not be needed. @OsherElhadad , FYI: it works well in my environment - when I port-forward 7878 from my WSL instance.
Thanks

PR #427, opened 2026-09-03
examples/tau2_airline: reach the model through the ETE gateway

aviweit:
## Summary

`examples/tau2_airline` reaches its model through the internal ETE LiteLLM gateway: an OpenAI-compatible endpoint with a fixed base URL and standard bearer auth, so there is no custom header and no patching of the pinned tau2 clone.

Evaluation runs **`aws/gpt-oss-120b`**, normalized once to `openai/aws/gpt-oss-120b` so litellm takes the OpenAI route rather than trying to reach AWS directly and failing auth.

It also fixes where the benchmark is vendored: `setup.sh` / `itest.sh` now clone tau2-bench into **`vendor/tau2-bench`** instead of a sibling of the repo, aligning this example with `examples/skillsbench` (`vendor/skillsbench`). One place for vendored benchmarks, removed by deleting `vendor/` — which is gitignored, so the checkout never appears as untracked noise.

## Changes

| File | What |
|---|---|
| `adapters/gateway.py` | New. Credentials from repo-root `.env` (`OPENAI_BASE_URL` / `OPENAI_API_BASE` + `OPENAI_API_KEY`), resolved LAZILY so `cap-evolve check` stays offline. |
| `adapters/adapter.py` | Imports it for `agent_model` / `user_model` / `llm_args_for`; call sites otherwise unchanged. |
| `capevolve.yaml` | `target_model: aws/gpt-oss-120b`. The two agentopt gptoss specs already named that id. |
| `setup.sh` | Warns per missing gateway variable before the check gate; clones into `vendor/tau2-bench`. |
| `itest.sh` | Clones into `vendor/tau2-bench`. |
| `PROMPT.md`, `DEMO.md` | Name the gateway and its credentials; `PROMPT.md` points intake at `vendor/tau2-bench` (`runner_repo_path: ../../vendor/tau2-bench`). |
| `docs/REPRODUCE_tau2.md` | Clone target follows the scripts — the only file outside `examples/tau2_airline/`. |

Three details the wiring has to get right:

- **`normalize()` is idempotent.** `agent_model()` is called more than once per run and `openai/openai/...` is a 404.
- **No `api_key` ever enters `llm_args`.** tau2 records `llm_args` verbatim into its results file — which is what `trajectories()` exposes, what cap-evolve copies into the optimizer's workdir each iteration, and what `store: git` **commits**. A key placed there becomes a committed secret. litellm reads `OPENAI_API_KEY` from the environment for the `openai/` route, so omitting it costs nothing.
- **The price entry is registered under the id litellm actually queries.** litellm has no entry for a gateway catalog id, and the lookup uses the *unprefixed* name, so both spellings are registered. Gateway spend is not metered here, so a `$0` runner figure means NOT MEASURED rather than free.

The normalization prefixes cover each catalog namespace the gateway serves (`aws/`, `azure/`, `Azure/`, `gcp/`, `GCP/`, `rits/`, `ibm/`, `openai/`); an id outside them is left for litellm to route by its own rules.

## Testing — 2-task smoke, end to end

Ran `examples/tau2_airline/smoke.sh` on the real gateway. Exit 0, all phases: check → baseline → optimize → gate → finalize → sealed test → report.

```json
{
  "run_dir": ".capevolve/run_smoke",
  "best_id": "seed",
  "finalized": true,
  "baseline_val": 0.5,
  "best_val": 0.5,
  "test_reward": 1.0,
  "test_baseline_reward": 1.0,
  "test_delta": 0.0,
  "iterations": 1
}
```

| Check | Result |
|---|---|
| Gateway probe (one real completion) | HTTP 200, content `ready` |
| `cap-evolve check` | `{"ok": true}` — 50 tasks, deterministic scorer |
| Rollouts | **6/6 clean, 0 errored** |
| Baseline val | 0.5 (task 1 scored a **full** 1.0) |
| Optimizer candidate | `cand_0001` scored val 1.0 |
| Gate | Correctly **rejected**: `paired Δ̄=+0.5000 <= 1.0·SE=0.5000 (n=2)` |
| Sealed test | 1.0 |
| Credential / model / transport errors | **0** — no 401, 403, 404, auth, not-found, rate-limit, timeout, or unmapped-model signatures |
| Artifacts | `report.md`, `dashboard.html`, `JOURNAL.md`, `META_INSIGHTS.md` written |

The gate rejection is the system working, not failing: on `n=2` with `SE=0.5`, a `+0.5` gain is exactly at the `1.0·SE` bar and cannot be distinguished from noise.

`core/tests`: 1117 passed, 12 skipped.


aviweit:
## Integration test run — tau2 airline, task 9

Ran the end-to-end optimization pipeline against this branch. Completed cleanly (exit 0), ~51 min wall-clock, run dir `.capevolve/run_itest`.

### Spec

```yaml
# INTEGRATION-TEST spec — taubench (tau2 airline) regression on ONE task (task 9).
# General end-to-end regression for the optimization pipeline on a real benchmark:
# existing seed (policy + tools), agent + user simulator = claude-haiku-4-5 via the IBM
# Anthropic-compatible gateway (set TAU2_AGENT_MODEL / TAU2_USER_MODEL), optimizer =
# claude-code @ claude-sonnet-4-6. Cheap: 1 task, 1 trial, 1 iteration.
capabilities: [system-prompt, tools]
capability_path: seed_capability
actions: [edit]
capability_sources: [seed_capability/reference/data_model.py]

optimizer_skill: claude-code
optimizer_model: claude-sonnet-4-6
optimizer_max_turns: 100
optimizer_usd_per_iter: 50.0

algorithm_skill: hill-climb
algorithm_focus: all

dataset_source: adapter
split_seed: 0
split_train: 0.5
split_val: 0.25
split_test: 0.25
split_ids_file: itest_split.json

target_model: aws/gpt-oss-120b            # the agent+user-sim under optimization (normalized by adapters/gateway.py)

num_trials: 10

gate_mode: paired
gate_k_se: 1.0

max_iterations: 2
stall: 0
max_metric_calls: 0
max_usd: 50.0
max_optimizer_usd: 50.0

store: git
```

Two values in the executed copy (`.capevolve/project/capevolve.itest.yaml`) differed from the checked-in example above: `optimizer_max_turns: 40` and `max_optimizer_usd: 8.0`. Neither was reached during the run.

### Result

| Field | Value |
|---|---|
| Status | Finalized, exit 0 |
| Winner | `cand_0001` |
| Test (optimized) | **0.700 ± 0.153** |
| Test (seed baseline) | 0.500 ± 0.167 |
| **Test delta** | **+0.200** |
| Iterations | 2 / 2 (cap reached) |
| Duration | ~51 min |

### Candidates

| Iter | Candidate | Parent | Parent val | val | Δ | Verdict | Opt cost | Opt tokens |
|---|---|---|---|---|---|---|---|---|
| — | `seed` | — | — | 0.700 | — | baseline | — | — |
| 1 | **`cand_0001`** | seed | 0.700 | **0.900** | +0.200 | accept | $1.337 | 23,496 |
| 2 | `cand_0002` | cand_0001 | 0.900 | 0.600 | −0.300 | reject | $1.572 | 34,654 |

### Evals (task `9`, 10 trials each)

| Split | Tag | Reward | Stderr | pass^1 | pass@2 | Seconds |
|---|---|---|---|---|---|---|
| val | seed | 0.700 | 0.153 | 0.700 | 0.933 | 467 |
| val | cand_0001 | 0.900 | 0.100 | — | — | 377 |
| val | cand_0002 | 0.600 | 0.163 | — | — | 494 |
| test | FINAL (cand_0001) | 0.700 | 0.153 | 0.700 | 0.933 | 385 |
| test | FINAL_seed | 0.500 | 0.167 | 0.500 | 0.778 | 347 |

All 50 rollouts scored with zero errored trials (`errored_trials: 0`, `valid_trials: 10` on every eval) — the gateway path held up end to end.

### Budget

| Limit | Configured (as run) | Spent | Used | Bound? |
|---|---|---|---|---|
| `max_iterations` | 2 | 2 | 100% | **yes — ended the run** |
| `max_optimizer_usd` | $8.00 | $2.909 | 36% | no |
| `max_usd` (total) | $50.00 | $2.909 | 6% | no |
| `max_metric_calls` | off (0) | 50 | — | no |
| `stall` | off (0) | 1 | — | no |
| Runner cost | — | $0.00 | — | not instrumented |
| Runner / optimizer time | — | 2,069 s / 911 s | — | — |

### Caveats on reading the +0.200

This run deliberately targets a single task, so `splits.json` is `train = val = test = ["9"]`. Two consequences:

1. **The test score is a fit metric, not a held-out one.** `report.md` describes the delta as "on held-out tasks the optimizer never saw" — that wording is generated unconditionally and is inaccurate for an overlapping-split run. The run did emit a `splits_warning` for this at startup.
2. **The delta is within trial noise at n=1 task.** Identical configurations re-rolled 10 trials moved by the same 0.200: `seed` scored 0.700 on val and 0.500 on test; `cand_0001` scored 0.900 on val and 0.700 on test. A `gate_warning` fired on both steps for the related reason — paired SE collapsed to 0 at n=1, so the significance gate degraded to a STRICT "accept any Δ>0" fallback.

`cand_0001` did beat `seed` in both evals, which is mild directional evidence, but the magnitude isn't resolvable at this task count. As a pipeline regression test the run is green; it isn't evidence about optimization quality.

examples/tau2_airline/itest.sh
```diff
@@ -42,7 +42,7 @@ fi
 "$PY" -c "import tau2" >/dev/null 2>&1 || die "tau2 import failed after install"
 
 say "2/4  Credentials (IBM Anthropic-compatible gateway)"
-# The adapter (rits.py) and the claude-code optimizer read these from the environment;
+# The adapter (gateway.py) and the claude-code optimizer read these from the environment;
 # we deliberately do NOT write them to a .env file (a self-hosted runner's workspace
 # may persist, so this test never leaves the token on disk). An existing .env is honored.
 if { [ -z "${ANTHROPIC_BASE_URL:-}" ] || [ -z "${ANTHROPIC_AUTH_TOKEN:-}" ]; } && [ ! -f "$REPO/.env" ]; then
```
OsherElhadad: `gateway.py` dropped the old `rits.py`'s claude/anthropic routing path entirely — `llm_args()`/`llm_args_for()` now unconditionally require `OPENAI_BASE_URL`/`OPENAI_API_KEY`, with no branch for `anthropic/*` models. But this script (and `capevolve.agentopt.yaml`) still runs `TAU2_AGENT_MODEL=anthropic/claude-haiku-4-5` and only checks/documents `ANTHROPIC_BASE_URL`+`ANTHROPIC_AUTH_TOKEN` here.

Verified locally: `gateway.normalize("anthropic/claude-haiku-4-5")` returns the string unchanged (it's not in `_GATEWAY_PREFIXES`), and `gateway_credentials()` then raises `RuntimeError: OPENAI_BASE_URL ... and OPENAI_API_KEY not set` even with `ANTHROPIC_BASE_URL`/`ANTHROPIC_AUTH_TOKEN` set — this itest will fail at step 4/4 as written. Either restore anthropic-model routing in `gateway.py` (as `rits.py`'s `_is_anthropic`/`_gateway_args` did) or update this script + `capevolve.agentopt.yaml` to the new credential/model scheme.

examples/tau2_airline/capevolve.agentopt.yaml
```diff
@@ -9,9 +9,8 @@
 #     index stride (val = 0,4,8…; test = 1,5,9…; train = the rest) so the three splits
 #     draw uniformly from the task list — the cheapest available defence against the
 #     split-difficulty mismatch a previous run hit (val 0.75, test 0.25 on 12+8 ids).
-#   * agent + user simulator run on claude-haiku-4-5 through the IBM Anthropic-compatible
-#     gateway (TAU2_AGENT_MODEL / TAU2_USER_MODEL, handled by adapters/rits.py), so no
-#     RITS_API_KEY is required.
+#   * agent + user simulator run on claude-haiku-4-5 through the ETE gateway
+#     (TAU2_AGENT_MODEL / TAU2_USER_MODEL, handled by adapters/gateway.py).
```
OsherElhadad: This comment (and `target_model: claude-haiku-4-5` below) claims the claude path is "handled by adapters/gateway.py", but `gateway.py`'s `llm_args()` unconditionally requires `OPENAI_BASE_URL`/`OPENAI_API_KEY` and has no branch for `anthropic/`-prefixed models (the old `rits.py` had `_is_anthropic`/`_gateway_args` for exactly this). Running this spec as-is will raise `RuntimeError: OPENAI_BASE_URL ... not set` rather than route through the anthropic-compatible gateway. Same issue as flagged on `itest.sh`, which uses this same claude-haiku path.

OsherElhadad (commented):
Reviewed the diff, `gateway.py`'s self-check, and cross-checked against the rest of the repo (CI workflows, `ci/benchmarks/lib/run_suite.sh`, docs). Overall the rename/relocation (RITS → generic ETE gateway, `../tau2-bench` → `vendor/tau2-bench`) is clean and well-motivated (no api_key in `llm_args`, idempotent `normalize()`, lazy credential resolution, zero-cost registration to avoid noisy ERROR logs — all verified against `gateway.py`'s own `__main__` self-check, which passes).

**Real bug — claude/anthropic routing was dropped, but is still referenced as working:**
The deleted `rits.py` had two distinct routes: litellm-proxy catalog ids (`aws/`, `azure/`, `gcp/`, `rits/`, ...) via `_proxy_args()`, and `anthropic/`/`claude-*` models via a separate `_gateway_args()` that read `ANTHROPIC_BASE_URL`/`ANTHROPIC_AUTH_TOKEN` and set an `Authorization: Bearer` header. The new `gateway.py` only implements the first route — every model, including `anthropic/claude-haiku-4-5`, now falls into `llm_args()`, which unconditionally requires `OPENAI_BASE_URL`/`OPENAI_API_KEY` and raises `RuntimeError` if they're absent.

But `examples/tau2_airline/itest.sh` (defaults `TAU2_AGENT_MODEL=anthropic/claude-haiku-4-5`, checks for `ANTHROPIC_BASE_URL`/`ANTHROPIC_AUTH_TOKEN`) and `capevolve.agentopt.yaml` (`target_model: claude-haiku-4-5`, comment says "handled by adapters/gateway.py") both still assume that path works. I reproduced it locally: `gateway.normalize("anthropic/claude-haiku-4-5")` passes the string through unchanged (not in `_GATEWAY_PREFIXES`), and `gateway_credentials()` then raises for missing `OPENAI_BASE_URL`/`OPENAI_API_KEY` even with the Anthropic env vars set. So `itest.sh` will fail at step 4/4 as shipped. Posted inline comments on both files.

I checked whether this breaks CI: it doesn't — `.github/workflows/integration-tests.yml`'s tau2 job runs through `ci/benchmarks/lib/run_suite.sh`, which uses a completely separate adapter template (`ci/benchmarks/templates/tau2_bench/adapter.py` + a `litellm_proxy/...` `.env`), not `examples/tau2_airline/adapters/gateway.py`. So this is scoped to the example's own on-demand `itest.sh`/`agentopt.yaml`, not a CI regression — but it is a real functional break for anyone who runs the documented claude-haiku itest path.

**Minor / non-blocking — stale references outside the changed files:**
- `docs/TROUBLESHOOTING.md:39-42` still documents `RITS_API_KEY`/`RITS_API_URL` for "the tau2 example."
- `examples/skillsbench/PROMPT.md:146` still points to `examples/tau2_airline/adapters/rits.py` as a loader-pattern example — that file no longer exists.

Neither is in the diff, so not blocking, but worth a follow-up cleanup pass.

**Verified:**
- `gateway.py`'s self-check (`python adapters/gateway.py`) passes: idempotent `normalize()`, `api_key` correctly excluded from `llm_args()`, zero-cost price registration lands under the unprefixed id litellm looks up.
- No other stale `rits.py`/`RITS_API_KEY` references inside `examples/tau2_airline/` itself — the rename is otherwise complete there.
- `vendor/tau2-bench` relocation is consistent across `setup.sh`, `itest.sh`, `PROMPT.md`, `docs/REPRODUCE_tau2.md`.

Not approving/requesting changes formally per instructions — just flagging the anthropic-routing regression as something to fix or explicitly scope out before merge.

aviweit:
@OsherElhadad, thanks for testing it. You're
right on both counts, and your second comment is the same root cause.

Reproduced: `RuntimeError: OPENAI_BASE_URL (or OPENAI_API_BASE) and OPENAI_API_KEY
not set`. I missed it because a repo-root `.env` supplied `OPENAI_*` on my box, so
the call returned 200 — the failure only shows on a runner with no `.env`, which is
exactly where `itest.sh` runs.

Took your second option, using `aws/claude-haiku-4-5`:

- `itest.sh` — exports `aws/claude-haiku-4-5`, preconditions on `OPENAI_BASE_URL` +
  `OPENAI_API_KEY`, and notes the claude-code optimizer authenticates separately.
- `capevolve.agentopt.yaml` — `target_model: aws/claude-haiku-4-5`; the "handled by
  adapters/gateway.py" claim replaced with what actually happens.
- `capevolve.itest.yaml` — header names the gateway and the model.

`gateway.py` unchanged: one route, one credential pair.

`aws/` vs `anthropic/` matters more than it looks. `aws/` is in
`_GATEWAY_PREFIXES`, so `normalize()` makes it `openai/aws/claude-haiku-4-5` and
litellm takes the OpenAI-compatible route — where the key's model scope is
enforced. The `anthropic/` spelling skips that: a key scoped to
`['aws/gpt-oss-120b']` is refused on `aws/claude-haiku-4-5` but answers for
`anthropic/claude-haiku-4-5` with a model the response can't identify. It was
passing for a reason we shouldn't rely on.

Consequence: the itest needs a haiku-scoped key. A narrow key surfaces as `key not
allowed to access model`, now documented in TROUBLESHOOTING.md as a scope issue
rather than a broken config.

Verified: 1117 core tests pass, your failure case is fixed, and two end-to-end
smokes through the gateway — `aws/gpt-oss-120b` (6/6 rollouts, 0 errored) and
`aws/claude-haiku-4-5` (3 eval passes) — clean on credentials, model scope and cost
mapping.

Both your options were viable — `ANTHROPIC_BASE_URL` does reach haiku; I chose the
single-route one for simplicity. Fixed in 6eaa327.

@OsherElhadad , please review it again and let me know if you have questions or issues.

Thanks.

OsherElhadad:
Fresh re-review against the current HEAD (`6eaa327b`, 2026-09-04) — checked out into an isolated worktree.

**The anthropic-route bug from the earlier review is fixed**, and verifiably so. To recap: the earlier review found that `itest.sh` / `capevolve.agentopt.yaml` targeted `anthropic/claude-haiku-4-5` with `ANTHROPIC_BASE_URL`/`ANTHROPIC_AUTH_TOKEN` credentials, but the new `gateway.py` only implements the litellm-proxy `OPENAI_BASE_URL`/`OPENAI_API_KEY` route and had no `anthropic/` handling — so that path would 401/misroute at runtime.

Commit `6eaa327b` ("tau2_airline: run claude-haiku through the gateway's own namespace") resolves this — not by adding an anthropic-native route, but by switching the model id itself to `aws/claude-haiku-4-5`, which is already in `gateway.py`'s `_GATEWAY_PREFIXES`, plus updating the credential names/docs in `itest.sh`, `capevolve.itest.yaml`, and `capevolve.agentopt.yaml` to `OPENAI_BASE_URL`/`OPENAI_API_KEY` to match. Verified directly in the worktree:

```
>>> gateway.normalize('aws/claude-haiku-4-5')
'openai/aws/claude-haiku-4-5'
```

and `python3 examples/tau2_airline/adapters/gateway.py` (the module's own self-check) passes cleanly. So both the gpt-oss and claude-haiku targets now route through the same OpenAI-compatible gateway path with consistent, correctly-named credentials — no more anthropic/-prefixed dead end.

One minor, non-blocking nit found in this pass: `docs/REPRODUCE_tau2.md` (touched by this PR only for the `vendor/tau2-bench` path change) still refers to `adapters/rits.py` as "the RITS shim" in two places (§3 and §8), which is now stale — the file was renamed to `gateway.py` and the RITS wiring removed entirely in this PR. Cosmetic only, doesn't affect runtime.

CI is green and the PR shows `MERGEABLE`.

aviweit:
@OsherElhadad , I handled the cosmetic change. Thanks.

OsherElhadad (approved):

PR #454, opened 2026-09-06
tau2: keep tau2's own simulation traces, at one path every adapter agrees on

aviweit:
## What this fixes

`examples/tau2_airline` threw away tau2's own trajectories — both `run_tasks` call sites
passed `save_path=None`, so the traces existed only in memory and were dropped after scoring.
Users got cap-evolve's rollout JSON, which `tau2 view` cannot read, while tau2 closed every
run by recommending exactly that command.

The adapters that *did* save them disagreed on where, so `tau2 view --dir` took a different
path shape depending on which arm produced the run. All adapters now write one format:

```
<run_dir>/native_sims/<tag>/<split>/results_<YYYYmmdd_HHMMSS>_<pid>.json
```

The timestamp is not decoration. tau2 treats an existing results file as a run to **resume** —
it prompts on stdin and raises `FileExistsError` otherwise, and an eval has no stdin. Since the
same tag+split really is written twice (baseline and finalize), a fixed filename would hang or
kill the run. A unique stamp removes that, and let the old `-2`/`-3` suffix walk be deleted.

## Also fixes: candidate evals wrote nothing

The guard only accepted `candidates/<tag>`, but that's just the baseline and the finalize —
**every candidate eval gets `work/<cid>`**. So the seed and the winner saved traces and no
candidate did, which is backwards: candidate traces are how you learn why an edit scored what
it did. The run dir is `parent.parent` under either name, so the fix is one line:

```python
if cand.parent.name not in ("candidates", "work"):
    return None
```

## Changed

| file | what |
|---|---|
| `examples/tau2_airline/adapters/adapter.py` | the save-path logic, both call sites wired |
| `templates/adapters/tau2_bench/adapter.py` | same, byte-identical |
| `core/tests/test_tau2_native_sims.py` | 25 tests, incl. a guard that fails if adapters drift apart |
| `ci/benchmarks/lib/run_suite.sh` | turns saving off in CI |
| `examples/tau2_airline/PROMPT.md` | documented a path nothing actually wrote |

Only 6 lines are deleted on the whole branch: the 4 `save_path=None` literals and 2 stale doc
lines. The adapters duplicate this code on purpose — each ships as one self-contained file
copied into a project — so a test holds them identical instead of a shared import.

## On by default, off in CI

`CAPEVOLVE_NATIVE_SIMS=0` disables saving; default is on, because a flag you must set to get a
feature is a flag nobody sets. CI sets `0`: the files reach no artifact upload there, and a full
tier writes ~500 sims per eval (~74 KB each) — this runner has lost a pilot to a full disk.

Saving is otherwise inert. Verified on live runs: identical reward, identical token count, and
identical rollout contents with the flag on and off. The optimizer reads cap-evolve's own
per-tag rollouts, never these files.

## Testing

Unit: full `core/tests` suite **1210 passed, 12 skipped**.

Live, from scratch — `.venv`, `.capevolve/project` and `vendor/` deleted, `setup.sh` re-run
(fresh venv, tau2-bench clone, `cap-evolve check` ok), then 2 tasks · 1 trial · 1 iteration:

| path | result |
|---|---|
| `cap-evolve run --spec …` | **pass** — 3 trace files, run finalized |
| `bash examples/tau2_airline/run.sh` | **pass** — 3 trace files, run finalized |
| CI path, saving off | **pass** — 0 files, no directory created, run finalized |

Both saving-on runs produced the same three files — baseline, **candidate** (the `work/` fix,
absent before this change), and finalize writing a *second* file into the same directory with no
resume prompt and no suffix walk. The adapter also prints `tau2 view --dir <path>` right after
tau2's own bare recommendation, so the traces are actually findable.


aviweit:
@OsherElhadad , can you please review ? Thanks.

OsherElhadad (approved):

PR #464, opened 2026-09-08
feat(harbor): HARBOR_LOCAL_ASIS, guarded against silent under-delivery

bcarmeli:
## Summary

Splits the harbor-adapter half out of #320, which bundled a shared-template change
together with parsec-specific benchmark wiring. This PR carries only the shared
`templates/adapters/harbor/adapter.py` change plus a new test; the parsec-specific
half is PR #<parsec-bench-local, to follow>.

- Adds `HARBOR_LOCAL_ASIS`, an opt-in env flag that skips `package_dataset` and
  delivers the candidate to the harbor container purely via
  `--extra-instruction-path` (i.e. as raw instruction text), for benches whose
  harness only understands a system-prompt/instruction-text capability.
- Adds a guard, `_assert_asis_can_deliver`, called from the ASIS branch before
  `harbor_run`. Because ASIS skips `package_dataset` entirely, and the template's
  own declared default capability is `[skill-package]`, a skill-package candidate
  run under `HARBOR_LOCAL_ASIS=1` would otherwise silently drop everything except
  the instruction-file text (`scripts/`, `references/`, ...) — the run completes,
  the gate runs, numbers get published, and the capability mismatch is invisible.
  The guard raises `ValueError` (matching this adapter's existing convention) naming
  the env var, the dropped paths, and the fix (drop ASIS, or reduce the candidate's
  capability to instruction text only).
- Adds `core/tests/test_harbor_local_asis.py` — the first test for this adapter
  template. Covers: ASIS + instruction-only candidate (no raise), ASIS + candidate
  with extra files (raises, message names the env var and the dropped path),
  non-ASIS + extra files (no raise, repack path delivers everything), and a static
  check that the guard call precedes `harbor_run` in `run_batch`'s source.

## Test plan

- [x] `python -m pytest core/tests/test_harbor_local_asis.py -q` passes
- [x] `python -m pytest core/tests -q` (full suite) passes
- [x] Confirmed the new test fails against the unguarded adapter (reverted the guard
      locally and re-ran), proving it demonstrably covers the bug
- [ ] Reviewer: confirm `ValueError` (not `RuntimeError`) matches this adapter's
      existing convention at the call sites you'd expect

### DIFF
--- a/core/tests/test_harbor_local_asis.py
+++ b/core/tests/test_harbor_local_asis.py
@@ -1,0 +1,133 @@
+"""HARBOR_LOCAL_ASIS must not silently under-deliver a skill-package candidate.
+
+``HARBOR_LOCAL_ASIS=1`` hands a hand-authored local dataset to ``harbor run`` verbatim so
+the default ``package_dataset`` repack cannot overwrite a task's real task.toml / tests/
+verifier / Dockerfile with generic templates. The cost of skipping that repack is that the
+candidate no longer travels in each task's ``environment/capability/`` dir: the ONLY channel
+left is ``--extra-instruction-path``, and ``_write_candidate_instruction`` puts nothing on it
+but the text of ``prompt.md``/``SKILL.md``.
+
+The harbor template's own declared default is ``capabilities: [skill-package]``
+(templates/adapters/harbor/capevolve.yaml), so the two features compose into a silent bug: a
+skill-package candidate under ASIS loses its bundled ``scripts/``/``references/`` while the
+run completes, the gate runs and the numbers get published — every edit outside the SKILL.md
+body was never delivered, and nothing says so. ``_assert_asis_can_deliver`` turns that into a
+loud failure before any container starts.
+"""
+
+import importlib.util
+import os
+import sys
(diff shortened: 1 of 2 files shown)


bcarmeli:
Correction to the test plan above: \`python -m pytest core/tests -q\` on this branch has 2 failures, not 0 — \`test_cli_surface.py::test_spec_defaults_relative_to_project_not_the_cwd\` and \`::test_spec_outside_the_project_is_refused_naming_both_paths\`, both \`FileNotFoundError: no manifest — run install.sh or skills/_registry/build_manifest.py first\`.

This is a local environment setup gap (missing generated skills manifest under \`~/.claude/skills/_registry/\`), unrelated to this PR's change — it reproduces identically on an unrelated branch (#465) that never touches \`templates/adapters/harbor/adapter.py\` or anything this PR touches. The harbor-specific test (\`test_harbor_local_asis.py\`, 6/6) and the rest of the suite (1258 passed, 2 skipped) are unaffected. Correcting "full suite passes" since that wasn't literally true.

OsherElhadad (approved):

PR #465, opened 2026-09-08
feat(parsec): add parsec as a local-only benchmark, plus a v2 tier

bcarmeli:
## Summary

Splits the parsec-specific half out of #320, which bundled parsec's benchmark
wiring together with a shared harbor-adapter change (that half is PR #464). This
PR is parsec's local-only integration under `ci/benchmarks/parsec/`, addressing
the review feedback on #320:

- **Removes the contamination**: deletes `pilot/split_ids.json`. `run_suite.sh`
  treats the presence of a tier's `split_ids.json` as opt-in to a genuine
  held-out split; parsec's file put `{0005, 0032, 0047, 0048, 0088, 0104}` in
  `test`, but the published N=30 pilot run optimized directly against
  `0047`/`0048` — i.e. against its own claimed held-out set. Deleting the file
  (rather than regenerating it) makes the pilot an honest no-holdout FIT tier,
  matching what `ci/benchmarks/parsec/README.md` already claims it is.
- **Removes the redundant override file**: deletes `pilot/overrides.env` — every
  key it set (`HARBOR_LOCAL_ASIS=1`, `BACKEND_MCP_URL=...`) is already exported
  unconditionally by `run_suite.sh`'s parsec case block with identical values.
- **Fixes two broken defaults** in `utils/patch-harbor-tasks.sh`: `SRC` no longer
  defaults to a personal laptop path (now unset-by-default, fails with a message
  naming `PARSEC_HARBOR_TASKS_SRC`); `DST` now matches what `run_suite.sh`
  actually reads (`$REPO/e2e/parsec/harbor-tasks-patched`) instead of a path
  under `ci/benchmarks/parsec/` that nothing consumes.
- **Adds a v2 tier** — a second, independently-authored parsec dataset (10
  hand-written `bench-aap2-*` tasks, one isolated simulator per task, vs. v1's 4
  shared simulator endpoints across 30 real-trace tasks). Wires a `v2` branch
  into `run_suite.sh`'s parsec case (`HARBOR_DATASET` defaults to
  `harbor-tasks-v2.1`), and promotes that tier's task-patching script
  (`utils/patch-harbor-tasks-v2.1.sh`) the same way v1's patcher is promoted.
- **Promotes two previously-gitignored files** that the README's own commands
  depended on but that shipped nowhere: `docker/parsec-agent-base/Dockerfile` and
  `utils/build-parsec-agent-base.sh` (updated to the new in-repo paths).
- **Rewrites `ci/benchmarks/parsec/README.md`**: fixes the `BENCH`-as-env-var
  error (it's positional — commands are now
  `TIER=<tier> bash ci/benchmarks/lib/run_suite.sh parsec`), adds a prominent
  internal-only section (the dataset and the four kaegis simulators are
  IBM/RH-internal — this benchmark cannot be run externally), states plainly
  that parsec is not CI-dispatchable and why, and explains the pilot is now a
  FIT tier with no committed split.
- **Adds `parsec` to `ci/benchmarks/lib/run_suite.sh`'s usage string** and to
  `ci/benchmarks/README.md`'s inventory (flagged local-only) — but deliberately
  **not** to the per-bench PR-label enumeration in that same file, since no label
  dispatches parsec.
- **Updates the one guard test this touches**:
  `test_which_benches_ship_a_pilot_tier` now includes `parsec` in the benches
  that ship a `pilot/tasks.json`, with an inline rationale: parsec ships a pilot
  tier so it's runnable locally, but stays out of `benchmarks.yml`'s `BENCHES`,
  so the `benchmark-pilot` label / `tier=pilot` fan-out this test otherwise pins
  is unaffected.

### Deliberately not wired (by design, not oversight)

`benchmarks.yml`'s `BENCHES`/dispatch `options`, `ALL_BENCHES` in
`test_benchmarks_plan_legs.py`, `BENCHES` in `test_site_live_panel_tiers.py`,
`JOB_RE` in `site/benchmarks.js`, and the case in `ci/benchmarks/lib/ci_setup.sh`
all describe *CI-dispatchable* benchmarks. parsec's dataset and simulators are
internal-only, so none of these should include it — adding it there would make
`bench=all` CI sweeps hard-exit or make the live "Running now" panel expect a
job that will never appear.

Results and recipes from parsec's actual runs (v1's 30-task pilot, v2's 10-task
run) live on the `parsec-history` branch, not in this PR.

## Test plan

- [x] `python -m pytest core/tests -q` — full suite passes on this branch
- [x] `bash -n` on `run_suite.sh`, `patch-harbor-tasks.sh`,
      `patch-harbor-tasks-v2.1.sh`, `build-parsec-agent-base.sh`
- [x] `patch-harbor-tasks.sh`'s `DST` default matches `run_suite.sh`'s
      `HARBOR_DATASET` default for the `pilot`/`smoke` tiers (grepped both)
- [x] Confirmed this branch does **not** touch
      `templates/adapters/harbor/adapter.py` (`git log --stat` / `git diff
      main...HEAD --stat`)
- [ ] Reviewer: this cannot be run end-to-end outside the IBM/RH network (needs
      the internal dataset + kaegis simulators) — README commands were checked
      far enough to confirm they parse and reach the dataset check, not a full run

### DIFF
--- a/ci/benchmarks/README.md
+++ b/ci/benchmarks/README.md
@@ -27,6 +27,18 @@ not a leaderboard.
 - Results are uploaded as an artifact and posted as a sticky PR comment (metrics table +
   optimized-capability diff).
 
+> **Also in this tree, but not part of the CI suite: [`parsec`](parsec/README.md)**
+> (**local-only / internal-only**). Red Hat's LLM-agentic troubleshooting tool; tiers
+> `smoke` (5) · `pilot` (30 v1 real-trace tasks) · `v2` (10 authored tasks, one isolated
+> simulator each). `run_suite.sh` accepts it as a `<bench>`, so it runs the same code path
+> locally, but it is deliberately absent from `benchmarks.yml`'s `BENCHES` and from every
+> other dispatch list: neither its task trees (internal RH, not in the public
+> `rhpds/parsec`) nor its kaegis simulators (`github.ibm.com/kaegis/simulation-harness`)
+> exist outside IBM/RH, so a dispatched leg could only ever fail for infrastructure
+> reasons. Nothing but per-tier `tasks.json` metadata is committed — the datasets are
+> regenerated locally by `parsec/utils/`. See its README for the setup pipelines and the
+> reasoning.
+
 ## Layout
 
 ```
--- a/ci/benchmarks/lib/run_suite.sh
+++ b/ci/benchmarks/lib/run_suite.sh
@@ -14,7 +14,7 @@
 set -uo pipefail
 LIB_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 REPO="$(cd "$LIB_DIR/../../.." && pwd)"
-BENCH="${1:?bench (tau2|swebench|skillsbench|spreadsheetbench)}"
+BENCH="${1:?bench (tau2|swebench|skillsbench|spreadsheetbench|parsec)}"
 PY="${CAPEVOLVE_PY:-$REPO/.venv-e2e/bin/python}"; [ -x "$PY" ] || PY="python3"
 TIER="${TIER:-smoke}"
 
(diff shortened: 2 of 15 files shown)


bcarmeli:
Correction to the test plan above: re-running \`python -m pytest core/tests -q\` just now surfaced 4 failures, not 0. Tracked both down before leaving this as-is:

- \`test_cli_surface.py::test_spec_defaults_relative_to_project_not_the_cwd\` and \`::test_spec_outside_the_project_is_refused_naming_both_paths\` — both fail with \`FileNotFoundError: no manifest — run install.sh or skills/_registry/build_manifest.py first\`. Reproduces identically on an unrelated branch (PR #464) with no parsec changes at all, so this is a local environment setup gap (missing generated manifest), not something this PR causes.
- \`test_eventstream.py::test_dashboard_stream_reexports_the_shared_helper\` and \`::test_follower_thread_reports_instead_of_dying_silently\` — both pass cleanly (38/38) when run in isolation; they only fail as part of the full 1256-test run, consistent with a thread-join timeout (\`t.join(timeout=5.0)\`) being flaky under load rather than a deterministic regression.

Net: no evidence any of the 4 are caused by this PR's changes, but "full suite passes" was not the right claim to make without qualification — correcting it here.

OsherElhadad (approved):

PR #485, opened 2026-09-09
De-duplicate the tau2-airline scoring/feedback stack (-565 lines across the two arms)

### LINKED ISSUE #479: Share or drift-guard the ~280 duplicated scoring/feedback lines across the two tau2 airline arms
Follow-up to @OsherElhadad's review of #424 — ["Duplication: `direct/adapters/adapter.py` and
`spa/adapters/adapter.py` share ~300 lines essentially
verbatim"](https://github.com/skillberry-ai/cap-evolve/pull/424#pullrequestreview-5110283384)

## The measurement

`examples/skillberry_benchmarks_tau2_airline/{direct,spa}/adapters/adapter.py` are 707 and 652
lines, with **572 identical lines**. Sixteen functions are byte-identical, totalling **341 lines**:

| lines | function |
|---|---|
| 64 | `_build_feedback` |
| 49 | `_sim_to_rollout` |
| 40 | `_sim_save_path` |
| 33 | `_derive_total_cost` |
| 28 | `score` |
| 22 | `_split_of` |
| 21 | `_localize_action` |
| 17 | `_localize_communicate` |
| 15 | `trajectories` |
| 11 | `_leaked_stop_continuation` |
| 11 | `_iter_agent_tool_calls` |
| 10 | `run_batch` |
| ≤6 each | `__init__`, `run_target`, `_tau2_tasks_by_id`, `_errored` |

Six functions genuinely differ per arm and should stay per-arm: `apply`, `run_trials`, `tasks`,
`runner_model`, `_shown_metrics`, `_user_profile_facts`.

## Why it is worth acting on

The concern is drift rather than aesthetics: a fix to the `_localize_communicate` heuristic or the
`_STOP_LEAK_RE` regex has to be applied twice, and nothing fails if someone updates only one copy.
The gold-safe feedback stack is the part most likely to be refined over time, and also the part
where a silent divergence between the arms would quietly make the two sets of numbers
non-comparable.

## Two thirds of this already has a pattern

`_sim_save_path` (40 lines) and `_split_of` (22) are duplicated **deliberately**, and their
alignment is enforced by `core/tests/test_tau2_native_sims.py`, which extracts that region from
every tau2 adapter, SHAs it, and fails naming the divergent files. That convention exists because
each adapter ships as one self-contained file copied into a project's `adapters/`, so a shared
import would break the copy-and-run property the examples aim for.

So the codebase already has an answer for "duplicated on purpose"; it is just not applied to the
remaining ~279 lines.

## Options

1. **Extend the drift guard** to cover the scoring/feedback stack — keeps the self-contained-file
   property and adds a test that fails loudly on divergence. Cheapest, and consistent with existing
   precedent.
2. **Shared module** under `examples/skillberry_benchmarks_tau2_airline/`, imported by both arms —
   removes the duplication outright, at the cost of single-file portability for these two arms
   specifically. They already share a parent directory, so this is less costly here than it would
   be for `templates/adapters/`.
3. **Both** — a shared module for the arm-local feedback stack, and the drift guard for the region
   that must stay inline across all four tau2 adapters.

Worth deciding which convention applies where, since #424's own stated goal is byte-identical
portability: option 1 honours that literally, option 2 trades it for a single source of truth
within one example.

aviweit:
Closes #479. Raised in @OsherElhadad's review of #424.

## The problem

`examples/skillberry_benchmarks_tau2_airline/{direct,spa}/adapters/adapter.py` were 707 and
652 lines with **572 identical lines** between them. Sixteen functions were byte-identical,
totalling **341 lines** — the whole scoring and gold-safe feedback stack:

| lines | function | | lines | function |
|---|---|---|---|---|
| 64 | `_build_feedback` | | 20 | `_localize_action` |
| 49 | `_sim_to_rollout` | | 16 | `_localize_communicate` |
| 32 | `_derive_total_cost` | | 14 | `trajectories` |
| 26 | `score` | | 10 | `_iter_agent_tool_calls` |
| 16 | `_shown_metrics` | | 9 | `_leaked_stop_continuation` |

A fix to the `_STOP_LEAK_RE` regex or to a `_localize_*` heuristic had to be applied twice,
and nothing failed if only one copy was updated. The feedback stack is the part most likely
to be refined over time, and a silent divergence between the arms would quietly make their
two sets of numbers non-comparable.

## What this changes

A new sibling module, `examples/skillberry_benchmarks_tau2_airline/scoring.py`, exposes
`Tau2ScoringMixin`; both arms' `Adapter` mix it in. It deploys exactly like `gateway.py` —
each `setup.sh` copies it into `$PROJECT/adapters/`, and `adapter.py` imports it after
putting its own directory on `sys.path`. These arms already shipped two files rather than
one, so a third sibling is the existing convention, not a departure from it.

    scoring.py                                   | +328 (new)
    direct/adapters/adapter.py                   | -281
    spa/adapters/adapter.py                      | -284
    direct/setup.sh, spa/setup.sh                | +5 / -1 each
    core/tests/test_tau2_arms_shared_scoring.py  | +314 (new, 13 tests)

**Net −565 lines across the two adapters.** The moved code was extracted programmatically
rather than retyped, so it is byte-identical to what shipped.

### Two things deliberately stay per-arm

* **`_user_profile_facts`** — genuinely different logic in each arm. The mixin calls it as
  `cls._user_profile_facts(...)`, so normal MRO picks up each override.
* **`DOMAIN`** — `"airline"` vs `"airline_skillberry"`. `_sim_to_rollout` is a `classmethod`
  reading `cls.DOMAIN`, and each `Adapter` binds it as a class attribute.

### And three things stay inline in every adapter

`_native_sims_enabled`, `_split_of` and `_sim_save_path` are **not** moved.
`core/tests/test_tau2_native_sims.py` extracts that region from each `adapter.py`
*textually* and asserts one digest across all four tau2 adapters, and the other two
(`examples/tau2_airline`, `templates/adapters/tau2_bench`) ship standalone and cannot import
this module. Moving them would break that guard. That test still passes unchanged.

## Verification

| gate | result |
|---|---|
| golden `score()` output, before vs after, both arms, 8 branches | byte-identical |
| the two arms score identically to each other | pass |
| `test_tau2_native_sims.py` drift guard, all 4 adapters | pass |
| full core suite | **1251 passed, 12 skipped** |
| `ruff --select F821` on all three changed Python files | clean |
| `cap-evolve check`, both arms, from a deleted `.venv` + re-cloned `vendor/` | `"ok": true`, `problems: []` |
| a test asserting `scoring.py` references no name that only `adapter.py` defines | pass |

## Live end-to-end runs

Both arms, task 9, 10 trials, 2 optimizer iterations, runner agent and user simulator
`aws/gpt-oss-120b` (the spa arm's agent is the `ibm/skillberry-local` sentinel, routed
through SPA to `openai/aws/gpt-oss-120b`).

| | **direct** | **spa** |
|---|---|---|
| baseline (val) | 0.500 ± 0.167 | 0.600 ± 0.163 |
| `cand_0001` | **accept** → 0.600 val, Δ +0.100 | **accept** → 0.800 val, Δ +0.200 |
| `cand_0002` | **accept** → 0.800 val, Δ +0.200 | **reject** → 0.500 val, Δ −0.300 |
| best | `cand_0002` (0.800) | `cand_0001` (0.800) |
| sealed test | **0.900** vs 0.500 baseline → **Δ +0.400** | **0.700** vs 0.600 baseline → **Δ +0.100** |
| optimizer spend | $11.16 / 113,648 tok | $8.86 / 60,457 tok |
| runner · optimizer time | 996 s · 1,895 s | 5,250 s · 1,601 s |

Both arms ran two full optimizer iterations with every rollout scored through the shared
`scoring.py` — accepts *and* a correctly-gated reject — which is the property these runs
exist to demonstrate: the shared module serves usable, localised feedback in production,
not merely importable code.

### Run acceptance checks

| check | direct | spa |
|---|---|---|
| only tools optimized | ✅ both candidates touched only `tools/tools.py` | ✅ both touched only `my_skill/scripts/*.py` (5 and 8 files) |
| benchmark `policy.md` unmodified | ✅ `vendor/skillberry-benchmarks` clean — read as context, never written | n/a |
| `my_skill/SKILL.md` still empty | n/a | ✅ 0 bytes in seed and both candidates |
| `primitive_tools/functions.py` unchanged | n/a | ✅ byte-identical across seed and both candidates |
| store serving | n/a | ✅ `my_skill` approved, 14 tools, re-imported during the run |
| proxy-agent serving | n/a | ✅ every `ChatRequest` carried `model: ibm/skillberry-local` |

## Run health

Both runs completed cleanly, with no infrastructure errors and no rollout excluded as
uncontrollable noise.

* **Unit tests pass.** The full core suite is green — 1251 passed, 12 skipped — including the
  13 new tests in `test_tau2_arms_shared_scoring.py` and the unchanged four-adapter drift
  guard in `test_tau2_native_sims.py`.
* **`cap-evolve check` passes on both arms** against the final code: `"ok": true`,
  `"stubs": []`, `"problems": []`, with `tasks('val') -> 50 task(s)`, a deterministic scorer
  probe and a callable `materialize()`.
* **The spa arm operated end to end.** The store was refreshed with each candidate skill —
  `my_skill` re-imported per evaluation and left `approved` with its 14 tools — the Skillberry
  proxy-agent served every agent turn, and every `ChatRequest` carried the
  `ibm/skillberry-local` sentinel, so each call was injected with the candidate rather than
  reaching the gateway unmediated.
* **The direct arm operated end to end.** Candidate tool code was loaded in-process for each
  evaluation, with the pinned benchmark checkout used as read-only context throughout and left
  unmodified.
* **Native trajectories persisted on both arms** — five
  `native_sims/<tag>/<split>/results_<ts>_<pid>.json` files each, covering the seed, both
  candidates and the finalize, so either run is inspectable with `tau2 view`.

### DIFF
--- a/core/tests/test_tau2_arms_shared_scoring.py
+++ b/core/tests/test_tau2_arms_shared_scoring.py
@@ -1,0 +1,314 @@
+"""The two tau2-airline arms share ONE scoring/feedback implementation (#479).
+
+`direct/adapters/adapter.py` and `spa/adapters/adapter.py` used to carry ~250 lines of
+scoring and gold-safe feedback byte-identically. A fix to the STOP-leak regex or to a
+`_localize_*` heuristic had to be applied twice and drifted silently if only one copy was
+touched. Both now mix in `Tau2ScoringMixin` from `examples/.../scoring.py`.
+
+This test pins the three properties that make that refactor safe:
+
+1. both arms really inherit the shared mixin, and no longer define the moved members;
+2. `score()` produces IDENTICAL output on both arms across the interesting branches —
+   the check that catches a hook accidentally binding to the shared copy instead of the
+   arm's override;
+3. the shared module contains no hard-coded ``Adapter.`` reference. That was a real bug
+   found while moving the code: `_derive_total_cost` was a ``@staticmethod`` calling
+   ``Adapter._iter_agent_tool_calls``, which resolved inside `adapter.py` and raised
+   ``NameError`` once moved — swallowed by `_build_feedback`'s ``except Exception`` and
+   silently degrading the feedback to its generic fallback.
+
+`_user_profile_facts` deliberately stays per-arm, and `_native_sims_enabled` /
+`_split_of` / `_sim_save_path` deliberately stay inline in every tau2 adapter — see
+`test_tau2_native_sims.py`, which extracts that region textually from each `adapter.py`.
+"""
+
+import ast
+import importlib
+import importlib.util
+import sys
+import types
+from pathlib import Path
+
+import pytest
+
+REPO = Path(__file__).resolve().parents[2]
+ARMS = REPO / "examples" / "skillberry_benchmarks_tau2_airline"
+SHARED = ARMS / "scoring.py"
+ARM_NAMES = ("direct", "spa")
+
+MOVED = ("score", "_sim_to_rollout", "_build_feedback", "_derive_total_cost",
+         "_localize_action", "_localize_communicate", "_iter_agent_tool_calls",
+         "trajectories")
+
+
+@pytest.fixture(autouse=True)
+def _isolated_modules():
+    """Restore every module name this test binds.
+
+    It stubs ``gateway`` and imports ``scoring`` plus two same-named adapters. Leaving any
+    of those in ``sys.modules`` would hand a later test the wrong module — the trap
+    `test_tau2_native_sims.py` documents for the shared ``tau2`` stubs.
+    """
+    names = ("gateway", "scoring", "tau2", "tau2.data_model",
+             "tau2.data_model.simulation")
+    before = {k: sys.modules.get(k) for k in names}
+    paths = list(sys.path)
+    try:
+        yield
+    finally:
+        for k in [k for k in sys.modules if k.startswith("_arm_adapter_")]:
+            del sys.modules[k]
+        for k, v in before.items():
+            if v is None:
+                sys.modules.pop(k, None)
+            else:
+                sys.modules[k] = v
+        sys.path[:] = paths
+
+
+def _load_arm(arm: str):
+    """Import an arm's adapter with ``gateway`` stubbed; no network, no credentials."""
+    adapters = ARMS / arm / "adapters"
+    # ARMS itself carries scoring.py in the repo; setup.sh copies it beside adapter.py in a
+    # deployed project, where the adapter's own directory already covers the import.
+    for p in (REPO / "core", adapters, ARMS):
+        if str(p) not in sys.path:
+            sys.path.insert(0, str(p))
+    gw = types.ModuleType("gateway")
+    gw.agent_model = gw.user_model = lambda: "aws/gpt-oss-120b"
+    for name in ("llm_args", "agent_llm_args", "upstream_llm_args"):
+        setattr(gw, name, lambda *a, **k: {})
+    gw.llm_args_for = lambda m: {}
+    gw.load_env = gw.register_zero_cost = lambda *a, **k: None
+    gw.gateway_credentials = lambda *a, **k: ("http://stub", "stub")
+    sys.modules["gateway"] = gw
+    # _sim_to_rollout imports tau2 lazily for TerminationReason; CI has no tau2 install.
+    for name, mod in (("tau2", types.ModuleType("tau2")),
+                      ("tau2.data_model", types.ModuleType("tau2.data_model")),
+                      ("tau2.data_model.simulation", types.ModuleType("tau2.data_model.simulation"))):
+        sys.modules.setdefault(name, mod)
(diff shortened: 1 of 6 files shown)


aviweit:
@OsherElhadad , I've been having second thoughts about this one, and I'd value your take.

One small thing I noticed: _shown_metrics was listed in #479 as one to keep per-arm, and the mixin ended up with SPA's docstring — so the direct arm now reads as though SPA reports no token usage, which doesn't quite fit it there.

The bigger thing is that #479 closed on "worth deciding which convention applies where," and I'm not sure we ever really made that call.

`setup.sh` presents itself as the transcript of what an agent following `PROMPT.md` does, and `PROMPT.md` treats each arm as its own onboarding — so a cross-arm module isn't something that path would produce.  I'm not certain sharing between these two closes the drift risk on its own.

Let's set this aside for now and settle the convention first. We can discuss that offline. For now I will move this PR to draft.

PR #489, opened 2026-09-14
spa: clear a skill's dependents before deleting it

aviweit:
# spa: clear a skill's dependents before deleting it

**Base:** `main` ← **Head:** `fix/spa-skill-delete-prereqs`

## Problem

`reset_store_to_skill` failed with

```
SPA deploy failed: RuntimeError: could not remove existing skill my_skill from the store
```

and took **every rollout** with it — 50 tasks × 10 trials, `coverage 0/50`. cap-evolve then reported
`baseline val=0.0`, `headroom=1.0`, `verdict "floor"` and carried on, so the run looked like a
legitimate zero rather than a broken one.

Reproducible: interrupt `run.sh` (Ctrl-C), then invoke it again without stopping the services. The
only known recovery was purging all data from the store by hand.

## Root cause

The store is an external service, so the code below is from **`skillberry-store` at tag `0.2.1`** —
the ref `spa_env.STORE_REF` pins — read from a clone checked out at that tag (`c5932cd`), not from
this repo. There, `skills_service.delete` refuses to delete a skill while anything depends on it:

```python
dependents = self.handler.dependency_manager.get_dependents(uuid)
if dependents:
    raise ObjectInUseError("skill", uuid, dependents)      # -> HTTP 409 in skills_api.py
```

Exactly two kinds of object register that dependency, both via
`get_service("skill").add_dependent(<kind>, <own uuid>, [skill_uuid])`:

| kind | registered at (skillberry-store @ 0.2.1) |
|---|---|
| vMCP server | `services/vmcp_service.py:191` |
| vNFS server | `services/vnfs_service.py:149` |

Established by enumerating **every** `add_dependent` call at that tag, not by sampling: the others
are skill→tool/snippet (`skills_service.py:223,226,564,567`) and tool→tool
(`tools_service.py:326,633`), and neither makes a *skill* undeletable.

Those objects live in the **store**, not in the run. An interrupted run leaves them behind in a
store that keeps running (the services are detached, so Ctrl-C never reaches them), and `spa_env`
had no concept of them at all — `grep vmcp` over the file returned nothing. Purging the store by
hand worked because it removed the dependents; this change does that part deliberately.

## The fix

`delete_skill`'s documented order gains one step:

```
1. GET the manifest to collect tool_uuids / snippet_uuids
2. DELETE anything that DEPENDS on the skill (vMCP / vNFS servers)   <-- new
3. DELETE the skill (no cascade)
4. DELETE each tool and snippet, now that nothing depends on them
```

A new `delete_skill_dependents(skill_uuid)` lists both kinds and deletes the rows whose `skill_uuid`
matches. It is safe to call unconditionally, which is what makes it correct on every entry path:

- **no recursion** — neither `vmcp_service.delete` nor `vnfs_service.delete` consults
  `get_dependents`, so a dependent has no prerequisites of its own;
- **the registration goes with the object** — both call `remove_dependent(...)` on their way out
  (`vmcp_service.py:629`, `vnfs_service.py:536`), so deleting the row cannot leave an entry behind
  that still blocks the skill;
- **it does not over-reach** — only rows matching this skill's UUID are touched, so a vMCP server
  belonging to a different skill is left alone; it cannot be what blocks this skill, and deleting it
  would break the other one.

It lives in `delete_skill` rather than in `reset_store_to_skill` because every import in this arm
goes through a delete first, so this covers the import path automatically as well as anyone calling
`delete_skill` directly.

## Also: stop discarding the reason

`delete_skill` returned a bare `False`, which `reset_store_to_skill` reported as
`could not remove existing skill <name>` — no status code, no named dependent. There are four
distinct ways that `False` could arise, and that message distinguishes none of them; it is why this
took two days to diagnose.

It now raises with the store's own 409 body — which names the dependents — and says to add the kind
to `_SKILL_DEPENDENT_KINDS`. Any other unexpected status is raised with its body too.

## Also: tool→tool dependencies could fail a delete on ordering

Found while auditing the store's delete path at 0.2.1 against `delete_skill`, and not previously
handled. **Tools carry the same guard as skills** — `tools_service.delete` raises `ObjectInUseError`
while a tool has dependents (`tools_service.py:672`) — and a tool **auto-registers a dependency on
every tool it calls by bare name** (`tools_service.py:326`).

`delete_skill` deleted a skill's tools in manifest order, so if one wrapper called another and the
callee came first, that delete returned 409 and the run failed on a store-ordering detail. It does
not bite today only because the wrappers call the *frozen primitives*, which are protected and never
deleted — but nothing stops a candidate from writing one tool that calls another by name.

The deletes now **retry until a pass frees nothing**, which resolves any dependency order without a
topological sort, and the no-progress exit keeps a genuine cycle from looping — it reports every
object that stayed stuck instead.

`purge_orphans` had the identical single-pass bug, and it runs on the same class of objects **one line
after `delete_skill`** in `reset_store_to_skill` — fixing one and not the other would only have moved
the failure. Both now share a single `_delete_all` helper, and `purge_orphans` raises naming what
stayed stuck rather than returning the bare `False` that surfaced as `"could not purge leftover
unprotected tools/snippets"` — the same unactionable message this PR exists to remove. Its
protected-substrate behaviour is unchanged and now pinned by a test.

## Test results

**Unit tests pass.** `1314 passed, 12 skipped` — `main`'s 1289 plus the 25 added here, skips
unchanged. They drive the real `delete_skill` against a fake store that **enforces the store's 409
rule** rather than recording calls, so they cannot pass vacuously — checked by reverting each fix in a
scratch copy and confirming the relevant tests then fail: 3 for the ordering fix, 3 for the review
follow-ups, 4 for the retry loop and snippet handling, 2 for `purge_orphans`. **Against this branch
all 25 pass.**
One test per entry path, plus: a dependent of another skill is left alone, 404 is the clean-slate
path, a surviving 409 names the unknown dependent, an unexpected 500 carries its body, matching is on
`skill_uuid` rather than name, and a failed dependent delete is named rather than swallowed. The
retry loop is covered by a two-tool case, a **three-deep chain** that a single retry cannot resolve,
and a cycle that must stop rather than loop; snippets by one test that they go with the skill and one
that a shared one is named; `purge_orphans` by four — retrying a blocked orphan, naming one that stays
wedged, keeping the protected substrate, and clearing snippets. The fake store returns `501` for anything it does not model, so an
unexercised path cannot pass on the "404 means already gone" rule.

**Manual verification on both arms:**

- **SPA arm** — the failure is gone. Across repeated runs the store **always started from the seed
  skill**, including the case that used to break it: interrupting `run.sh` and re-invoking it without
  stopping the services. No `could not remove existing skill` at all.
- **Direct arm** — run as well, to confirm nothing regressed. `delete_skill` is shared, so the direct
  arm exercises the same code path without the vMCP/vNFS objects present; it behaves as before.

**Contract verified against the running store**, since the fake store's shapes were read from the
store's source rather than observed:

| checked | result |
|---|---|
| `GET /vmcp_servers/?fields=wide` | `[]` — a bare JSON list, which is the shape `_list` takes |
| `GET /vnfs_servers/?fields=wide` | `[]` — same |
| `GET /skills/<name>?fields=wide` | exposes `uuid` at top level, which is what `delete_skill` reads |
| the dependent row's field name | **`skill_uuid`** — confirmed from the API contract rather than a sample row: `VmcpSchema.skill_uuid` (`schemas/vmcp_schema.py:24`) and `VnfsSchema.skill_uuid` (`vnfs_schema.py:22`) are declared fields, so that is the serialized name by construction |

Worth being precise about what the live runs do and do not prove: both dependent lists were **empty**
at inspection time, so the absence of the failure is consistent with the fix but is not by itself
evidence that the cleanup fired. The field name — the one real inference in the unit tests — is
settled by the schema above instead, which is stronger than a sample row would have been.

## Follow-ups (not in this PR)

**Dangling dependents are not cleaned.** A vMCP/vNFS row whose `skill_uuid` points at an
already-deleted skill matches no live skill, so `delete_skill_dependents` skips it and `purge_orphans`
has no concept of those kinds. It cannot block a future import — a re-imported skill gets a fresh
UUID — so it is store litter rather than a bug, but it accumulates one row per interrupted run, which
is exactly the scenario this PR addresses.

**The store's own cascade should be fixed upstream.** `DELETE /skills/<name>?delete_tools=true` exists
and would make steps 3–4 above a single call, but at 0.2.1 the cascade runs *before* the skill's
dependency on its tools is released, so every `tools_service.delete` inside it raises
`ObjectInUseError` and the cascade swallows it as a `logger.warning`: HTTP 200, `deleted_tools: []`,
tools still there. Moving the cascade after `remove_dependent`, or having it report failures, would
retire this workaround.

When **100% of rollouts fail with the same infrastructure error**, cap-evolve still reported
`baseline val=0.0`, `coverage 0.0`, `headroom=1.0` and `verdict "floor"` rather than aborting. A
total infra failure producing a plausible-looking floor is arguably more dangerous than the bug it
was hiding, since a partially-broken run could skew a real experiment without anyone noticing. Worth
raising separately — it is not SPA-specific.


skills/interventions/llm-proxies/spa/scripts/spa_env.py
```diff
@@ -1104,9 +1162,26 @@ def delete_skill(skill_name: str, protect: Optional[Protection] = None) -> bool:
...
+            f"clearing {' and '.join(k for k, _ in _SKILL_DEPENDENT_KINDS)}"
+            f"{' (and some of those could not be deleted)' if not deps_ok else ''}.\n"
+            f"HTTP 409 from DELETE /skills/{skill_name}: {body.strip()[:400]}\n"
+            "Add the dependent kind named above to _SKILL_DEPENDENT_KINDS.")
     if status not in (200, 204, 404):
-        return False
+        raise RuntimeError(
+            f"DELETE /skills/{skill_name} returned HTTP {status}: {body.strip()[:400]}")
 
     ok = True
```
OsherElhadad, line 1186: `deps_ok` (from `delete_skill_dependents`) is only read inside the `status == 409` branch to enrich the error message — it's never combined into the `ok` this function returns. If a dependent delete fails (e.g. a transient 500 deleting a vMCP row) but the subsequent `DELETE /skills/<name>` still returns 200/204 for some other reason, `delete_skill` returns `True` while an orphaned vMCP/vNFS row is left behind in the store, undetected.

Given the store's own precondition ("can't delete a skill while a dependent row referencing it exists"), this is probably unreachable in practice — a failed delete of a *blocking* dependent should make the skill DELETE 409 anyway, which is caught. But it's worth either asserting that invariant explicitly (`ok &= deps_ok`) or adding a one-line comment noting why it's safe to ignore, since right now it reads like an oversight rather than a deliberate choice. None of the 9 new tests exercise "dependent delete fails AND the skill delete still succeeds" to confirm the assumption either way.

skills/interventions/llm-proxies/spa/scripts/spa_env.py
```diff
@@ -1104,9 +1162,26 @@ def delete_skill(skill_name: str, protect: Optional[Protection] = None) -> bool:
             continue                      # protected -> keep
         deletable.append(tu)
 
-    status, _ = _curl(["-X", "DELETE", _store_url(f"/skills/{skill_name}")])
+    # Clear the skill's dependents FIRST: the store refuses to delete a skill while anything
+    # depends on it, and vMCP/vNFS servers registered against this skill are exactly that. See
+    # delete_skill_dependents for why this is safe to do unconditionally.
+    deps_ok = delete_skill_dependents(manifest.get("uuid") or skill_name)
```
OsherElhadad, line 1168: `manifest.get("uuid") or skill_name` — if the store's manifest ever lacks `uuid` (contrary to the PR's own contract check, but not guaranteed by a type system here), this silently falls back to matching `skill_uuid` against the skill's *name* instead of its UUID. Since real dependent rows always carry an actual UUID in `skill_uuid`, that fallback can never match anything — `delete_skill_dependents` becomes a silent no-op instead of erroring, and the failure only surfaces indirectly via the 409 message a few lines down (which is still fine, just a step removed from the actual cause). Not urgent given the PR's own verification that `uuid` is always present at top level, but a fallback with no code path that actually exercises it is a bit of a smell — either drop it (let a missing `uuid` raise/KeyError loudly) or add a test for it.

OsherElhadad (commented):
Reviewed by cloning the repo into a separate workdir, fetching the PR ref, and running the new test suite directly (both against this branch and against `main` to confirm they actually exercise the fix rather than passing vacuously).

**Verified:**
- `python -m pytest core/tests/test_spa_skill_dependents.py` — 9/9 pass against this branch.
- Against `main`'s `spa_env.py` (pre-fix), the tests that depend on `delete_skill_dependents` fail (that function doesn't exist yet), confirming the suite does test real behavior, not mocks recording calls.
- The ordering fix (`delete_skill_dependents` before `DELETE /skills/<name>`) and the `skill_uuid`-based filtering (leaving a dependent of another skill alone) both check out logically against the described store contract.
- `vmcp_service.py`, `vnfs_service.py`, `skills_service.py` (quoted in the PR body/docstrings as the source of the 409 rule) aren't in this repo — the "store" is evidently an external service. That's fine, and the PR is explicit about this being inferred from schema/API contract rather than observed source, but it does mean the "exactly two kinds" and field-name (`skill_uuid`) claims can't be verified from this codebase — a future third dependent kind or a renamed field would only surface at runtime via the new 409 error. Given the PR already builds that escape hatch in (`_SKILL_DEPENDENT_KINDS` message), this seems like an acceptable, already-mitigated risk rather than a blocker.

Two inline notes below — neither blocking, both about a corner the new tests don't cover.

aviweit:
@OsherElhadad — thanks for the careful review, the corner cases especially. All three points are
addressed in `4adcccfc` and the commits after it.

**`deps_ok` never combined into the return** ([r4006951242](https://github.com/skillberry-ai/cap-evolve/pull/489#discussion_r4006951242)) — you were right that it read as an
oversight. `delete_skill_dependents` now returns the dependents it could **not** delete
(`["<kind>/<uuid>", …]`) instead of a bool, and those are threaded into the same failure path as a
stuck tool, so they are named rather than reduced to a flag. The case you noted as untested —
dependent delete fails *and* the skill delete still succeeds — is now
`test_a_dependent_that_will_not_delete_is_reported_even_if_the_skill_deletes`.

**`manifest.get("uuid") or skill_name`** ([r4006951252](https://github.com/skillberry-ai/cap-evolve/pull/489#discussion_r4006951252)) — dropped, per your first option. A dependent's
`skill_uuid` always holds a UUID, so matching it against a name could never hit anything; a manifest
without `uuid` now raises with the body. Covered by
`test_a_manifest_without_a_uuid_raises_instead_of_matching_on_the_name`.

**The store isn't in this repo** — fair, and it made the central claims uncheckable. Every citation
now names the ref: **`skillberry-store` @ tag `0.2.1` (`c5932cd`)**, the tag `spa_env.STORE_REF` pins,
read from a clone at that tag. "Exactly two dependent kinds" is now stated as established by
enumerating every `add_dependent` call at that tag — the others are skill→tool/snippet and tool→tool,
neither of which makes a *skill* undeletable. Your point that a third kind or a renamed field would
surface only at runtime still stands, and the escape hatch is unchanged.

**A wider ordering issue, from the same audit.** Store objects can depend on each other — a tool
registers a dependency on any tool it calls by bare name, and an object with dependents cannot be
deleted — so deleting a set in listing order fails whenever a callee comes first. Both `delete_skill`
and `purge_orphans` now share one `_delete_all` helper that retries until a pass frees nothing, and
`purge_orphans` names what stayed stuck instead of returning a bare `False`. It runs one line after
`delete_skill` on the same class of objects, so fixing one and not the other would only have moved the
failure.

**Left as a follow-up:** a vMCP/vNFS row whose `skill_uuid` points at an already-deleted skill matches
no live skill, so nothing here clears it. It cannot block a future import, but it accumulates one row
per interrupted run. Happy to fold it in if you would rather it were handled now.

**Tests 9 → 25.** Effectiveness checked the way you checked it — reverting each fix in a scratch copy
and confirming the relevant tests then fail (3 + 3 + 4 + 2); against this branch all 25 pass, suite
1314 passed / 12 skipped. Two notes since you ran the suite yourself: the fake store now models
`DELETE /snippets/`, and returns `501` for anything it does not model — it previously fell through to
`404`, which `_delete_object` treats as "already gone", so an unexercised path could have passed
vacuously.

OsherElhadad (approved):

PR #495, opened 2026-09-19
parsec v4 task-by-task optimization (v4_t2_e1): adapter, scaffolding, single-task runner

bcarmeli:
## Summary

Builds the cap-evolve infrastructure for `v4_t2_e1` — task-by-task optimization
of the 21 sub-1.0 parsec v4 tasks identified by the `v4_t1_e1` baseline —
per the design in `docs/superpowers/specs/2026-09-17-parsec-v4-task-by-task-optimization-design.md`.

- **Harbor adapter** (`scripts/v4_t2_e1/common/adapters/adapter.py`): injects a
  candidate by overwriting parsec's live `config/prompts/*.md` files (picked up
  on the next request via their mtime-keyed cache, no restart needed), with
  snapshot/restore around each evaluation so a crash can't leave one
  candidate's prompts serving the next trial.
- **Per-task scaffolder** (`scripts/v4_t2_e1/scaffold_projects.py`): creates one
  cap-evolve project per task (21 total), each with its own frozen seed
  snapshot, `capevolve.yaml`, and `split_ids.json` pinning train=val=test to
  that one task.
- **Single-task runner** (`scripts/v4_t2_e1/run_one_task.py`) and **pre-flight
  check** (`scripts/v4_t2_e1/preflight_check.py`), plus an optimizer
  `INSTRUCTIONS.md` scoped to system-prompt-only edits.
- **Docs**: the 2026-09-09 intake design, the 2026-09-17 task-by-task design
  (now including a section on why v1/v2's cap-evolve intake pattern doesn't
  carry over to v4's live-process architecture), the implementation plan, and
  the 2026-09-09 session handoff notes.

This is infrastructure only — one task (`icinga-011-aap2-job-status-alert`)
has been smoke-tested end-to-end (1 capped iteration; no improving candidate
found). The full 21-task sweep has not yet been run.

## Test plan

- [x] `uv run --no-project --with-editable "./core[dev]" --with pytest --python 3.13 python3 -m pytest scripts/v4_t2_e1/tests -q`
- [x] `uv run --no-project --with-editable "./core[dev]" --with-editable "./dashboard/backend[dev]" --with pytest --python 3.13 python3 -m pytest dashboard/backend/tests -q` (38 passed)
- [ ] Full 21-task optimization sweep (tracked separately, not part of this PR)


OsherElhadad:
Reviewed. This is infra-only (as the PR says) and it's heavily documented, but one real portability/maintainability issue stands out in the code (not just the docs):

**The `PARSEC_V4N` default is a hardcoded personal absolute path, duplicated in two source files instead of coming from one shared constant.**

- `scripts/v4_t2_e1/common/adapters/adapter.py`:
  ```python
  V4N = Path(os.environ.get("PARSEC_V4N", "/Users/boazc/workarea/Python/rhdp-parsec/v4_2026-09-16"))
  ```
- `scripts/v4_t2_e1/scaffold_projects.py`:
  ```python
  PARSEC_V4N = Path(
      os.environ.get("PARSEC_V4N", "/Users/boazc/workarea/Python/rhdp-parsec/v4_2026-09-16")
  )
  ```

`scaffold_projects.py` even calls this out in its own comment right above the constant:

> Same env var, same default, as common/adapters/adapter.py's V4N. Two sources of truth for "which parsec checkout" is how a seed snapshot gets frozen from one tree while the trials that score it run against another.

So the drift risk is correctly identified but not fixed — the two defaults will silently diverge the moment one file's default is bumped (e.g. to a new dated snapshot dir) and the other isn't. Given the adapter is loaded fresh per-instance specifically so an env override always takes effect (per its own docstring), the natural fix is to have `scaffold_projects.py` import `V4N`/`PARSEC_V4N` from `common/adapters/adapter.py` (or a small shared `common/` constants module) rather than re-declaring the same default independently.

Separately, this also means the *default* value (used whenever `PARSEC_V4N` isn't exported) is a specific developer's home-directory path with a dated snapshot folder name (`v4_2026-09-16`) baked into a committed file. That's fine as a personal dev convenience today, but it's a landmine for anyone else who runs `scaffold_projects.py`/`run_one_task.py` without first discovering they need to set `PARSEC_V4N` — they'll get a confusing downstream `FileNotFoundError` (from the `bench-v4-<id>` directory check in `adapter.py.__init__`) rather than a clear "PARSEC_V4N is not set" error. Worth either requiring the env var explicitly (same pattern already used for `TASK_ID`, which raises a clear `RuntimeError` when missing) or at least validating `V4N.exists()` up front with a message naming the env var.

One smaller note: the module docstring for `common/adapters/adapter.py` says trial isolation relies on "`harbor run --n-concurrent 1`, plus `live()`'s own advisory lock" — but the actual advisory `flock` lives in `run_one_task.py`'s `run_task()`, not inside `Adapter.live()` (which only snapshots/restores the prompt files, with no locking of its own). Minor, but worth fixing the comment so a future edit to `live()` doesn't assume it's already lock-protected.

bcarmeli:
## Update: CCC/LSF port of the `v4_t2_e1` benchmark harness

This PR now also includes the port of the `v4_t2_e1` harness from Mac-local
execution to CCC/LSF, per
`docs/superpowers/plans/2026-09-23-parsec-v4-ccc-port.md` (15 commits,
`da299d99..5a0ce2d3`).

- **`scripts/ccc/parsec_stack.sh`**: brings up/down the 5 MCP simulator
  containers (podman-managed, image `simulation-harness:bench`) and
  `parsec-live` on a CCC compute node, with activation + readiness polling
  (`GET /api/v1/simulation`) instead of a bare port check, and a
  host-scoped PID directory so concurrent jobs on different nodes don't
  collide.
- **`scripts/ccc/run_ccc_parsec.sh`**: drives one `v4_t2_e1` task through
  `cap-evolve` on LSF, with an EXIT-trap teardown so the stack (and
  `parsec-live`) always comes down when the job ends — including on an
  LSF wall-clock kill or `bkill`, not just a clean exit.
- **`docs/how-to/ccc/CCC_PODMAN_SETUP.md`**: setup/troubleshooting for the
  podman-based container stack on CCC.

This work went through 4 rounds of whole-branch review (each round's
findings fixed and re-verified before the next) before merge-readiness;
notably round 3 caught a regression that round 2's own fix had introduced
(a job killed *during* stack startup could leak the whole container stack —
fixed by re-arming the teardown trap before startup instead of after).
Final round-4 review returned zero findings.

**Known gap, not yet exercised in review**: nothing in this branch has been
run against real podman containers or a real `parsec-live` process — no CCC
allocation with the `simulation-harness:bench` image was available in any
review session. The first real run on a CCC node is the true acceptance
test for the container/`parsec-live` lifecycle.

**Also outstanding, not part of this PR**: the LiteLLM API key in
`_run/parsec-live/config/config.local.yaml` on the CCC trees was chmod'd to
`600` in an earlier session but still needs to be rotated — that's a
pending action for whoever holds that key, not something this branch does.

PR #496, opened 2026-09-19
ci(benchmarks): run the tau2-airline direct and spa arms as benchmark legs

aviweit:
# ci(benchmarks): run the tau2-airline direct and spa arms as benchmark legs

Adds CI support for the two delivery arms of the tau2-airline onboarding, so the delivery path
itself gets measured.

**The user experience.** The Benchmarks form gains an **`intervention`** input — `direct` | `spa`,
defaulting to `direct` — and the selected benchmark runs under the intervention chosen.
A new **`tau2-custom`** picker option is created by this PR. It is the tau2-airline onboarding,
and `intervention` selects which of its two arms runs.

**`direct` is the default, and `spa` is ignored where a benchmark does not support it** — that
benchmark simply runs `direct`. Today only `tau2-custom` obeys both values; for every other option
the input has no effect. Both arms run the same airline task ids with the same tools-only surface,
sourced from `examples/skillberry_benchmarks_tau2_airline/`; only delivery differs.

## A decision worth reviewing

**`tau2-custom` installs into its own CI venv** (`venv-skillberry-tau2`); every other benchmark
keeps the shared one.

The cause is **two tau2 flavours in one tree.** `tau2-custom` installs `tau2` from
`skillberry-benchmarks` — the pinned fork, with the `[skillberry]` extra and the
`airline_skillberry` domain — while the `tau2` leg installs the public `sierra-research` checkout.
Same package name, two sources, both editable, so one venv can only hold whichever installed last,
and the loser fails at *run* time with a missing domain rather than at install time. Before this
branch no two benchmarks shared a package name, so one venv was safe.

**It is defensive, not a fix for a live bug.** Legs serialise on a single runner, and `ci_setup.sh`
installs the right `tau2` immediately before each one, so the shared venv would work today. What it
protects against is a second runner — and a runner is a *process*, not a machine: two runners on the
same VM is the cheap way to add throughput, and they share `$HOME`, so they share that venv. Legs
would then overlap and one would get the other's `tau2`. One extra venv is cheap insurance against a
fleet change nobody would think of as risky.

Unrelated but easy to confuse with the above: the **Store and Proxy-Agent** each get their own
py3.11 venv inside their own clone, created and installed by `spa_env` in the intervention skill.
That is how the skill has always worked and this PR does not change it.

## Per-arm behaviour

| | `direct` | `spa` |
|---|---|---|
| delivery | in-process | Store serves the skill, Proxy-Agent injects it |
| agent model | the gateway model | `ibm/skillberry-local` sentinel; `SPA_MODEL_NAME` derived from `AGENT_MODEL` |
| services | none | Environment Manager `:8004`, Store, Proxy-Agent — torn down on an `EXIT` trap |
| rollout concurrency | 10 | 4 — the adapter's own default, matching what the arm runs with locally |

Both install `skillberry-benchmarks` at the pin the arms' own `setup.sh` uses; a test keeps the two
pins equal. Service logs are read-only — each service rotates its own, so the leg copies bounded
tails into the artifacts.

## Also here

- **Raw tau2 trajectories persisted** for all three tau2 legs, uploaded from the run dir — which
  sits outside the previously-uploaded output dir, so they were being written and discarded.
- **`_spa_env()` no longer locates the intervention library by parent count.** `parents[3]` reaches
  the repo root only for `<repo>/.capevolve*/project`; under CI's deeper layout it resolved to the
  work dir and every rollout failed to deploy. `CAPEVOLVE_SKILLS_DIR` first, parent count as
  fallback, so local runs are unchanged.
- **The live panel shows the arms — but only after this merges.** See the note below; that is
  expected, not an outstanding bug.
- **The Config pane gains a Delivery section** and renders `intervention=direct` when the spec omits
  the key. Previously a run recorded nothing about how it delivered. `run_config` carries the
  resolved value too.

## Reviewing the live panel: it cannot work from a branch

**Expected behaviour, worth knowing before you test it.** `pages.yml` pins `ref: main` — a
deliberate choice, commented there — so the benchmarks page always serves `main`'s
`site/benchmarks.js`, whatever branch a run was dispatched from. Pushing this branch or dispatching
from it cannot change what that page can read.

`main`'s copy enumerates five bench names:

```js
/^([a-z][a-z0-9-]*) \/ (tau2|swebench|skillsbench|spreadsheetbench|rfe-creator)$/
```

and a leg's job name is `<tier> / <matrix.bench>`, so:

| dispatched | job name | matched by main's regex |
|---|---|---|
| `tau2` | `smoke / tau2` | yes |
| `skillsbench` | `smoke / skillsbench` | yes |
| `tau2-custom` + direct | `smoke / skillberry_tau2_direct` | **no** |

That is why the arms have been invisible in that panel since they were added, while every other
benchmark appears. This PR widens the regex to any bench token and maps the arms to
`tau2-custom (direct|spa)` for display, keeping the raw leg name in the data URL so the "Watch live"
link still resolves. Its test hardcoded the same five names — which is why nothing caught this — and
now derives `BENCHES` from the workflow.

**So: the panel will not show an arm until this is merged.** Verify the fix by reading
`site/benchmarks.js` and `core/tests/test_site_live_panel_tiers.py`, or by dispatching after merge —
not by dispatching from the branch.

## What this touches outside CI

A branch named for CI changes four things that are not CI. Each is small and each is there for a
defect the CI work exposed, but they need eyes beyond the benchmark suite:

| area | files | effect |
|---|---|---|
| core | `core/cap_evolve/cli.py`, `dashboard.py` (+12 lines) | every run records and displays its delivery path |
| example | `examples/…/spa/adapters/adapter.py` | local spa runs keep working via the `parents[3]` fallback |
| site | `site/benchmarks.js` | the public Pages live panel |
| tests | 4 of 6 test files are pre-existing corrections, not new coverage | two encoded stale facts and would have failed or stayed blind |

## Verification

**1355 passed, 12 skipped.** Mostly static analysis of the committed wiring plus a behavioural test
of the library lookup: picker/planner agreement, the input's values and resolution, identical task
ids across both arms, parsed spec keys, an absolute `runner_repo_path`, both project layouts
resolving.

Key invariants mutation-tested — breaking spa concurrency, the benchmark pin, the planner list, the
trajectory default, the env-var lookup and the heredoc guard each failed exactly the intended test.

### On the self-hosted runner

**`tau2-custom` + `direct` — complete and published.** Dispatched at `smoke`, 1 iteration, 1 trial:
reward `0.400 -> 0.400`, optimizer `$4.5942`, 21m55s, result `success`. The row reached the
benchmark-history page, which also confirms the artifact shape the aggregate job depends on: a
second `path:` on the upload had re-rooted the artifact at `.work/`, so `record.py` logged
`no runmeta for <slug>, skipping` and published nothing for any tau2-family leg. Fixed here by
keeping the upload to one path and leaving raw trajectories in the run dir.

**`tau2-custom` + `spa` —  complete and published.** The full SPA path ran: the tau2 Environment
Manager serving `:8004` with the `airline_skillberry` domain registered, the Store on `:8000`/`:8002`
holding the candidate skill, and the Proxy-Agent on `:7000`/`:7001` injecting it, with agent turns
routed through the `ibm/skillberry-local` sentinel to the gateway model resolved from
`SPA_MODEL_NAME`.

So both values of `intervention` have now been exercised on the runner through this branch's
`tau2-custom` picker option.

## Follow-ups, not in this PR

- **The direct arm should report tokens and does not.** It hardcodes `tokens=0` (adapter.py:403)
  while the `tau2` leg reports 1.08M for the same rollouts — the template's `_cost_and_tokens(sim)`
  sums `prompt_tokens + completion_tokens` from the very `sim` objects the arm already has, so this
  is fixable there. It predates this PR, which only makes it visible by running the arm in CI, but
  it leaves cost — one of the three headline metrics — structurally zero for that leg. The spa arm
  is a different case: SPA surfaces no upstream usage, so its 0 means *not measured* and wants
  labelling rather than computing.
- No `max-parallel` on the bench matrix; serialization comes from having one runner, not from the
  workflow.
- The `all` benchmark option will be removed: running every benchmark takes a very long time and
  there is no need to offer it. Git issue: #497


.github/workflows/benchmarks.yml
```diff
@@ -29,7 +29,12 @@ on:
         description: "Which benchmark to run"
         type: choice
         default: all
-        options: [all, tau2, swebench, skillsbench, spreadsheetbench, rfe-creator]
+        options: [all, tau2, swebench, skillsbench, spreadsheetbench, rfe-creator, tau2-custom]
```
eranra, line 32: we need to decide tau2-custom or tau2-balackbox ... in any event the names od the folders and the code and the name of the benchmarks need to be aligned. both are good but we need to choose one that will be all over.

.github/workflows/benchmarks.yml
```diff
@@ -29,7 +29,12 @@ on:
...
+      intervention:
+        description: "How the candidate reaches the model."
+        type: choice
+        default: direct
+        options: [direct, spa]
```
eranra, line 37: please --- no spa --- let's say blackbox

.github/workflows/benchmarks.yml
```diff
@@ -188,7 +191,20 @@ jobs:
...
+          # Store + Proxy-Agent. They stay separate legs so each keeps its own tier task lists,
+          # its own history row and its own concurrency group.
+          BENCHES = ["tau2", "swebench", "skillsbench", "spreadsheetbench", "rfe-creator",
+                     "skillberry_tau2_direct", "skillberry_tau2_spa"]
+
+          # `tau2-custom` is the picker's name for "the arms"; `intervention` chooses which one.
+          # A benchmark token is what the rest of the pipeline keys off — tasks.json paths, the
+          # work dirs, the history rows — so resolve it here, once.
+          CUSTOM = "tau2-custom"
+          ARM_OF = {"direct": "skillberry_tau2_direct", "spa": "skillberry_tau2_spa"}
```
eranra: this naming peoblem is all over ...

Also -- the Claude name for slection or options is "arm" but this is not common and not good technical english --- option is a simple term that I propse to use.

ci/benchmarks/lib/ci_setup.sh
```diff
@@ -10,7 +10,14 @@
 set -euo pipefail
 BENCH="${1:?bench}"
 CACHE="${CAPEVOLVE_CI_CACHE:-$HOME/.cache/capevolve-ci}"
-VENV="$CACHE/venv"
+# The arms get their OWN venv. They install tau2 from skillberry-benchmarks; the `tau2` leg
+# installs the public sierra-research checkout. Same package name, two sources, both editable —
+# in one venv whichever ran last wins, and the loser fails with a missing domain rather than an
+# install error. Sequencing on a single runner hides it; a second runner would not.
+case "$BENCH" in
+  skillberry_tau2_*) VENV="$CACHE/venv-skillberry-tau2" ;;
```
eranra: Can we change something that we will not need a "corenr case" handling in the code ... we can change the benchmarks or something but I hope we can change this to be generic and not need a corner case handling here.

ci/benchmarks/lib/ci_setup.sh
```diff
@@ -146,6 +153,62 @@ case "$BENCH" in
   tau2)
     [ -d "$CACHE/tau2-bench/.git" ] || git clone --depth 1 https://github.com/sierra-research/tau2-bench "$CACHE/tau2-bench"
     uv pip install -p "$CAPEVOLVE_PY" -q $IDX -e "$CACHE/tau2-bench" ;;
+  skillberry_tau2_direct|skillberry_tau2_spa)
```
eranra: It makes things over complicated here because of this anomally that the name of the benchmark is the same ... maybe we can change at the source or something ... I really think we should not create this complexity to the code if we do not have to.

ci/benchmarks/lib/make_arm_tiers.py
eranra, line 1: can we stop using the term arm

OsherElhadad:
Reviewed this PR together with #495, including the pre-existing SPA (Store + Proxy-Agent) blackbox code it wires into CI (`skills/interventions/llm-proxies/spa/scripts/spa_env.py`, `examples/skillberry_benchmarks_tau2_airline/spa/`). Overall the CI plumbing is very carefully reasoned (the venv-collision defense, the log-rotation-at-start discipline, the absolute-path fixes for `optimizer_instructions_file`/`runner_repo_path`). Two things worth fixing before merge:

**1. `ci/benchmarks/README.md` documents the wrong concurrency for the `spa` arm.**

The new table says:

> | rollout concurrency | 10 | **1** — the proxy binds one skill at a time, so parallel candidates would serve the wrong tools |

But the actual code (`ci/benchmarks/lib/run_suite.sh`, in the `skillberry_tau2_spa` branch) sets:

```sh
export TAU2_MAX_CONCURRENCY="${TAU2_MAX_CONCURRENCY:-4}"
```

which matches the adapter's own default (`TAU2_MAX_CONCURRENCY` defaults to `"4"` in `examples/skillberry_benchmarks_tau2_airline/spa/adapters/adapter.py`), and is explicitly asserted by the new test `test_the_spa_arm_matches_the_adapters_own_concurrency` in `core/tests/test_ci_skillberry_tau2_arms.py`. So the code and the test agree on **4**, but the README says **1**, with a rationale ("parallel candidates would serve the wrong tools") that doesn't actually apply here — within one evaluation all 4 concurrent rollouts hit the *same* deployed candidate/skill, they aren't different candidates racing each other. Looks like the README is a stale leftover from an earlier revision where concurrency really was 1. Please update the doc to match the code (or vice versa if 4 is actually wrong and the test/PR description are also wrong).

**2. New fixed-port service (`tau2 Environment Manager`, port 8004) has no protection against the same "second runner on the same VM" scenario the PR otherwise carefully defends against.**

The PR description calls out, as a deliberate defensive measure, that two runner *processes* sharing one VM/`$HOME` would otherwise stomp on each other's `tau2` venv — hence the new `venv-skillberry-tau2`. But the new Environment Manager start-up logic in `run_suite.sh` (the `skillberry_tau2_spa` branch) hardcodes `ENV_PORT="${ENV_PORT:-8004}"`, and only checks "is anything answering on 8004" before deciding to start vs. reuse it:

```sh
if curl -sf -o /dev/null --max-time 3 "http://127.0.0.1:$ENV_PORT/docs" \
   || curl -sf -o /dev/null --max-time 3 "http://127.0.0.1:$ENV_PORT/"; then
  echo "  tau2 environment service already up on $ENV_PORT"
```

If two `skillberry_tau2_*` legs (or a `spa` leg and something else that happens to bind 8004) run concurrently on the same self-hosted runner box, the second one will either fail to bind the port or — worse — silently treat the first leg's already-running Environment Manager as "healthy" and reuse it across two unrelated runs. That's the exact class of cross-run contamination the venv fix exists to prevent, just left open here. Given the PR already reasons carefully about this failure mode elsewhere, it'd be worth either namespacing `ENV_PORT` per leg (e.g. derived from `$BENCH`/`$TIER`) or explicitly documenting that this leg must never run with a sibling on the same host.

Nothing else jumped out — the shell heredoc quoting guard (no backticks/`$` in the `EXTRA_YAML` comments) and the `_spa_adapter_skills_lookup` regression test for the CI vs. local project-nesting path are both nice, targeted fixes.

.github/workflows/benchmarks.yml
aviweit: Aligned on `tau2_custom`.

Everything now uses that one name:

- picker option: `tau2-custom`
- leg names: `tau2_custom_direct` / `tau2_custom_spa`
- CI tier dirs: `ci/benchmarks/tau2_custom/{direct,spa}/<tier>/`
- examples: `examples/tau2_custom/{direct,spa}/`
- the CI venv: `venv-tau2-custom`

.github/workflows/benchmarks.yml
aviweit: Agreed on the term — `blackbox` reads better than `spa` for a delivery mode.

Deferring it to a successive PR rather than doing it here: `spa` is not just this input's value, it is also the pre-existing intervention skill (`skills/interventions/llm-proxies/spa/`), its `spa_env` module and the example directories this PR only wires into CI. Renaming it properly means touching that pre-existing code, which would enlarge a PR whose subject is the CI wiring, and `direct`/`spa` is a user-visible dispatch value so the change is worth making once, deliberately.

So this PR keeps `direct` | `spa`, and the rename to `blackbox` lands separately.

ci/benchmarks/lib/make_arm_tiers.py
aviweit: Removed — the file is deleted in this branch.

It was a hand-run generator: it wrote the six committed `tasks.json` files and nothing in CI ever called it (no workflow, script or test referenced it). Its purpose was to stop the two options' task lists from drifting apart, but a generator nobody runs cannot enforce that — editing a `tasks.json` by hand bypassed it silently.

Two tests now enforce the invariant instead of documenting it:

- `test_the_two_arms_run_identical_task_ids` — per tier, both lists must be identical
- `test_smoke_matches_the_plain_tau2_leg` — smoke must match `ci/benchmarks/tau2/smoke`

So the property the script existed for is now actually checked, and the term is gone with the file.

aviweit:
@OsherElhadad, thank you for this review. Both findings were real.

## 1. README concurrency — fixed

You were right, and the number you questioned (4) was the correct one. Fixed in `fd85ca86`: the row now reads `10 | 4`, matching `run_suite.sh` (10 for direct, 4 for blackbox) and the adapter's own default.

## 2. `ENV_PORT` 8004 — you are right, and it is one symptom of a larger one

Agreed: Added to the PR's follow-ups, and tracked separately as #509.

**The general shape.** Port 8004 is one of several identifiers that would have to become per-instance before **two blackbox optimizations could run concurrently at all** — on this benchmark or any other. The others are in the delivery path itself, not in the CI wiring:

- **the skill name is a constant.** `SKILL_NAME = "my_skill"` in the arm's adapter, matched by `skill_name: my_skill` in the generated spec and `spa_env.start_spa("my_skill")`. Two concurrent blackbox optimizations would be deploying candidates under one name.
- **the store holds one skill by design.** `reset_store_to_skill()` makes the candidate *the only* skill in the store, from a clean slate, on every `apply()` — which is what guarantees a run starts from a known state. Two concurrent optimizations would wipe each other's candidate even with distinct names, so distinct names alone would not be enough without a per-instance store.

So concurrent blackbox optimization is not a missing `ENV_PORT` variable — it needs instance-scoped identity across the stack: env-manager port, skill name, store instance. Your comment is the visible end of that.

**Why it works today.** The repo has exactly one self-hosted runner (`skillberry-1-capevolve`, labels `self-hosted,Linux,X64,ibm-vpc`) and the bench job targets `[self-hosted, ibm-vpc]`, so the matrix legs serialise and nothing is contended. That is a property of the fleet, not of the code — nothing in the workflow enforces it (there is no `max-parallel`, and the `concurrency` group is per-leg, so different legs are explicitly allowed to overlap). A second runner process on the same VM, sharing `$HOME`, exposes all of the above — which is the same argument that justified the separate venv in this PR.

**An aside comment on the failure mode**, having gone through `tau2/orchestrator/environment_manager.py`: reuse of the manager itself does **not** cause cross-run state contamination. It is multi-tenant by construction — `self.environments` is keyed by `env_id`, each conversation gets a fresh `uuid4()`, and every `start_environment` builds a new Environment, which for airline means `FlightDB.load(...)` parsed fresh from disk. Trajectories and routes are per-`env_id`, and `stop_environment` deletes only that key. So two runs sharing one manager see disjoint databases; the manager is the one service in this stack that is genuinely safe to share.

What is at risk there is process lifetime: the blackbox leg ends with `pkill -f "EnvironmentManager(host='127.0.0.1', port=$ENV_PORT)"`, so whoever finishes first kills the manager the other leg is mid-rollout on. And the liveness probe does not check that the adopted manager registered the domain this leg needs.

OsherElhadad (approved):
LGTM

